PR #2, opened 2017-07-24
[BPK-551] Add travis integration

k0nserv:
Adds initial travis CI. We'll do the docs deploy later

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: node_js
+node_js:
+  - "7.5.0"
+env:
+  - NPM_CONFIG_LOGLEVEL=warn
+install:
+  - npm install
+  - npm run build
+script:
+  - npm test
+  - npm run docs:dist:gzip
+  - test -e dist/index.html
+  - test -e dist/sassdoc/index.html
+cache:
+  directories:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: node_js
+node_js:
+  - "7.5.0"
+env:
+  - NPM_CONFIG_LOGLEVEL=warn
+install:
+  - npm install
+  - npm run bootstrap
```
tvooo: Do we need this? `bootstrap` is run as postinstall script, meaning `npm install` should be sufficient. However, for some reason, Drone doesn't do it. Is that the same for Travis?

k0nserv: Good shout I just copied it over from `.drone.yml`. Let's see what happens in the [build](https://travis-ci.org/Skyscanner/backpack/builds/256812600)

k0nserv: You're right postinstall does work correctly and `npm run bootstrap` ends up running twice. I'll remove one entry

k0nserv:
I added caching for `node_modules` hopefully that will just work even with Lerna creating symlinks

PR #4, opened 2017-07-24
[BPK-551] Remove comments from compiled CSS and add license header

georgegillams:


packages/bpk-stylesheets/webpack.config.js
```diff
@@ -106,6 +114,10 @@ module.exports = {
   },
 
   plugins: [
+    new WrapperPlugin({
+      test: /\.css$/,
+      header: license,
```
k0nserv: Would it be possible to reuse the text blob in [license-header.js](https://github.com/Skyscanner/backpack/blob/master/packages/bpk-tokens/formatters/license-header.js) instead of introducing another place were the comment is present?

georgegillams: Absolutely -  just didn't realise it was already available somewhere!

k0nserv: 👍 We need to maintain the normalizer license too I believe...

tvooo:
@k0nserv could you review this again? I made changes so am not fit to be a reviewer.

georgegillams:
If it wasn't valid, would the docs page not look broken?

k0nserv:
Yeah that'd be one way to verify it 👍 I take it that the docs don't look broken then?

georgegillams:
Just double-checking, but I'm pretty sure I would have noticed!

PR #6, opened 2017-07-24
[BPK-693] Add gulp step to set correct SVG permission

georgegillams:


k0nserv:
Shouldn't this also include a bunch of changes to the svgs?

k0nserv:
I ran this locally and it didn't produce a change which brings the question why is this needed? cc @matthewdavidson

georgegillams:
The icon svgs all have the correct permissions, so this should not change them. If you 'chmod +x *' in one of the directories and then rebuild you should find the output ones lose executableness

tvooo:
To clarify: as this is async, on some runs the output will have bad permissions, and sometimes it won't, which then ends as a change in git even though nothing changed in the source.

k0nserv:
Right that makes sense then, LGTM 👍

PR #11, opened 2017-07-25
[BPK-814] Change alignment HOC behaviour

georgegillams:


matthewdavidson:
I noticed that some of the examples on the alignment page are missing / incomplete:
screen shot 2017-08-01 at 14 22 40
The large button snippet should be like this:
```
<BpkButton large>
  Button Text
  &nbsp;
  <AlignedLargeButtonArrow fill={colors.colorWhite} />
</BpkButton>
```

PR #13, opened 2017-07-26
[BPK-816] text component

k0nserv:
This can be reviewed. WIP because I still need to write up the docs

### DIFF
--- a/packages/bpk-component-text/package.json
+++ b/packages/bpk-component-text/package.json
@@ -1,0 +1,20 @@
+{
+  "name": "bpk-component-text",
+  "version": "0.0.1",
+  "description": "Backpack text component.",
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^15.0.0"
+  },
+  "dependencies": {
+    "bpk-mixins": "^15.7.1",
+    "bpk-react-utils": "^2.1.1",
+    "prop-types": "^15.5.8"
+  }
+}
--- a/packages/bpk-component-text/readme.md
+++ b/packages/bpk-component-text/readme.md
@@ -1,0 +1,60 @@
+# bpk-component-text
+
+> Backpack text component.
+
+## Installation
+
+```sh
+npm install bpk-component-text --save-dev
+```
+
+## Usage
+
+```javascript
+import React from 'react';
+import BpkText from 'bpk-component-text';
+
+export default () => (
+  <BpkText tagName="h1" textStyle="xl">My Heading</BpkText>
+);
+```
+
+When using the same style in many places repeating the `textStyle` and `tagName` props can become tedious in this case you can use `withTextStyle`
+
+```javascript
+import React from 'react';
+import BpkText, { withTextStyle } from 'bpk-component-text';
+
+const LargeParagraph = withTextStyle(BpkText, 'lg', 'p');
+const TinySpan = withTextStyle(BpkText, 'xs', 'span');
+
+export default () => (
+  <div>
+    <LargeParagraph>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+      Pellentesque imperdiet lobortis tellus, non rhoncus erat tincidunt id.
+      Pellentesque consectetur, dolor nec vulputate vehicula, ex metus mattis ante,
+      non dictum mi ante eu arcu.
+    </LargeParagraph>
+    <LargeParagraph>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+        Pellentesque imperdiet lobortis tellus, non rhoncus erat tincidunt id.
+        Pellentesque consectetur, dolor nec vulputate vehicula, ex metus mattis ante,
+        non dictum mi ante eu arcu.
+    </LargeParagraph>
+    <TinySpan>
+      Side effects of Backpack include euphoria, happiness, and increased develpoment velocity.
+    </TinySpan>
+  </div>
+);
+```
+
+## Props
+
+| Property  | PropType                                 | Required | Default Value |
+| --------- | ---------------------------------------- | -------- | ------------- |
+| children  | -                                        | true     | -             |
+| textStyle | 'xs', 'sm', 'base', 'lg', 'xl', 'xxl'    | false    | 'base'        |
+| tagName   | 'span', 'p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' | false    | 'span'        |
+| className | string                                   | false    | null          |
+| bold      | bool                                     | false    | false         |
--- a/packages/bpk-component-text/src/BpkText.js
+++ b/packages/bpk-component-text/src/BpkText.js
@@ -1,0 +1,62 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import { cssModules } from 'bpk-react-utils';
+
+import STYLES from './BpkText.scss';
+
+const getClassName = cssModules(STYLES);
+
+const TEXT_STYLES = ['xs', 'sm', 'base', 'lg', 'xl', 'xxl'];
+
+const classes = {};
+TEXT_STYLES.forEach((textStyle) => { classes[textStyle] = getClassName(`bpk-text--${textStyle}`); });
+
+
+const BpkText = (props) => {
+  const { bold, className, children, tagName: TagName, textStyle, ...rest } = props;
+  const classNames = [getClassName('bpk-text'), classes[props.textStyle]];
+
+  if (bold) { classNames.push(getClassName('bpk-text--bold')); }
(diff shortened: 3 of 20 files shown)


packages/bpk-component-text/readme.md
```diff
@@ -0,0 +1,61 @@
...
+	      Pellentesque consectetur, dolor nec vulputate vehicula, ex metus mattis ante,
+	      non dictum mi ante eu arcu.
+	  </LargeParagraph>
+  	  <TinySpan>
+  		Side effects of Backpack include euphoria, happiness, and increased develpoment velocity.
+      </TinySpan> 
+  </div>
+);
+```
+
```
matthewdavidson: Can we have one new line between text blocks? :)

packages/bpk-component-text/src/BpkText.js
```diff
@@ -0,0 +1,62 @@
...
+  TYPES.map(type => [type, getClassName(`bpk-text--${type}`)]),
+);
+
+const BpkText = (props) => {
+  const classNames = [classes.get(props.textStyle)];
+
+  if (props.className) { classNames.push(props.className); }
+
+  return (
+    <props.tagName className={classNames.join(' ')}>
```
matthewdavidson: How do I as a consumer add arbitrary props? i.e. `<BpkText id="my-id">Lorem ipsum ...<BpkText/>` wont work...

packages/bpk-component-text/src/BpkText.js
```diff
@@ -0,0 +1,62 @@
...
+import React from 'react';
+import { cssModules } from 'bpk-react-utils';
+
+import STYLES from './BpkText.scss';
+
+const getClassName = cssModules(STYLES);
+
+const TYPES = ['xs', 'sm', 'base', 'lg', 'xl', 'xxl'];
+
+// FIXME: Should we prefer `Map` or Object?
```
matthewdavidson: I'd always trend towards boring plain old objects unless we get explicit value of using a `Map`. I don't think we do in this case.

packages/bpk-component-text/src/BpkText.scss
```diff
@@ -0,0 +1,27 @@
+@import '~bpk-mixins';
+
+.bpk-text {
```
matthewdavidson: So technically this isn't strictly bem right ...the only classes that are exposed here are modifier classes, i.e. the html will look like:
```
<span class="bpk-text--sm" />
```
Whereas strict BEM would be:
```
<span class="bpk-text bpk-text--sm" />
```

This comment is here more as an observation as opposed to a requests to change ... thoughts?

packages/bpk-component-text/src/withTextStyle.js
```diff
@@ -0,0 +1,46 @@
...
+
+const withTextStyle = (WrappedComponent, type, tagName) => {
+  const component = (props) => {
+    const { children, ...rest } = props;
+
+    return (
+      <WrappedComponent
+        textStyle={type}
+        tagName={tagName}
+        {...rest}
```
matthewdavidson: if consumers apply a `textStyle` or `tagName` it will override the `textStyle` or `tagName` being provided by the HOC. Is this expected?

packages/bpk-component-text/src/BpkText.js
k0nserv: Yeah the only reason I favoured `Map` is because it has a constructor that takes key, value pairs and that can easily be constructued with `map` when the value is derivable from the key

packages/bpk-component-text/readme.md
k0nserv: Yeah this is just down to my markdown editor not being great at displaying source level whitespace

packages/bpk-component-text/src/BpkText.js
k0nserv: I looked at `BpkHeading` which does support id, but doesn't pass down all other properties. I'm open to doing that though

packages/bpk-component-text/src/BpkText.scss
k0nserv: Yeah I know, feels kinda iffy to have `bpk-text` as a class just for the sake of it. Would `bpk-text-sm`, `bpk-text-base` be better or worse?

packages/bpk-component-text/src/withTextStyle.js
k0nserv: Good point, I guess it could be considered both a bug and a feature. Do we have any precedence we can align the behaviour with?

packages/bpk-component-text/src/BpkText.js
matthewdavidson:
```js
const classes = {};
TYPES.map(type => classes[type] = getClassName(`bpk-text--${type}`))
```

packages/bpk-component-text/src/BpkText.js
matthewdavidson: Yeah as a general rule we should allow arbitrary props to be added as well as custom classNames.

packages/bpk-component-text/src/BpkText.scss
matthewdavidson: I reckon it would be better - they are distinct, right?

packages/bpk-component-text/src/withTextStyle.js
matthewdavidson: I think the HOC styles should win in this case - consumers will be forced to consider why thigns aren't working and realise that they probably shouldn't be using the HOC or vice versa.

matthewdavidson: There might be another example in the codebase somewhere...

packages/bpk-component-text/src/BpkText.js
k0nserv: Yeah, not as nice though and the `map` should be `forEach`. But yeah a plain object is better

packages/bpk-component-text/src/withTextStyle.js
k0nserv: > I think the HOC styles should win in this case - consumers will be forced to consider why thigns aren't working and realise that they probably shouldn't be using the HOC or vice versa.

Yeah I like that line of reasoning 👍

packages/bpk-component-text/src/BpkText.js
matthewdavidson: oh gosh 😳

packages/bpk-component-text/src/BpkText.js
k0nserv: Done

packages/bpk-component-text/src/BpkText.js
k0nserv: Done

packages/bpk-component-text/src/BpkText.scss
k0nserv: Yeah, just did it

k0nserv:
We should remove the global style rules that are bleeing through here to prevent people from misunderstanding what the component does
image

k0nserv:
This is what happens in the storybook and the reason for the margins are [normalize.css](https://github.com/necolas/normalize.css/blob/master/normalize.css#L49). I think we should set `margin: 0` for `BpkText`.
image

jamesf3rguson:
I agree we should zero them @k0nserv

k0nserv:
Pushed documentation changes. I suggest we review and merge this and then add the bold prop after we merge #15

k0nserv:
Hmm since #15 is merged I'll rebase this and add bold then

matthewdavidson:
@k0nserv dont forget a changelog entry ;)

k0nserv:
Changelog and bold added. Ready for review

packages/bpk-component-text/src/BpkText.scss
k0nserv: This reverted back to having `bpk-text` and the styles as modifiers because we need to remove margin for all text styles

packages/bpk-component-text/src/BpkText.js
```diff
@@ -0,0 +1,62 @@
...
+const BpkText = (props) => {
+  const { bold, className, children, tagName: TagName, textStyle, ...rest } = props;
+  const classNames = [getClassName('bpk-text'), classes[props.textStyle]];
+
+  if (bold) { classNames.push(getClassName('bpk-text--bold')); }
+  if (className) { classNames.push(className); }
+
+  return (
+    <TagName className={classNames.join(' ')} {...rest}>
+      {props.children}
```
tvooo: Why not use `children` directly, as we pluck it out of `props` anyway

packages/bpk-docs/src/components/Paragraph/Paragraph.scss
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-paragraph {
```
tvooo: This should probably be `.bpk-docs-paragraph`

packages/bpk-mixins/src/mixins/_typography.scss
```diff
@@ -59,6 +59,17 @@
 ///
 /// @example scss
 ///   .selector {
+///     @include bpk-text-xs();
+///   }
+
+@mixin bpk-text-xs {
```
tvooo: That should already be part of the change I did in BPK-800

packages/bpk-component-text/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "bpk-component-text",
+  "version": "0.0.1",
```
k0nserv: Yeah 👍

packages/bpk-component-text/src/BpkText.js
k0nserv: This is left from previous iterations

packages/bpk-mixins/src/mixins/_typography.scss
k0nserv: Yeah you're right

packages/bpk-docs/src/components/Paragraph/Paragraph.scss
k0nserv: Sure

packages/bpk-component-text/src/BpkText.js
k0nserv: Interesting that `eslint` didn't catch that I was already destructing `children` without using it

PR #14, opened 2017-07-26
[WIP] [BPK-558] Upgrade theo

matthewdavidson:
Breaking changes from `theo@^4.0.0` => `theo@^5.0.0`:

- Breaking change: removed the color/hex8 transform. Instead, use color/hex8argb in Android, and color/hex8rgba in, for example, CSS level 4 values
- Breaking change: Node.js 6 and up is required
- Dropped support for Node.js v4
- Theo v5 is now compatible with Node.js v6.3 and up, dropping support for Node.js v4.
- Pointing to a non-existing alias now throws an error instead of failing silently.
- Lodash's implementation of kebabCase was dropped because it separated numbers as words:
-- Lodash: A1 -> a-1
-- Theo: A1 -> a1

(See https://github.com/salesforce-ux/theo/blob/master/CHANGELOG.md).

### DIFF
--- a/packages/bpk-tokens/tokens/base.android.xml
+++ b/packages/bpk-tokens/tokens/base.android.xml
@@ -24,11 +24,11 @@
   <property name="DURATION_SM" category="animations">200ms</property>
   <property name="DURATION_BASE" category="animations">400ms</property>
   <color name="AUTOSUGGEST_LIST_BACKGROUND_COLOR" category="autosuggest">#ffffffff</color>
-  <color name="AUTOSUGGEST_LIST_ITEM_ACTIVE_BACKGROUND_COLOR" category="autosuggest">#e6e4ebff</color>
-  <color name="AUTOSUGGEST_LIST_ITEM_HIGHLIGHTED_BACKGROUND_COLOR" category="autosuggest">#f3f2f5ff</color>
+  <color name="AUTOSUGGEST_LIST_ITEM_ACTIVE_BACKGROUND_COLOR" category="autosuggest">#ffe6e4eb</color>
+  <color name="AUTOSUGGEST_LIST_ITEM_HIGHLIGHTED_BACKGROUND_COLOR" category="autosuggest">#fff3f2f5</color>
   <property name="BADGE_PADDING_Y" category="badges">0</property>
   <property name="BADGE_PADDING_X" category="badges">6</property>
-  <color name="BADGE_BACKGROUND_COLOR" category="badges">#ffbb00ff</color>
+  <color name="BADGE_BACKGROUND_COLOR" category="badges">#ffffbb00</color>
   <property name="BADGE_CENTERED_VERTICAL_ALIGN" category="badges">text-bottom</property>
   <property name="BORDER_RADIUS_XS" category="radii">3</property>
   <property name="BORDER_RADIUS_SM" category="radii">6</property>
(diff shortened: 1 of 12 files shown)


packages/bpk-tokens/tokens/base.android.xml
```diff
@@ -24,11 +24,11 @@
   <property name="DURATION_SM" category="animations">200ms</property>
   <property name="DURATION_BASE" category="animations">400ms</property>
   <color name="AUTOSUGGEST_LIST_BACKGROUND_COLOR" category="autosuggest">#ffffffff</color>
-  <color name="AUTOSUGGEST_LIST_ITEM_ACTIVE_BACKGROUND_COLOR" category="autosuggest">#e6e4ebff</color>
-  <color name="AUTOSUGGEST_LIST_ITEM_HIGHLIGHTED_BACKGROUND_COLOR" category="autosuggest">#f3f2f5ff</color>
+  <color name="AUTOSUGGEST_LIST_ITEM_ACTIVE_BACKGROUND_COLOR" category="autosuggest">#ffe6e4eb</color>
```
k0nserv: The alpha value moving in these seem like a breaking change. We don't have anyone consuming these values yet though?

packages/bpk-tokens/tokens/base.raw.json
```diff
@@ -1,96 +1,278 @@
...
-    "BREAKPOINT_TABLET": "50.25rem",
-    "BREAKPOINT_ABOVE_TABLET": "50.3125rem",
-    "BREAKPOINT_DESKTOP": "71.25rem",
-    "ONE_PIXEL_REM": ".0625rem",
-    "ZINDEX_AUTOSUGGEST": "900",
-    "ZINDEX_POPOVER": "900",
-    "ZINDEX_TOOLTIP": "900",
-    "ZINDEX_MODAL_SCRIM": "1000",
-    "ZINDEX_MODAL": "1100"
+    "WHITE": {
```
k0nserv: Do we have anything that relies on this structure, this is also breaking if so

packages/bpk-tokens/tokens/base.android.xml
matthewdavidson: Yeah tbh this is still WIP. Do you know if that still works btw - the alpha value moving?

k0nserv: Per the [changelog](https://github.com/salesforce-ux/theo/blob/master/CHANGELOG.md#500) it seems like it was using a different transform before. Per the [Android docs](https://developer.android.com/guide/topics/resources/more-resources.html#Color) it seems that `#RRGGBBAA` was never supported and I assume in that case the parse would just not parse the alpha. `#AARRGGBB` is a supported format

packages/bpk-tokens/tokens/base.raw.json
matthewdavidson: AFAIK we have no consumers of ios, raw or android tokens yet.

k0nserv: Cool 👍

matthewdavidson: @k0nserv see my changelog entry - I have decided to make it a breaking change as folks could be relying on the `.raw.json` format - even we rely on it in the codebase.

k0nserv: Yeah sounds like a good idea. Better safe than sorry

PR #15, opened 2017-07-27
[BPK-800] Typography and spacing tokens for iOS theme

tvooo:
The new text tokens were also backported to the base theme, and I adapted the text mixins.


packages/bpk-component-modal/src/bpk-modal-dialog.scss
```diff
@@ -98,9 +98,11 @@
 
   &__dialog-heading {
     margin: 0;
-    font-weight: 700;
 
     @include bpk-text-base;
+
+    // Temporary override as part of BPK-800; to be refactored after BPK-816 is done
+    font-weight: 700; // stylelint-disable-line order/order
```
matthewdavidson: For info, the suggested refactoring is something along the lines of:

```scss
  &__dialog-heading {
    @include bpk-text-base;
    @include bpk-text-base--bold;
    // or...
    @include bpk-text--bold;
  }
```

packages/bpk-tokens/src/base/typography.json
```diff
@@ -262,6 +262,126 @@
...
+      "type": "number",
+      "category": "line-heights"
+    },
+    "TEXT_SM_FONT_WEIGHT": {
+      "value": "400",
+      "type": "number",
+      "category": "font-weights"
+    },
+    "TEXT_SM_LETTER_SPACING": {
+      "value": "normal",
```
k0nserv: `"normal"` is a value that's specific to the `letter-spacing` css attribute. How will using this value on Android/iOS/RN work?

packages/bpk-component-modal/src/bpk-modal-dialog.scss
```diff
@@ -98,9 +98,11 @@
 
   &__dialog-heading {
     margin: 0;
-    font-weight: 700;
 
     @include bpk-text-base;
```
matthewdavidson: To be clear this _is_ a breaking change for `@include bpk-text-base;` right? we'll have to provide upgrade instructions along the lines of:
```
// before
.my-style {
  font-weight: bold;
  
  @include bpk-text-base;
}

// after
.my-style {
  @include bpk-text-base;
  @include bpk-text-base--bold;
}
```

Right?

packages/bpk-component-modal/src/bpk-modal-dialog.scss
tvooo: Should I add this as part of this PR? And: `bpk-text-bold` or `bpk-text--bold`? (there's no `bpk-text` so the latter seems wrong-ish)

packages/bpk-tokens/tokens/ios.android.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="utf-8"?>
```
k0nserv: `ios.android.xml` is interesting

packages/bpk-component-modal/src/bpk-modal-dialog.scss
k0nserv: Are we doing `bpk-text-${textStyle}--bold` or just `bpk-text--bold`? Do we envision having different weights for bold per textStyle?

packages/bpk-tokens/src/base/typography.json
jamesf3rguson: Could we use '0'?

packages/bpk-tokens/tokens/ios.android.xml
tvooo: Just the name? This file is just generated, it well never actually be used...I hope :D

jamesf3rguson: I hope so too 😂

packages/bpk-component-modal/src/bpk-modal-dialog.scss
matthewdavidson: Given that we currently have each text mixin as distinct BEM blocks i.e:
```scss
@include bpk-text-sm;
@include bpk-text-base;
@include bpk-text-lg;
@include bpk-text-xl;
```
as opposed to BEM modifiers i.e:
```scss
@include bpk-text;
@include bpk-text--sm;
@include bpk-text--base;
@include bpk-text--lg;
@include bpk-text--xl;
```
I reckon @tvooo's suggestion of  `@include bpk-text-bold;` makes more sense ... :/

packages/bpk-tokens/src/base/typography.json
k0nserv: Nope

> The spacing is the normal spacing for the current font. Unlike a value of 0, this value allows the user agent to alter the space between characters in order to justify text.

packages/bpk-tokens/tokens/ios.android.xml
k0nserv: Haha yup it was just amusing

packages/bpk-tokens/src/base/typography.json
tvooo: Hmm. Can we find out what the equivalent for Android/iOS would be and create a formatter for it?

packages/bpk-component-modal/src/bpk-modal-dialog.scss
k0nserv: Feel like this is a question for @jamesf3rguson ultimately.

packages/bpk-component-modal/src/bpk-modal-dialog.scss
k0nserv: Yeah I'd agree. Good catch

tvooo: @jamesf3rguson should the `bpk-text-bold` mixin always set the font-weight to "700"? What should it do in the iOS theme? (Option: not set anything at all)

packages/bpk-tokens/src/base/typography.json
k0nserv: Sure, but I think the source of truth should be platform agnostic. Same concern I have about all our spacing values being `rem`.

k0nserv: [iOS](https://developer.apple.com/documentation/uikit/nskernattributename?language=objc) for reference

tvooo: So what's a platform-agnostic source of truth for something that is ultimately handled very differently across platforms? I'm not sure, for example, if setting `letter-spacing: 1px` on web is equivalent to setting kerning to `1pt` on iOS.

packages/bpk-component-modal/src/bpk-modal-dialog.scss
jamesf3rguson: Yep always 700 for Web, iOS I wouldn't set anything

jamesf3rguson: As in iOS they will have their own weights

tvooo: Ok, how should I solve this from a technical aspekt @matthewdavidson?

I'll have to create a token which is the font weight that this mixin sets. In the base theme it will be 700. In the iOS theme, can I make it "false" and check in the mixin if it's a number? I dislike this approach again because it's so specific to the mixin where it's being used.

matthewdavidson: If a consumer uses `@include bpk-text-bold;` in an ios webview then thats a mistake right? I.e. the designer shouldn't have given them that design?

Otherwise the value for `font-weight` in a `bpk-text-bold` mixin should be the `FONT_WEIGHT_BOLD` token?

packages/bpk-tokens/src/base/typography.json
k0nserv: Yeah it's definitely confusing and the more I think of it it feels like we're trying to shoehorn two distinct dimensions into one. What we really have is theme and platform which are distinct dimensions and maybe having a single value for every token in a theme isn't enough. 

Using spacing as an example

| Platform                  | base         | iOS  |
| ------------------------- | ------------ | ---- |
| Web(iOS/Android webviews) | .375rem(6px) | 4px  |
| iOS(Native and RN)        | -            | 4pt  |
| Android(Native and RN)    | -            | -    |



Is platform specific formatters enough to transform a platform agnostic value into a platform specific one? In the above example `SPACING_XS` would be a `4pt` for `iOS` and we'd use a `1px = 1pt` transformation to convert to pixels for web and then possibly another calculation to get a rem value which works. But like you mention to we need platform specific tokens for values such as kerning/letter spacing because the semantics of the different platforms are different enough to make it impossible to transform an agnostic value into a specific one?

packages/bpk-component-modal/src/bpk-modal-dialog.scss
tvooo: Adressed this

tvooo:
I addressed Matt's comments, we still don't really have a solution for the letter-spacing/kerning tokens.

tvooo:
We decided to go with this solution for now and add different tokens for platform-specific features (kerning?) when appropriate.

PR #18, opened 2017-07-28
Bpk 817 deprecate headings and paragraphs

k0nserv:
This removes the usage of `BpkHeading` and `BpkParagraph` in the docs and as part of stories.


packages/bpk-docs/src/components/Heading/Heading.scss
```diff
@@ -0,0 +1,36 @@
...
+@import '~bpk-mixins/index';
+
+.bpk-docs-heading {
+  margin-top: 0;
+  margin-bottom: $bpk-spacing-sm;
+
+  &--h1,
+  &--h2,
+  &--h3 {
+    font-weight: 400;
```
k0nserv: Good point

k0nserv:
This also regress mobile heading responsiveness. Do we wanna use `bpk-component-breakpoint` to achieve that or just a CSS media query?

matthewdavidson:
> This also regress mobile heading responsiveness. Do we wanna use bpk-component-breakpoint to achieve that or just a CSS media query?

I like the declarative nature of using `BpkBreakpoint`, but behind the scenes it uses the `window.matchMedia` api which I suspect isn't as performant as plain ole' media queries. 

@tvooo, thoughts?

tvooo:
Agree with both points. Like using BpkBreakpoint, but it's probably not as performant. Then again, how often does the window size change?

tvooo:
Happy with the changes. I can imagine that the duplicate mixins will confuse people, not sure if we can/should do something about that?

matthewdavidson:
> Happy with the changes. I can imagine that the duplicate mixins will confuse people, not sure if we can/should do something about that?

Yeah Im not particularly happy with them either - i guess you understand why i did it though?

tvooo:
Yep sure. Otherwise I'd have "requested changes" :trollface: Go merge!

PR #20, opened 2017-07-31
Enable BpkTableCell custom styling

robinknox:
Ability to apply a CSS class to a table cell BpkTableCell component through the className property

### DIFF
--- a/packages/bpk-component-table/src/BpkTableCell-test.js
+++ b/packages/bpk-component-table/src/BpkTableCell-test.js
@@ -25,4 +25,9 @@ describe('BpkTableCell', () => {
     const tree = renderer.create(<BpkTableCell>Tabular data</BpkTableCell>).toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  it('should render correctly with custom class', () => {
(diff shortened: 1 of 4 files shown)


packages/bpk-component-table/src/BpkTableCell-test.js
```diff
@@ -22,7 +22,7 @@ import BpkTableCell from './BpkTableCell';
 
 describe('BpkTableCell', () => {
   it('should render correctly', () => {
-    const tree = renderer.create(<BpkTableCell>Tabular data</BpkTableCell>).toJSON();
+    const tree = renderer.create(<BpkTableCell className="test">Tabular data</BpkTableCell>).toJSON();
```
tvooo: Can you add an additional test case (`it(...)`) for this configuration? See for example https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-table/src/BpkTable-test.js#L33

robinknox: Hi, that should be the additional test added. I've also added the changelog entry as well.

tvooo:
Looks good, will merge tomorrow morning!

PR #21, opened 2017-08-01
[BPK-797] Unpin react-responsive version

georgegillams:


matthewdavidson:
@georgegillams can you double check the behaviour manually?

PR #22, opened 2017-08-02
[BPK-817] Opt for css media queries over BpkBreakpoint

matthewdavidson:
BpkBreakpoint uses `window.matchMedia` based mechanism to render "media queries" declaratively. We use SSR on the docs site an thus we need to media queries to work on first paint.

### DIFF
--- a/packages/bpk-docs/src/components/Heading/Heading.js
+++ b/packages/bpk-docs/src/components/Heading/Heading.js
@@ -18,158 +18,39 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import BpkText from 'bpk-component-text';
-import BpkBreakpoint, { BREAKPOINTS } from 'bpk-component-breakpoint';
-
-import { cssModules, withDefaultProps } from 'bpk-react-utils';
+import { cssModules } from 'bpk-react-utils';
 
 import STYLES from './Heading.scss';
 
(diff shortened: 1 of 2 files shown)


packages/bpk-docs/src/components/Heading/Heading.js
```diff
@@ -18,158 +18,39 @@
...
+  h6: 'h6',
 };
 
 const Heading = (props) => {
-  const { level, ...rest } = props;
+  const { level: TagName, className, ...rest } = props;
 
-  const Component = HEADINGS[level];
+  const classNames = [getClassName('bpk-docs-heading')];
+  classNames.push(getClassName(`bpk-docs-heading--${TagName}`));
```
tvooo: This could be part of the initialization already

packages/bpk-docs/src/components/Heading/Heading.js
```diff
@@ -18,158 +18,39 @@
...
-      BpkText,
-      {
-        textStyle: 'sm',
-        tagName: 'h6',
-        bold: false,
-        className,
-      },
-    ),
-  },
+const HEADINGS_LEVELS = {
```
tvooo: This can now be an array of strings. Also avoids the `Object.keys` awkwardness further down

PR #23, opened 2017-08-02
[BPK-799] Fix popover behaviour when mouse-down inside, drag out…

georgegillams:
…side, and mouse-up


packages/bpk-react-utils/src/Portal.js
```diff
@@ -30,10 +30,13 @@ class Portal extends Component {
     super();
 
     this.portalElement = null;
+    this.shouldClose = false;
```
tvooo: Maybe add a one-line comment here to explain what this is used for

PR #25, opened 2017-08-03
[BPK-826] Increase the tappable area around barchart bars

tvooo:


changelog.md
```diff
@@ -2,7 +2,9 @@
 
 ## UNRELEASED
 
-_Nothing yet..._
+**Fixed:**
```
tvooo: Not a 100% sure how to classify this. It doesn't change anything from a cosmetical perspective, however, if someone would have used the barchart with a different component you could argue that this _might_ break because `paddingInner` is not set to 0 on the xScale.

packages/bpk-component-barchart/src/BpkBarchartBar.js
```diff
@@ -51,35 +51,53 @@ const BpkBarchartBar = (props) => {
     onClick,
     outlier,
     selected,
+    padding,
```
k0nserv: I was looking for an explination of what this value is but I couldn't find one because the readme.md isn't updated

changelog.md
matthewdavidson: Let's favor strict semver 👍

tvooo: ...which means...?

packages/bpk-component-barchart/src/BpkBarchartBar.js
tvooo: "updated" you're funny... this component isn't even documented ;)

changelog.md
matthewdavidson: It means that if this change might break consumers build / design then it's a major bump.

However that being said .. if consumers are using `innerPadding` and it has a similar visual effect then it's not a breaking change.

packages/bpk-component-barchart/src/BpkBarchartBar.js
k0nserv: Ah okay so this is internal only?

tvooo: Nope. Well, kind of. It doesn't really make sense to use it outside of BpkBarchart, but who knows? We export it, so we should probably document it. What do you think @matthewdavidson ? Also, with all the other components in the barchart...

matthewdavidson: `BpkBarchartBar` isn't exported, but `BpkBarchartBars` is (see https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-barchart/index.js#L20).

I think it's too early to make a decision with all the other components. This will become clearer once a lien chart gets added into the mix.

tvooo: Fair enough. So as the API between bars and bar is private, I guess it's ok to postpone documenting it for now?

tvooo:
Updated the changelog. @k0nserv could you approve the changes?

PR #26, opened 2017-08-03
[BPK-796] Enable styling tooltip portal

k0nserv:


k0nserv:
I think we should also change `className` to `portalClassName` while we're at it. This would be more inline with popover

k0nserv:
Rename pushed in f4ee712

packages/bpk-component-tooltip/src/BpkTooltipPortal-test.js
```diff
@@ -52,7 +52,7 @@ describe('BpkTooltipPortal', () => {
       <BpkTooltipPortal
         id="my-tooltip"
         target={<div>target</div>}
-        className="my-custom-class"
+        portalClassName="my-custom-class"
```
matthewdavidson: There should be two tests for this - one for custom `className` and one for `portalClassName`.

This current one is a mixture of the two (from a test name and component prop point of view)

k0nserv: Good point 👍

matthewdavidson: looks good :shipit:

k0nserv:
Hmmm
image

PR #27, opened 2017-08-03
# [NO-JIRA] Adding three icons for leg room

jamesf3rguson:
Added legroom-normal, legroom-extra, legroom-reduced


k0nserv:
Do we not want a changelog entrry for this?

jamesf3rguson:
Oops, my bad. Yes I’ll sort that now.

PR #28, opened 2017-08-05
[BPK-827] Fix logic in NavSelect when refreshing

georgegillams:


packages/bpk-docs/src/layouts/SideNavLayout/NavSelect.js
```diff
@@ -57,22 +57,26 @@ NavSelectCategory.propTypes = {
...
-      {props.links.map(
-        link => (link.category
-          ? <NavSelectCategory key={link.id} link={link} />
-          : <NavSelectItem key={link.id} link={link} />
-      ))}
-    </BpkSelect>
-  </BpkFieldset>
-);
+const NavSelect = (props) => {
+  const pathname = props.location.pathname.replace(/\/$/, '');
```
k0nserv: Could we add a comment to explain why this is necessary please?

PR #32, opened 2017-08-07
[BPK-799] Fix popover behaviour when dragging for mobile

georgegillams:
+ [x] If I've updated `npm-shrinkwrap.json` those chagnes are intentional.


packages/bpk-react-utils/src/Portal.js
```diff
@@ -107,6 +108,10 @@ class Portal extends Component {
     }
   }
 
+  onMouseMove() {
```
tvooo: For consistency's sake this should be called `onDocumentMouseMove`

PR #36, opened 2017-08-08
[BPK-784] Align postcss config across webpack configs

matthewdavidson:
Configs exist for docs site, storybook and base stylesheet.
Removed the `*.css` loader in the storybook config as this appears unused.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,13 +1,12 @@
 const fs = require('fs');
 const path = require('path');
-const autoprefixer = require('autoprefixer');
-const sassFunctions = require('../packages/bpk-mixins/sass-functions');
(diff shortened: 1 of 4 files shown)


.storybook/webpack.config.js
```diff
@@ -53,39 +41,6 @@ module.exports = {
           },
         ],
       },
-      {
```
k0nserv: Why is this removed? It doesn't seem related to the required change

matthewdavidson: I put this in the description:
> Removed the *.css loader in the storybook config as this appears unused.

Happy to put back, but it does seem to be redundant.

k0nserv: Ah right yeah no keep it in then

PR #37, opened 2017-08-08
[BPK-837] Rename 'Notifications' to 'Banner alerts' in docs

tvooo:

### DIFF
--- a/packages/bpk-docs/src/constants/redirect-routes.js
+++ b/packages/bpk-docs/src/constants/redirect-routes.js
@@ -16,5 +16,9 @@
  * limitations under the License.
  */
 
-// Deprecated routes, kept here for forwarding
-export const NOTIFICATIONS = '/components/molecules/notifications'; // eslint-disable-line
+import * as ROUTES from './routes';
+
+// Deprecated routes, kept here for redirects
+export default {
+  '/components/molecules/notifications': ROUTES.BANNER_ALERTS,
+};
--- a/packages/bpk-docs/src/constants/redirect-routes.js
+++ b/packages/bpk-docs/src/constants/redirect-routes.js
@@ -16,6 +16,5 @@
  * limitations under the License.
  */
 
-import page from './NotificationsPage';
-
-export default page;
+// Deprecated routes, kept here for forwarding
+export const NOTIFICATIONS = '/components/molecules/notifications'; // eslint-disable-line
--- a/packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
+++ b/packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
@@ -34,25 +34,15 @@ const richMessage = <span style={{ fontWeight: 700 }}>Successful alert with cust
 
 const components = [
   {
-    id: 'banner-alerts',
-    title: 'Banner alerts',
+    id: 'default',
+    title: 'Default',
     blurb: [
       <Paragraph>
-        These are displayed to the user to provide feedback when an action has been performed. They are available in
-        three styles to indicate success, warning or error and can be configured to display further information to
-        user in the form of a collapsible panel.
+        Banner alerts come in three styles to indicate success, warning or error.
       </Paragraph>,
     ],
     examples: [
       <BpkBannerAlert message="Successful alert." type={ALERT_TYPES.SUCCESS} />,
-      <br />,
-      <BpkBannerAlert
-        message="Successful alert with more information."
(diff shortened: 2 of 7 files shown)


packages/bpk-docs/src/constants/redirect-routes.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Deprecated routes, kept here for forwarding
+export const NOTIFICATIONS = '/components/molecules/notifications'; // eslint-disable-line
```
matthewdavidson: Prefer explicit rule i.e. `// eslint-disable-line import/prefer-default-export`

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -75,26 +84,19 @@ const components = [
...
+  title="Banner alerts"
   blurb={[
     <Paragraph>
-      Backpack supports a number of different types of notification styles, which can be used in different scenarios
-      to provide messaging to the user.
+      Banner alerts provide feedback to the user when an action has been performed.
     </Paragraph>,
   ]}
   components={components}
   sassdocId="notifications"
```
matthewdavidson: Stills say's `notifications` - is this ok? I realise that we'd have to change token level groups / categories. Perhaps that ship has sailed.

webpack.config.babel.js
```diff
@@ -34,7 +35,10 @@ const WrapperPlugin = require('wrapper-webpack-plugin');
 
 
 const staticSiteGeneratorConfig = {
-  paths: Object.keys(ROUTES).map(key => ROUTES[key]),
+  paths: [
+    ...Object.keys(ROUTES).map(key => ROUTES[key]),
```
matthewdavidson: Is this ordering conscious? What happens when someone tries to add a route that already exists in `REDIRECT_ROUTES` .. they'll get confused because it redirects to somewhere else. Perhaps this is a good thing - i.e. we wont break existing redirects.

packages/bpk-docs/src/constants/redirect-routes.js
tvooo: I did that, and then the line was too long, so I got tired 😂 I can disable both rules though (prefer-default-export and max-line-length)

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
tvooo: Yes, that's why I left it. Wouldn't go and change all the tokens categories now, I don't see much of a value in there.

packages/bpk-docs/src/constants/redirect-routes.js
matthewdavidson: 😂

matthewdavidson: You can put it above the line you know ... `/* eslint-disable-next-line import/prefer-default-export */`

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
matthewdavidson: coolio

webpack.config.babel.js
tvooo: Why would the order here matter? This only collects the paths which are covered by SSR. If a path exists twice, that shouldn't make a difference. If this even is relevant needs to be shown once I've solved the SSR problem...

packages/bpk-docs/src/constants/redirect-routes.js
tvooo: Oooooh I didn't know that! Awesome! :)

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
jamesf3rguson: Looks good to me

k0nserv:
Can't we do something like returning 

```html
<html>
<head>
  <meta http-equiv="refresh" content="0; url=/components/molecules/banner-alerts"/>
</html>
```

for `/components/molecules/notifications`

PR #40, opened 2017-08-10
[No Jira] Allow custom className and other arbitrary props for bpk-component-banner-alert

badsyntax:


k0nserv:
@badsyntax I think you forgot to update some snapshots for the arbitrary props test

badsyntax:
@k0nserv woops, i have pushed those changes.

PR #42, opened 2017-08-11
[BPK-860] Remove unused output formats

k0nserv:


k0nserv:
cc @matthewdavidson This is the general approach I figure we'd take. Haven't removed `raw.json` yet working on that still

k0nserv:
Hmm the `raw.json` format actually conveys more information than `es6.js`. We could change the `es6.js` format, but that'd be a breaking change. Maybe we should keep the the raw format anyway? 

bpk-svgs relies on category and type as seen here https://github.com/Skyscanner/backpack/blob/master/packages/bpk-tokens/tokens/base.raw.json#L1304-L1311

PR #52, opened 2017-08-17
[BPK-836] Used medium weights for xl and base

k0nserv:
In line with material design.


packages/bpk-tokens/src/android/base/typography.json
```diff
@@ -179,7 +179,7 @@
       "category": "font-sizes"
     },
     "TEXT_XL_FONT_WEIGHT": {
-      "value": "{!FONT_WEIGHT_REGULAR}",
+      "value": "{!FONT_WEIGHT_MEDIUM}",
```
georgegillams: The PR title talks about `lg` and `base`, but you seem to have changed `xl` and `base`??

georgegillams: Also, in iOS the font weight for `base` and `lg` are still `NORMAL`, but I assume this difference between platforms is deliberate?

k0nserv: Sorry yeah should be xl and base

PR #53, opened 2017-08-17
pass correct arguments to onClick in BpkButton

avivian:
Came across this while trying to stop event propagation. The `...arguments` parameter is actually that of the `BpkButton` class rather the function, so the onClick handler being passed was receiving garbage. This fixes it. Not sure the best method of testing it, I wasn't able to import a SyntheticEvent from react to assert the type. Open to suggestions.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,9 @@
 
 ## UNRELEASED
 
-__Nothing yet...__
+**Fixed:**
+- bpk-component-button:
+  - Fixed SyntheticEvent not being passed to the onClick handler.
 
 ## 2017-08-17 (2) - Fix a missing dependency in several components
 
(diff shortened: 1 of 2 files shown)


changelog.md
```diff
@@ -16,6 +16,8 @@
     **NOTE:** This may break your snapshot tests and thus your build. This is nothing to worry about and the
     snapshots can safely be updated. Just make sure to perform the update in isolation so that all the
     snapshot updates relate to this issue and don't introduce unrelated regressions.
+- bpk-component-button:
```
k0nserv: Could you rebase this as these changes have been released

avivian:
@k0nserv rebase done

k0nserv:
LGTM

PR #56, opened 2017-08-17
[BPK-846] Add Danger to workflow

shaundon:

Couple of things I'd like to mention:
- I didn't add the licence header to `dangerfile.js` because I looked at other config type files and they don't have it, e.g. `gulpfile.babel.js`, so I assumed that's the convention.
- I also didn't see any non `package/*` changes highlighted in the changelog, so I didn't update it. I can change that if we think it's necessary.
- There was some work on the Github side to create a bot account, this is documented in the JIRA ticket.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ script:
   - npm run docs:dist:gzip
   - test -e dist/index.html
   - test -e dist/sassdoc/index.html
+  - npm run danger
 cache:
   directories:
   - node_modules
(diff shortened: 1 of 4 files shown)


shaundon:
Works locally, needs hooking up to Github now.

shaundon:
@georgegillams I'm not sure - I looked at other config type files like the gulpfile and it wasn't on there, so I decided to follow that convention. Having said that, it can't hurt to have it, so I'll add it now.

shaundon:
It's giving me so many ideas for other checks to add! They [recommend building up gradually](http://danger.systems/js/usage/culture.html), which sounds like a good way of doing it too.

PR #62, opened 2017-08-24
[NOJIRA] Fix table components to accept arbitrary props and custom cl…

matthewdavidson:
…ass names

### DIFF
--- a/packages/bpk-component-table/src/BpkTableHead.js
+++ b/packages/bpk-component-table/src/BpkTableHead.js
@@ -19,7 +19,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const BpkTableHead = props => <thead>{props.children}</thead>;
+const BpkTableHead = props => <thead {...props} />;
 
 BpkTableHead.propTypes = {
   children: PropTypes.node.isRequired,
--- a/changelog.md
+++ b/changelog.md
@@ -16,6 +16,9 @@
 - bpk-component-button:
   - Fixed SyntheticEvent not being passed to the onClick handler.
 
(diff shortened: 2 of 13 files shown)


packages/bpk-component-table/src/BpkTableHead.js
```diff
@@ -19,7 +19,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const BpkTableHead = props => <thead>{props.children}</thead>;
+const BpkTableHead = props => <thead {...props} />;
```
k0nserv: Does passing `children` like this have any caveats or does it just compile to identical code?

k0nserv: Just checked

```javascript
const test1 = <b>Hello</b>
const test2 = <b children={["hello"]} />
```
Compiles to

```javascript
"use strict";

var test1 = React.createElement(
  "b",
  null,
  "Hello"
);
var test2 = React.createElement("b", { children: ["hello"] });
```

k0nserv: BUt that probably just takes the same code paths anyway

matthewdavidson: We use this pattern elsewhere and _it works_... not sure of the implications.

k0nserv: https://github.com/facebook/react/blob/master/src/isomorphic/classic/element/ReactElement.js#L214-L230 it gets added to props when passed in third argument

georgegillams:
...except for the "failed build" thing...

PR #70, opened 2017-08-28
[BPK-231] Add docs for srcset use in Image component

georgegillams:


packages/bpk-component-image/src/__snapshots__/BpkImage-test.js.snap
```diff
@@ -545,3 +545,107 @@ exports[`BpkImage should render correctly 1`] = `
   </div>
 </div>
 `;
+
+exports[`BpkImage should support srcSet 1`] = `
```
matthewdavidson: this snapshot doesn't assert against any sort of `srcset` support

packages/bpk-component-image/stories.js
```diff
@@ -39,14 +39,35 @@ storiesOf('bpk-component-image', module)
...
+      height={408}
+      src={image}
+    />
+  ))
+  .add('Using SrcSet', () => (
+    <BpkImage
+      altText="image"
+      width={612}
+      height={408}
+      srcSet={`${image}?resize=280px:280px&quality=100 280w,
```
matthewdavidson: AFAIK you still need to supply a `src` attribute when using `srcset`...

packages/bpk-docs/src/pages/ImagesPage/ImagesPage.js
```diff
@@ -48,7 +50,24 @@ const components = [
         altText="Lake"
         width={816}
         height={544}
-        src={`/${image1}`}
+        src={image1}
+        srcSet={`${image1}?resize=280px:280px&quality=100 280w,
+          ${image1}?resize=320px:320px&quality=100 320w,
```
matthewdavidson: All these examples contain quite a lot of different options - two or three should be adequate, no?

packages/bpk-component-image/stories.js
georgegillams: FYI it works without, but probably best practice to!

matthewdavidson: Yup, as a fallback for older browsers 👍

packages/bpk-component-image/src/__snapshots__/BpkImage-test.js.snap
georgegillams: Ah - `inView` was set to false!

packages/bpk-docs/src/pages/ImagesPage/ImagesPage.js
georgegillams: Fair point - have reduced to 4

packages/bpk-docs/src/pages/ImagesPage/ImagesPage.js
```diff
@@ -48,7 +50,18 @@ const components = [
...
+        src={image1}
+        srcSet={`${image1}?resize=320px:320px&quality=100 320w,
+          ${image1}?resize=520px:520px&quality=100 520w,
+          ${image1}?resize=1024px:1024px&quality=100 1024w,
+          ${image1}?resize=1920px:1920px&quality=100 1920w`}
+        // If the viewport is wider than 1140px, then this image will
+        // take up ≈765px of the viewport. If the viewport
+        // is wider than 805px it will take up ≈ all but 292px
+        // Otherwise, it will take up ≈ all but 72px.
+        sizes={`(min-width: 1140px) 765px,
```
matthewdavidson: use tokens for these breakpoint values

PR #71, opened 2017-08-28
Bpk 886 calendar focus tracking

georgegillams:


packages/bpk-component-calendar/src/Week.js
```diff
@@ -87,6 +102,30 @@ class Week extends Component {
...
+  onMouseClick(date) {
+    this.props.onDateClick(date);
+    this.setState({ savedDate: null });
+  }
+
+  onMouseDown(date, month) {
+    if (isSameMonth(date, month)) {
+      this.props.onDateMouseDown(null, { date });
+    } else {
+      this.setState({ savedDate: date });
```
k0nserv: I'm trying to understand why this should be different. Can you expand on the reason?

packages/bpk-component-calendar/src/BpkCalendarGrid-test.js
```diff
@@ -129,4 +129,29 @@ describe('BpkCalendarGrid', () => {
...
+      formatMonth={formatMonth}
+      formatDateFull={formatDateFull}
+      DateComponent={BpkCalendarDate}
+      daysOfWeek={weekDays}
+      weekStartsOn={0}
+      showWeekendSeparator
+      onDateMouseDown={onDateMouseDown}
+    />);
+
+    expect(onDateMouseDown.mock.calls.length).toBe(0);
```
k0nserv: This is redundant really

georgegillams: You mean checking that the `length` is `0`? or the whole test?

packages/bpk-component-calendar/src/Week.js
```diff
@@ -87,6 +102,30 @@ class Week extends Component {
...
+    }
+  }
+
+  onMouseUp(event) {
+    if (this.state.savedDate != null) {
+      this.props.onDateMouseDown(event, { date: this.state.savedDate });
```
matthewdavidson: Calling a "mouse down" event handler on a mouse up event is surprising.

georgegillams: I agree - the 'mouseDown` and 'onClick' events should really be renamed if we are going with this solution, but that would presumably break a lot of stuff!

matthewdavidson: Yup, but I mean more fundamentally... they are not the same right? It's surprising that to bind a callback to a `mousedown` event for it to be called on `mouseup`.

k0nserv: That's a good point, maybe they sould have abstract names in week and above?

georgegillams: Something like `onDateFocused` and `onDateSelected`???

packages/bpk-component-calendar/src/BpkCalendarGrid-test.js
k0nserv: Checking that the length is zero. That's guaranteed by the mock implementation

packages/bpk-component-calendar/src/Week.js
k0nserv: Yeah something to that effect

PR #72, opened 2017-08-29
BPK-870 react native docs

matthewdavidson:


native/react-native-bpk-component-text/index.js
```diff
@@ -0,0 +1,3 @@
+import BpkText from './src/BpkText';
```
georgegillams: Shouldn't all new files have the license header?

packages/bpk-docs/src/components/Header/Header.js
```diff
@@ -35,7 +35,8 @@ const getClassName = cssModules(STYLES);
 const headerLinks = [
   { to: ROUTES.USING_BACKPACK, children: 'Using Backpack' },
   { to: ROUTES.STYLE, children: 'Style' },
-  { to: ROUTES.DOCS, children: 'Components' },
+  { to: ROUTES.COMPONENTS, children: 'Components' },
+  // { to: ROUTES.NATIVE, children: 'Native' },
```
georgegillams: Is this needed or not?

native/react-version-test.js
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+it('native/* should be ^16.0.0', () => {
```
k0nserv: Should we keep these?

packages/react-version-test.js
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+it('packages/* should be ^15.0.0', () => {
+  expect(React.version).toMatch(/^15/);
```
k0nserv: Ditto

native/react-version-test.js
matthewdavidson: So I pondered this too ... whilst the shrinkwrap determines what version we use, it does not guarantee that node resolves the right version. I kinda like them. They are cheap and will act as a failsafe.

packages/bpk-docs/src/components/Header/Header.js
matthewdavidson: Yeah this is the "hide" part.

native/react-native-bpk-component-text/index.js
matthewdavidson: Good catch, fixed

native/react-version-test.js
k0nserv: That shounds good to me :)

matthewdavidson:
@georgegillams you happy?

PR #79, opened 2017-09-04
[BPK-908] Fix lazy loading HOC not responding to div scroll events

georgegillams:

### DIFF
--- a/packages/bpk-component-image/src/withLazyLoading.js
+++ b/packages/bpk-component-image/src/withLazyLoading.js
@@ -36,9 +36,10 @@ export default function withLazyLoading(Component, document) {
     }
 
     componentDidMount() {
-      const passiveArgs = this.supportsPassiveEvents() ? { passive: true } : false;
-      document.addEventListener('scroll', this.checkInView, passiveArgs);
+      const passiveArgs = this.supportsPassiveEvents() ? { passive: true } : {};
+      document.addEventListener('scroll', this.checkInView, { capture: true, ...passiveArgs });
       document.addEventListener('resize', this.checkInView);
+      document.addEventListener('orientationchange', this.checkInView);
       document.addEventListener('fullscreenchange', this.checkInView);
       // call checkInView immediately incase the
       // component is already in view prior to scrolling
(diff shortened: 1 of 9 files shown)


packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -57,8 +58,10 @@ export default function withLazyLoading(Component, document) {
     }
 
     removeEventListeners() {
-      document.removeEventListener('scroll', this.checkInView);
+      const passiveArgs = this.supportsPassiveEvents() ? { passive: true } : {};
+      document.removeEventListener('scroll', this.checkInView, { capture: true, ...passiveArgs });
```
matthewdavidson: When removing an event listener, do you have to provide the options? What happens if you omit them?

packages/bpk-mixins/src/mixins/_images.scss
```diff
@@ -33,8 +33,14 @@
   position: relative;
   display: block;
   max-width: 100%;
+  transition: background-color $bpk-duration-base ease-in-out;
   background-color: $bpk-color-gray-100;
   overflow: hidden;
+  transition-delay: $bpk-duration-base;
+}
+
+@mixin bpk-image--show {
```
matthewdavidson: This doesn't need to be a mixin - just have it inline. In fact the mixins for the image component are realistically never going to be used by consumers. We should remove them (perhaps outside the scope of this PR).

georgegillams: Done here https://github.com/Skyscanner/backpack/pull/82

packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -36,9 +38,11 @@ export default function withLazyLoading(Component, document) {
     }
 
     componentDidMount() {
-      const passiveArgs = this.supportsPassiveEvents() ? { passive: true } : false;
-      document.addEventListener('scroll', this.checkInView, passiveArgs);
+      const passiveArgs = this.supportsPassiveEvents() ? { passive: true } : {};
+      this.scrollEventArgs = { capture: true, ...passiveArgs };
```
matthewdavidson: Hmm we should have a test for this behaviour.

I.e. a test that asserts that `this.scrollEventArgs` is the same after `componentDidMount ` and during `componentWillUnmount`.

georgegillams: I feel like this is just testing javascript?? Maybe worthwhile in-case someone changes the args?

georgegillams: Actually, come to think of it, we don't want to test that they are the same - as this is internal workings of the HOC. What we want to test is that the event listeners are, indeed, added and removed correctly

georgegillams: IMHO

PR #83, opened 2017-09-06
[BPK-869] Button component for iOS

shaundon:


native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -0,0 +1,86 @@
...
+  android: () => ANDROID_TOKENS,
+})();
+
+const styles = {
+  base: {
+    underlayColor: 'rgba(0, 0, 0, 0.15)',
+    text: {
+      backgroundColor: 'transparent',
+      color: tokens.colorWhite,
+      fontSize: tokens.textSmFontSize,
```
k0nserv: Create a layer of indirection for tokens don't directly reference the bonds. This should be `tokens.buttonTextFontSize`and so on for all other tokens in this component

packages/bpk-tokens/tokens/ios/base.react.native.common.js
```diff
@@ -1,21 +1,21 @@
 /*
- *
```
matthewdavidson: Looks like you've saved this auto-generated file - needs reverting.

native/packages/react-native-bpk-component-button/readme.md
```diff
@@ -0,0 +1,34 @@
...
+```js
+import React, { Component } from 'react';
+import { View } from 'react-native';
+import BpkButton from 'react-native-bpk-component-button';
+
+export default class App extends Component {
+  render() {
+    return (
+      <View>
+        <BpkButton title="Book flight" onPress={() => {}} />
```
matthewdavidson: I'm keen for these readme code examples to match the screenshot we have in the docs as much as possible. I.e. if you take the text component for example - it has a container with some padding etc (see https://github.com/Skyscanner/backpack/tree/master/native/packages/react-native-bpk-component-text). I guess this can be refined when you approach the docs for this component.

k0nserv:
For this change we probably can't get away without shipping native libraries to go along. We need to explore creating a cocoapod podspec and publishing that as well as publishing for the android eco system too

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -0,0 +1,120 @@
...
+       <TouchableHighlight
+         style={getButtonStyle(props)}
+         disabled={disabled}
+         selected={selected}
+         onPress={onPress}
+         underlayColor={styles.base.underlayColor}
+         {...rest}
+       >
+         <View>
+           <Text style={getTextStyle(props)}>{title}</Text>
```
matthewdavidson: I'm not sure about this api, `<BpkButton title="Search" />`, it's too limiting.

I'd have expected something like:
```
<BpkButton>
  Search <BpkIcon />
</BpkButton>
```

I'm aware that in RN all text must be wrapped in a `Text` component, but we should be able to do that i.e. `<BpkText>{children}</BpkText>`. `Text` supports nesting (see https://facebook.github.io/react-native/docs/text.html).

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -0,0 +1,86 @@
...
+
+const IOS_TOKENS = require('bpk-tokens/tokens/ios/base.react.native.common.js');
+const ANDROID_TOKENS = require('bpk-tokens/tokens/android/base.react.native.common.js');
+
+const tokens = Platform.select({
+  ios: () => IOS_TOKENS,
+  android: () => ANDROID_TOKENS,
+})();
+
+const styles = {
```
matthewdavidson: Use `StyleSheet.create` here.

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -0,0 +1,120 @@
...
+ } from 'react-native';
+ import React from 'react';
+ import PropTypes from 'prop-types';
+ import LinearGradient from 'react-native-linear-gradient';
+
+ import styles from './BpkButton-styles';
+
+ const getTextStyle = ({ type, large, disabled }) => {
+   const textStyle = [styles.base.text];
+   if (styles[type].text) {
```
matthewdavidson: Will throw is someone passes an unknown type.

k0nserv: That's fine though? It's enforced via propTypes seems expected that it would trhow

native/packages/react-native-bpk-component-button/src/BpkButton.js
k0nserv: Surely the path of least surprise would be 

```
<BpkButton>
  <BpkText>Search <BpkIcon /></BpkText>
</BpkButton>
```
?

shaundon: That's fine - I was in two minds about which way to do it, and went with the existing approach because that's how the RN button component did it. I can change that shortly.

native/packages/react-native-bpk-component-button/src/BpkButton.js
shaundon: I haven't put it in yet, but I was thinking it could use `PropTypes.oneOf` to enforce it

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
shaundon: Will do - it's on my todo list

k0nserv:
> For this change we probably can't get away without shipping native libraries to go along. We need to explore creating a cocoapod podspec and publishing that as well as publishing for the android eco system too

Actually regarding this it looks like the method of choice in the React Native community is not to publish for the other platforms, but instead use a path reference.

```
pod 'BpkButton', :path => 'node_modules/react-native-bpk-component-button'
````

At the root of the package we should include a `BpkButton.podspec`. I'm not sure for Android, but I believe we need to create a similar gradle

File: native/packages/react-native-bpk-component-button/src/BpkButton.js (hunk 4 above)
Comment (matthewdavidson) [reply native/packages/react-native-bpk-component-button/src/BpkButton.js]: @k0nserv yup perhaps, but we need control over the text styling too.

@shaundon Ah, it's interesting that the RN component follows that api - perhaps it may be worth conforming to that (for reference https://facebook.github.io/react-native/docs/button.html).

File: native/packages/react-native-bpk-component-button/src/BpkButton.js (hunk 6 above)
Comment (matthewdavidson) [reply native/packages/react-native-bpk-component-button/src/BpkButton.js]: PropTypes only throw a warning - it's a bad developer experience if someone adds a button component to their view and it throws because they have a typo e.g. 
```
<BpkButton type="primay" />
```
The above would throw a big red screen with a strange error message about accessing an unknown member and any warnings would be hidden.

Comment (k0nserv) [reply native/packages/react-native-bpk-component-button/src/BpkButton.js]: I think that's great, fail early and fail hard after all. The other option is silently accepting errors and making it more likely that the developer misses the issue

Comment (matthewdavidson) [reply native/packages/react-native-bpk-component-button/src/BpkButton.js]: Or even better.. throw with a nice error message ;)

i.e. 
```
if (TYPES.indexOf(props.type) === -1) {
  throw new Error(`Unknown button type: ${props.type}`);
}
```

File: native/packages/react-native-bpk-component-button/src/BpkButton.js (hunk 4 above)
Comment (k0nserv) [reply native/packages/react-native-bpk-component-button/src/BpkButton.js]: @shaundon That's a good point that I failed to recognize. The api on iOS is

```
func setTitle(_ title: String?, 
          for state: UIControlState)
```

Arbitrary children is a web thing. See the docs here https://developer.apple.com/documentation/uikit/uibutton/1624018-settitle

File: native/packages/react-native-bpk-component-button/src/BpkButton.js (hunk 6 above)
Comment (k0nserv) [reply native/packages/react-native-bpk-component-button/src/BpkButton.js]: Yeah I'm all for that

File: native/packages/react-native-bpk-component-button/package.json [hunk 7]
```diff
@@ -0,0 +1,25 @@
+{
+  "private": true,
+  "name": "react-native-bpk-component-button",
+  "version": "0.0.0",
+  "main": "index.js",
+  "description": "Backpack button component.",
```
georgegillams: Should the description say that this is a RN component?

native/packages/react-native-bpk-component-button/BpkButton.podspec
```diff
@@ -0,0 +1,15 @@
...
+  s.name             = package['name']
+  s.version          = package['version']
+  s.summary          = package['description']
+  s.license          = package['license']
+  s.homepage         = 'https://github.com/skyscanner/backpack'
+  s.authors          = package['author']
+  s.source           = { :git => 'https://github.com/skyscanner/backpack.git', :tag => "v#{s.version}" }
+  s.source_files     = 'src/*[^\-test].js'
+  s.dependency         'React'
+end
```
shaundon: @k0nserv does this look ok to you? I looked at some stuff online and had a go at making a podspec.

shaundon: Also, seeing as we're not supporting Android yet, maybe we can leave the gradle part for now - what do you think?

native/ios/native.xcodeproj/project.pbxproj
```diff
@@ -25,7 +25,7 @@
 		2D02E4BC1E0B4A80006451C7 /* AppDelegate.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB01A68108700A75B9A /* AppDelegate.m */; };
```
shaundon: I have no idea what this is.

native/ios/native/Images.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -2,6 +2,16 @@
...
+      "size" : "20x20",
+      "scale" : "2x"
+    },
+    {
+      "idiom" : "iphone",
+      "size" : "20x20",
+      "scale" : "3x"
+    },
+    {
+      "idiom" : "iphone",
```
shaundon: ¯\_(ツ)_/¯

k0nserv: Did you change the app icon?

shaundon: Not knowingly! Do you think it's worth reverting these files or leaving them be?

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -0,0 +1,194 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Platform, StyleSheet } from 'react-native';
+
+const IOS_TOKENS = require('bpk-tokens/tokens/ios/base.react.native.common.js');
+const ANDROID_TOKENS = require('bpk-tokens/tokens/android/base.react.native.common.js');
```
georgegillams: Matt and I thought it was best to in-line these so that it doesn't actually import everything for both platforms. We're not even sure if this optimises everything but it might?!

packages/bpk-docs/src/pages/NativeButtonPage/NativeButtonPage.js
```diff
@@ -0,0 +1,143 @@
...
+        src: `/${iosScreenshotPrimary}`,
+        altText: 'iOS Primary Button Component',
+        subText: '(iPhone 7 Simulator)',
+      },
+      {
+        title: 'Android',
+        width: 1080,
+        height: 1920,
+        src: `/${androidEmpty}`,
+        altText: 'Android Primary Text Component',
```
georgegillams: Should be "Button Component"

packages/bpk-docs/src/pages/NativeButtonPage/NativeButtonPage.js
```diff
@@ -0,0 +1,143 @@
...
+        src: `/${iosScreenshotSecondary}`,
+        altText: 'iOS Secondary Button Component',
+        subText: '(iPhone 7 Simulator)',
+      },
+      {
+        title: 'Android',
+        width: 1080,
+        height: 1920,
+        src: `/${androidEmpty}`,
+        altText: 'Android Secondary Text Component',
```
georgegillams: Should be "Button Component"

packages/bpk-docs/src/pages/NativeButtonPage/NativeButtonPage.js
```diff
@@ -0,0 +1,143 @@
...
+        src: `/${iosScreenshotDestructive}`,
+        altText: 'iOS Destructive Button Component',
+        subText: '(iPhone 7 Simulator)',
+      },
+      {
+        title: 'Android',
+        width: 1080,
+        height: 1920,
+        src: `/${androidEmpty}`,
+        altText: 'Android Destructive Text Component',
```
georgegillams: Should be "Button Component"

packages/bpk-docs/src/pages/NativeButtonPage/NativeButtonPage.js
```diff
@@ -0,0 +1,143 @@
...
+        src: `/${iosScreenshotFeatured}`,
+        altText: 'iOS Featured Button Component',
+        subText: '(iPhone 7 Simulator)',
+      },
+      {
+        title: 'Android',
+        width: 1080,
+        height: 1920,
+        src: `/${androidEmpty}`,
+        altText: 'Android Featured Text Component',
```
georgegillams: Should be "Button Component"

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -0,0 +1,194 @@
...
+    button: {
+      borderColor: 'transparent',
+    },
+    text: {
+      color: tokens.colorGray300,
+    },
+  }),
+  iconOnly: StyleSheet.create({
+    container: {
+      minWidth: tokens.spacingXl,
```
georgegillams: Use `width` instead of `minWidth`

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -0,0 +1,194 @@
...
+    },
+  }),
+  iconOnlyLarge: StyleSheet.create({
+    container: {
+      minWidth: largeHeight,
```
georgegillams: Use `width` instead of `minWidth`

native/packages/react-native-bpk-component-button/stories.js
```diff
@@ -0,0 +1,257 @@
...
+
+import BpkButton, { BUTTON_TYPES } from './src/BpkButton';
+
+import ArrowImageSrc from './rightarrow_360.png';
+
+const IOS_TOKENS = require('bpk-tokens/tokens/ios/base.react.native.common.js');
+const ANDROID_TOKENS = require('bpk-tokens/tokens/android/base.react.native.common.js');
+
+const tokens = Platform.select({
+  ios: () => IOS_TOKENS,
```
k0nserv: Why aren't we requring in here? The primary reason I started using comon JS was so that we could dynimcally require per platform. I see that this has also regressed for `BpkText`

native/storybook/storybook.js
```diff
@@ -5,12 +5,12 @@ import { getStorybookUI, configure } from '@storybook/react-native';
...
 // import stories
 configure(() => {
-  // require('./stories');
   require('../packages/react-native-bpk-component-text/stories');
+  require('../packages/react-native-bpk-component-button/stories');
 }, module);
 
 // This assumes that storybook is running on the same host as your RN packager,
 // to set manually use, e.g. host: 'localhost' option
-const StorybookUI = getStorybookUI({ port: 7007, onDeviceUI: true });
```
k0nserv: This will conflict with Matt's branch which changes it to be `onDeviceUI: false`

native/packages/react-native-bpk-component-button/stories.js
shaundon: Now changed. I changed it in the first place because ESLint didn't like it - will change it for BpkText in a future PR too.

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -0,0 +1,194 @@
...
+    minHeight: tokens.spacingXl,
+    minWidth: 150,
+  },
+
+  // Applied to the TouchableHighlight element.
+  button: {
+    // Must be half the height to get rounded corners.
+    borderRadius: tokens.spacingXl / 2,
+    minHeight: tokens.spacingXl,
+    paddingTop: tokens.spacingMd,
```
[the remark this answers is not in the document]
shaundon: Done - just an artefact of when all values weren't the same.

native/storybook/storybook.js
shaundon: Now changed.

packages/bpk-docs/src/pages/NativeButtonPage/NativeButtonPage.js
shaundon: Good spot!

native/packages/react-native-bpk-component-button/src/BpkButton.js
shaundon: My solution:

``` 
if (!BUTTON_TYPES.includes(type)) {
     throw new Error(`"${type}" is not a valid button type. Valid types are ${BUTTON_TYPES.join(', ')}`);
}
```

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -0,0 +1,194 @@
...
+const largeHeight = tokens.spacingXl * 1.5;
+const smallBorderWidth = tokens.spacingSm / 2;
+
+// The base styles that are initially applied to all buttons.
+const base = StyleSheet.create({
+
+  // Applied to the outer LinearGradient element.
+  container: {
+    // Must be half the height to get rounded corners.
+    borderRadius: tokens.spacingXl / 2,
```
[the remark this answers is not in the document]
shaundon: It doesn't accept percentages, unfortunately 😢 Only numbers.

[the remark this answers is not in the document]
shaundon: Setting it to an arbitrarily high number (100) did the trick.

native/packages/react-native-bpk-component-button/readme.md
```diff
@@ -0,0 +1,67 @@
...
+        <BpkButton large type="primary" title="Book flight" onPress={() => {}} />
+
+        <BpkButton type="primary" title="Book flight" icon={ArrowImage} onPress={() => {}} />
+        <BpkButton type="featured" icon={ArrowImage} onPress={() => {}} />
+      </View>
+    );
+
+    const ArrowImage = <Image source="require(./rightarrow_360.png)" style={{ height: 14, width: 16 }}/>;
+
+    const styles = StyleSheet.create({
```
georgegillams: Move this to above `export default class App...` as EsLint would complain

native/packages/react-native-bpk-component-button/readme.md
```diff
@@ -0,0 +1,67 @@
...
+        <BpkButton selected type="secondary" title="Book flight" onPress={() => {}} />
+        <BpkButton disabled type="destructive" title="Book flight" onPress={() => {}} />
+        <BpkButton large type="primary" title="Book flight" onPress={() => {}} />
+
+        <BpkButton type="primary" title="Book flight" icon={ArrowImage} onPress={() => {}} />
+        <BpkButton type="featured" icon={ArrowImage} onPress={() => {}} />
+      </View>
+    );
+
+    const ArrowImage = <Image source="require(./rightarrow_360.png)" style={{ height: 14, width: 16 }}/>;
```
georgegillams: Move this to above `export default class App...` as EsLint would complain

native/packages/react-native-bpk-component-button/readme.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+## Props
+
+| Property    | PropType                                                  | Required | Default Value |
+| ----------- | --------------------------------------------------------- | -------- | ------------- |
+| type        | oneOf('primary', 'featured', 'secondary', 'destructive')  | false    | null          |
+| title       | string                                                    | false    | null          |
+| icon        | element                                                   | false    | null          |
+| onPress     | function                                                  | true     | -             |
```
georgegillams: we tend to put required ones first, then the other, and sort them all alphabetically

native/storybook/storybook.js
```diff
@@ -5,11 +5,13 @@ import { getStorybookUI, configure } from '@storybook/react-native';
 configure(() => {
   require('../packages/react-native-bpk-component-text-input/stories');
   require('../packages/react-native-bpk-component-text/stories');
+  require('../packages/react-native-bpk-component-button/stories');
 }, module);
 /* eslint-enable global-require */
 
+// This assumes that storybook is running on the same host as your RN packager,
+// to set manually use, e.g. host: 'localhost' option
```
georgegillams: We did a bad rebase here - we overwrote changes that we shouldn't have.
My fault as much as anything!

packages/bpk-docs/src/pages/NativeButtonPage/NativeButtonPage.js
```diff
@@ -0,0 +1,143 @@
...
+      },
+    ],
+  },
+];
+
+const NativeTextPage = () => <DocsPageBuilder
+  title="Buttons"
+  blurb={[
+    <Paragraph>
+      The Backpack button component is for all your button pressing needs. It&#39;s a wrapper around the React
```
matthewdavidson: 😂

PR #89, opened 2017-09-07
Correct font for iOS devices

georgegillams:


native/packages/react-native-bpk-component-text/src/BpkText.js
```diff
@@ -46,14 +46,28 @@ const styleForTextStyle = (textStyle) => {
     }
   };
 
+  const determineIosFontFamily = (fontWeight) => {
+    let fontName = '.SFUIDisplay';
+    if (Platform.Version >= 11) {
+      fontName = '.SFProDisplay';
+    }
+    switch (fontWeight) {
+      case '700':
```
k0nserv: Is this needed? From my tests the platform correctly handles weights on its own?

native/packages/react-native-bpk-component-text/src/BpkText.js
```diff
@@ -46,14 +46,28 @@ const styleForTextStyle = (textStyle) => {
     }
   };
 
+  const determineIosFontFamily = (fontWeight) => {
+    let fontName = '.SFUIDisplay';
+    if (Platform.Version >= 11) {
+      fontName = '.SFProDisplay';
```
k0nserv: Does setting `sans-serif` like on Android not do the correct thing for 10 and 11?

k0nserv: Just had a look and on iOS the font `System` is what we'd want instead of `sans-serif` in this case

k0nserv:
Can we use iPhone 7 for the iOS screenshot please?

shaundon:
@k0nserv maybe iPhone 8 after today 😉

packages/bpk-tokens/src/android/base/typography.json
```diff
@@ -63,36 +63,6 @@
       "type": "number",
       "category": "typesettings"
     },
-    "LETTER_SPACING_XS": {
```
matthewdavidson: Remember and add a changelog entry! (calling this out as this is a breaking change)

PR #96, opened 2017-09-13
add theming to bpk-link

ejulianova:
+ [ ] For any new components I've made sure that the style can be extended by the consumer using a `className` override.

### DIFF
--- a/packages/bpk-mixins/src/mixins/_buttons.scss
+++ b/packages/bpk-mixins/src/mixins/_buttons.scss
@@ -30,27 +30,6 @@
 ///     @include bpk-button();
 ///   }
 
-$vars: (
-  primary-color: $bpk-button-background-color,
-  primary-contrast-color: $bpk-button-color,
-  primary-gradient-start-color: $bpk-color-green-500,
-  primary-gradient-end-color: $bpk-color-green-600,
-  primary-light-color: $bpk-color-green-600,
-  primary-light-contrast-color: $bpk-button-color,
-  primary-dark-color: $bpk-color-green-700,
-  primary-dark-contrast-color: $bpk-button-color,
-);
-
-@mixin gradient($startVarName, $endVarName) {
-  background-image: linear-gradient(-180deg, map-get($vars, $startVarName) 0%, map-get($vars, $endVarName) 100%);
-  background-image: linear-gradient(-180deg, var(--#{$startVarName}, map-get($vars, $startVarName)) 0%, var(--#{$endVarName}, map-get($vars, $endVarName)) 100%);
-}
-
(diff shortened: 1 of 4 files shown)


packages/bpk-mixins/src/mixins/_buttons.scss
```diff
@@ -30,27 +30,6 @@
 ///     @include bpk-button();
 ///   }
 
-$vars: (
```
k0nserv: Please don't remove this I still need it

packages/bpk-mixins/src/mixins/_typography.scss
```diff
@@ -478,22 +478,23 @@
   padding: 0;
   border: 0;
   background-color: transparent;
-  color: $bpk-link-color;
   text-decoration: $bpk-link-text-decoration;
   cursor: pointer;
   appearance: none;
 
+  @include prop(color, secondary-color);
```
k0nserv: Should this really be the secondary color?

packages/bpk-mixins/src/mixins/_buttons.scss
ejulianova: moved it to tokens/base.default.scss
is that okay?

packages/bpk-mixins/src/mixins/_typography.scss
ejulianova: I used secondary color, cause we use the primary one for the "call-to-action" color (bpk green).

packages/bpk-mixins/src/mixins/_buttons.scss
k0nserv: Don't put it in that file, it gets autogenerated and will be overwitten

packages/bpk-mixins/src/mixins/_typography.scss
k0nserv: Yeah you're right our color scheme is confusing. We might need more variables than Primary and Secondary in the end

packages/bpk-mixins/src/mixins/_buttons.scss
ejulianova: Moved it to a separate file - `_theming.scss` that's included in `_index.scss`.

packages/bpk-mixins/src/mixins/_theming.scss
```diff
@@ -0,0 +1,28 @@
+$vars: (
+  primary-color: $bpk-button-background-color,
```
k0nserv: Given how generic this now is we should directly refernce the lowelst level abstractions i.e color, but I can fix that late

PR #99, opened 2017-09-13
theme button--selected and bpk-progress

ejulianova:
+ [ ] For any new components I've made sure that the style can be extended by the consumer using a `className` override.


packages/bpk-component-nudger/src/bpk-nudger.scss
```diff
@@ -21,6 +21,11 @@
 .bpk-nudger {
   &__button {
     position: relative;
+    @include prop(color, secondary-color);
+
+    &[disabled] {
```
k0nserv: Why did you change this to not be based on class name?

ejulianova: The classname was being set to the icon instead of the button itself. And now we set the fill to currentColor which is the button font color.

PR #104, opened 2017-09-19
Release tweaks

georgegillams:


native/packages/react-native-bpk-component-text-input/stories.js
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
-import { StyleSheet, View, Platform } from 'react-native';
-
 import { storiesOf } from '@storybook/react-native';
+import { StyleSheet, View, Platform } from 'react-native';
 
-import BpkTextInput from '../../packages/react-native-bpk-component-text-input';
+import BpkTextInput from './index';
 import BpkText from '../../packages/react-native-bpk-component-text';
```
matthewdavidson: This should just be `import BpkText from 'react-native-bpk-component-text';`

PR #105, opened 2017-09-20
[BPK-906] Separate layout docs into 3 pages

georgegillams:

### DIFF
--- a/packages/bpk-docs/src/pages/BreakpointsPage/BreakpointsPage.js
+++ b/packages/bpk-docs/src/pages/BreakpointsPage/BreakpointsPage.js
@@ -1,0 +1,80 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import BpkBreakpoint, { BREAKPOINTS } from 'bpk-component-breakpoint';
+import { cssModules } from 'bpk-react-utils';
+import breakpointReadme from 'bpk-component-breakpoint/readme.md';
+
+import DocsPageBuilder from './../../components/DocsPageBuilder';
+import Paragraph from './../../components/Paragraph';
+import STYLES from './breakpoints-page.scss';
+
+const getClassName = cssModules(STYLES);
+
+const MediaQueryStatus = (props) => {
+  const className = getClassName(props.isActive ? 'bpk-breakpoints-demo--active' : 'bpk-breakpoints-demo--inactive');
+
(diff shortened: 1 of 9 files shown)


packages/bpk-docs/src/pages/BreakpointsPage/BreakpointsPage.js
```diff
@@ -0,0 +1,43 @@
...
+const breakpoints = pickBy(TOKENS, (value, key) => includes(key, 'breakpoint') && !includes(key, 'breakpointQuery'));
+
+const BreakpointsPage = () => <DocsPageBuilder
+  title="Breakpoints"
+  blurb={[
+    <Paragraph>
+      To simplify things, Backpack uses only three breakpoints optimised for mobile, tablet and desktop viewports.
+    </Paragraph>,
+  ]}
+  tokenMap={breakpoints}
```
matthewdavidson: A suggestion: perhaps instead of showing the token map - can we show an example of the breakpoint component in action - perahps similar to the example we use in the storybook https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-breakpoint/stories.js#L43

packages/bpk-component-breakpoint/readme.md
```diff
@@ -14,10 +14,34 @@ npm install bpk-component-breakpoint --save-dev
...
+  backgroundColor: colorGray100,
+};
+
 export default () => (
-  <BpkBreakpoint query={BREAKPOINTS.MOBILE}>
-    {isActive => (isActive ? <span>Mobile viewport is active</span> : <span>Mobile viewport is inactive</span>)}
-  </BpkBreakpoint>
+  <div>
+    <BpkBreakpoint query={BREAKPOINTS.MOBILE}>
+      {isActive => <div style={isActive ? activeStyle : inactiveStyle}>{'MOBILE'}</div>}
```
matthewdavidson: This is an interesting one, because I'd immediately looked at that code and thought that `<div style={isActive ? activeStyle : inactiveStyle}>` should be encapsulated in a component i.e.

```
const BreakpointIndicator = (props) => (
  <BpkBreakpoint query={props.query}>
    <div style={props.isActive ?{...} : {...}}>{props.children}</div>
  </BpkBreakpoint>
);
```

However...

It's perhaps a little too abstract for the purposes of demonstrating the usage of `BpkBreakpoint`. Perhaps 

```
const BreakpointIndicator = (props) => (
    <div style={props.isActive ? {...} : {...}}>{props.children}</div>
);
```

packages/bpk-docs/src/pages/BreakpointsPage/BreakpointsPage.js
```diff
@@ -0,0 +1,43 @@
...
+  readme={breakpointReadme}
+  sassdocId="breakpoints"
```
matthewdavidson: Consider adding `showMenu=false` here to hide the menu given that it only has one item ("Default").

packages/bpk-docs/src/pages/BreakpointsPage/BreakpointsPage.js
```diff
@@ -0,0 +1,74 @@
...
+const BreakpointsPage = () => <DocsPageBuilder
+  title="Breakpoints"
+  blurb={[
+    <Paragraph>
+      To simplify things, Backpack uses only three breakpoints optimised for mobile, tablet and desktop viewports.
+    </Paragraph>,
+  ]}
+  components={components}
+  readme={breakpointReadme}
+  sassdocId="breakpoints"
```
matthewdavidson: Consider adding showMenu=false here to hide the menu given that it only has one item ("Default").

georgegillams: Did we agree against this?

PR #113, opened 2017-09-21
Bpk 879 react native card component

matteo-hertel:


native/packages/react-native-bpk-component-card/readme.md
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+```
+
+## Props
+
+| Property        | PropType  | Required | Default Value |
+| -----------     | --------- | -------- | ------------- |
+| children        | node      | true     | -             |
+| padded          | boolean    | false   | true          |
```
k0nserv: A bar is misaligned here

native/packages/react-native-bpk-component-card/src/BpkCard.ios-test.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import { Text } from 'react-native';
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import BpkCard from './BpkCard';
+
+describe('iOS', () => {
+  describe('BpkCard', () => {
```
k0nserv: Should iOS not be using the common tests?

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -0,0 +1,80 @@
...
+    shadowOffset: { width: 0, height: 1 },
+    shadowOpacity: 0.3,
+    shadowRadius: 3,
+    backgroundColor: colorWhite,
+  },
+});
+
+const BpkCard = (props) => {
+  const { padded } = props;
+  const paddingStyle = { padding: spacingBase };
```
k0nserv: We should err on the side of using `Stylesheet.create` so can't we do this by adding 

```
padded: {
  padding: spacingBase,
}
```

when creating the styles and conditionally applying it?

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -0,0 +1,80 @@
...
+
+  if (!padded) {
+    paddingStyle.padding = 0;
+  }
+
+  return (
+    <View
+      style={[styles.card, paddingStyle]}
+    >
+      {props.children}
```
k0nserv: Can we destruct children above and also extract out the `rest` and spread it on the View?

native/packages/react-native-bpk-component-card/stories.js
```diff
@@ -0,0 +1,61 @@
...
+);
+
+storiesOf('BpkCard', module)
+  .addDecorator(getStory =>
+    <View style={styles.centered}>
+      {getStory()}
+    </View>,
+)
+  .add('Default', () => (
+    <View>
```
k0nserv: We should use the content described in the JIRA card for the docs here so that we can use these stories to generate the appropriate screenshots.

native/packages/react-native-bpk-component-card/readme.md
matteo-hertel: Cheers, fixed it

native/packages/react-native-bpk-component-card/src/BpkCard.js
matteo-hertel: absolutely right, the previous revision had `nopadding` as prop rather than `padded` now what you're suggesting makes more sense, I'll amend

native/packages/react-native-bpk-component-card/src/BpkCard.js
matteo-hertel: sure, will do

matteo-hertel:
As discuss during the stand up the docs will go in a separate PR in order to avoid bloated PRs

native/packages/react-native-bpk-component-card/stories.js
matteo-hertel: Not sure I'm following here, are you saying name the stories based on the different configuration I've got in the JIRA ticket? or something else?

matteo-hertel: Ignore my comment, I've been told what to do 😄

native/packages/react-native-bpk-component-card/stories.js
```diff
@@ -0,0 +1,60 @@
...
+      <BpkCard>{copy}</BpkCard>
+    </View>
+  ))
+  .add('Without Padding', () => (
+    <View>
```
georgegillams: The card doesn't need to be wrapped in a `View`

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -0,0 +1,85 @@
...
+const styles = StyleSheet.create({
+  card: {
+    // TODO: update once tokens are available
+    borderRadius: 8,
+    shadowColor: colorGray900,
+    shadowOffset: { width: 0, height: 1 },
+    shadowOpacity: 0.3,
+    shadowRadius: 3,
+    backgroundColor: colorWhite,
+  },
```
georgegillams: shaddow is not showing up on android

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -0,0 +1,85 @@
...
+  const { padded, children, ...rest } = props;
+
+  const cardStyle = [styles.card];
+
+  if (padded) {
+    cardStyle.push(styles.padded);
+  }
+
+  return (
+    <View
```
georgegillams: When there is no padding, the child element could overlay this view and hide the borderRadius

matteo-hertel: This is all sorted now, thanks for the help fixing it!

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -72,7 +73,10 @@ const BpkCard = (props) => {
   }
 
   return (
-    <View
+
+    <TouchableHighlight
```
k0nserv: If this is pressable should it have a "button" role for accessibility? cc @shaundon

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -41,7 +43,6 @@ const {
 const styles = StyleSheet.create({
   card: {
     // TODO: update once tokens are available
-    borderRadius: 8,
```
k0nserv: No more borders?

native/packages/react-native-bpk-component-card/src/BpkCard.js
shaundon: It should yeah - @matteo-hertel happy to give you a hand if you like

native/packages/react-native-bpk-component-card/src/BpkCard.js
matteo-hertel: it has been moved to the common styles because both the TouchableHighlight and View are using it

native/packages/react-native-bpk-component-card/src/BpkCard.js
matteo-hertel: @shaundon yes, please!

changelog.md
```diff
@@ -2,7 +2,10 @@
 
 ## UNRELEASED
 
-_Nothing yet!_
+**Added:**
+- react-native-bpk-component-card
+  - New component BpkCard
```
matthewdavidson: Please follow the structure of previous "new component" changelogs

See https://github.com/Skyscanner/backpack/blob/master/changelog.md#2017-09-20---new-react-native-button-component:
```md
- react-native-bpk-component-text-input: 0.0.1 => 1.0.0
  - New component, see http://backpack.prod.aws.skyscnr.com/components/native/text-input
```

matthewdavidson: Please follow the structure of previous "new component" changelogs

See https://github.com/Skyscanner/backpack/blob/master/changelog.md#2017-09-20---new-react-native-button-component:
```md
- react-native-bpk-component-text-input: 0.0.1 => 1.0.0
  - New component, see http://backpack.prod.aws.skyscnr.com/components/native/text-input
```

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -0,0 +1,107 @@
...
+  const { padded, children, onPress, accessibilityLabel, ...rest } = props;
+
+  const cardStyle = [styles.card, styles.common];
+
+  if (padded) {
+    cardStyle.push(styles.padded);
+  }
+
+  return (
+
```
matthewdavidson: Extraneous newline

native/packages/react-native-bpk-component-card/stories.js
```diff
@@ -0,0 +1,72 @@
...
+  colorBlue700,
+  colorGray500,
+} = tokens;
+
+const styles = StyleSheet.create({
+  centered: {
+    flex: 1,
+    flexDirection: 'column',
+    justifyContent: 'center',
+    padding: spacingSm,
```
matthewdavidson: Can we make sure that the padding around stories is consistent across components (especially for screenshots) - see https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-text/stories.js#L12

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -0,0 +1,107 @@
...
+        {children}
+      </View>
+    </TouchableHighlight>
+  );
+};
+
+BpkCard.propTypes = {
+  accessibilityLabel: PropTypes.string,
+  children: PropTypes.node,
+  onPress: PropTypes.func.isRequired,
```
matthewdavidson: I can see your motivation for making `onPress` required, by in my experience it's better to be less opinionated about what is strictly required - many times we've had to make breaking changes to relax required props down the line because there has been a valid use case that we haven't thought of.

The web counterpart doesn't have `onClick` required for instance (see http://backpack.prod.aws.skyscnr.com/components/web/cards/).

Happy to get more input on this too btw.

native/packages/react-native-bpk-component-card/src/BpkCard.android-test.js
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import commonTests from './BpkCard.commontests';
+
+describe('Android', () => {
+  commonTests();
```
matthewdavidson: Don't we need to mock `Platform` from `react-native` here?

native/packages/react-native-bpk-component-card/src/BpkCard.ios-test.js
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import commonTests from './BpkCard.commontests';
+
+
+describe('iOS', () => {
+  commonTests();
```
matthewdavidson: And here...

native/packages/react-native-bpk-component-card/readme.md
```diff
@@ -0,0 +1,57 @@
...
+  container: {
+    flex: 1,
+    justifyContent: 'center',
+    padding: TOKENS.spacingBase,
+  }
+});
+
+export default class App extends Component {
+  render() {
+    const copy = (
```
shaundon: Very small thing but I'm wondering if something like `cardContent` would be more explicit as a variable name?

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -0,0 +1,107 @@
...
+        style={[styles.overflowHidden, styles.common]}
+      >
+        {children}
+      </View>
+    </TouchableHighlight>
+  );
+};
+
+BpkCard.propTypes = {
+  accessibilityLabel: PropTypes.string,
```
matthewdavidson: Strictly speaking, we are doing extra work here by accepting this prop and passing it through i.e. if we didn;t do this it would still be handled by the spread `{...rest}`. Im in favor of removing it - we can still have it in the readme if we want to promote it and can add a test for doing so.

native/packages/react-native-bpk-component-card/readme.md
```diff
@@ -0,0 +1,57 @@
...
+        <BpkText textStyle="xl">Flights to Edinburgh</BpkText>
+        <BpkText textStyle="lg">Flights to Edinburgh</BpkText>
+        <BpkText textStyle="base">Flights to Edinburgh</BpkText>
+        <BpkText textStyle="sm">Flights to Edinburgh</BpkText>
+        <BpkText textStyle="xs">Flights to Edinburgh</BpkText>
+      </View>
+    );
+    return (
+      <View style={styles.container}>
+        <BpkCard onPress={() => { } accessibilityLabel="Example Card"}>{copy}</BpkCard>
```
shaundon: There's a typo here, the closing `}` for the `onPress` prop is at the end of `Example Card`. It should be:

```<BpkCard onPress={() => {}} accessibilityLabel="Example Card">{copy}</BpkCard>```

File: native/packages/react-native-bpk-component-card/src/BpkCard.commontests.js [hunk 20]
```diff
@@ -0,0 +1,51 @@
+/*
```
shaundon: In other components, the filename is `componentName-test.common.js`. I think the filename here should change to fit that convention.

native/packages/react-native-bpk-component-card/stories.js
matteo-hertel: sure, will update the story and the screenshots as well

changelog.md
matteo-hertel: Yes, not a problem, can I omit the docs link as it'll not be published by the time the PR will be merged?

native/packages/react-native-bpk-component-card/stories.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+import { StyleSheet, View, Platform } from 'react-native';
+import BpkCard from './index';
+import BpkText from '../react-native-bpk-component-text';
```
shaundon: Better to import this as a module, i.e. add `react-native-bpk-component-text` as a `devDependency` and import it that way. Lerna does some clever symlinking so that it doesn't actually install the dependency, it just references it inside the monorepo.

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -0,0 +1,107 @@
...
+    cardStyle.push(styles.padded);
+  }
+
+  return (
+
+    <TouchableHighlight
+      accessibilityComponentType="button"
+      accessibilityLabel={accessibilityLabel}
+      onPress={onPress}
+      style={cardStyle}
```
matthewdavidson: We should allow userland style's to be applied here.

changelog.md
matthewdavidson: yeah sure

native/packages/react-native-bpk-component-card/src/BpkCard.js
shaundon: Good idea, I didn't think of the `...rest` argument when I suggested the card accept this prop.

native/packages/react-native-bpk-component-card/src/BpkCard.js
shaundon: Should `BpkButton` mirror this too?

native/packages/react-native-bpk-component-card/src/BpkCard.js
matteo-hertel: vs code auto formatter failed me 😢

changelog.md
georgegillams: Although obviously you won't yet have version numbers to include 😄

native/packages/react-native-bpk-component-card/readme.md
k0nserv: Isn't your convention to just return null in example callbacks. i.e `onPress={() => null}`?

native/packages/react-native-bpk-component-card/src/BpkCard.commontests.js
k0nserv: 👍

changelog.md
matteo-hertel: so that leaves what I've put there? 😄 

```
+- react-native-bpk-component-card
+  - New component BpkCard
```

matthewdavidson: ...  i suppose it does! 😳 😂

native/packages/react-native-bpk-component-card/src/BpkCard.js
matteo-hertel: I totally agree with that, we had a discussion yesterday about it and the decision was to put it, if @matthewdavidson is in favour of removing it I'll get rid of it and add a section in the readme for accessibility

native/packages/react-native-bpk-component-card/src/BpkCard.js
matteo-hertel: I agree with the not making the `onPress` mandatory, off the back of a discussion from yesterday the end result was "a card should be clickable so onPress is required", that leaves to the point tough if there is no onPress should a card use TouchHighlight? it'll give the pressed feedback when nothing it's actually happening

native/packages/react-native-bpk-component-card/src/BpkCard.android-test.js
matteo-hertel: yep, I'll add that in, thanks

changelog.md
georgegillams: 😆 😆 😆 😆

matteo-hertel: I'll leave it like that then!

native/packages/react-native-bpk-component-card/src/BpkCard.js
matteo-hertel: After a brief discussion, we came to an agreement to leave the onPress required, for now, it'll be easy to change in the future and making it nor required will not be a breaking change

k0nserv:
LGTM, make sure you squash the commits before merging

PR #115, opened 2017-09-21
Add touch and hover event support to BpkBarchart

tomcorke:


packages/bpk-component-barchart/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bpk-component-barchart",
-  "version": "2.0.12",
+  "version": "2.1.0",
```
k0nserv: Please don't update these version numbers. Lerna takes care of that for us

changelog.md
```diff
@@ -13,6 +13,8 @@
 ## 2017-09-20 - New React Native Button component
 
 **Added:**
+- bpk-component-barchart: 2.0.10 => 2.1.0
```
k0nserv: Don't reference the new version number here, we do this as part of the release

tomcorke:
Thanks @k0nserv, I've removed those changes.

packages/bpk-component-barchart/package.json.orig
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "bpk-component-barchart",
+<<<<<<< HEAD
```
georgegillams: Unresolved conflict

georgegillams:
I feel like we should include examples of the different interaction-types in storybook. We can do that  ourselves as a separate PR, though 😸

PR #120, opened 2017-09-22
[BPK-972] Make height of textinput 100pc

georgegillams:


georgegillams:
before
after

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -79,6 +79,7 @@ const styles = StyleSheet.create({
     // TODO Replace with fontFamilyLg token once available
     fontFamily: Platform.OS === 'android' ? 'sans-serif-medium' : 'System',
     lineHeight: textLgLineHeight,
+    height: '100%',
```
matthewdavidson: I'd expect to see a snapshot update for this.

shaundon: Soon Danger will do this!

georgegillams: But Travis already does!

georgegillams: I feel like instead of doing this, we should have a touch-listener on the entire component which activated the TextInput.
Thoughts?

PR #121, opened 2017-09-22
[BPK-903] Move bonds to tokens section

matthewdavidson:
This is the first PR for moving the tokens pages to their own section. I will follow up with the restructure of the actual pages themselves in separate PR's

### DIFF
--- a/packages/bpk-docs/src/layouts/TokensLayout/TokensLayout.js
+++ b/packages/bpk-docs/src/layouts/TokensLayout/TokensLayout.js
@@ -1,0 +1,48 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import SideNavLayout from './../SideNavLayout';
+import * as routes from './../../constants/routes';
+
+const links = [
+  {
+    id: 'Tokens',
+    category: 'Tokens',
+    sort: true,
+    links: [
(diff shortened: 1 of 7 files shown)


matthewdavidson:
*homepage before:*
before homepage
*homepage after:*
after homepage
*components section before:*
before components
*components section after:*
after components
*new tokens section:*
new tokens section

packages/bpk-docs/src/layouts/TokensLayout/TokensLayout.js
```diff
@@ -0,0 +1,48 @@
...
+import SideNavLayout from './../SideNavLayout';
+import * as routes from './../../constants/routes';
+
+const links = [
+  {
+    id: 'Tokens',
+    category: 'Tokens',
+    sort: true,
+    links: [
+      { id: 'COLORS', route: routes.COLORS, children: 'Colors' },
```
georgegillams: It bugs me that these are not sorted alphabetically in the code. As this is an entirely new file can we at least do it here?

packages/bpk-docs/src/routes/Routes.js
```diff
@@ -97,15 +98,19 @@ const Routes = (
...
-      <Route path={ROUTES.TYPESETTING} component={TypesettingPage} />
+      <Route path={ROUTES.LAYOUT} component={LayoutPage} />
       <Route path={ROUTES.RADII} component={RadiiPage} />
       <Route path={ROUTES.SHADOWS} component={ShadowsPage} />
-      <Route path={ROUTES.BORDERS} component={BordersPage} />
-      <Route path={ROUTES.LAYOUT} component={LayoutPage} />
-      <Route path={ROUTES.ANIMATION} component={AnimationPage} />
+      <Route path={ROUTES.TYPESETTING} component={TypesettingPage} />
+    </Route>
+
```
georgegillams: The additional whitespace here is inconsistent with the rest of the file :)

matthewdavidson: Actually, each top level route (i.e. using, style, tokens, components & resources) has new lines around it.

packages/bpk-docs/src/layouts/TokensLayout/TokensLayout.js
matthewdavidson: If we sorted alphabetically in the code then there would be no use in the `sort: true` functionality - I dunno about you but when I add a new route I am too lazy to put it in alphabetical order ;)

packages/bpk-docs/src/routes/Routes.js
georgegillams: My bad!

matthewdavidson: Not at all - the diff doesn't do it justice!

PR #122, opened 2017-09-22
[BPK-925] Add API for setting opacity of colour tokens

georgegillams:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,9 @@
 - react-native-bpk-component-card:
   - New component BpkCard
 
+- bpk-tokens:
+  - Color tokens can now be made opaque with `setOpacity`
+
 ## 2017-09-25 - Bar chart bars now have hooks for hover and touch events
 
 **Added:**
--- a/packages/bpk-tokens/index.js
+++ b/packages/bpk-tokens/index.js
@@ -1,0 +1,22 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import setOpacity from './src/setOpacity';
+
+export { setOpacity };
+export default { setOpacity };
--- a/packages/bpk-tokens/package.json
+++ b/packages/bpk-tokens/package.json
@@ -11,5 +11,8 @@
     "url": "git@github.com:Skyscanner/backpack.git"
   },
   "author": "Backpack Design System <backpack@skyscanner.net>",
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "dependencies": {
+    "color": "^2.0.0"
+  }
 }
--- a/packages/bpk-tokens/src/setOpacity-test.js
+++ b/packages/bpk-tokens/src/setOpacity-test.js
@@ -1,0 +1,57 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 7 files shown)


packages/bpk-tokens/package.json
```diff
@@ -11,5 +11,8 @@
     "url": "git@github.com:Skyscanner/backpack.git"
   },
   "author": "Backpack Design System <backpack@skyscanner.net>",
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "devDependencies": {
```
matthewdavidson: This should be a `"dependency"` - this will fail otherwise when a consumer `npm install`'s bpk-tokens as only `"dependency"`'s are installed.

packages/bpk-tokens/src/setOpacity.js
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Color from 'color';
+
+const setOpacity = function (colorToken, opacity) {
```
matthewdavidson: What happens if i pass `colorToken` as `null` ?

matthewdavidson: (or any other unexpected value)

packages/bpk-tokens/index.js
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import setOpacity from './src/setOpacity';
+
+export default { setOpacity };
+export { setOpacity };
```
matteo-hertel: Quick question, why exporting twice when you're exporting default already?

changelog.md
```diff
@@ -5,10 +5,14 @@
 **Fixed:**
 - react-native-bpk-component-text-input:
   - Made hit area larger to make activation of the Text Input easier on a touch-device
+
```
matteo-hertel: whats the policy for new lines? yes or no? 😄

changelog.md
```diff
@@ -5,10 +5,14 @@
...
 **Added:**
 - react-native-bpk-component-card
   - New component BpkCard
 
+- bpk-tokens
+  - New `Color` class for setting opacity of Bpk colours
+
```
matteo-hertel: Same as above

changelog.md
```diff
@@ -5,10 +5,14 @@
 **Fixed:**
 - react-native-bpk-component-text-input:
   - Made hit area larger to make activation of the Text Input easier on a touch-device
+
 **Added:**
 - react-native-bpk-component-card
   - New component BpkCard
 
+- bpk-tokens
+  - New `Color` class for setting opacity of Bpk colours
```
matthewdavidson: This is incorrect, you have actually added `setOpacity` ;)

Suggestion:
```
- Color tokens can now be made opaque with `setOpacity`
```

matteo-hertel: Color should be `setOpacity`

matteo-hertel: Sorry this has been picked up already, ignore me 😄

packages/bpk-tokens/src/setOpacity.js
```diff
@@ -0,0 +1,30 @@
...
+  if (!parseFloat(opacity)) {
```
matthewdavidson: So I know I mentioned not writing tests for third party code - but we have now introduced our own error checking therefore there should be tests for it :)

packages/bpk-tokens/src/setOpacity.js
georgegillams: An error is thrown by the underlying library

georgegillams: `setOpacity(null, 0.5) => "rgba(0, 0, 0, 0.5)"`

matteo-hertel: we should probably prevent that, having that default doesn't look correct to me

packages/bpk-tokens/src/setOpacity.js
georgegillams: done 😸

georgegillams:
We don't want to go too overkill when it comes to handling edge cases as we could get seriously carried away. We've done the main ones, so I think it's best to leave any others to fail gracefully according to the `color` library

packages/bpk-tokens/src/setOpacity-test.js
```diff
@@ -0,0 +1,57 @@
...
+    // eslint-disable-next-line max-len
+    expect(() => setOpacity('rgb(119, 67, 206)', null)).toThrow('Invalid arg `opacity` of type `object` supplied to `setOpacity`, expected a numeric value between 0.0 and 1.0.');
+  });
+
+  it('should error if color token is invalid', () => {
+    expect(() => setOpacity('ASDFG', 0.5)).toThrow('Unable to parse color from string: ASDFG');
+  });
+
+  it('should error if color token is null', () => {
+    expect(setOpacity(null, 0.5)).toEqual('rgba(0, 0, 0, 0.5)');
```
shaundon: I'm wondering if the description of this test should change, as it's not really errorring, more falling back.

georgegillams: Good point!

PR #124, opened 2017-09-22
[BPK-966] Make Dangerfile more opinionated

shaundon:
Danger will now run at the beginning of a build, before tests run.
If a package source files have changed but snapshots haven't, Danger
will post a warning.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -25,12 +25,26 @@ import { danger, warn, message } from 'danger';
 
 const fileChanges = [...danger.git.modified_files, ...danger.git.created_files];
 
+// Always be nice.
+message('Thanks for the PR 🎉.');
+
 // If any of the packages have changed, the changelog should have been updated.
 const changelogModified = includes(fileChanges, 'changelog.md');
-const packagesModified = fileChanges.filter(filePath => filePath.startsWith('packages/')).length > 0;
+const packagesModified = fileChanges.some(filePath => (
+  filePath.startsWith('packages/') || filePath.startsWith('native/packages/')
+));
 if (packagesModified && !changelogModified) {
   warn('One or more packages have changed, but `changelog.md` wasn\'t updated.');
 }
 
-// Always be nice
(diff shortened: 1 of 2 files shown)


dangerfile.js
```diff
@@ -25,12 +25,26 @@ import { danger, warn, message } from 'danger';
 
 const fileChanges = [...danger.git.modified_files, ...danger.git.created_files];
 
+// Always be nice.
+message('Thanks for the PR 🎉.');
+
 // If any of the packages have changed, the changelog should have been updated.
 const changelogModified = includes(fileChanges, 'changelog.md');
-const packagesModified = fileChanges.filter(filePath => filePath.startsWith('packages/')).length > 0;
+const packagesModified = fileChanges.filter(filePath => (
```
k0nserv: Could use `Array.prototype.some` here instead

dangerfile.js
```diff
@@ -25,12 +25,26 @@ import { danger, warn, message } from 'danger';
...
 if (packagesModified && !changelogModified) {
   warn('One or more packages have changed, but `changelog.md` wasn\'t updated.');
 }
 
-// Always be nice
-message('Thanks for the PR 🎉.');
+// If source files have changed, the snapshots should have been updated.
+const sourceFilesModified = fileChanges.filter(filePath => (
+  // packages/(one or more chars)/src/(one or more chars).js
+  filePath.match(/packages\/.+\/src\/.+\.js/) && !includes(filePath, '-test.')
```
georgegillams: What if the changed file(s) are in `/native/packages/...`??

georgegillams: Ignore me - it would still be a match!!

shaundon: It still works, because the regex doesn't have the start-of-line thing and will therefore match any part of the path.

dangerfile.js
shaundon: I didn't know about `some`! That's awesome

k0nserv:
LGTM

PR #126, opened 2017-09-22
[NO-JIRA] docs pluralisation

georgegillams:
At the moment our docs use inconsistent singlular / plural names for components. I thought we should make RN pages be consistent before we get too stuck in!

+ [ ] For any new web components I've made sure that the style can be extended by the consumer using a `className` override.
+ [ ] For any new native components I've made sure that the style can be extended by the consumer using a `style` override.


georgegillams:
Maybe it would be better to make everything singular instead

"Texts" and "Horizontal navigations" is gonna sound a bit weird!

PR #128, opened 2017-09-25
WIP: Added bpk-component-toast.

clolov:
+ [x] ~~For any new native components I've made sure that the style can be extended by the consumer using a `style` override.~~
+ [x] ~~If I've updated `npm-shrinkwrap.json` those changes are intentional.~~

### DIFF
--- a/packages/bpk-component-toast/readme.md
+++ b/packages/bpk-component-toast/readme.md
@@ -1,0 +1,28 @@
+# bpk-component-toast
+
+> Backpack toast component.
+
+## Installation
+
+```sh
+npm install bpk-component-toast --save-dev
+```
+
+## Usage
+
+```javascript
+import React from 'react';
+import BpkToast from 'bpk-component-toast';
+
+export default () => (
+  <BpkToast message="My sample toast" isVisible />
+);
+```
+
+## Props
+
+| Property  | PropType                                 | Required | Default Value |
+| --------- | ---------------------------------------- | -------- | ------------- |
+| message   | string                                   | true     | -             |
+| isVisible | bool                                     | false    | false         |
+| onClose   | function                                 | false    | () => {}      |
--- a/packages/bpk-component-toast/src/BpkToast.js
+++ b/packages/bpk-component-toast/src/BpkToast.js
@@ -1,0 +1,98 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import BpkCard from 'bpk-component-card';
+import BpkLargeExclamationCircle from 'bpk-component-icon/lg/exclamation-circle';
+import BpkCloseButton from 'bpk-component-close-button';
+import { cssModules } from 'bpk-react-utils';
+
+import STYLES from './bpk-toast.scss';
+
+const getClassName = cssModules(STYLES);
+
+class BpkToast extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { isOpen: true };
+
+    this.onClose = this.onClose.bind(this);
+    this.setAnimation = this.setAnimation.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.props.message !== nextProps.message || nextProps.isVisible) {
+      this.setState({ isOpen: true });
+    }
+  }
+
+  onClose() {
+    this.setState({ isOpen: false });
+    this.props.onClose();
+  }
+
+  setAnimation() {
+    return this.state.isOpen ? 'bpk-toast-animation-forward' : 'bpk-toast-animation-backward';
+  }
+
+  render() {
+    const classNames = [];
+    classNames.push(getClassName('bpk-toast'));
+    classNames.push(getClassName(this.setAnimation()));
+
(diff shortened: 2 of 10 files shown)


packages/bpk-component-toast/src/BpkToast.js
```diff
@@ -0,0 +1,98 @@
...
+
+  setAnimation() {
+    return this.state.isOpen ? 'bpk-toast-animation-forward' : 'bpk-toast-animation-backward';
+  }
+
+  render() {
+    const classNames = [];
+    classNames.push(getClassName('bpk-toast'));
+    classNames.push(getClassName(this.setAnimation()));
+
```
georgegillams: If the consumer provides style (or any other props) these should be applied to the component (usually to the outermost-div)

clolov: I believe this is done on the next line (https://github.com/Skyscanner/backpack/pull/128/files#diff-1b87f02ec1a0abe7a5a6fd05e0ff09beR60). Let me know if that is what you meant or if I have misunderstood something.

packages/bpk-component-toast/src/BpkToast.js
```diff
@@ -0,0 +1,98 @@
...
+  closeLabel: PropTypes.string,
+  message: PropTypes.string.isRequired,
+};
+
+BpkToast.defaultProps = {
+  onClose: () => {},
+  isVisible: false,
+  className: null,
+  closeLabel: null,
+  message: undefined,
```
georgegillams: This line doesn't need to be here

packages/bpk-component-toast/readme.md
```diff
@@ -0,0 +1,28 @@
...
+export default () => (
+  <BpkToast message="My sample toast" isVisible />
+);
+```
+
+## Props
+
+| Property  | PropType                                 | Required | Default Value |
+| --------- | ---------------------------------------- | -------- | ------------- |
+| message   | string                                   | true     | undefined     |
```
georgegillams: Replace `undefined` with `-`

georgegillams:
Generally looking good. Thanks for contributing 😸

clolov:
Glad I could help :).

clolov:
@georgegillams who should I contact for a design review?

packages/bpk-component-toast/src/BpkToast.js
georgegillams: That'll do it. Maybe best to check that `this.props.className !== null` too, but if it works without that's fine

georgegillams:
I can get @jamesf3rguson to look over it 👍

jamesf3rguson:
happy to have a look. leave it with me.

jamesf3rguson:
Firstly, thanks so much for looking this, its generally looking and behaving as expected. Just a few (hopefully small) tweaks from me:

### Visual
- [ ] The blue icon container width should be: (rem equivalent of 48px)
- [ ] The ‘card’ container should use a box-shadow of box-shadow-lg
- [ ] Toast message text should be right aligned
- [ ] Toast message text should use color-gray-700
- [ ] The gradient ‘scrim’ should be a little taller which should make it look more gradual: (rem equivalent of 114px)
- [ ] The toast notification should be positioned bpk-spacing-base from the bottom of the browser to ensure it is not too close to the bottom of the screen. 

### Interaction
None :)

### Motion
- [ ] The toast notification should bounce similar to this prototype: https://codepen.io/jamesferguson/pen/YZOGEd?editors=1100
- [ ] The gradient ‘scrim’ should move in (and fade) from the bottom of the screen slightly ahead of the bounce animation above.

If you’d like to pair on the motion stuff, give me a shout, it might be easier. Describing motion can sometimes be a tricky one. Any questions on anything else, just ping me.

packages/bpk-component-toast/src/BpkToast.js
```diff
@@ -0,0 +1,98 @@
...
+
+  render() {
+    const classNames = [];
+    classNames.push(getClassName('bpk-toast'));
+    classNames.push(getClassName(this.setAnimation()));
+
+    if (this.props.isVisible) { classNames.push(this.props.className); }
+
+    return this.props.isVisible ? (
+      <div className={classNames.join(' ')}>
```
matthewdavidson: Please spread any other arbitrary props that consumers pass to `<BpkToast>` onto the outer most element.

i.e. if consumers provide:
```

      <BpkToast
        id="my-toast"
        message="Sample toast message"
        className="sample-class-name"
        closeLabel="Close toast"
        isVisible
        data-my-custom-prop="my-data"
      />
```

then the custom prop `data-my-custom-prop` shoudl be applied to the outer `div` as follows:
```
<div className={classNames.join(' ')} {...rest}>
```

`rest` can be obtained from props using [object rest spread](https://github.com/tc39/proposal-object-rest-spread) syntax:
```
const { props, i, am, interested, in, ...rest } = this.props;
```

packages/bpk-component-toast/stories.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import BpkToast from './index';
+
+storiesOf('bpk-component-toast', module)
+  .add('Default', () => (
+    <BpkToast
+      id="my-toast"
```
matthewdavidson: Related to https://github.com/Skyscanner/backpack/pull/128/files#r141286889, this `id` won't actually be applied to anything.

matthewdavidson:
Hey @clolov! Thanks so much for the PR. From my point of view the code looks good - aside from James' design tweaks and the other small issues i'd probably merge this as is. There are some fundamental considerations we need to address in follow up PR's though:

1. Documentation on http://backpack.prod.aws.skyscnr.com/
2. a11y considerations (see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role for inspiration). Essentially this boils down to testing the component using a screen reader
3. From a keyboard navigation POV, should a user be able to focus on other elements on the page whilst the toast is open? Have a look at the modal component as an example of where we "scope"  focus to the dialogue only http://backpack.prod.aws.skyscnr.com/components/web/modals
4. I think this toast needs to be rendered using a portal (see this for explanation https://stackoverflow.com/questions/28802179/how-to-create-a-react-modalwhich-is-append-to-body-with-transitions). The backpack modal, popover and tooltip components make use of portals for a11y, styling and positioning benefits.

matthewdavidson:
Closing whilst there is inactivity on this PR. We can reopen in thefuture.

PR #132, opened 2017-09-26
Bpk 879 shadow fix

matteo-hertel:


native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -46,9 +47,9 @@ const styles = StyleSheet.create({
     backgroundColor: colorWhite,
     elevation: 3,
     shadowColor: colorGray900,
-    shadowOffset: { width: 0, height: 1 },
+    shadowOffset: { width: 0, height: 1 / PixelRatio.get() },
```
k0nserv: Hmm so this will produce half points. Interesting, but if it does what we want I guess 🤷‍♂️

jamesf3rguson:
Looks good to me

PR #136, opened 2017-09-26
Allow bpk ticket stub to take custom props

ksiwei:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,9 @@
 
 ## UNRELEASED
 
-_Nothing yet..._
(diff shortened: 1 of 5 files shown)


changelog.md
```diff
@@ -2,7 +2,9 @@
 
 ## UNRELEASED
 
-_Nothing yet..._
+**Added:**
+- react-native-bpk-component-card: 0.0.3 => 1.0.0
```
matthewdavidson: @ksiwei i've just realised what you put in this changelog 😂

PR #138, opened 2017-09-26
[BPK-903] Add token switcher

matthewdavidson:


packages/bpk-docs/src/components/DocsPageBuilder/TokenSwitcher.js
```diff
@@ -0,0 +1,105 @@
...
+  },
+  android: {
+    id: 'android',
+    name: 'Android',
+  },
+};
+
+class TokenSwitcher extends Component {
+  constructor() {
+    super();
```
matteo-hertel: Just a minor thing, but we should accept props in the constructor signature and pass it to `super`

```
constructor(props) {
  super(props);
```

packages/bpk-docs/src/pages/ZIndexesPage/ZIndexesPage.js
```diff
@@ -0,0 +1,32 @@
...
+import TOKENS from 'bpk-tokens/tokens/base.raw.json';
+import IOS_TOKENS from 'bpk-tokens/tokens/ios/base.raw.json';
+import ANDROID_TOKENS from 'bpk-tokens/tokens/android/base.raw.json';
+
+import DocsPageBuilder from './../../components/DocsPageBuilder';
+import { getPlatformTokens } from './../../helpers/tokens-helper';
+
+const ZIndexesPage = () => <DocsPageBuilder
+  title="Z-Indexes"
+  tokenMap={getPlatformTokens(TOKENS, IOS_TOKENS, ANDROID_TOKENS, ({ category }) => category === 'z-indices')}
```
matteo-hertel: @shaundon indices or indexes? 😄

matthewdavidson: Ha so I stumbled across this http://grammarist.com/usage/indexes-indices/.

Indexes sounds better - nobody says indices 😂 .

However we cannot change the token category which unfortunately is called indices.

shaundon: Yeah I just googled it and apparently both are good :P

packages/bpk-docs/src/components/DocsPageBuilder/TokenSwitcher.js
matthewdavidson: Technically you only need to pass props to super if you intend to access `this.props` in the constructor.

However, I reckon you probably just want us to be consistent, in that case I reckon we should make that change across the board in a separate PR - currently we only pass props to super when required.

matteo-hertel: Make sense, cheers

PR #139, opened 2017-09-26
Rn code coverage

matteo-hertel:

This brings the code coverage flag in and I've improved the coverage to 100% on RN components, the super annoying thing is: when testing errors, those are going to `console.error` and displayed in the console even if an error is expected, any way to prevent this?

the end result is quite lovely
screen shot 2017-09-26 at 2 20 15 pm

### DIFF
--- a/native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
+++ b/native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
@@ -16,96 +16,123 @@
  * limitations under the License.
  */
 
- import { Image } from 'react-native';
- import React from 'react';
- import renderer from 'react-test-renderer';
+import { Image } from 'react-native';
+import React from 'react';
+import renderer from 'react-test-renderer';
 
- import BpkButton, { BUTTON_TYPES } from './BpkButton';
+import BpkButton, { BUTTON_TYPES } from './BpkButton';
 
- const commonTests = () => {
-   jest.mock('Image', () => 'Image');
+const onPressFn = jest.fn();
+const commonTests = () => {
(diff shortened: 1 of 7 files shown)


native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
```diff
@@ -16,96 +16,127 @@
...
+    it('should throw an error for invalid button type', () => {
+      try {
+        renderer.create(
+          <BpkButton
+            title="Lorem ipsum"
+            type="silly"
+            onPress={onPressFn}
+          />,
+        ).toJSON();
+      } catch (exc) {
```
georgegillams: This is putting a lot of red output in the console when we run `test`.
Can we use `expect(…).toThrow(…)` instead?

matteo-hertel: unfortunately it does the very same, I think is because it returns the error rather than throw it

georgegillams: It does as well. That's annoying
I still think this would be better though:
```
    it('should throw an error for invalid button type', () => {
      expect(() => renderer.create(
        <BpkButton
          title="Lorem ipsum"
          type="silly"
          onPress={onPressFn}
        />,
        )).toThrow('"silly" is not a valid button type. Valid types are primary, featured, secondary, destructive');
    });
```

matteo-hertel: yeah much cleaner, to be honest it was like this in the beginning, but I tried to have the sterr not showing up so I ended up with try/catch

PR #141, opened 2017-09-27
[WIP] [BPK-969] Enhance Dangerfile checks

shaundon:
Danger will now check that shrinkwrap changes are intentional.
Also Danger will ensure new files are licensed.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -20,8 +20,9 @@
 
 // See http://danger.systems/js if you're not sure what this is.
 
+import fs from 'fs';
 import includes from 'lodash.includes';
-import { danger, warn, message } from 'danger';
+import { danger, fail, warn, message } from 'danger';
 
 const createdFiles = danger.git.created_files;
 const modifiedFiles = danger.git.modified_files;
@@ -30,27 +31,28 @@ const fileChanges = [...modifiedFiles, ...createdFiles];
 // Always be nice.
 message('Thanks for the PR 🎉.');
 
+// Ensure new web components are extensible by consumers.
 const webComponentIntroduced = createdFiles.some(filePath => (
   filePath.match(/packages\/bpk-component.+\/src\/.+\.js/)
 ));
 
 if (webComponentIntroduced) {
-  message('It looks like you are introducing a web component');
-  warn('Ensure the component style is extensible via `className`');
+  message('It looks like you are introducing a web component.');
+  warn('Ensure the component style is extensible via `className`.');
 }
 
+// Ensure new native components are extensible by consumers.
 const nativeComponentIntroduced = createdFiles.some(filePath => (
   filePath.match(/native\/packages\/react-native-bpk-component.+\/src\/.+\.js/)
 ));
 
 if (nativeComponentIntroduced) {
-  message('It looks like you are introducing a native component');
-  warn('Ensure the component style is extensible via `style`');
+  message('It looks like you are introducing a native component.');
+  warn('Ensure the component style is extensible via `style`.');
 }
 
-
 // If any of the packages have changed, the changelog should have been updated.
-const changelogModified = includes(fileChanges, 'changelog.md');
+const changelogModified = includes(modifiedFiles, 'changelog.md');
 const packagesModified = fileChanges.some(filePath => (
   filePath.startsWith('packages/') || filePath.startsWith('native/packages/')
 ));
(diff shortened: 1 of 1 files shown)


dangerfile.js
```diff
@@ -23,34 +23,37 @@
 import includes from 'lodash.includes';
 import { danger, warn, message } from 'danger';
 
+const fs = require('fs');
```
matthewdavidson: use es2015 module syntax:
```
import fs from 'fs';
```

dangerfile.js
```diff
@@ -71,3 +74,21 @@ const snapshotsModified = fileChanges.some(filePath => (
...
+// New files should include the Backpack license heading.
+const unlicensedFiles = createdFiles.filter((filePath) => {
+  if (filePath.match(/\.(js|css|scss|sh)/)) {
+    const fileContent = fs.readFileSync(filePath);
+    return !fileContent.includes('Backpack - Skyscanner\'s Design System');
+  }
+  return false;
+});
+if (unlicensedFiles.length > 0) {
+  warn(`These new files are do not include the license heading: ${unlicensedFiles.join(', ')}`);
```
matthewdavidson: I think this should fail the build.

dangerfile.js
```diff
@@ -71,3 +74,21 @@ const snapshotsModified = fileChanges.some(filePath => (
...
+const shrinkwrapUpdated = includes(modifiedFiles, 'changelog.md');
+if (shrinkwrapUpdated) {
+  warn('`npm-shrinkwrap.json` was updated. Ensure that this was intentional.');
+}
+
+// New files should include the Backpack license heading.
+const unlicensedFiles = createdFiles.filter((filePath) => {
+  if (filePath.match(/\.(js|css|scss|sh)/)) {
+    const fileContent = fs.readFileSync(filePath);
+    return !fileContent.includes('Backpack - Skyscanner\'s Design System');
```
matthewdavidson: Hmm nitpick: i'd have chosen something a little more obvious such as `Licensed under the Apache License, Version 2.0 (the "License");`.

dangerfile.js
shaundon: Yep sorry - I was on autopilot!

shaundon:
My previous comment that it won't let me edit should have said "addressed".

shaundon:
Couple of things I'd like to highlight for reviewers:

* The 'encourage smaller PRs' rule has a threshold of 8 files. This was just an arbitrary decision by me, we can change it if the squad decides to.
* We only check for licences in new files that end in .js, .css, .scss and .sh. I'm basing this on a search I did on the codebase to see which file types had licences.

dangerfile.js
```diff
@@ -71,3 +73,27 @@ const snapshotsModified = fileChanges.some(filePath => (
...
+const shrinkwrapUpdated = includes(modifiedFiles, 'changelog.md');
+if (shrinkwrapUpdated) {
+  warn('`npm-shrinkwrap.json` was updated. Ensure that this was intentional.');
+}
+
+// New files should include the Backpack license heading.
+const unlicensedFiles = createdFiles.filter((filePath) => {
+  if (filePath.match(/\.(js|css|scss|sh)/)) {
+    const fileContent = fs.readFileSync(filePath);
+    return !fileContent.includes('Licensed under the Apache License, Version 2.0 (the "License")');
```
georgegillams: `license-header.js` exports the license for uses like this. Better to check against this if possible (but I don't know if that is possible)

shaundon: I did think about that, but I have concerns for a couple of reasons:

* I feel like doing `String.includes` on such a large text snippet might be slower than doing it on a small one.
* `dangerfile.js` exports an XML style comment and a block style one. We'd need to introduce logic to `Dangerfile.js` to know which one of these to use in the check based on the file extension. I dunno - feels a bit like overkill to me for a simple check!

What do you reckon?

georgegillams: Sure - I'm a fan of keeping it simple 😸

dangerfile.js
```diff
@@ -30,27 +31,28 @@ const fileChanges = [...modifiedFiles, ...createdFiles];
...
+// Ensure new native components are extensible by consumers.
 const nativeComponentIntroduced = createdFiles.some(filePath => (
   filePath.match(/native\/packages\/react-native-bpk-component.+\/src\/.+\.js/)
 ));
 
 if (nativeComponentIntroduced) {
-  message('It looks like you are introducing a native component');
-  warn('Ensure the component style is extensible via `style`');
+  message('It looks like you are introducing a native component.');
+  warn('Ensure the component style is extensible via `style`.');
```
matteo-hertel: I feel like we could do some fancy checking on `propTypes` of the native component and break the build if style is not there, wdyt?

shaundon: We probably could, but I feel like that should be done in a further PR so as not to conflate this one

dangerfile.js
```diff
@@ -71,3 +73,27 @@ const snapshotsModified = fileChanges.some(filePath => (
...
+// New files should include the Backpack license heading.
+const unlicensedFiles = createdFiles.filter((filePath) => {
+  if (filePath.match(/\.(js|css|scss|sh)/)) {
+    const fileContent = fs.readFileSync(filePath);
+    return !fileContent.includes('Licensed under the Apache License, Version 2.0 (the "License")');
+  }
+  return false;
+});
+if (unlicensedFiles.length > 0) {
+  fail(`These new files are do not include the license heading: ${unlicensedFiles.join(', ')}`);
```
matteo-hertel: what if we're adding a new file in e.g. new file in the dist folder in `bpk-svgs`?

matteo-hertel: oh merged already! I guess we'll find out 😄

shaundon: Hmm yeah good point. Maybe it should have just been a `warn` in the first place.

matteo-hertel: I'm adding some files in, if it's breaking something we'll fix it

shaundon: So I think we have two options here:

1) Change `fail` to `warn`.
2) Change Danger to disregard files inside `dist` folders.

Option 1 is my preference - I see it like the changelog where Danger will be right most of the time but sometimes we disregard the warning for good reason.

@matteo-hertel @matthewdavidson @georgegillams thoughts?

PS this is already merged so any changes will come in another one.

matthewdavidson: It should be impossible to add a new file without the license header. If there are directories of files that we need to ignore then that's fine (analogous to `.gitignore` / `.eslintignore`)

matthewdavidson: If we only have danger emitting warnings then eventually they'll become visual noise - humans will ignore them and eventually they'll cease to be of use.

georgegillams: Yeah I think we should disregard anything autogenerated (which I think is just inside `dist`??)

shaundon: https://github.com/Skyscanner/backpack/pull/144 sorts this out

PR #145, opened 2017-09-27
[BPK-903] Add correct unit formatting

matthewdavidson:
We now rely on token meta data to apply correct formatting on the docs site. This requires that the token `type` is actually correct. Some were not, so i fixed them.

+ [ ] For any new web components I've made sure that the style can be extended by the consumer using a `className` override.
+ [ ] For any new native components I've made sure that the style can be extended by the consumer using a `style` override.

### DIFF
--- a/packages/bpk-docs/src/helpers/tokens-helper.js
+++ b/packages/bpk-docs/src/helpers/tokens-helper.js
@@ -19,10 +19,14 @@
 import union from 'lodash/union';
 import kebabCase from 'lodash/kebabCase';
 
+// We don't set the root font size in the backpack base stylesheet, which means that the root font size falls back to
+// the browser default - typically 16px;
+const ROOT_FONT_SIZE = 16;
+
 export const formatTokenName = name => kebabCase(name);
 
 export const toPx = (value) => {
-  const parsed = parseFloat(value) * 16;
+  const parsed = parseFloat(value) * ROOT_FONT_SIZE;
   return parsed ? `${parsed}px` : null;
 };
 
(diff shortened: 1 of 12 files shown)


packages/bpk-docs/src/helpers/tokens-helper.js
```diff
@@ -19,32 +19,56 @@
...
-    parsed = parseFloat(value.replace(/rem/, '')) * 16;
-  }
-
-  if (/%$/.test(value)) {
-    parsed = parseFloat((value.replace(/%/, '')) / 100) * 16;
-  }
+export const formatTokenName = name => kebabCase(name);
 
+export const toPx = (value) => {
+  const parsed = parseFloat(value) * 16;
```
shaundon: `16` is here because that's 1rem right? I'm wondering if that merits a const/comment just to make it totally clear.

k0nserv: The best solution would be to have a token for this right? `BASE_FONT_SIZE` or so?

matthewdavidson: Not really - that would mean we introduce a token specifically for converting rem's to pixels in our docs site. It'll be used nowehre else. Plus we already have `FONT_SIZE_ROOT` with is set to `100%` (see https://github.com/Skyscanner/backpack/blob/master/packages/bpk-tokens/src/web/base/typography.json#L41)

k0nserv: But this has a relation to all our rem values. If we change the base font size to say 18px all our rem values have to be recalculated

matthewdavidson: If we do do that (which I doubt we will) then we'll notice that all the values are incorrect in the docs and we'll update this constant

k0nserv: Fair enough then :)

PR #148, opened 2017-09-28
[BPK-939] Upgrade storybook to support setting root

georgegillams:


package.json
```diff
@@ -50,9 +50,9 @@
   "author": "Backpack Design System <backpack@skyscanner.net>",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@storybook/addon-actions": "^3.1.9",
-    "@storybook/addon-links": "^3.1.6",
-    "@storybook/react": "^3.1.9",
+    "@storybook/addon-actions": "^3.2.11",
```
matthewdavidson: Eye spy with my little eye that we have some duplication of dependencies on the go.

If the root `package.json` lists `@storybook/addon-actions` & `@storybook/addon-links` and our `native/package.json` lists the exact same package versions then we can get rid of them in the native one.

georgegillams: Yup that seems to all work fine.
But I removed the duplicates from `native/package.json`. Were you expecting me to remove these instead??

georgegillams: Done 😸

PR #149, opened 2017-09-29
[BPK-841] Add rn icon component

georgegillams:


native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const characterCode = iconMappings[iconName];
+
+  const textStyleFinal = [styles.icon, small && styles.small];
+  textStyleFinal.push({ color });
+  if (style) {
+    textStyleFinal.push(style);
+  }
+
+  return (<Text style={textStyleFinal} {...rest} >{String.fromCharCode(parseInt(characterCode, 16))}</Text>);
```
matteo-hertel: any particular reason for nor using BpkText?

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -0,0 +1,83 @@
...
+
+BpkIcon.propTypes = {
+  iconName: iconNamePropType,
+  color: PropTypes.string.isRequired, // TODO Color Prop Type???
+  small: PropTypes.bool,
+  style: View.propTypes.style,
+};
+
+BpkIcon.defaultProps = {
+  iconName: '',
```
matteo-hertel: with am empty string as default it'll Throw an error if no icon is passed, is this intended behaviour?

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -19,17 +19,12 @@
...
   StyleSheet,
 } from 'react-native';
 import React from 'react';
 import PropTypes from 'prop-types';
-
-const tickSmIcon = require('./icons/sm/tick-circle.png'); // eslint-disable-line import/no-unresolved
-const tickLgIcon = require('./icons/lg/tick-circle.png'); // eslint-disable-line import/no-unresolved
-const exclaimationSmIcon = require('./icons/sm/exclaimation-circle.png'); // eslint-disable-line import/no-unresolved
-const exclaimationLgIcon = require('./icons/lg/exclaimation-circle.png'); // eslint-disable-line import/no-unresolved
+import BpkIcon from 'react-native-bpk-component-icon';
```
matteo-hertel: I'm assuming this PR is introducing both the component and the implementation in BpkTextInput, is that correct?

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
georgegillams: Good point

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
matthewdavidson: Can we extract the encoding to a separate function?

georgegillams: I chose `Text` as our usage is very specific and not really related to typesetting. Any thoughts?

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -0,0 +1,83 @@
...
+
+const BpkIcon = (props) => {
+  const { iconName, color, small, style, ...rest } = props;
+
+  const characterCode = iconMappings[iconName];
+
+  const textStyleFinal = [styles.icon, small && styles.small];
+  textStyleFinal.push({ color });
+  if (style) {
+    textStyleFinal.push(style);
```
matthewdavidson: this means consumers can do this:
```
<BpkIcon color={TOKENS.colorWhite} style={{ color: TOKENS.colorBlue500 }} />
```
What should win?

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -0,0 +1,83 @@
...
+};
+
+const {
+  spacingBase,
+  spacingLg,
+} = tokens;
+
+const styles = StyleSheet.create({
+  icon: {
+    width: spacingLg,
```
matthewdavidson: Why is width being set here? (not saying it's wrong, im just trying to understand these styles)

package.json
```diff
@@ -27,7 +27,7 @@
     "jest": "jest --coverage",
     "build:tokens": "lerna run tokens --scope bpk-tokens",
     "build:svgs": "lerna run svgs --scope bpk-svgs",
-    "build": "npm run build:tokens && npm run build:svgs && lerna run build",
+    "build": "npm run build:tokens && npm run build:svgs && lerna run build && cd native/ && react-native link && cd -",
```
matthewdavidson: surely `react-native link` only has to be run once?

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
georgegillams: Yeah - I implemented its use where appropriate. It does mean I've somewhat bloated the PR, but I wanted to make the changes in order to test it and I thought I might as well include what I'd done

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
matthewdavidson: Yeah no I think your right to use `Text` here - it's a completely different use case.

We want consumers to use BpkText whenever they are displaying text on the screen to users. The icon component isn't doing that. 

Rule of thumb is that if you find yourself having to override the styles of a backpack component, then you probably shouldnt be using it.

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
matthewdavidson: Yeah actually we have a separate card for consuming the BpkIcon component in other Bpk components - i'd take the BpkTextInput & BpkButton changes out of this MR - it's too early.

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
georgegillams: If we don't set width, then the text component will automatically be "1 character" wide, which doesn't work very well with wider icons

screen shot 2017-09-29 at 16 28 42

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
georgegillams: I'm not sure who should win here. It's a bit of an edge case

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
georgegillams: Done

package.json
georgegillams: It should be run every-time a new font file is generated. If we add an icon, generate a new BpkIcon.ttf, and don't run `react-native link` then the new font will not be made available to our storybook app

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
georgegillams: @matthewdavidson also made a good point that if we make use of `BpkIcon` before it is finished, we will actually break the existing react-native components as the font won't yet be installed in the apps

package.json
matthewdavidson: Ah ok - then:
```
cd native/ && react-native link && cd -
```
can be simplified to:
```
(cd native && react-native link)
```

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+  if (style) {
+    textStyleFinal.push(style);
+  }
+
+  return (<Text style={textStyleFinal} {...rest} >{mapCharacterCode(characterCode)}</Text>);
+};
+
+BpkIcon.propTypes = {
+  iconName: iconNamePropType,
```
matthewdavidson: no need for this custom prop type, it can be simplified to:
```
PropTypes.oneOf(Object.keys(iconMappings))
```

matthewdavidson: Also, any reason why not just `icon`?

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -0,0 +1,80 @@
...
+
+  if (style) {
+    textStyleFinal.push(style);
+  }
+
+  return (<Text style={textStyleFinal} {...rest} >{mapCharacterCode(characterCode)}</Text>);
+};
+
+BpkIcon.propTypes = {
+  icon: PropTypes.oneOf(Object.keys(iconMappings)),
```
shaundon: Could we make this a required prop? (There might be a reason why we aren't doing that, just checking)

matteo-hertel: nice one, is supposed to be required, I'll update now

shaundon: Cool - you'll be able to remove the `defaultProps` entry then too :)

PR #151, opened 2017-09-29
[no-jira] Make Danger licence check ignore snapshots

shaundon:


shaundon:
@georgegillams who doesn't love a one character PR!

PR #152, opened 2017-09-29
[BPK-996] add fixed timestamp for font generation

matteo-hertel:


packages/bpk-svgs/gulpfile.babel.js
```diff
@@ -283,6 +283,7 @@ gulp.task('icons-font', ['icons-common'], () => {
        */
       normalize: true,
       fontHeight: 1001,
+      timestamp: 1436442578, // Thu Jul 9 12:49:38 2015 GMT+1
```
matthewdavidson: Haha love it. I reckon you'll need to put a better comment than that tho.

Perhaps: `A static timestamp to prevent changes showing up in git`

matteo-hertel: fair enough 😄

matteo-hertel:
backpack's first commit!

shaundon:
❤️

packages/bpk-svgs/gulpfile.babel.js
```diff
@@ -283,6 +283,7 @@ gulp.task('icons-font', ['icons-common'], () => {
        */
       normalize: true,
       fontHeight: 1001,
+      timestamp: 1436442578, // A static timestamp to prevent changes showing up in git, backpack's first commit!
```
k0nserv: Could you please expand this comment here and also add some more description of this in commit message? It's not clear why this is needed

matteo-hertel: It's to prevent git from thinking it's a changed file, if timestamp is not defined the file itself will have the current timestamp and during the build the  file's timestamp will change, what do you think would be a better message?

georgegillams: tbh I think the message is ok for now, and it's blocking me so can we just merge @k0nserv

PR #160, opened 2017-10-04
add gradle/pod

matteo-hertel:


native/packages/react-native-bpk-component-icon/BpkSvgsFont.podspec
```diff
@@ -0,0 +1,16 @@
...
+
+Pod::Spec.new do |s|
+
+  s.name           = "BpkSvgsFont"
+  s.version        = version
+  s.summary        = "Backpack's collection of SVGs."
+  s.homepage       = "https://github.com/Skyscanner/backpack/tree/master/packages/react-native-bpk-component-icon"
+  s.license        = "Apache License 2.0"
+  s.author         = { "Backpack" => "backpack@skyscanner.net" }
+  s.resources      = "./node_modules/bpk-svgs/dist/font/*.ttf"
```
georgegillams: I might be wrong, but the way I saw this working is that the scripts will point to the `.ttf`s in the`react-native-bpk-component-icon` package, and the icon package will have a dependency on a specific version of `bpk-svgs`

k0nserv: This seems correct given that it does work :)

matthewdavidson: Note: this _could_ break if consumers install a _different_ version of `bpk-svgs` than the one depended on by `react-native-bpk-component-icon` because npm will hoist the top level dependency to that path over the dependency specified by the icon component. E.g.:
```
node_modules/
    react-native-bpk-component-icon/
    bpk-svgs/
```

User runs `npm install bpk-svgs --save-dev`, which results in:
```
node_modules/
    react-native-bpk-component-icon/
        node_modules/
            bpk-svgs/
    bpk-svgs/ // <- our podspec now points to the newer version which could be breaking
```

matteo-hertel: not sure we can do anything about it, although the icon component is importing both the font and the mapping, so even if an old version is added as devDeps the icon component will not "break"

k0nserv: Yeah I was concerned about the implications of this too. I guess we'll see how it behaves in the wild

matthewdavidson: No you are correct the icon component will be fine - just the rendering might be affected because the app manifest points to the wrong `.ttf`. I reckon it's a low risk - why would consumers depend on `bpk-svgs` directly?

native/packages/react-native-bpk-component-icon/readme.md
```diff
@@ -8,9 +8,20 @@
...
 
-<!-- // TODO ADD INSTRUCTIONS FOR ADDING THE PODSPEC / GRADLE SCRIPT TO THE iOS / Android PROJECT IN WHICH THIS IS NOW BEING USED! -->
+## Native usage
 
-## Usage
+### IOS
+
+Add the following to your `Podfile` and run `pod update`:
+
+`pod 'BpkIcons', :path => '../react-native/node_modules/react-native-bpk-component-icon'`
```
matteo-hertel: @k0nserv could you please double check this?

native/packages/react-native-bpk-component-icon/BpkSvgsFont.podspec
```diff
@@ -0,0 +1,16 @@
+require 'json'
+version = JSON.parse(File.read('package.json'))["version"]
+
+Pod::Spec.new do |s|
+
```
k0nserv: Can we remove this newline?

native/packages/react-native-bpk-component-icon/BpkSvgsFont.podspec
```diff
@@ -0,0 +1,16 @@
...
+  s.name           = "BpkSvgsFont"
+  s.version        = version
+  s.summary        = "Backpack's collection of SVGs."
+  s.homepage       = "https://github.com/Skyscanner/backpack/tree/master/packages/react-native-bpk-component-icon"
+  s.license        = "Apache License 2.0"
+  s.author         = { "Backpack" => "backpack@skyscanner.net" }
+  s.resources      = "./node_modules/bpk-svgs/dist/font/*.ttf"
+  s.preserve_paths = "**/*.js"
+  s.dependency 'React'
+
```
k0nserv: and this one

native/packages/react-native-bpk-component-icon/readme.md
k0nserv: I think the pod name should match what you call the pod here. i.e if it's `BpkIcons` here then the podspec should be `BpkIcons.podspec` and the name in the podspec should be `BpkIcons`

matteo-hertel: nice one, I thought I renamed it, and I obviously didn't!

matthewdavidson: Why:
```
'../react-native/node_modules/react-native-bpk-component-icon'
```
and not:
```
'node_modules/react-native-bpk-component-icon'
```

I.e. surely it depends on the consumer's project to determine the correct path?

(i'm referencing https://github.com/oblador/react-native-vector-icons#option-with-cocoapods)

matteo-hertel: I followed the same logic as the other RN components used in the app, both IOS and Android have a react-native folder in the root so that's why the `../`

k0nserv: Yeah but that's specific to our app, we shouldn't be encoding that here IMO

matteo-hertel: ok no problem, I'll amend that

PR #161, opened 2017-10-04
[BPK-1011] - fix ticket notch safari display issue

matthewdavidson:
**Before:** 
before
**After:**
after


packages/bpk-component-ticket/src/bpk-ticket.scss
```diff
@@ -173,7 +173,8 @@
       display: block;
       width: $notch-size * 4;
       height: $notch-size * 4;
-      border: $notch-size solid white;
+      transform: translate3d(0, 0, 0); // BPK-1011 fix rendering bug in safari 11
```
k0nserv: Can we link to an explanation for this fix?

PR #163, opened 2017-10-04
[BPK-953] Allow BpkButton to be themed

shaundon:
screen shot 2017-10-04 at 15 21 47

### DIFF
--- a/native/packages/react-native-bpk-component-button/package.json
+++ b/native/packages/react-native-bpk-component-button/package.json
@@ -15,8 +15,10 @@
   },
   "dependencies": {
     "bpk-tokens": "^26.3.0",
+    "lodash": "^4.17.4",
     "prop-types": "^15.5.8",
     "react-native-bpk-component-text": "^2.1.9",
+    "react-native-bpk-theming": "^1.0.0",
     "react-native-linear-gradient": "^2.3.0"
   }
 }
--- a/native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
+++ b/native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
@@ -172,12 +172,27 @@ const gradientColors = {
   disabled: [tokens.colorGray100, tokens.colorGray100],
 };
 
+const themeMappings = {
+  text: {
+    color: {
+      primary: 'buttonPrimaryTextColor',
+    },
+  },
+  gradient: {
+    primary: {
+      startColor: 'buttonPrimaryGradientStartColor',
+      endColor: 'buttonPrimaryGradientEndColor',
+    },
+  },
+};
+
 const styles = {
   base,
   types,
   modifiers,
   gradientColors,
   underlayColor,
+  themeMappings,
 };
 
 export default styles;
--- a/native/packages/react-native-bpk-component-button/src/BpkButton.js
+++ b/native/packages/react-native-bpk-component-button/src/BpkButton.js
@@ -23,7 +23,9 @@
  import React from 'react';
  import PropTypes from 'prop-types';
  import LinearGradient from 'react-native-linear-gradient';
+ import xor from 'lodash/xor';
 
+ import { withTheme } from 'react-native-bpk-theming';
  import BpkText from 'react-native-bpk-component-text';
 
  import styles from './BpkButton-styles';
@@ -64,8 +66,24 @@
    return styleForElement;
  };
 
- const getGradientColors = ({ type, disabled, selected }) => {
-   let gradientColors = styles.gradientColors[type];
+ const getThemingForElement = (elementType, theme, { type, selected, disabled }) => {
+   const themeForElement = {};
+   if (theme && !selected && !disabled && styles.themeMappings[elementType]) {
+     Object.keys(styles.themeMappings[elementType]).forEach((key) => {
+       const values = styles.themeMappings[elementType][key];
+       if (values[type]) {
+         themeForElement[key] = theme[values[type]];
+       }
+     });
+   }
+   return themeForElement;
+ };
+
+ const getGradientColors = (theme, { type, disabled, selected }) => {
+   let gradientColors = theme ?
+   [theme.buttonPrimaryGradientStartColor, theme.buttonPrimaryGradientStartColor] :
+   styles.gradientColors[type];
+
    if (selected) {
      gradientColors = styles.gradientColors.selected;
    }
@@ -97,9 +115,29 @@
      style,
      ...rest
    } = props;
+   let theme = props.theme;
 
+   // Validate the button type.
    if (!BUTTON_TYPES.includes(type)) {
      throw new Error(`"${type}" is not a valid button type. Valid types are ${BUTTON_TYPES.join(', ')}`);
+   }
+
+   // Validate that correct theming attributes have been provided.
+   if (theme) {
+     // Must provide disabled and selected states regardless of theme.
+     const attributesToSupply = [
+       // TODO this should support multiple types in future with some logic.
+       styles.themeMappings.text.color.primary,
+       styles.themeMappings.gradient.primary.startColor,
+       styles.themeMappings.gradient.primary.endColor,
+     ];
+
+     // If all attributes were not supplied, turn off theming.
+     // TODO allow all types to be themed.
+     const allAttributesSupplied = xor(attributesToSupply, Object.keys(theme)).length === 0;
+     if (!allAttributesSupplied || type !== 'primary') {
+       theme = null;
+     }
    }
 
    const accessibilityTraits = ['button'];
@@ -110,7 +148,10 @@
    // Note that TouchableHighlight isn't on Android, so TouchableFeedback
    // will need to be used to support it.
    return (
-     <LinearGradient style={[getStyleForElement('container', props), style]} colors={getGradientColors(props)}>
+     <LinearGradient
+       colors={getGradientColors(theme, props)}
+       style={[getStyleForElement('container', props), style]}
+     >
        <TouchableHighlight
          style={getStyleForElement('button', props)}
          disabled={disabled}
@@ -127,7 +168,7 @@
              <BpkText
                textStyle={large ? 'lg' : 'sm'}
                emphasize
-               style={getStyleForElement('text', props)}
+               style={[getStyleForElement('text', props), getThemingForElement('text', theme, props)]}
              >
                {title}
              </BpkText>
@@ -139,7 +180,7 @@
    );
  };
 
- BpkButton.propTypes = {
+ const propTypes = {
    title: PropTypes.string.isRequired,
    icon: iconPropType,
    iconOnly: PropTypes.bool,
@@ -150,7 +191,14 @@
    selected: PropTypes.bool,
    accessibilityLabel: PropTypes.string,
    style: View.propTypes.style,
+   theme: PropTypes.shape({
+     gradientStartColor: PropTypes.string.isRequired,
+     gradientEndColor: PropTypes.string.isRequired,
+     textColor: PropTypes.string.isRequired,
+   }),
  };
+
+ BpkButton.propTypes = propTypes;
 
  BpkButton.defaultProps = {
    icon: null,
@@ -161,7 +209,11 @@
    selected: false,
    accessibilityLabel: null,
    style: null,
+   theme: null,
  };
 
- export default BpkButton;
- export { BUTTON_TYPES };
+ export default withTheme(BpkButton);
+ export {
+   propTypes,
+   BUTTON_TYPES,
+  };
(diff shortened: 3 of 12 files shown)


native/packages/react-native-bpk-component-button/package.json
```diff
@@ -15,8 +15,10 @@
   },
   "dependencies": {
     "bpk-tokens": "^26.2.0",
+    "lodash": "^4.17.4",
```
shaundon: For `xor`.

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -168,12 +168,37 @@ const gradientColors = {
...
+      startColor: 'disabledBackgroundColor',
+      endColor: 'disabledBackgroundColor',
+    },
+    selected: {
+      startColor: 'selectedGradientStartColor',
+      endColor: 'selectedGradientEndColor',
+    },
+  },
+};
+
```
shaundon: I'm concerned that I've fallen into an over-engineering trap here. However, we do need a place where theme attributes are translated to the element they apply to. So on line 174, we're saying that the `color` style property for the `text` component uses the `textColor` theme attribute when the type is primary.

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -64,13 +66,35 @@
...
+         themeForElement[key] = theme[values.disabled];
+       } else if (selected && theme[values.selected]) {
+         themeForElement[key] = theme[values.selected];
+       } else if (values[type]) {
+         themeForElement[key] = theme[values[type]];
+       }
+     });
+   }
+   return themeForElement;
+ };
```
shaundon: Here we iterate over the mappings (set in `BpkButton-styles.js`) and apply them to the component, based on disabled, selected status etc.

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -64,13 +66,35 @@
...
-     gradientColors = styles.gradientColors.selected;
+     gradientColors = theme ?
+     [theme.selectedGradientStartColor, theme.selectedGradientEndColor] : styles.gradientColors.selected;
    }
    if (disabled) {
-     gradientColors = styles.gradientColors.disabled;
+     gradientColors = theme ?
+     [theme.disabledBackgroundColor, theme.disabledBackgroundColor] : styles.gradientColors.disabled;
    }
    return gradientColors;
```
shaundon: Once we support more types this will need to become more complex, but for now it's sufficient.

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -161,7 +225,13 @@
    selected: false,
    accessibilityLabel: null,
    style: null,
+   theme: null,
  };
 
+ const BpkButtonThemed = withTheme(BpkButton);
```
shaundon: Originally I wanted to just export only `BpkButton` and have it take theming if it's inside a `BpkThemeProvider`. This doesn't work, because `withTheme` throws an error if it's not inside `BpkThemeProvider`, so it won't work for non-themed buttons.

Assuming we can't find a way around this, I think we should standardise on how to export themed variants of components.

native/packages/react-native-bpk-component-button/stories.js
```diff
@@ -100,92 +101,109 @@ ArrowImage.defaultProps = {
...
-        iconOnly
-        onPress={action(`${type} icon only button clicked`)}
-        style={styles.buttonStyles}
-      />
+  selectedGradientStartColor: '#FF5722',
+  selectedGradientEndColor: '#E64A19',
+  selectedTextColor: '#FFEB3B',
+
+  disabledBackgroundColor: '#B2FF59',
+  disabledTextColor: '#b71c1c',
```
shaundon: See screenshot attached to PR for how these look. Right now I took them Material, but I anticipate that they'll be changed based on Design's feedback and possibly tokenised.

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -150,6 +204,16 @@
    selected: PropTypes.bool,
    accessibilityLabel: PropTypes.string,
    style: View.propTypes.style,
+   theme: PropTypes.shape({
+     gradientStartColor: PropTypes.string.isRequired,
```
k0nserv: I think we might want a custom prop validation here that looks at the type of the button before determining what is required

jamesf3rguson: @k0nserv @shaundon Yes, lets remove the ability to theme disabled styles. Sorry my bad as its on the card.

jamesf3rguson: I'm also tempted to ditch the selected button state from native all together as its not being used and its a strange one which we should probably not be promoting. Maybe raise a separate card for this?

shaundon: Happy to remove both selected and disabled states - means less logic and fewer lines of code on our part 😄 

I think we should make a card to remove selected from the existing button, and in the meantime I'll remove selected and disabled from theming.

native/packages/react-native-bpk-component-button/package.json
matthewdavidson: I was gonna say we can rely on `lodash.xor`, but then I read [a comment from](https://github.com/lodash/lodash/issues/2385#issuecomment-221890679) the creator of lodash himself.

native/packages/react-native-bpk-component-button/src/__snapshots__/BpkButton-test.android.js.snap
```diff
@@ -209,13 +212,16 @@ exports[`Android BpkButton should render correctly with type="destructive" 1`] =
...
             },
             Array [
-              Object {
-                "backgroundColor": "transparent",
-                "color": "rgb(255, 255, 255)",
-              },
-              Object {
-                "color": "rgb(255, 84, 82)",
-              },
+              Array [
```
matthewdavidson: Is this correct? `style` now has `[{}, {}, [[{}, {}, {}]]]`? 😱

native/packages/react-native-bpk-component-button/stories.js
```diff
@@ -215,6 +222,14 @@ storiesOf('BpkButton', module)
...
     </View>
   ))
+  .add('docs:withTheme', () => (
+    <BpkThemeProvider theme={themeAttributes}>
+      <ScrollView>
+        <BpkText textStyle="xxl">Primary</BpkText>
+        {generateButtonStoryForType('primary')}
+      </ScrollView>
+    </BpkThemeProvider>
+  ))
```
shaundon: Notice that it now uses the same method as other button types - the only difference is it's wrapped in `BpkThemeProvider` 😄

shaundon:
I've got theming working with a single component now. Previously we had `BpkButton` and `BpkButtonThemed`, now we just have `BpkButton`.

To achieve this, I wrote our own versions of `withTheme` and `createThemeListener`. They're essentially the same as the one in `theming`, just they don't error when no theme context is present.

What do people think? @matteo-hertel and I discussed it and think it's probably the best solution (of a bad bunch), where the other options are fork it and use a different library respectively.

native/packages/react-native-bpk-component-button/package.json
```diff
@@ -15,8 +15,10 @@
   },
   "dependencies": {
     "bpk-tokens": "^26.3.0",
+    "lodash": "^4.17.4",
     "prop-types": "^15.5.8",
     "react-native-bpk-component-text": "^2.1.9",
+    "react-native-bpk-theming": "^0.0.1",
```
georgegillams: Seems that this should be `1.0.2`?

native/packages/react-native-bpk-component-button/stories.js
```diff
@@ -215,6 +222,14 @@ storiesOf('BpkButton', module)
       {generateButtonStoryForType('featured')}
     </View>
   ))
+  .add('docs:withTheme', () => (
+    <BpkThemeProvider theme={themeAttributes}>
+      <ScrollView>
```
georgegillams: This puts the themed buttons at the top of the page whereas other buttons pages are centred vertically.

native/packages/react-native-bpk-component-button/stories.js
georgegillams: 👍

native/packages/react-native-bpk-component-button/stories.js
shaundon: Thanks, now fixed.

native/packages/react-native-bpk-theming/src/BpkThemeProvider.js
```diff
@@ -16,11 +16,70 @@
  * limitations under the License.
  */
 
- import { ThemeProvider, withTheme } from 'theming';
+ import React from 'react';
+ import PropTypes from 'prop-types';
+ import { ThemeProvider } from 'theming';
+
+ // This is the same as https://github.com/iamstarkov/theming/blob/master/src/create-theme-listener.js
```
matteo-hertel: I think it's worth discuss with @matthewdavidson and @k0nserv about this approach, I personally think is the best compromise of going around an issue that the 3rd party library is causing without forking the library or changing to something else, and if the main library get update we can remove this

What do you guys think?

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -168,12 +168,27 @@ const gradientColors = {
   disabled: [tokens.colorGray100, tokens.colorGray100],
 };
 
+const themeMappings = {
+  text: {
+    color: {
+      primary: 'textColor',
```
matthewdavidson: I think the names should be a lot more button specific:
```
buttonPrimaryTextColor
buttonPrimaryGradientStartColor
buttonPrimaryGradientEndColor
```

native/packages/react-native-bpk-theming/src/BpkThemeProvider.js
```diff
@@ -16,11 +16,70 @@
...
+ // This is the same as https://github.com/iamstarkov/theming/blob/master/src/create-with-theme.js
+ // but using the custom themeListener from above.
+ const createWithTheme = () => {
+   const themeListener = createThemeListener();
+   const getDisplayName = Component => (
+     Component.displayName || Component.name || 'Component'
+   );
+   return Component => (
+     class WithTheme extends React.Component {
+       static displayName = `WithTheme(${getDisplayName(Component)})`;
```
k0nserv: This is a HoC so it should be `wtihTheme` not `WithTheme`

native/packages/react-native-bpk-theming/src/BpkThemeProvider.js
```diff
@@ -16,11 +16,70 @@
...
+     contextTypes,
+     initial,
+     subscribe,
+   };
+ };
+
+ // This is the same as https://github.com/iamstarkov/theming/blob/master/src/create-with-theme.js
+ // but using the custom themeListener from above.
+ const createWithTheme = () => {
+   const themeListener = createThemeListener();
```
k0nserv: In `theming` the pass a channel to the themeListener. Should we also be doing this?

shaundon: I copied it from the original library, but I can change it sure.

native/packages/react-native-bpk-theming/src/BpkThemeProvider.js
shaundon: I copied it from the original library, but I can change it sure.

native/packages/react-native-bpk-theming/src/BpkThemeProvider.js
shaundon: So, I looked into it and it just always gets defaulted to `__THEMING__`. So I decided to just hard code that in, instead.

native/packages/react-native-bpk-theming/src/BpkThemeProvider.js
```diff
@@ -16,11 +16,70 @@
...
  */
 
- import { ThemeProvider, withTheme } from 'theming';
+ import React from 'react';
+ import PropTypes from 'prop-types';
+ import { ThemeProvider } from 'theming';
+
+ // This is the same as https://github.com/iamstarkov/theming/blob/master/src/create-theme-listener.js
+ // but returns null instead of an error when context[CHANNEL] is undefined.
+ const createThemeListener = () => {
```
matthewdavidson: Can we split these functions into separate files, similar to the upstream example?

native/packages/react-native-bpk-theming/src/BpkThemeProvider.js
matthewdavidson: I agree with this approach 👍

k0nserv:
See https://github.com/iamstarkov/theming/issues/13 about the errors. We should raise this workaround in an issue I guess

PR #165, opened 2017-10-04
[No_Jira] update story and doc based on feedback

matteo-hertel:
Update the icon storybook and the docs based on the feedback received from the design point of view


native/packages/react-native-bpk-component-icon/stories.js
```diff
@@ -18,7 +18,7 @@
 
 import React from 'react';
 import { storiesOf } from '@storybook/react-native';
-import { Platform, View, StyleSheet } from 'react-native';
+import { Platform, View, StyleSheet, Text } from 'react-native';
```
georgegillams: Can we use `BpkText` for the story titles?
I realise that's kinda ironic as you made the exact opposite comment to me the other day 😆

matteo-hertel: Point taken, in this one I don't have BpkText as devDependency, in your bit you had it 😄

native/packages/react-native-bpk-component-icon/stories.js
```diff
@@ -42,13 +47,102 @@ const styles = StyleSheet.create({
...
+      />
+      <BpkIcon
+        style={styles.singleIcon}
+        icon="hotels"
+      />
+    </View>
+  </View>);
+}
+
+function getColouredIcons() {
```
k0nserv: Any particular reason we aren't using an arrow function here?

matteo-hertel: none, i didn't felt the need for it! it'll be more consistent with the rest of the file tough so I'll update, thanks

native/packages/react-native-bpk-component-icon/stories.js
```diff
@@ -42,13 +48,96 @@ const styles = StyleSheet.create({
...
+  bottomMargin: {
+    marginBottom: spacingSm,
   },
 });
 
+/*
+ * the following functions should go on the bottom
+ * but eslint complains about it -_-
+ * @return
+ */
```
matthewdavidson: > the following functions should go on the bottom but eslint complains about it -_-

Why should those functions go on the bottom?

matthewdavidson: Otherwise i'd remove that comment :).

matteo-hertel: oh good catch, well I was using regular functions before and es-lint was complaining about them being declared after the use, function hoisting is not a thing according to es-lint 😄

PR #166, opened 2017-10-05
[NO-JIRA] Remove `.enable-jp-font-feature-settings` class

matthewdavidson:
We added this a while back and AFAIK it's not being used anymore (Its been cleaned up in the only known internal usage).


k0nserv:
This isn't a depreaction. It's a removal

matthewdavidson:
@k0nserv fixed.

PR #173, opened 2017-10-06
Bpk 937 touchable highlight

matthewdavidson:
Added iOS only `UNDERLAY_COLOR` & `UNDERLAY_OPACITY` tokens and consumed them in the button component.

### DIFF
--- a/native/packages/react-native-bpk-theming/src/BpkThemeProvider-test.common.js
+++ b/native/packages/react-native-bpk-theming/src/BpkThemeProvider-test.common.js
@@ -36,6 +36,7 @@ const commonTests = () => {
     });
 
     it('should error when theme is not provided', () => {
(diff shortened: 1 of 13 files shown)


matthewdavidson:
**Note:** It's interesting how none of the button tests were updated as part of this. `TouchableHighlight` doesn't show up in snapshot output 🤔

native/packages/react-native-bpk-theming/src/BpkThemeProvider-test.common.js
```diff
@@ -36,6 +36,7 @@ const commonTests = () => {
     });
 
     it('should error when theme is not provided', () => {
+      jest.spyOn(console, 'error').mockImplementation(() => jest.fn());
```
shaundon: ❤️ this

PR #174, opened 2017-10-06
[BPK-1006] Add Android elevation tokens

matteo-hertel:


matteo-hertel:
The elevation changed from the hardcoded 3 to elevationSm (4), not a huge difference but still different

shaundon:
@k0nserv I'm pretty sure they are different!

matteo-hertel:
There is a small change that needs to be made so don't merge it

matteo-hertel:
The card elevation has been updated to elevationXs (2 dp)

PR #175, opened 2017-10-06
[BPK-1016] Add radii tokens

matthewdavidson:


shaundon:
Diff looks good, although Travis is failing on a test. I imagine it's just a snapshot that needs updating?

packages/bpk-tokens/src/react-native-output-test.js
```diff
@@ -32,6 +32,8 @@ const ANDROID_ONLY_PROPS = [
   'ELEVATION_XL',
 ];
 const IOS_ONLY_PROPS = [
+  'BORDER_RADIUS_SM',
+  'BORDER_RADIUS_PILL_LG',
```
matteo-hertel: This one should be  `BORDER_RADIUS_PILL` I think!

matthewdavidson: Good catch 👍 I actually just needed to remove both of them from iOS only props - I negotiated with James to consolidate them. 😇

PR #177, opened 2017-10-06
[BPK-1017] iOS shadow tokens

matteo-hertel:


native/packages/react-native-bpk-component-card/src/__snapshots__/BpkCard-test.android.js.snap
```diff
@@ -27,13 +27,13 @@ exports[`Android BpkCard should render correctly 1`] = `
           Object {
             "backgroundColor": "rgb(255, 255, 255)",
             "elevation": 2,
-            "shadowColor": "rgb(37, 32, 51)",
+            "shadowColor": undefined,
```
matthewdavidson: This doesn't look good - perhaps we should conditionally apply these styling props based on `Platform.os`? 

Saying that I reckon we probably do this another places ... thoughts?

packages/bpk-tokens/src/ios/base/aliases.json
```diff
@@ -28,6 +28,18 @@
     "SPACING_XXL": 40,
     "BORDER_SIZE_SM": 1,
     "BORDER_SIZE_LG": 2,
-    "BORDER_SIZE_XL": 3
+    "BORDER_SIZE_XL": 3,
+    "SHADOW_OFFSET_HEIGHT_SM": 1,
```
matthewdavidson: Hmm, so you have opted for {TOKEN_NAME}_{PROPERTY_NAME}_{MODIFIER}`

Interestingly we use `{TOKEN_NAME}_{MODIFIER}_{PROPERTY_NAME}` elsewhere (see text tokens https://github.com/Skyscanner/backpack/blob/master/packages/bpk-tokens/tokens/ios/base.raw.json#L1028).

matthewdavidson: We use `{TOKEN_NAME}_{MODIFIER}_{PROPERTY_NAME}` with buttons too, see https://github.com/Skyscanner/backpack/blob/master/packages/bpk-tokens/src/web/base/buttons.json

native/packages/react-native-bpk-component-card/src/__snapshots__/BpkCard-test.android.js.snap
matteo-hertel: I think we could go with one of three ways:
1. create a getter function that based on the OS returns the correct properties
2. have a default for everything if a value is not passed through e.g. `{shadowColor: token || defaultValue}`
3. leave it as it is, Android will ignore that property anyway as it's iOS only (https://facebook.github.io/react-native/docs/shadow-props.html#shadowcolor)

wdyt?

packages/bpk-tokens/src/ios/base/aliases.json
matteo-hertel: I'll update to be consistent with the rest of the tokens!

native/packages/react-native-bpk-component-card/src/__snapshots__/BpkCard-test.android.js.snap
matthewdavidson: Yeah perhaps option 1 & 2 are doing more work than is actually required - especially given that they are ignored on android. Ok lets leave as as (option 3) :)

packages/bpk-tokens/src/ios/base/shadows.json
```diff
@@ -0,0 +1,82 @@
+{
+  "imports": [
+    "./aliases.json"
+  ],
+  "props": {
+    "SHADOW_COLOR_SM": {
+      "value": "{!GRAY_900}",
+      "type": "color",
+      "category": "shadowsettings"
```
matthewdavidson: This category field is used (amongst other things) to pull values into the tokens page (see web example with `"box-shadows"` category value http://backpack.prod.aws.skyscnr.com/tokens/shadows).

2 options:

1. Use `"box-shadows"` - the same as web
1. Use a different value (perhaps just `"shadows"` tho :P) and update the `ShadowsPage` (https://github.com/Skyscanner/backpack/blob/master/packages/bpk-docs/src/pages/ShadowsPage/ShadowsPage.js#L29) to include them

matteo-hertel: I like option one the most because is the same as web, unless we want to behave differently from native vs web, I'd keep it consistent; if we want to start to have different behaviour for RN option 2 will be a good starting point

matthewdavidson: Yeah I agree - lets go with option 1. 

Related to this, I think we should update the shadows page to include `"elevation"` to surface the android tokens?

PR #181, opened 2017-10-09
[BPK-1019] Add theming support for secondary button

shaundon:
screen shot 2017-10-09 at 13 44 33

In addition to secondary theming, this PR also includes better propTypes validation, so that a warning will be printed if:

a) The user is trying to theme an unthemeable button type
b) The user hasn't included all the necessary theme attributes for the type of button they're trying to theme.

### DIFF
--- a/native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
+++ b/native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
@@ -173,12 +173,22 @@ const themeMappings = {
   text: {
     color: {
       primary: 'buttonPrimaryTextColor',
+      secondary: 'buttonSecondaryTextColor',
+    },
+  },
+  button: {
+    borderColor: {
+      secondary: 'buttonSecondaryBorderColor',
     },
   },
   gradient: {
     primary: {
       startColor: 'buttonPrimaryGradientStartColor',
       endColor: 'buttonPrimaryGradientEndColor',
+    },
+    secondary: {
+      startColor: 'buttonSecondaryBackgroundColor',
+      endColor: 'buttonSecondaryBackgroundColor',
     },
   },
 };
--- a/native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
+++ b/native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
@@ -125,6 +125,25 @@ const commonTests = () => {
         iconOnly: true,
       }, 'icon', 'BpkButton').toString()).toEqual('Error: Invalid prop `icon` supplied to `BpkButton`. When `iconOnly` is enabled, `icon` must be supplied.'); // eslint-disable-line max-len
     });
+
+    it('should reject theme property when some theme attributes are omitted', () => { // eslint-disable-line max-len
+      expect(propTypes.theme({
+        type: 'primary',
+        theme: {},
+      }, 'theme', 'BpkButton').toString()).toEqual('Error: Invalid prop `theme` supplied to `BpkButton`. For buttons of type `primary`, the `theme` prop must include `buttonPrimaryTextColor, buttonPrimaryGradientStartColor, buttonPrimaryGradientEndColor`'); // eslint-disable-line max-len
+    });
+
+    it('should accept theme property when correct attributes are supplied', () => { // eslint-disable-line max-len
+      expect(propTypes.theme({
+        type: 'primary',
+        theme: {
+          buttonPrimaryGradientStartColor: 'red',
+          buttonPrimaryGradientEndColor: 'green',
+          buttonPrimaryTextColor: 'blue',
+        },
+      }, 'theme', 'BpkButton')).toBeFalsy();
+    });
+
     it('should throw an error for invalid button type', () => {
       jest.spyOn(console, 'error').mockImplementation(() => jest.fn());
       expect(() => renderer.create(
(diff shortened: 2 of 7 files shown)


native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
```diff
@@ -125,6 +125,32 @@ const commonTests = () => {
         iconOnly: true,
       }, 'icon', 'BpkButton').toString()).toEqual('Error: Invalid prop `icon` supplied to `BpkButton`. When `iconOnly` is enabled, `icon` must be supplied.'); // eslint-disable-line max-len
     });
+
+    it('should reject theme property when button type is not themeable', () => {
+      expect(propTypes.theme({
+        type: 'featured',
+        theme: {},
+      }, 'theme', 'BpkButton').toString()).toEqual('Error: Invalid prop `theme` supplied to `BpkButton`. `type` is `featured`, but themeable types are `primary, secondary`.'); // eslint-disable-line max-len
```
matthewdavidson: Hmm, i think this might be a little overzealous. Imagine a scenario where an app has multiple button types inside. A `<ThemeProvider theme={myTheme} />` is wrapped around the entire app. This would throw a warning when in fact the use case is valid.

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -173,13 +173,23 @@ const themeMappings = {
...
     },
   },
   gradient: {
     primary: {
       startColor: 'buttonPrimaryGradientStartColor',
       endColor: 'buttonPrimaryGradientEndColor',
     },
+    secondary: {
+      startColor: 'buttonSecondaryBackgroundColor',
+      endColor: 'buttonSecondaryBackgroundColor',
```
matthewdavidson: I understand the convenience here - ideally we wouldn't wrap the button in a linear gradient component if we arent't using it. Perhaps we can revisit this in a later PR (especially when we remove selected state etc).

native/package-lock.json
```diff
@@ -13,7 +13,7 @@
     "@storybook/addon-actions": {
       "version": "3.2.11",
       "resolved": "https://registry.npmjs.org/@storybook/addon-actions/-/addon-actions-3.2.11.tgz",
-      "integrity": "sha512-WbB+KNn+4BzsBzHDR4fdE33YXoC/etSrBqwz3RM4unJlgRSiJxWFyv0aKrMQA5IrfJPyFudI8wum1asgBATf7Q==",
+      "integrity": "sha1-TfXtFTac0/vhYe+vnh+Izn1shjE=",
```
matthewdavidson: these shouldn't have changed :/

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
```diff
@@ -138,15 +164,9 @@ const commonTests = () => {
...
-    const themeAttributes = {
-      gradientStartColor: '#CE93D8',
-      gradientEndColor: '#AB47BC',
-      textColor: 'rgba(255, 255, 255, 0.8)',
-    };
-
     it('should render correctly', () => {
       const tree = renderer.create(
-        <BpkThemeProvider theme={themeAttributes}>
+        <BpkThemeProvider theme={{}}>
```
matthewdavidson: Why are we not testing this anymore?

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
shaundon: That's a good point yeah. Shall we just have it 'fail' silently?

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
shaundon: When I looked back at this, it wasn't actually testing it very accurately as the theme names changed (e.g. `textColor` became `buttonPrimaryTextColor`. So I figured it was less misleading to have it take an empty object.

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
shaundon: Yeah that makes sense, it's a bit of a hack having a non-changing gradient.

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
matthewdavidson: We need a test case to show that when we wrap a themeable component in a themeprovider with a theme object that it works though, right?

shaundon: Now addressed

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
shaundon: I've removed warning when the type isn't themeable. Currently the warning when they fail to provide all the necessary theme attributes is still there - can remove it if we think that's overzealous too. I'm just concerned that failing silently (by falling back to no theme) will be a bad UX for consumers. What do you think?

matthewdavidson: So we defo had to remove the other one because it would throw a warn when the consumer use case was completely valid. Our incomplete theme object warning shouldn't suffer the same problem as an incomplete theme is...well just that - an incomplete theme. I cannot think of a case where supplying an incomplete theme is beneficial, therefore I think our warning is good.

matthewdavidson:
@shaundon we still need a changelog actually.

shaundon:
@matthewdavidson I was gonna add a changelog entry, but the previous one that's still unreleased covers this too, I think.

```
- react-native-bpk-component-button
  - Added theming support
```

PR #182, opened 2017-10-09
[BPK-434] Add datatable component from react-virtualized

jslvtr:
Added datatable with a subset of the suggested features:

- Sortable columns.
- RTL and accessibility support.
- Support for _many_ rows via react-virtualized.

Also includes a readme and example in the storybook.

backpack-table-v0 2

### DIFF
--- a/packages/bpk-component-datatable/src/BpkDataTable.js
+++ b/packages/bpk-component-datatable/src/BpkDataTable.js
@@ -16,24 +16,32 @@
  * limitations under the License.
  */
 
+import PropTypes from 'prop-types';
 import React, { Component } from 'react';
-import ReactDOM from 'react-dom';
-import { defaultTableRowRenderer, Column, Table, AutoSizer } from 'react-virtualized';
+import { Table, AutoSizer } from 'react-virtualized';
 import { cssModules } from 'bpk-react-utils';
 import _ from 'lodash';
 
 import STYLES from './bpk-data-table.scss';
-import BpkColumn from './BpkColumn';
+import { BpkColumn } from './BpkColumn';
 
 const getClassName = cssModules(STYLES);
 
-export default class BpkDataTable extends Component {
+const sortList = ({ sortBy, sortDirection, list }) => {
+  const sorted = _.sortBy(list, sortBy);
+  if (sortDirection === 'DESC') {
+    return _.reverse(sorted);
+  }
+  return sorted;
+};
+
+class BpkDataTable extends Component {
   constructor(props) {
     super(props);
 
     const sortBy = 'name';
     const sortDirection = 'ASC';
-    const sortedList = this._sortList({ sortBy, sortDirection, list: props.rows });
+    const sortedList = sortList({ sortBy, sortDirection, list: props.rows });
     let columns = _.cloneDeep(this.props.children);
 
     if (this.props.dir === 'rtl') {
@@ -46,17 +54,16 @@ export default class BpkDataTable extends Component {
       sortDirection,
       columns,
       rows: props.rows,
-      rowSelected: undefined
+      rowSelected: undefined,
     };
 
-    this.table = null;
     this.rowClicked = this.rowClicked.bind(this);
     this.rowStyle = this.rowStyle.bind(this);
-    this._sort = this._sort.bind(this);
+    this.sort = this.sort.bind(this);
   }
 
-  rowClicked({ event, index, rowData }) {
-    if (this.state.rowSelected == index) {
+  rowClicked({ index }) {
+    if (this.state.rowSelected === index) {
       this.setState({ rowSelected: undefined });
     } else {
       this.setState({ rowSelected: index });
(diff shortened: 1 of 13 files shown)


georgegillams:
From initially playing with this it's looking good. Just a couple of things:
 - When a column header or row is selected, the outline shown in chrome is partially obscured.
 - Also when a column is clicked to make it sort, the arrow icon displaces other content causing it to jump left / right. Maybe the space for this icon should be always reserved to avoid this:

data_table

jslvtr:
Hey @georgegillams, good find! Let me correct those things. Any ideas about why the outline might be partially hidden in Chrome?

georgegillams:
Often this happens if you are using `overflow: hidden` as it treats the outline as part of the component, and is effectively clipping the edge off.

jslvtr:
Hmm 🤔 it's probably part of `react-virtualized`. I'll have a look!

georgegillams:
If it is then it may be out of your control 😢

jslvtr:
Yes 😢

jslvtr:
@georgegillams indeed, `overflow: hidden` is hard-coded as an inline style 😱 . I think we'll have to stick with it unless we want to do `overflow: inherit !important;` in the stylesheet 🔫

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,136 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Table, AutoSizer } from 'react-virtualized';
+import { cssModules } from 'bpk-react-utils';
+import _ from 'lodash';
```
azylinski: It would be better, to import only what's needed - so sortBy and cloneDeep

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,136 @@
...
+
+import STYLES from './bpk-data-table.scss';
+import { BpkColumn } from './BpkColumn';
+
+const getClassName = cssModules(STYLES);
+
+const sortList = ({ sortBy, sortDirection, list }) => {
+  const sorted = _.sortBy(list, sortBy);
+  if (sortDirection === 'DESC') {
+    return _.reverse(sorted);
```
azylinski: No need for lodash: `sorted.reverse()`

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,136 @@
...
+  constructor(props) {
+    super(props);
+
+    const sortBy = 'name';
+    const sortDirection = 'ASC';
+    const sortedList = sortList({ sortBy, sortDirection, list: props.rows });
+    let columns = _.cloneDeep(this.props.children);
+
+    if (this.props.dir === 'rtl') {
+      columns = _.reverse(columns);
```
azylinski: Same here: `sorted.reverse()`

packages/bpk-component-datatable/src/BpkDataTable.js
janeklb: should also slice first, since reverse is (and `_.reverse`) are mutable operations: `sorted.slice().reverse()` :)

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+class BpkDataTable extends Component {
+  constructor(props) {
+    super(props);
+
+    const sortBy = 'name';
+    const sortDirection = 'ASC';
+    const sortedList = sortList({ sortBy, sortDirection, list: props.rows });
+    let columns = _.cloneDeep(this.props.children);
```
janeklb: `_.cloneDeep` causes JS heap errors - if this is just to get a copy of the children for rtl reversing purposes, could you use `this.props.children.slice()` instead?

IllegalCactus:
Hi,

Quick comment regarding the DataTable component: would it be possible perhaps to add an optional `onRowClick` function prop, to handle when rows are clicked? I.e. if you wanted to pass down a custom handler. We're currently integrating a copy of this component into our app and would find this functionality quite useful :)

We've just copied the module directly into our repo, but got the functionality above working with this change:

```diff
diff --git a/client/src/components/analytics/market-comparison-table/bpk-component-datatable/src/BpkDataTable.js b/client/src/components/analytics/market-comparison-table/bpk-component-datatable/src/BpkDataTable.js
index 706df06..c10451f 100644
--- a/client/src/components/analytics/market-comparison-table/bpk-component-datatable/src/BpkDataTable.js
+++ b/client/src/components/analytics/market-comparison-table/bpk-component-datatable/src/BpkDataTable.js
@@ -63,6 +63,9 @@ class BpkDataTable extends Component {
     } else {
       this.setState({ rowSelected: index });
     }
+    if (this.props.onRowClick) {
+      this.props.onRowClick(this.state.rows[index]);
+    }
   }
 
   rowStyle({ index }) {
@@ -120,6 +123,7 @@ BpkDataTable.propTypes = {
   rows: PropTypes.arrayOf(Object).isRequired,
   children: PropTypes.arrayOf(PropTypes.element).isRequired,
   height: PropTypes.number.isRequired,
+  onRowClick: PropTypes.func,
 };
 
 export default BpkDataTable;
```

Would that fit with your intentions behind this component? Is there a better approach you would prefer?

Thanks :)

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,153 @@
...
+
+    if (this.props.dir === 'rtl') {
+      columns = columns.reverse();
+    }
+
+    this.state = {
+      sortedList,
+      sortBy,
+      sortDirection,
+      columns,
```
azylinski: columns doesn't have to be stored in the state, could be rendered from props

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,153 @@
...
+      rows: props.rows,
+      rowSelected: undefined,
+    };
+
+    this.rowClicked = this.rowClicked.bind(this);
+    this.rowStyle = this.rowStyle.bind(this);
+    this.sort = this.sort.bind(this);
+  }
+
```
azylinski:
```
  componentWillReceiveProps(nextProps) {
    const sortedList = sortList({ sortBy: this.state.sortBy, sortDirection: this.state.sortDirection, list: nextProps.rows });
    this.setState({ rows: nextProps.rows, sortedList, rowSelected: undefined });
  }
```

Without that, there would be no rerender on props change.

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,153 @@
...
+    this.setState({ sortBy, sortDirection, sortedList });
+  }
+
+  headerRenderer(args) {
+    return <SearchableHeaderRenderer {...args} />;
+  }
+
+  render() {
+    const { sortedList, sortDirection, sortBy } = this.state;
+    const { height } = this.props;
```
azylinski:
```
const { height, children } = this.props;
```

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,153 @@
...
+            rowClassName={this.rowStyle}
+            onRowClick={this.rowClicked}
+            sort={this.sort}
+            sortBy={sortBy}
+            sortDirection={sortDirection}
+            gridStyle={{ direction: this.props.dir }}
+            headerRowRenderer={this.headerRenderer}
+          >
+            {
+              this.state.columns.map((child, index) => (
```
azylinski:
```
children.map(...
```

matthewdavidson:
Closing for now whilst this PR remains inactive. We can reopen it in the future.

PR #185, opened 2017-10-10
[BPK-1024] Upgrade react-native to ^0.49.0

matthewdavidson:
**NOTE:** our components are still at `>= 0.47.0` for peer dependency.
Upgraded `babel-preset-react-native` to `^4.0.0`.
I also removed stage 3 TC39 proposal syntax from create-with-theme.js


native/package.json
```diff
@@ -17,12 +17,12 @@
...
-    "@storybook/react-native": "^3.2.11",
-    "babel-preset-react-native": "3.0.2",
-    "react-dom": "^16.0.0",
-    "react-test-renderer": "^16.0.0",
+    "babel-preset-react-native": "^4.0.0",
+    "@storybook/react-native": "^3.2.12",
     "react": "^16.0.0",
-    "react-native": "^0.47.0"
+    "react-dom": "^16.0.0",
+    "react-native": "^0.49.3",
```
matteo-hertel: Just a quick one on this, will lock this down on the full version or we allow patch level update (0.49.*)?  there has been much discussion about this but no action yet

native/package.json
```diff
@@ -17,12 +17,12 @@
...
   "devDependencies": {
-    "@storybook/react-native": "^3.2.11",
-    "babel-preset-react-native": "3.0.2",
-    "react-dom": "^16.0.0",
-    "react-test-renderer": "^16.0.0",
+    "babel-preset-react-native": "^4.0.0",
+    "@storybook/react-native": "^3.2.12",
     "react": "^16.0.0",
-    "react-native": "^0.47.0"
+    "react-dom": "^16.0.0",
```
k0nserv: The peer for `0.49.3` is `16.0.0.beta-5` for `react`, `react-dom`, `react-test-renderer`. We should specify those versions IMO

native/package.json
matthewdavidson: Have a look at this: https://runkit.com/embed/umidffgbnot2

(The caret has different behaviour depending on pre `1.0.0`)

native/package.json
matthewdavidson: cool will do

matthewdavidson: Fixed - here is an example https://runkit.com/embed/umidffgbnot2

matthewdavidson: Please approve when happy :)

matthewdavidson: Ah you mean in our root dependencies? I guess so ...

matthewdavidson:
@k0nserv because JDI and it's a small PR

native/packages/react-native-bpk-theming/package.json
```diff
@@ -10,7 +10,7 @@
   "author": "Backpack Design System <backpack@skyscanner.net>",
   "license": "Apache-2.0",
   "peerDependencies": {
-    "react": "^16.0.0",
+    "react": "^16.0.0-beta.1",
```
k0nserv: Shouldn't we still have alpha too? Until the apps are both upgraded?

PR #189, opened 2017-10-10
[BPK-1004] Add check to ensure all bpk dependencies are latest version

georgegillams:
I decided this should be dealt with sooner rather than later as we are adding a lot of new components (and therefore new packages which is where these errors creep in).
Currently, in our master branch, we have 3 outdated references!

success

screen shot 2017-10-11 at 12 12 46


scripts/npm/update-bpk-dependencies.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+};
+
+const installNcu = () => {
+  try {
+    execSync('npm list -g npm-check-updates');
+  } catch (e) {
+    console.log('NCU not installed 😱');
+    console.log('Installing `npm-check-updates` now! 🙉 ');
+    execSync('npm install -g npm-check-updates');
```
matthewdavidson: This is naughty! Can we find a solution that doesn't rely on a third party tool - especially one that we have to install globally? (don't let hugo see this 😂 )

matthewdavidson: Perhaps something along the lines of:

1. grab a list of `package.json`'s
2. create a map of `"package-name": version,`
3. iterate through each package and check `dependencies` & `devDependencies` against package version map
4. if they don't match, throw error

georgegillams: I thought about that, but I didn't like the manual feeling of it.
Happy to go down that route instead if you think it's better

.travis.yml
```diff
@@ -15,6 +15,7 @@ script:
   - npm run docs:dist:gzip
   - test -e dist/index.html
   - test -e dist/sassdoc/index.html
+  - npm run update-bpk-dependencies
```
matthewdavidson: I reckon this should be a lot earlier in the pipeline - even part of `npm test`?

Currently it'll only fail on CI.

package.json
```diff
@@ -43,7 +43,8 @@
     "android": "(cd ./native && npm run android)",
     "check-owners": "npm whoami && grep -E $(npm whoami) NPM_OWNERS && node scripts/npm/check-owners.js",
     "publish": "npm run check-owners && npm run build && git pull --rebase && npm run test && lerna publish",
-    "danger": "danger"
+    "danger": "danger",
+    "update-bpk-dependencies": "node scripts/npm/update-bpk-dependencies.js"
```
matthewdavidson: this doesn't actually update anymore does it? I suggest `"check-bpk-dependencies"` && `check-bpk-dependencies.js` respectively.

scripts/npm/update-bpk-dependencies.js
```diff
@@ -0,0 +1,89 @@
...
+  packageFiles.forEach((pf) => {
+    if (pf !== '') {
+      const pfContent = JSON.parse(fs.readFileSync(pf));
+      bpkPackageVersions[pfContent.name] = pfContent.version;
+    }
+  });
+  return bpkPackageVersions;
+};
+
+console.log('Checking dependencies on bpk packages');
```
matthewdavidson: How about `Checking Backpack cross dependencies...` ?

matthewdavidson:
Those screenshots have quite verbose output, but a realise that they might be old and based on `ncu` - whats the output like now? Can you update the screenshots?

.travis.yml
georgegillams: The reason I put it so late before is because it was such an expensive script. Now that it only takes a few seconds I agree 👍

package.json
georgegillams: Cool!

scripts/npm/update-bpk-dependencies.js
georgegillams: Agreed!

package.json
```diff
@@ -37,8 +37,9 @@
     "docs:dist:gzip": "npm run docs:dist && gulp",
     "start": "npm run build && npm run storybook",
     "native": "(cd ./native && npm run storybook)",
-    "test": "npm run lint && npm run jest && npm run test:native",
-    "test:native": "(cd ./native && npm test)",
+    "check-bpk-dependencies": "node scripts/npm/check-bpk-dependencies.js",
+    "test": "npm run lint && npm run check-bpk-dependencies && npm run jest && npm run test:native",
+    "test:native": "(npm run check-bpk-dependencies && cd ./native && npm test)",
```
matthewdavidson: erm this runs the check twice now?

georgegillams: If it only runs it within `./native` it won't check them all, though.

PR #190, opened 2017-10-11
[no jira] Change theme story to use BpkButton

shaundon:
Previously the example used `BpkThemeableText`, which is something that was created especially for the storybook. Now that buttons are themeable, I've changed the storybook to use that instead.

I also changed the green example to yellow, as our standard button is green and the difference wasn't so apparent.

screen shot 2017-10-11 at 14 40 32
screen shot 2017-10-11 at 14 37 15
screen shot 2017-10-11 at 14 40 35

### DIFF
--- a/native/packages/react-native-bpk-theming/package-lock.json
+++ b/native/packages/react-native-bpk-theming/package-lock.json
@@ -1,174 +1,283 @@
 {
-	"requires": true,
-	"lockfileVersion": 1,
-	"dependencies": {
-		"asap": {
-			"version": "2.0.6",
-			"resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
-			"integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY="
-		},
-		"brcast": {
-			"version": "2.0.2",
-			"resolved": "https://registry.npmjs.org/brcast/-/brcast-2.0.2.tgz",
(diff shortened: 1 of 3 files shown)


native/packages/react-native-bpk-theming/package-lock.json
```diff
@@ -143,15 +203,66 @@
...
+			"dev": true,
+			"requires": {
+				"is-arrayish": "0.3.1"
+			}
+		},
 		"theming": {
 			"version": "1.1.0",
 			"resolved": "https://registry.npmjs.org/theming/-/theming-1.1.0.tgz",
-			"integrity": "sha512-YRhQV+eqeuef70eqfMGKGR+3DBXoWJ4lsrSSuj8dzrv0/EsnVuDmZAjIifESvJARtuP1KbMpdl8qbXdLAz4t/w==",
+			"integrity": "sha1-BWJ2C1WhuRnC1e65QTA1H4lY4To=",
```
matthewdavidson: why are these sha1's whereas most of them are sha512? 🤔

shaundon: Honestly no idea, something in my NPM install seems to keep doing this though. I'll see if I can find anything.

shaundon: Looks like it's a known issue https://github.com/npm/npm/issues/17749. Not that it helps us to know that!

PR #191, opened 2017-10-11
[BPK-1068] update iOS instruction

matteo-hertel:
So as mentioned the font file it's working fine, I just had to update the instruction for a more manual approach, also I've dropped the podfile to prevent confusions 

icon


native/packages/react-native-bpk-component-icon/readme.md
```diff
@@ -10,11 +10,26 @@ npm install react-native-bpk-component-icon --save-dev
...
+    ```
+    if the entry `UIAppFonts` it's already there, just add `<string>BpkIcon.ttf</string>` inside the `<array>` like so
+    ```
+    <array>
+        ... existing entries
+        <string>BpkIcon.ttf</string>
+      </array>
+    ```
+2. In the `Build Phases` of your project, in the section `Copy Bundle Resources` add a reference to the      `BpkIcon.ttf` file path like `/path/to/node_modules/bpk-svgs/dist/font/BpkIcon.ttf`, the end result        should look like this
+
```
shaundon: Very nitpicky I know, but there's a load of superfluous spaces between `result` and `should`

shaundon: Very nitpicky I know, but there's a load of superfluous spaces between `result` and `should`

PR #194, opened 2017-10-12
[BPK-1004] Add Chris' icon changes

georgegillams:
Chris and I had a chat this afternoon and he felt that it would be best for us to make icon changes to our codebase ourselves as he is not familiar with our codebase / git etc, and he has enough on his plate without learning git just for the sake of contributing icons to our codebase.

So he raised this ticket for me, and here it is 😸 
https://gojira.skyscanner.net/browse/BPK-1070


native/packages/react-native-bpk-component-icon/src/__snapshots__/BpkIcon-test.android.js.snap
```diff
@@ -66,6 +66,6 @@ exports[`Android BpkIcon should render small icon correctly 1`] = `
     ]
   }
 >
-  
+  
```
shaundon: Is it me or has nothing changed here?

matteo-hertel: I think the unicode actually changed but the browser can't render it

shaundon: Ahh ok

PR #195, opened 2017-10-13
[BPK-999] Upgrade recompose to 0.26.x

shaundon:
This is so it'll work with React 16.x.

I'm not certain of what needed to be done here, so here's a summary of what I did:

- Searched for all `package.json` files that mention recompose. The only result was `bpk-react-utils'.
- Upgraded from 0.24.x to 0.26.x in `bpk-react-utils`.
- `npm run clean`
- `npm install`
- `npm run build`
- `npm test`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -32,6 +32,10 @@
     - ticket
     - time
 
+**Fixed:**
+- bpk-react-utils
+ - Upgraded [recompose](https://github.com/acdlite/recompose) dependency to support React 16.
+
 ## 2017-10-12 - New React Native icon component installation instructions and simpler React Native button API
 
 **Breaking:**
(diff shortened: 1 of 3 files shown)


shaundon:
@k0nserv should work fine with React 14, 15 or 16 according to their `package.json`: https://github.com/acdlite/recompose/blob/master/src/packages/recompose/package.json#L21

As for the lockfiles, I have no effing idea tbh, they seem to randomly change every time I run `npm install` lately. I think I'll just remove them from the PR to be safe.

k0nserv:
Cool, the lock files might just be out to date so maybe we should commit those in a separate PR. I'll check and see if I too have diffs

k0nserv:
Should we have a chanelog entry?

matteo-hertel:
I have changes in the package.lock as well on a fresh npm install, we can commit all the changes in a separate pr

shaundon:
Changelog updated.

k0nserv:
LGTM

PR #198, opened 2017-10-16
[EXPS-614]: Add className prop to BpkList and BpkListItem

sbennett117:
Prop contains same functionality as it does elsewhere, using BpkText as an example of how to implement. Has no effect on tests. Has been linted.

Pull Request also contains update to the docs for List and Badge to include className prop, since I noticed it was missing from Badge last week.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,14 @@
 ## UNRELEASED
 _Nothing Yet_
 
+
+## 2017-10-16 - Reack Lists and List Items can now have a className attribute added
+
+**Updated:**
+- bpk-component-list
+  - The `className` attribute has been added to BpkList and BpkListItem, mirroring functionality from other props that implement it
+
+
 ## 2017-10-13 - 17 new icons added and 5 existing icons changed
 
 **Added:**
--- a/changelog.md
+++ b/changelog.md
@@ -1,12 +1,9 @@
 # Backpack changelog
 
 ## UNRELEASED
-_Nothing Yet_
+### 2017-10-16 - React Lists and List Items can now have a className attribute added
 
-
-## 2017-10-16 - Reack Lists and List Items can now have a className attribute added
-
-**Updated:**
+**Added:**
 - bpk-component-list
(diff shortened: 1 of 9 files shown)


sbennett117:
Got that ready to push @k0nserv, should I update `changelog.md` too?

k0nserv:
Yeah please do :)

changelog.md
```diff
@@ -3,6 +3,14 @@
 ## UNRELEASED
 _Nothing Yet_
 
+
+## 2017-10-16 - Reack Lists and List Items can now have a className attribute added
```
k0nserv: Please don't add this line yet, put it under unreleased for now. We'll update this when we publish.

changelog.md
```diff
@@ -3,6 +3,14 @@
...
+
+**Updated:**
```
k0nserv: This should be `**Added**`

changelog.md
sbennett117: Ah, sorry! Never noticed that bit

sbennett117:
Just realised I have no idea if you get pinged whenever I commit, but that should be the changelog fixed

changelog.md
```diff
@@ -1,7 +1,12 @@
 # Backpack changelog
 
 ## UNRELEASED
-_Nothing Yet_
+### 2017-10-16 - React Lists and List Items can now have a className attribute added
+
+**Added:**
+- bpk-component-list
+  - The `className` attribute has been added to BpkList and BpkListItem, mirroring functionality from other props that implement it
```
k0nserv: Can we call props "props" instead of attributes to be consistent with React and the rest of the codebase? I believe the second part should say "mirroring functionallity in other components."

sbennett117: Not sure why I said attributes there, I wasn't sure if 'attributes' or 'props' was the correct terminology

sbennett117: There we are, fixed

sbennett117:
Sweet, thanks! I can't merge because I don't have write access, so could you push the button?

PR #199, opened 2017-10-16
[BPK-1072] Add animation to expandable banner alerts

georgegillams:


packages/bpk-component-banner-alert/src/BpkBannerAlert.js
```diff
@@ -128,11 +130,9 @@ class BpkBannerAlert extends Component {
             </span>
           ) : null}
         </header>
-        {
-          showChildren ?
-            <div className={getClassName('bpk-banner-alert__children-container')}>{children}</div>
-            : null
-        }
+        <BpkAnimateHeight duration={parseInt(durationSm, 10)} height={showChildren ? 'auto' : 0}>
```
matteo-hertel: not sure we need `parseInt` there the value is from the tokens, so we should be confident it's the correct format, there won't be any performance impact tough so you can leave it there

georgegillams: `BpkAnimateHeight` expects `duration` to be a number, and the animation tokens are strings (eg `"200ms"`)

matteo-hertel: I see, so we're relying on `parseInt` to return the 200 from "200ms", what if for any reason the string can't be parsed and returns a `NaN` will the component blow up?

georgegillams: I would hope that if we make breaking changes to our tokens we will test any components that use them

matteo-hertel: 🤞

PR #205, opened 2017-10-17
[BPK-1073] Add neutral banner alert configuration (native)

georgegillams:


native/packages/react-native-bpk-component-banner-alert/src/__snapshots__/BpkBannerAlert-test.ios.js.snap
```diff
@@ -302,14 +304,15 @@ exports[`iOS BpkBannerAlert should render correctly dismissable 1`] = `
...
+              "color": "rgb(82, 76, 97)",
               "flex": 1,
               "flexGrow": 1,
               "paddingLeft": 4,
             },
           ]
         }
       >
-        Successful alert.
+        Neutral alert.
```
matteo-hertel: This snapshot looks weird, if Neutral has been added, how come the diff has `successful` replaced with neutral and no new tree for the one added?

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert-test.common.js
```diff
@@ -44,6 +48,11 @@ const commonTests = () => {
     it('should render correctly dismissable', () => {
       const tree = renderer.create(
         <BpkBannerAlert
+          type={ALERT_TYPES.NEUTRAL}
```
matteo-hertel: not a deal breaker but neutral is missing in the other tests further down, we should have consistency in the tests files

georgegillams: Done :)

native/packages/react-native-bpk-component-banner-alert/src/__snapshots__/BpkBannerAlert-test.ios.js.snap
georgegillams: It would seem that the tests were never creating snapshots for anything other than the first component.

Have wrapped them all inside a view so that they actually render!

PR #207, opened 2017-10-18
Fixed propType warning by updating react-responsive to the latest version

ksiwei:
image
the browser warning above was fixed in the latest version of react-responsive. https://github.com/contra/react-responsive/commit/49919d09b7a4e984873b570132ada5554300f071

### DIFF
--- a/packages/bpk-component-breakpoint/package.json
+++ b/packages/bpk-component-breakpoint/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "bpk-mixins": "^17.0.7",
     "prop-types": "^15.5.8",
-    "react-responsive": "^1.3.2"
+    "react-responsive": "^3.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


packages/bpk-component-breakpoint/package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "bpk-mixins": "^17.0.7",
     "prop-types": "^15.5.8",
-    "react-responsive": "^1.3.2"
+    "react-responsive": "^3.0.0"
```
k0nserv: This looks like a significant change. Two new major versions, did we review the CHANELOG.md for react-responsive and determine there are no breaking changes we need to address?

ksiwei: Had a look at CHANGELOG.md, don't think there is any breaking change. Also ran storybook locally, everything seems fine.

PR #208, opened 2017-10-18
[BPK-934] Add BpkIcon support to BpkButton

shaundon:
- Users can now pass in a string to the `icon` prop if they wish. This will be used to make a `BpkIcon` inside the button.
- Users can still pass arbitrary elements in, but they won't be styled or themed, it's up to them.
- We don't validate that the string passed in is a valid icon, because `BpkIcon` does this and the error will bubble up.

Screenshots:

screen shot 2017-10-18 at 17 08 08

screen shot 2017-10-18 at 17 08 41


native/packages/react-native-bpk-component-button/src/__snapshots__/BpkButton-test.android.js.snap
```diff
@@ -1257,7 +1351,10 @@ exports[`Android BpkButton should support the "iconOnly" and "large" property 1`
               "paddingLeft": 16,
               "paddingRight": 16,
             },
-            undefined,
+            Object {
```
matteo-hertel: what's this bit here? from undefined to an object?

native/packages/react-native-bpk-component-button/src/__snapshots__/BpkButton-test.ios.js.snap
```diff
@@ -778,7 +809,10 @@ exports[`iOS BpkButton should support having only an icon 1`] = `
               "paddingRight": 12,
               "paddingTop": 8,
             },
-            undefined,
+            Object {
```
matteo-hertel: same as above, what's this bit?

native/packages/react-native-bpk-component-button/src/__snapshots__/BpkButton-test.android.js.snap
shaundon: I'm not 100% but I think it's some styling that wasn't rendered properly before and came out like that, which is working now.

k0nserv:
I prefer passing an Element over the string approach. This is more consistent with web, leaves open the possibility to use arbitrary images(like logos), and removes the need to have `react-native-bpk-component-button` depend on `react-native-bpk-component-icon`. Making it a string is also a breaking change whereas an Element wouldn't be. We still have the option to regress to a string in the future with a breaking change

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -128,11 +128,19 @@ const modifiers = {
...
+      paddingLeft: 0,
+      paddingRight: 0,
+    },
   }),
   iconOnlyLarge: StyleSheet.create({
     container: {
       width: largeHeight,
     },
+    button: {
+      paddingLeft: 0,
```
k0nserv: `paddingHorizontal` could be used here right?

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -128,11 +128,19 @@ const modifiers = {
     container: {
       width: tokens.spacingXl,
     },
+    button: {
+      paddingLeft: 0,
```
k0nserv: `paddingHorizontal` could be used here right?

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
```diff
@@ -26,8 +26,7 @@ import BpkButton, { propTypes, BUTTON_TYPES } from './BpkButton';
 
 const onPressFn = jest.fn();
 const commonTests = () => {
-  jest.mock('Image', () => 'Image');
-
+  jest.mock('Text', () => 'Text');
```
k0nserv: Why is this line indented like this?

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
shaundon: Didn't know about that prop, thanks!

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
shaundon: Just tried it and it didn't have the same effect, despite what the docs said 😢

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
shaundon: 2 spaces indent because it's in a function - I don't see the issue?

screen shot 2017-10-19 at 11 34 47

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
```diff
@@ -16,155 +16,173 @@
...
-          onPress={onPressFn}
-        />,
+       expect(tree).toMatchSnapshot();
+     });
+
+     it('should support elements as icons', () => {
+       const tree = renderer.create(
+         <BpkButton
+           title="Lorem ipsum"
+           icon={<Text>foo</Text>}
```
matteo-hertel: do you think is a good idea to add in the test a behaviour that we don't want to support? I'd use BpkIcon as element here, for people like me that look at the test to see how to use something, it'll encourage to use BpkIcon

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
k0nserv: Strange

k0nserv: What happens instead?

shaundon: Ah. I'm being dumb. Because of the cascade, it was inheriting `paddingLeft` and `paddingRight` from another style. I've changed everywhere to use `paddingHorizontal` and `paddingVertical` and it works like a charm now!

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
shaundon: I see what you're saying, but then also, we do technically *support* it, we just don't *recommend* it. I'd want to know if a change we made stopped this working correctly, for example.

For me, it's the same as how we have tests for having both text and an icon in the button, even though it's not something we display in the docs.

However if we want to steer users away from this edge case it should be removed from the code sample in `readme.md`, I think.

native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js
k0nserv: Hmm sorry, I'm blind it seems

shaundon: It's ok, the glasses you're wearing in your photo do look quite dark 😆

k0nserv: image

native/packages/react-native-bpk-component-button/readme.md
```diff
@@ -47,7 +45,7 @@ export default class App extends Component {
         <BpkButton disabled type="destructive" title={translationHelper.translate('BOOK_FLIGHT')} onPress={() => {}} />
         <BpkButton large type="primary" title={translationHelper.translate('BOOK_FLIGHT')} onPress={() => {}} />
 
-        <BpkButton type="featured" title={translationHelper.translate('BOOK_FLIGHT')} icon={ArrowImage} iconOnly onPress={() => {}} />
+        <BpkButton type="featured" title={translationHelper.translate('BOOK_FLIGHT')} icon="baggage" iconOnly onPress={() => {}} />
```
k0nserv: It would still be good to have a constant for the icons so instead of `"baggage"` we could have `ICONS.BAGGAGE`

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -128,11 +122,17 @@ const modifiers = {
     container: {
       width: tokens.spacingXl,
     },
+    button: {
+      paddingHorizontal: 0,
```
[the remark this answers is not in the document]
shaundon: Nah I was wrong - see this comment! https://github.com/Skyscanner/backpack/pull/208#discussion_r145715893

native/packages/react-native-bpk-component-button/readme.md
shaundon: Yeah that's a good point. I'm wondering if it should be a separate PR though as it's something `BpkIcon` would export and I don't want to conflate this PR though..

georgegillams: It absolutely should be. Is there a jira card open for this?

shaundon: There is now: https://gojira.skyscanner.net/browse/BPK-1078

PR #209, opened 2017-10-19
update emphasize behaviour in android

matteo-hertel:
update emphasize behaviour in android

Before
before

After
after

### DIFF
--- a/native/packages/react-native-bpk-component-text/src/__snapshots__/BpkText-test.android.js.snap
+++ b/native/packages/react-native-bpk-component-text/src/__snapshots__/BpkText-test.android.js.snap
@@ -40,7 +40,6 @@ exports[`Android BpkText should render correctly with \`emphasize\` prop 1`] = `
       },
       Object {
         "fontFamily": "sans-serif-medium",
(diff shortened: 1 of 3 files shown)


native/packages/react-native-bpk-component-text/src/__snapshots__/BpkText-test.android.js.snap
```diff
@@ -40,7 +40,6 @@ exports[`Android BpkText should render correctly with \`emphasize\` prop 1`] = `
       },
       Object {
         "fontFamily": "sans-serif-medium",
-        "fontWeight": "500",
```
shaundon: So, if this is removed and not replaced, is it just using the default font weight rather than explicitly setting something?

matteo-hertel: it's using `sans-serif-medium` as font family which is bold

shaundon: Ah ok, cool!

PR #210, opened 2017-10-19
[BPK-1076] `BpkIcon` implementation in banner-alerts

georgegillams:
Are we changelogging these @shaundon


shaundon:
It won't be visible to consumers so I reckon no changelog. Might be worth seeing what everybody else thinks though to get a consensus.

I should note that I have added a changelog for the button, but that's because the actual API is changing.

shaundon:
Shouldn't the test snapshots be changing if the internal workings of the component have changed?

georgegillams:
Good point. They're in a separate commit 😸

PR #211, opened 2017-10-19
[BPK-1031]android primary button support

matteo-hertel:
Add Android support for buttons

For the reviewers: 
- It was a bad idea focusing on iOS first and then add Andoird later, so what I've done is try to leave iOS as untouched as possible and heavily user `Platform.select` wrapping the button content in different containers
- I've declared two functions inside the component, it's like declaring functions inside the `render` method, I needed access to the lexical scope to avoid having massive and meaningless function signatures, however I'm not sure if there will be any performance impact because of it, if you know more about it, feel free to start a discussion
- Screenshots needs to be validated from the design pov
- looks like there is an issue with the diff of the main `BpkButton.js` if you can't see the diff append `?w=1` at the end of the url


matteo-hertel:
Do not merge until [https://github.com/Skyscanner/backpack/pull/208](https://github.com/Skyscanner/backpack/pull/208) is merged and the merge conflicts will be resolved

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -33,27 +33,42 @@ const underlayColor = Platform.select({
 // These should probably be their own tokens.
 // For now they are derived from existing tokens.
 const largeHeight = tokens.spacingSm * 12;
-const buttonBorderWidth = tokens.borderSizeLg;
+const buttonBorderWidth = Platform.OS === 'android' ? tokens.borderSizeSm : tokens.borderSizeLg;
```
georgegillams: Maybe this could be implemented in the tokens themselves, by adding aliases for `buttonBorderWidth` to the platform-specific tokens?

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -33,27 +33,42 @@ const underlayColor = Platform.select({
...
+      paddingLeft: tokens.spacingSm * 3,
+      paddingRight: tokens.spacingSm * 3,
+    }),
+    android: () => ({
+      borderRadius: tokens.borderRadiusPill,
+      height: tokens.spacingXl + tokens.spacingSm,
+      paddingTop: tokens.spacingMd,
+      paddingBottom: tokens.spacingMd,
+      paddingLeft: tokens.spacingSm * 3,
+      paddingRight: tokens.spacingSm * 3,
```
k0nserv: Theres a lot of room to use the button tokens as the layer of indirection here. I don't know what we've done previosuly for web with values that are multiple of tokens as we cannot express that in Theo. @jamesf3rguson would you know?

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -16,215 +16,269 @@
...
+  /**
+   * function to get the iOS container for the button
+   * it's declared here rather than outside the component
+   * so it has access to the lexical scope of the component
+   * and all the prop, theme etc is available
+   *
+   * @param {ReactElement} content button's content
+   * @returns {ReactElement} container iOS specific container
+   */
+  function getIOSContainer(content) {
```
k0nserv: Why are we using `function` over arrow function + const var here?

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -16,215 +16,269 @@
...
+    );
+  }
+
+  const textStyle = [
+    getStyleForElement('text', props),
+    getThemingForElement('text', theme, props),
+  ];
+
+  const buttonContent = (
+
```
k0nserv: Remove this line please

native/packages/react-native-bpk-component-button/src/BpkButton.js
matteo-hertel: It's just personal choice, keeps it more functional and will leverage hoisting (it's declared on top anyway tough)

matteo-hertel:
@georgegillams Button with Icon only will not be officially supported by this component, it'll be replaced with `Floating Action Button`

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
matteo-hertel: we don't have button tokens, I can create an alias for it and move this logic away, although the iOS style is using the same logic/syntax, so I think it's best to do a refactor for both in a separate PR

matteo-hertel:
@k0nserv the function signature for the containers will look like
`function getIOSContainer(buttonContent, theme, props, style, disabled, onPress, accessibilityLabel, title, accessibilityTraits, rest) {
`
I honestly don't want to add something like that

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
matteo-hertel: needs to be refactored for both iOS and Android, so I'll do that in a separate PR as it goes beyond the scope of this one

native/packages/react-native-bpk-component-button/src/BpkButton.js
k0nserv: Okay I guess it's down to persoanl choice. I would have used const + arrow function. Hoisting is a bad thing in my head

k0nserv:
> @k0nserv the function signature for the containers will look like
function getIOSContainer(buttonContent, theme, props, style, disabled, onPress, accessibilityLabel, title, accessibilityTraits, rest) {
I honestly don't want to add something like that

Aren't those alls just derived from props though? So we could get away with only passing props. Maybe a better way to go is creating bespoke components for the containers? i.e `src/ButtonContainer.android.js` and `src/ButtonContainer.ios.js` that way we can use props to express the signature and we can rely on React Native to load the platform specific code

native/packages/react-native-bpk-component-button/src/BpkButton.js
matteo-hertel: I disagree on hoisting being a bad thing, but for consistency's sake I've updated to use const + arrow function

matteo-hertel:
> Aren't those alls just derived from props though? So we could get away with only passing props. Maybe a better way to go is creating bespoke components for the containers? i.e src/ButtonContainer.android.js and src/ButtonContainer.ios.js that way we can use props to express the signature and we can rely on React Native to load the platform specific code

i like the container component idea, but if was done upfront while creating the component, atm with the theming and helper functions, it's all too  tightly coupled
I'm in favour of finding a solution that is shippable, and then revisit the component and change/fix our approach to it

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -61,10 +74,19 @@ const base = StyleSheet.create({
...
+  text: Platform.select({
+    ios: () => ({
+      backgroundColor: 'transparent',
+      color: tokens.colorWhite,
+
+    }),
+    android: () => ({
+      backgroundColor: 'transparent',
+      color: tokens.colorWhite,
+      // padding: tokens.spacingSm,
```
matteo-hertel: ignore this 😄

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
```diff
@@ -154,7 +176,7 @@ const modifiers = {
 
 const gradientColors = {
   primary: [tokens.colorGreen500, tokens.colorGreen600],
-  featured: ['rgb(250, 72, 138)', 'rgb(217, 43, 107)'], // TODO [tokens.colorPink500, tokens.colorPink600]
+  featured: [tokens.colorPink500, tokens.colorPink600],
```
shaundon: Thanks for doing this, been meaning to for ages!

native/packages/react-native-bpk-component-button/src/BpkButton-styles.js
shaundon: Should this be removed, as it's commented out?

matteo-hertel: yes I saw that, and I've fixed it 😄

PR #216, opened 2017-10-20
[NoJira] refactor RN button component

matteo-hertel:
I've slayed the beast by creating two components, one for iOS and one for Android for the two containers, I've moved away all the accessory functions in a utils file in order to have some sort of structure in place

### DIFF
--- a/native/packages/react-native-bpk-component-button/src/BpkButton.js
+++ b/native/packages/react-native-bpk-component-button/src/BpkButton.js
@@ -16,154 +16,47 @@
  * limitations under the License.
  */
 
+/* eslint-disable import/no-unresolved */
+/* eslint-disable import/extensions */
 import {
   View,
-  TouchableHighlight,
-  TouchableNativeFeedback,
   Platform,
 } from 'react-native';
 import React from 'react';
 import PropTypes from 'prop-types';
-import LinearGradient from 'react-native-linear-gradient';
-import difference from 'lodash/difference';
 
 import { withTheme } from 'react-native-bpk-theming';
 import BpkIcon from 'react-native-bpk-component-icon';
 import BpkText from 'react-native-bpk-component-text';
 
-import styles, { colorWhite } from './BpkButton-styles';
+import {
+  isTypeThemeable,
+  themeAttributesSupplied,
+  getStyleForElement,
+  iconPropType,
+  themePropType,
+  textStyle,
+} from './utils';
+
+import BpkButtonContainer from './layout/BpkButtonContainer';
 
 const BUTTON_TYPES = ['primary', 'featured', 'secondary', 'destructive'];
-const REQUIRED_THEME_ATTRIBUTES = {
-  primary: [
-    styles.themeMappings.text.color.primary,
-    styles.themeMappings.gradient.primary.startColor,
-    styles.themeMappings.gradient.primary.endColor,
-  ],
-  secondary: [
-    styles.themeMappings.text.color.secondary,
-    styles.themeMappings.button.borderColor.secondary,
-    styles.themeMappings.gradient.secondary.startColor,
-  ],
-};
-const THEMEABLE_TYPES = Object.keys(REQUIRED_THEME_ATTRIBUTES);
 
-const getStyleForElement = (elementType, { type, title, icon, iconOnly, large, disabled }) => {
-  // Start with base style.
-  const styleForElement = [styles.base[elementType]];
-
-  // Add styles for the button type (primary, secondary etc).
-  if (styles.types[type] && styles.types[type][elementType]) {
-    styleForElement.push(styles.types[type][elementType]);
-  }
-
-  // Add modifiers. Disabled comes last to override other styles.
-  if (large) {
-    let largeModifier = 'large';
-    if (['secondary', 'destructive'].includes(type)) {
-      largeModifier = 'largeWithOutline';
-    }
-    styleForElement.push(styles.modifiers[largeModifier][elementType]);
-  }
-
-  if (disabled) {
-    styleForElement.push(styles.modifiers.disabled[elementType]);
-  }
-
-  if (iconOnly) {
-    styleForElement.push(large ? styles.modifiers.iconOnlyLarge[elementType] : styles.modifiers.iconOnly[elementType]);
-  } else if (title && icon) {
-    // If it has a title and icon, get the style for that.
-    styleForElement.push(styles.modifiers[large ? 'textAndIconLarge' : 'textAndIcon'][elementType]);
-  }
-
-  return styleForElement;
-};
-
-const getThemingForElement = (elementType, theme, { type, disabled }) => {
-  const themeForElement = {};
-  if (theme && !disabled && styles.themeMappings[elementType]) {
-    Object.keys(styles.themeMappings[elementType]).forEach((key) => {
-      const values = styles.themeMappings[elementType][key];
-      if (values[type]) {
-        themeForElement[key] = theme[values[type]];
-      }
-    });
-  }
-  return themeForElement;
-};
-
-const getGradientColors = (theme, { type, disabled }) => {
-  let gradientColors = styles.gradientColors[type];
-
-  if (theme) {
-    const gradientThemeProps = styles.themeMappings.gradient[type];
-    gradientColors = [theme[gradientThemeProps.startColor], theme[gradientThemeProps.endColor]];
-  }
-
-  if (disabled) {
-    gradientColors = styles.gradientColors.disabled;
-  }
-  return gradientColors;
-};
-
-const iconPropType = (props, propName, componentName) => {
-  const icon = props[propName];
-  if (props.iconOnly && !icon) {
-    return new Error(`Invalid prop \`${propName}\` supplied to \`${componentName}\`. When \`iconOnly\` is enabled, \`${propName}\` must be supplied.`); // eslint-disable-line max-len
-  }
-  return false;
-};
-
-const isTypeThemeable = type => THEMEABLE_TYPES.includes(type);
-
-const themeAttributesSupplied = (type, theme) => (
-  difference(REQUIRED_THEME_ATTRIBUTES[type], Object.keys(theme)).length === 0
-);
-
-const themePropType = (props, propName, componentName) => {
-  const type = props.type;
-  const theme = props.theme;
-  if (!theme) {
-    return false;
-  }
-  if (!themeAttributesSupplied(type, theme)) {
-    return new Error(`Invalid prop \`${propName}\` supplied to \`${componentName}\`. For buttons of type \`${type}\`, the \`theme\` prop must include \`${REQUIRED_THEME_ATTRIBUTES[type].join(', ')}\``); // eslint-disable-line max-len
-  }
-  return false;
-};
-
-const getAndroidBackgroudnColour = (theme, props) => {
-  const style = {
-    backgroundColor: 'transparent',
-  };
-  /**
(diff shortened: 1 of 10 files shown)


native/packages/react-native-bpk-component-button/src/BpkButton-utils.js
```diff
@@ -0,0 +1,146 @@
+/*
```
matthewdavidson: This file can just be called `utils.js` and it's co-located with BpkButton.js.

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -18,152 +18,44 @@
...
+import {
+  isTypeThemeable,
+  themeAttributesSupplied,
+  getStyleForElement,
+  iconPropType,
+  themePropType,
+  textStyle,
+} from './BpkButton-utils';
+
+import BpkButtonAndroidContainer from './layout/BpkButtonContainer.android';
```
matthewdavidson: you should technically be able to avoid the `Platform.OS` check because the RN packager already conditionally imports file based on it's extension. I.e:

```
import BpkButtonContainer from './layout/BpkButtonContainer';
```
This will import `BpkButtonContainer.{platform}.js` based on current `Platform.OS`

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.android.js
```diff
@@ -0,0 +1,101 @@
...
+  getStyleForElement,
+  getAndroidBackgroundColour,
+  iconPropType,
+  themePropType,
+} from './../BpkButton-utils';
+
+const BpkButtonContainer = (props) => {
+  const {
+     type,
+    title,
```
matthewdavidson: A lot of these props aren't used? I am surprised the lint passed.

(`icon` & `iconOnly` for instance)

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -177,76 +69,39 @@ const BpkButton = (props) => {
...
-  const getIOSContainer = buttonContent => (<LinearGradient
-    colors={getGradientColors(theme, props)}
-    style={[getStyleForElement('container', props), style]}
-  >
-    <TouchableHighlight
-      style={[getStyleForElement('button', props), getThemingForElement('button', theme, props)]}
-      disabled={disabled}
+  return (
+    <BpkButtonContainer
+      {...props}
```
matthewdavidson: Usually when we destructure `props` and spread the remainder to `rest` we are explicit about what gets applied to underlying component. 

I can see some of the appeal of just spreading `props` onto the component but it's a bit of a sledgehammer approach - i.e. not very clear or explicit. It's asking for (unknown prop type warnings)

Also when ever I see a variable being destructured I don't expect it to ever be referenced again - is this just me?

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.android.js
```diff
@@ -0,0 +1,101 @@
...
+          style={[getStyleForElement('button', props), getAndroidBackgroundColour(theme, props)]}
+        >
+          {children}
+        </View>
+      </TouchableNativeFeedback>
+    </View>
+  );
+};
+
+const propTypes = {
```
matthewdavidson: these prop types can be assigned directly to `BpkButtonContainer.propTypes`

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.ios.js
```diff
@@ -0,0 +1,104 @@
...
+  themePropType,
+} from './../BpkButton-utils';
+import styles from './../BpkButton-styles';
+
+const BpkButtonContainer = (props) => {
+  const {
+    type,
+    title,
+    icon,
+    iconOnly,
```
matthewdavidson: Again some unused props.

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.ios.js
```diff
@@ -0,0 +1,104 @@
...
+        accessibilityLabel={accessibilityLabel || title}
+        accessibilityTraits={accessibilityTraits}
+        {...rest}
+      >
+        {children}
+      </TouchableHighlight>
+    </LinearGradient>);
+};
+
+const propTypes = {
```
matthewdavidson: and again this can be assigned to the class member directly.

native/packages/react-native-bpk-component-button/src/BpkButton.js
shaundon: Didn't know about that, that's really cool. I guess it's the same as how storybook does it.

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.android.js
shaundon: Agreed - I think this is just an oversight as in the original `BpkButton.js` file we had to do this so it could be exported and accessed from a test.

native/packages/react-native-bpk-component-button/src/BpkButton.js
matteo-hertel: Forgot a bout that, thanks!

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.android.js
matteo-hertel: that's a good edge case for the lint then, you're absolutely right, I've copy-pasted from the main file, lint was passing so I thought it was all good

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.android.js
matteo-hertel: In `BpkButton.js` we're exporting those propTypes, in the container no, so it was my oversight 😄

native/packages/react-native-bpk-component-button/src/BpkButton.js
```diff
@@ -18,152 +18,45 @@
...
+  getStyleForElement,
+  iconPropType,
+  themePropType,
+  textStyle,
+} from './utils';
+
+import BpkButtonAndroidContainer from './layout/BpkButtonContainer.android';
+import BpkButtonIosContainer from './layout/BpkButtonContainer.ios';
+
+const BpkButtonContainer = Platform.OS === 'android' ? BpkButtonAndroidContainer : BpkButtonIosContainer;
```
k0nserv: Shouldn't we be relying on the RN packager to ensure that the right code is loaded?

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.android.js
```diff
@@ -0,0 +1,96 @@
...
+import PropTypes from 'prop-types';
+
+import {
+  getStyleForElement,
+  getAndroidBackgroundColour,
+  themePropType,
+} from './../utils';
+
+const BpkButtonContainer = (props) => {
+  const theme = props.theme;
```
k0nserv: The new eslitn rules will complain about this not being destructured. Why aren't we doing all the destructing in a single place anyway?

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.ios.js
```diff
@@ -0,0 +1,99 @@
...
+import {
+  getStyleForElement,
+  getThemingForElement,
+  getGradientColors,
+  themePropType,
+} from './../utils';
+import styles from './../BpkButton-styles';
+
+const BpkButtonContainer = (props) => {
+  const theme = props.theme;
```
k0nserv: Same as above

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.ios.js
```diff
@@ -0,0 +1,99 @@
...
+        style={[buttonStyle, buttonTheme]}
+        underlayColor={styles.underlayColor}
+        {...rest}
+      >
+        {children}
+      </TouchableHighlight>
+    </LinearGradient>);
+};
+
+BpkButtonContainer.propTypes = {
```
k0nserv: Should we have a test that ensures these two have the same `PropTypes` exported?

native/packages/react-native-bpk-component-button/src/utils.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+  if (!themeAttributesSupplied(type, theme)) {
+    return new Error(`Invalid prop \`${propName}\` supplied to \`${componentName}\`. For buttons of type \`${type}\`, the \`theme\` prop must include \`${REQUIRED_THEME_ATTRIBUTES[type].join(', ')}\``); // eslint-disable-line max-len
+  }
+  return false;
+};
+
+export const getAndroidBackgroundColour = (theme, props) => {
+  const style = {
+    backgroundColor: 'transparent',
```
k0nserv: This part of the style isn't cached since it's not created with `StyleeSheet` create right?

native/packages/react-native-bpk-component-button/src/BpkButton.js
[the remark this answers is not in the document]
matteo-hertel: this is embarrassing, I've fixed it and somehow it's back

native/packages/react-native-bpk-component-button/src/utils.js
matteo-hertel: Oh, nice catch! I'll amend this!

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.ios.js
matteo-hertel: I would agree usually, but in this case there is a `large` prop used by iOS but not android

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.android.js
matteo-hertel: After a chat with @matthewdavidson it looks better is the props object is not used after has been destructured, if the new eslint will complain about it, we'll have to trade in some readability, to make the linter happy 😄

native/packages/react-native-bpk-component-button/src/BpkButton.js
matteo-hertel: yes indeed, I've changed it but I revert it back somehow 😅

native/packages/react-native-bpk-component-button/src/layout/BpkButtonContainer.ios.js
k0nserv: We could take a similar approach as we do with tokens where we support platform specific tokens.

matteo-hertel: like https://github.com/Skyscanner/backpack/blob/master/packages/bpk-tokens/src/react-native-output-test.js

k0nserv: Yep

matteo-hertel: that's a really good idea, I want to improve the code coverage anyway, I'll add that in!

PR #220, opened 2017-10-24
[NoJira] import proptypes from view to prevent broken build

matteo-hertel:
Import prop type to prevent the build to break in the native app

see 
https://github.com/necolas/react-native-web/issues/423

### DIFF
--- a/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
+++ b/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
@@ -20,6 +20,7 @@ import {
   View,
   Platform,
   Animated,
+  ViewPropTypes,
 } from 'react-native';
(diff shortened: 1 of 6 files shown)


matteo-hertel:
We're using it in `Cards` and `AnimateHeight`, I'll update the PR

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -82,7 +83,7 @@ class BpkAnimateHeight extends React.Component {
         toValue: finalValue,
         duration: animationDuration,
       },
-      ).start();
```
matteo-hertel: sorry, auto fix 😕

k0nserv:
Missing CHANGELOG.md entry

PR #225, opened 2017-10-25
[BPK-971] Update jest and lerna and misc

matthewdavidson:
This is one of the many PR's aimed at upgrading dependencies in the backpack monorepo. I focus on updating lerna & jest based dependencies. Here are the list of changelogs:

- TODO


packages/bpk-component-image/src/BpkImage-test.js
```diff
@@ -68,8 +74,10 @@ describe('BpkImage', () => {
         src="./path/to/image.jpg"
       />,
     ).toJSON();
+
```
k0nserv: Where these changes actually related to an upgrade or something you just did to make Arrange, Act, Assert more clear?

matthewdavidson: I just added new lines between tests and statements tbh. yet another call for prettier

k0nserv: 👍

PR #226, opened 2017-10-25
[BPK-964] Upgrade eslint-config-skyscanner

shaundon:
Upgrade `eslint-config-skyscanner` and remove rule opt-outs as they're in there now.


matthewdavidson:
The `package-lock.json` should have changed... I am surprised Danger never noticed this.

shaundon:
@matthewdavidson you're right - not sure why it wasn't, but I've fixed it now.

PR #227, opened 2017-10-26
Added bpk-component-slider.

raullm7:


packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,121 @@
...
+  className: '',
+  large: false,
+  valueTemplate: (min, max) => (
+    <div className={getClassName('bpk-slider__value')}>
+      <div className={getClassName('bpk-slider__value--min')}>{min}</div>
+      <span> - </span>
+      <div className={getClassName('bpk-slider__value--max')}>{max}</div>
+    </div>
+  ),
+  onChange: () => {},
```
matthewdavidson: Just make these default `null`?

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,121 @@
...
+  valueTemplate: PropTypes.func,
+  onAfterChange: PropTypes.func,
+};
+
+BpkSlider.defaultProps = {
+  step: 1,
+  minDistance: 1,
+  className: '',
+  large: false,
+  valueTemplate: (min, max) => (
```
matthewdavidson: We should avoid using the terminology `template` in react - everything should be a component. 

Can we rename this? Perhaps `ValueComponent` ?

matthewdavidson: Also the signature should be updated to conform to an actual component i.e. `(props) => <Component />`

matthewdavidson: In fact... I actually think that this "value template" area shouldn't be part of this component and should instead live in userland.

Apologies haha.

packages/bpk-component-slider/src/bpk-slider.scss
```diff
@@ -0,0 +1,89 @@
...
+$slider-control-inset-size: 20 * $bpk-one-pixel-rem;
+$slider-control-size-large: 42 * $bpk-one-pixel-rem;
+$slider-control-inset-size-large: 30 * $bpk-one-pixel-rem;
+
+.bpk-slider {
+  margin-bottom: $bpk-spacing-xxl;
+
+  &__range {
+    // eslint-disable-next-line selector-attribute-quotes
+    div[class$='1'] {
```
matthewdavidson: Can you explain these styles? Are you not able to apply classes to these elements you are targeting? I there any other way you can achieve the same result?

packages/bpk-component-slider/src/bpk-slider.scss
```diff
@@ -0,0 +1,89 @@
...
+.bpk-slider-bar {
+  position: relative;
+  top: $slider-control-size / 2.5;
+  height: $slider-track-size;
+  border-radius: $slider-track-size / 2;
+  background-color: $bpk-color-gray-100;
+}
+
+.bpk-slider-wrapper {
+  margin-bottom: $bpk-spacing-lg;
```
matthewdavidson: I would avoid imposing margin bottom to all instances of a slider - let userland control this behaviour.

packages/bpk-component-slider/stories.js
```diff
@@ -0,0 +1,75 @@
...
+ * limitations under the License.
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import BpkSlider from './index';
+
+const SliderContainer = props => (
```
matthewdavidson: This slider container appears to be redundant.

packages/bpk-docs/src/pages/SlidersPage/SlidersPage.js
```diff
@@ -0,0 +1,132 @@
...
+        )}
+      />,
+    ],
+  },
+];
+
+
+const SlidersPage = () => (
+  <DocsPageBuilder
+    title="Slider bars"
```
matthewdavidson: Can we align the page title to the nav used in the docs? i.e. `Sliders`

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,121 @@
...
+
+import STYLES from './bpk-slider.scss';
+
+const getClassName = cssModules(STYLES);
+
+class BpkSlider extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { value: props.value };
```
matthewdavidson: This component shouldn't have `state` - leave this to userland. As such, this should be refactored into a stateless functional component.

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,121 @@
...
+    const sliderClass = [];
+    const handleClass = [];
+    const barClass = [];
+
+    sliderClass.push(getClassName('bpk-slider'));
+    sliderClass.push(getClassName(this.state.value.length
+      ? 'bpk-slider__range'
+      : 'bpk-slider__simple'));
+    handleClass.push(getClassName('bpk-slider-handle'));
+    barClass.push(getClassName('bpk-slider-bar'));
```
matthewdavidson: This classnames logic can be simplified to something along the lines of this:
```
    const sliderClass = [getClassName('bpk-slider')];
    const handleClass = [getClassName('bpk-slider-handle')];

    sliderClass.push(getClassName(this.state.value.length ? 'bpk-slider__range' : 'bpk-slider__simple'));
    if (large) handleClass.push(getClassName('bpk-slider-handle--large'));
    if (className) sliderClass.push(getClassName(className));
```

In short... 
```
const myArray = [];
myArray.push(item)
```
can be:
```
const myArray = [item];
```

etc.

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,121 @@
...
+      onAfterChange,
+    } = this.props;
+
+    const sliderClass = [];
+    const handleClass = [];
+    const barClass = [];
+
+    sliderClass.push(getClassName('bpk-slider'));
+    sliderClass.push(getClassName(this.state.value.length
+      ? 'bpk-slider__range'
```
matthewdavidson: these classnames should be modifiers i.e. `bpk-slider--range` & `bpk-slider--simple`. Also you can default to simple and only add one class based on whether it's a range or not.

For reference:
```
.calculator {} // block 
.calculator--alternate {} // block modifier 
.calculator__definitions {} // element 
.calculator__definitions--expanded {} // element modifier
```

packages/bpk-component-slider/readme.md
```diff
@@ -0,0 +1,43 @@
...
+
+```js
+import React, { Component } from 'react';
+import BpkSlider from 'bpk-component-slider';
+
+const Slider = () => (
+  <BpkSlider
+    min={0}
+    max={100}
+    value={[20, 80]}
```
shaundon: What happens if `value` is given an array with more than two entries?

packages/bpk-component-slider/readme.md
```diff
@@ -0,0 +1,43 @@
...
+
+```
+
+## Props
+
+| Property                 | PropType                      | Required | Default Value |
+| ------------------------ | ----------------------------- | -------- | ------------- |
+| min                      | number                        | true     | -             |
+| max                      | number                        | true     | -             |
+| value                    | number or list                | true     | -             |
```
shaundon: I think this should say `number or array` to match the official names in PropTypes.

packages/bpk-component-slider/src/BpkSlider.js
raullm7: Done

packages/bpk-component-slider/src/BpkSlider.js
raullm7: Done

packages/bpk-component-slider/src/bpk-slider.scss
raullm7: Cool, I have removed it

packages/bpk-component-slider/src/bpk-slider.scss
raullm7: The problem here is that the slider component adds an index to the end of the class which I pass. So the actual class would be `bpk-slider--range + hash + index`
Using css modules I cannot think of a better solution to identify the ranges (the horizontal bars in the slider) individually.

packages/bpk-component-slider/src/BpkSlider.js
raullm7: Done

packages/bpk-component-slider/readme.md
raullm7: It would have more than 3 handlers, I have not made CSS style for the ranges (horizontal bars in the slider) in case there are more than 3 (See picture). Should we make CSS for this situation or just avoid the user from using arrays with more than two entries?

screen shot 2017-10-30 at 1 56 09 pm

packages/bpk-component-slider/readme.md
```diff
@@ -0,0 +1,43 @@
...
+| min                      | number                        | true     | -             |
+| max                      | number                        | true     | -             |
+| value                    | number or array               | true     | -             |
+| step                     | number                        | true     | -             |
+| minDistance              | number                        | false    | 1             |
+| className                | string                        | false    | null          |
+| large                    | bool                          | false    | false         |
+| onChange                 | func                          | false    | null          |
+| onAfterChange            | func                          | false    | null          |
+| valueTemplate            | func                          | false    | null          |
```
matthewdavidson: Remember and keep this up to date with the `propTypes`

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,91 @@
...
+  const handleClass = [getClassName('bpk-slider-handle')];
+  const barClass = [getClassName('bpk-slider-bar')];
+
+  if (value.length) sliderClass.push(getClassName('bpk-slider--range'));
+  if (className) sliderClass.push(getClassName(className));
+  if (large) handleClass.push(getClassName('bpk-slider-handle--large'));
+
+  const handleActiveClass = getClassName('bpk-slider-handle--active');
+  return (
+    <div className={getClassName('bpk-slider-wrapper')}>
```
matthewdavidson: Is this wrapper element required? what happens if you remove it?

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,91 @@
...
+
+  const sliderClass = [getClassName('bpk-slider')];
+  const handleClass = [getClassName('bpk-slider-handle')];
+  const barClass = [getClassName('bpk-slider-bar')];
+
+  if (value.length) sliderClass.push(getClassName('bpk-slider--range'));
+  if (className) sliderClass.push(getClassName(className));
+  if (large) handleClass.push(getClassName('bpk-slider-handle--large'));
+
+  const handleActiveClass = getClassName('bpk-slider-handle--active');
```
matthewdavidson: this classname can be inlined on line 61

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,91 @@
...
+    step,
+    large,
+    minDistance,
+    className,
+    onChange,
+    onAfterChange,
+  } = props;
+
+  const sliderClass = [getClassName('bpk-slider')];
+  const handleClass = [getClassName('bpk-slider-handle')];
```
matthewdavidson: the handle and bar elements seem to be children of `bpk-slider` - this means that they should have classnames of `bpk-slider__handle` & `bpk-slider__bar` respectively

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,91 @@
...
+  onAfterChange: PropTypes.func,
+};
+
+BpkSlider.defaultProps = {
+  step: 1,
+  minDistance: 1,
+  className: '',
+  large: false,
+  onChange: null,
+  onAfterChange: () => {},
```
matthewdavidson: This default should also be `null`

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,91 @@
...
+  className: PropTypes.string,
+  large: PropTypes.bool,
+  onChange: PropTypes.func,
+  onAfterChange: PropTypes.func,
+};
+
+BpkSlider.defaultProps = {
+  step: 1,
+  minDistance: 1,
+  className: '',
```
matthewdavidson: this default should be `null`

packages/bpk-component-slider/readme.md
shaundon: Good question! I think it's probably best to give users the freedom to use more than two for any edge cases where that's necessary, in which case there should probably be some styles applied to the track between them. I think keeping it blue is fine though.

matthewdavidson: I agree with @shaun - if it's easy to add and support is there then we should do it. However I wouldn't necessarily block this PR on this - it can be added at a later date/PR.

matthewdavidson:
@raullm7 remember your danger notices - one file doesn't have the license header and you still need to add a changelog entry :)

packages/bpk-component-slider/src/BpkSlider.js
raullm7: If I set to `null` I get `_onAfterChange is not a function`

matthewdavidson: Looks like it's actually required then ... ? Did you mean to make it required? :)

Perhaps if it's not really required then you should fix the usage of `onAfterChange` and check if it is defined.

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,91 @@
...
+        min={min}
+        step={step}
+        value={value}
+        minDistance={minDistance}
+        withBars
+        className={sliderClass.join(' ')}
+        handleClassName={handleClass.join(' ')}
+        handleActiveClassName={handleActiveClass}
+        barClassName={barClass.join(' ')}
+        onChange={v => onChange(v)}
```
k0nserv: These two extra anonymous functions are redundant. Could just be

```javascript
onChange={onChange}
onAfterChange={onAfterChange}
```

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,91 @@
...
+    className,
+    onChange,
+    onAfterChange,
+  } = props;
+
+  const sliderClass = [getClassName('bpk-slider')];
+  const handleClass = [getClassName('bpk-slider-handle')];
+  const barClass = [getClassName('bpk-slider-bar')];
+
+  if (value.length) sliderClass.push(getClassName('bpk-slider--range'));
```
k0nserv: Don't use bracket-less ifs

packages/bpk-component-slider/src/BpkSlider.js
```diff
@@ -0,0 +1,91 @@
...
+      />
+    </div>
+  );
+};
+
+BpkSlider.propTypes = {
+  min: PropTypes.number.isRequired,
+  max: PropTypes.number.isRequired,
+  minDistance: PropTypes.number,
+  value: PropTypes.oneOfType([PropTypes.array, PropTypes.number]).isRequired,
```
k0nserv: `PropTypes.array` should be `PropTypes.arrayOf(PropTypes.number)` right?

raullm7: Yes, I have changed it

packages/bpk-component-slider/src/BpkSlider.js
raullm7: Done

matthewdavidson:
Once we fix the build this is good to merge.

packages/bpk-component-slider/package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^15.0.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "bpk-mixins": "^17.0.7",
+    "bpk-react-utils": "^2.3.6",
+    "react-addons-shallow-compare": "^15.4.2",
```
matthewdavidson: this is unused!

packages/bpk-component-slider/src/BpkSlider-test.js
```diff
@@ -0,0 +1,107 @@
...
+        max={100}
+        value={[10, 90]}
+        minDistance={20}
+        large
+      />,
+    ).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('should render correctly with a value template', () => {
```
matthewdavidson: this test is redundant now!

PR #229, opened 2017-10-27
WIP: [BPK-509] Add dismissible config to web banner alerts

georgegillams:
dismissable
fade_in

### DIFF
--- a/packages/bpk-animate-height/readme.md
+++ b/packages/bpk-animate-height/readme.md
@@ -5,6 +5,8 @@
 Note: This was forked from https://github.com/Stanko/react-animate-height. We have added functionality to
 set the display property of the container to `display: none;` when the height is equal to 0.
 
+More information on the easing values can be found at [http://easings.net/](http://easings.net/)
+
 ## Installation
 
 ```sh
@@ -43,10 +45,11 @@ class AnimateHeightContainer extends Component {
 
 ## Props
 
-| Property    | PropType         | Required | Default Value |
-| ----------- | ---------------- | -------- | ------------- |
-| children    | node             | true     | -             |
-| duration    | number           | true     | -             |
-| height      | string or number | true     | -             |
-| easing      | string           | false    | 'ease'        |
-| style       | object           | false    | {}            |
+| Property            | PropType         | Required | Default Value |
+| ------------------- | ---------------- | -------- | ------------- |
+| children            | node             | true     | -             |
+| duration            | number           | true     | -             |
+| height              | string or number | true     | -             |
+| easing              | string           | false    | 'ease'        |
+| onAnimationComplete | function         | false    | null          |
+| style               | object           | false    | {}            |
--- a/packages/bpk-animate-height/src/AnimateHeight.js
+++ b/packages/bpk-animate-height/src/AnimateHeight.js
@@ -123,6 +123,9 @@ class AnimateHeight extends Component {
   onTransitionEnd() {
     if (this.contentElement && this.props.height === 0) {
       this.contentElement.style.display = 'none';
+      if (typeof this.props.onAnimationComplete === 'function') {
+        this.props.onAnimationComplete();
+      }
     }
   }
 
@@ -131,6 +134,7 @@ class AnimateHeight extends Component {
       children,
       duration,
       easing,
+      onAnimationComplete,
       style,
       ...rest
     } = this.props;
@@ -177,12 +181,14 @@ AnimateHeight.propTypes = {
   ]).isRequired,
   easing: PropTypes.string,
   transitionOverflow: PropTypes.string,
+  onAnimationComplete: PropTypes.func,
   style: PropTypes.object, // eslint-disable-line react/forbid-prop-types
 };
 
 AnimateHeight.defaultProps = {
   easing: 'ease',
   transitionOverflow: 'hidden',
+  onAnimationComplete: null,
   style: {},
 };
 
--- a/packages/bpk-component-banner-alert/package-lock.json
+++ b/packages/bpk-component-banner-alert/package-lock.json
@@ -31,7 +31,7 @@
 				"object-assign": "4.1.1",
 				"promise": "7.3.1",
 				"setimmediate": "1.0.5",
-				"ua-parser-js": "0.7.14"
+				"ua-parser-js": "0.7.17"
 			}
 		},
 		"iconv-lite": {
@@ -104,9 +104,9 @@
 			"integrity": "sha1-KQy7Iy4waULX1+qbg3Mqt4VvgoU="
 		},
 		"ua-parser-js": {
-			"version": "0.7.14",
-			"resolved": "https://registry.npmjs.org/ua-parser-js/-/ua-parser-js-0.7.14.tgz",
-			"integrity": "sha1-EQ1T+kw/MmwSEpK76skE0uAzh8o="
+			"version": "0.7.17",
+			"resolved": "https://registry.npmjs.org/ua-parser-js/-/ua-parser-js-0.7.17.tgz",
+			"integrity": "sha512-uRdSdu1oA1rncCQL7sCj8vSyZkgtL7faaw9Tc9rZ3mGgraQ7+Pdx7w5mnOSF3gw9ZNG6oc+KXfkon3bKuROm0g=="
 		},
 		"whatwg-fetch": {
 			"version": "2.0.3",
--- a/packages/bpk-component-banner-alert/readme.md
+++ b/packages/bpk-component-banner-alert/readme.md
@@ -34,5 +34,9 @@ export default () => (
 | message           | node                 | true     | -             |
 | ariaLive          | ARIA_LIVE (one of)   | false    | 'assertive'   |
 | children          | node                 | false    | null          |
-| toggleButtonLabel | string               | false    | null          |
 | className         | string               | no       | null          |
+| dismissable       | bool                 | false    | false         |
+| dismissLabel      | string               | false    | null          |
+| fadeIn            | bool                 | false    | false         |
+| onAction          | func                 | false    | null          |
+| toggleButtonLabel | string               | false    | null          |
--- a/packages/bpk-component-banner-alert/src/AnimateAndFade.js
+++ b/packages/bpk-component-banner-alert/src/AnimateAndFade.js
@@ -1,0 +1,136 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { cssModules } from 'bpk-react-utils';
+import BpkAnimateHeight from 'bpk-animate-height';
+import { durationSm } from 'bpk-tokens/tokens/base.es6';
+import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup';
+
+import STYLES from './animate-and-fade.scss';
+
+const getClassName = cssModules(STYLES);
+
+const animationDuration = parseInt(durationSm, 10);
+
+class AnimateAndFade extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isExpanded: this.props.show,
+      visible: this.props.show,
+      hideAnimationInProgress: false,
+      inDom: true,
+    };
+
+    this.onFadeComplete = this.onFadeComplete.bind(this);
+    this.onAnimateHeightComplete = this.onAnimateHeightComplete.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.show === this.props.show) {
+      return;
+    }
+
+    if (this.state.visible && this.state.isExpanded) {
+      this.setState({
+        hideAnimationInProgress: true,
+        visible: false,
+      });
+    } else if (!this.state.visible && !this.state.isExpanded) {
+      this.setState({
+        inDom: true,
+        isExpanded: true,
+        visible: true,
+      });
+    }
+  }
+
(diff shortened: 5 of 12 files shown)


packages/bpk-component-banner-alert/package-lock.json
```diff
@@ -1,17 +1,153 @@
 {
-	"requires": true,
+	"name": "bpk-component-banner-alert",
```
matthewdavidson: There should be no change to this file as you haven't modified `package.json`. Can you perhaps attempt to diagnose why this happens?

cc @matteo-hertel i've seen this in other MR's and I suspect that it'll come up as part of the node 8 / npm5 migration that we'll need to have an explanation / workaroudn for... 🤔

packages/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -0,0 +1,162 @@
...
+import BpkAnimateHeight from 'bpk-animate-height';
+import { durationBase } from 'bpk-tokens/tokens/base.es6';
+import { cssModules } from 'bpk-react-utils';
+import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup';
+
+import STYLES from './bpk-banner-alert.scss';
+
+const getClassName = cssModules(STYLES);
+
+const animateHeightDuration = parseInt(durationBase, 10);
```
matthewdavidson: Could just be `const duration = parseInt(durationBase, 10);` and reused in both places?

packages/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  render() {
+    const { children, onEnter, onLeave } = this.props;
+    const showPlaceholder = !this.state.visible && !this.state.hiding;
+
+    const fadingComponent = (
+      <CSSTransitionGroup
+        transitionName={{
+          leave: getClassName('bpk-banner-alert--leave'),
```
matthewdavidson: These class names don't match the component, they should be something along the lines of `bpk-animate-and-fade--leave`

packages/bpk-component-banner-alert/src/AnimateAndFade.js
matthewdavidson: Actually, shouldn't we be getting the animate height duration from the animate height component? Ideally through a token? At present it's best guess and could change in either instance.

georgegillams: Yes we could.
Our value for `animateHeightDuration` seen here is passed to the animate-height component, so unless we change the value here it won't change!

packages/bpk-component-banner-alert/package-lock.json
matteo-hertel: Very good point, let me investigate

packages/bpk-animate-height/readme.md
```diff
@@ -43,10 +43,11 @@ class AnimateHeightContainer extends Component {
...
-| duration    | number           | true     | -             |
-| height      | string or number | true     | -             |
-| easing      | string           | false    | 'ease'        |
-| style       | object           | false    | {}            |
+| Property                 | PropType         | Required | Default Value |
+| -----------              | ---------------- | -------- | ------------- |
+| children                 | node             | true     | -             |
+| duration                 | number           | true     | -             |
+| height                   | string or number | true     | -             |
+| easing                   | string           | false    | 'ease'        |
```
matteo-hertel: Do we have a list of accepted values for the `easing` prop somewhere? It's worth adding it to the docs/readme IMHO

packages/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -0,0 +1,157 @@
...
+  onToggle() {
+    if (this.state.visible && this.state.atHeight) {
+      this.hide();
+    } else if (!this.state.visible && !this.state.atHeight) {
+      this.show();
+    }
+  }
+
+  onAnimateHeightComplete() {
+    console.log('onAnimateHeightComplete()');
```
matteo-hertel: this should probably go

packages/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -0,0 +1,157 @@
...
+    this.setState({
+      inDom: !this.state.hideAnimationInProgress,
+      hideAnimationInProgress: false,
+    });
+  }
+
+  onFadeComplete() {
+    console.log('onFadeComplete()');
```
matteo-hertel: same as above

packages/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -0,0 +1,157 @@
...
+    if (!this.state.visible && this.state.hideAnimationInProgress) {
+      this.setState({
+        atHeight: false,
+      });
+    }
+  }
+
+  show() {
+    console.log('show()');
```
matteo-hertel: this one as well

packages/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -0,0 +1,157 @@
...
+    this.setState({
+      inDom: true,
+      atHeight: true,
+      visible: true,
+    });
+  }
+
+  hide() {
+    console.log('hide()');
```
matteo-hertel: image

packages/bpk-animate-height/src/AnimateHeight.js
```diff
@@ -123,6 +123,9 @@ class AnimateHeight extends Component {
   onTransitionEnd() {
     if (this.contentElement && this.props.height === 0) {
       this.contentElement.style.display = 'none';
+      if (this.props.onAnimationComplete !== null) {
```
matteo-hertel: so this is technically valid because we're using react's `PropTypes.func` , but a better check would be to use `typeof`, you'll invoke `onAnimationComplete` so check if it's a function and invoke it

packages/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -0,0 +1,157 @@
...
+            onTransitionEnd={this.onFadeComplete}
+          >
+            {this.state.visible &&
+                  children
+                }
+          </CSSTransitionGroup>
+        </BpkAnimateHeight>
+        : null
+    );
+    /* eslint-enable */
```
matteo-hertel: is es-lint disabled for this file?

packages/bpk-component-banner-alert/package-lock.json
matteo-hertel: see https://github.com/Skyscanner/backpack/pull/247#discussion_r149629803 not sure what's causing it tough

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -32,6 +35,62 @@ sapien, et dapibus mi aliquet non. Pellentesque auctor sagittis lectus vitae rho
 ante in, vestibulum nulla.`;
 const richMessage = <span style={{ fontWeight: 700 }}>Successful alert with custom rendered message</span>;
 
+const topBannerAlertStyle = {};
+const bannerAlertStyle = { marginTop: spacingLg };
```
matthewdavidson: Can we be a bit smarter about these styles? I.e. separate out to separate `.scss` stylesheet and prevent the style from being applied on the first banner alert (i.e. `:not(:first)`)

packages/bpk-component-banner-alert/stories.js
```diff
@@ -39,6 +39,9 @@ storiesOf('bpk-component-banner-alert', module)
   .add('Success (long message)', () => (
     <BpkBannerAlert message={longMessage} type={ALERT_TYPES.SUCCESS} />
   ))
+  .add('Success (dismissable)', () => (
+    <BpkBannerAlert dismissable message={message} type={ALERT_TYPES.SUCCESS} />
+  ))
   .add('Success with React rendered message', () => (
     <BpkBannerAlert message={<span style={{ fontWeight: 700 }}>{message}</span>} type={ALERT_TYPES.SUCCESS} />
```
matteo-hertel: Can you use the token `FONT_WEIGHT_BOLD` for the `fontWeight`

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -32,6 +35,62 @@ sapien, et dapibus mi aliquet non. Pellentesque auctor sagittis lectus vitae rho
 ante in, vestibulum nulla.`;
 const richMessage = <span style={{ fontWeight: 700 }}>Successful alert with custom rendered message</span>;
```
matteo-hertel: same here, can you please use the token `FONT_WEIGHT_BOLD` for the `fontWeight`?

packages/bpk-animate-height/readme.md
georgegillams: Are you thinking something like this?
https://github.com/Skyscanner/backpack/pull/229/files#diff-9e048b7a268bb81c6f466782a011779dR8

matteo-hertel: Sorry I didn't see that bit, ignore me!

georgegillams: That's because I only just added it!

matteo-hertel: ah! In which case don't ignore me 😄 but yes perfect, that's what I was talking about

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
```diff
@@ -79,42 +82,98 @@ const ToggleButton = (props) => {
...
+  componentDidMount() {
+    this.onShown();
+  }
+
+  onToggleExpand() {
     this.setState(state => ({
       expanded: !state.expanded,
     }));
+    if (this.props.onAction !== null) {
+      this.props.onAction();
```
matteo-hertel: for consistency's sake if you're using the `typeof  === 'function'` check, use it everywhere there is a similar call

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
```diff
@@ -150,6 +210,12 @@ class BpkBannerAlert extends Component {
         </BpkAnimateHeight>
       </section>
     );
+
+    return (
+      <AnimateAndFade className={className} animateOnEnter={fadeIn} animateOnLeave={dismissable} show={shown}>
+        {bannerAlert}
+      </AnimateAndFade>
+    );
     /* eslint-enable */
```
matteo-hertel: Not sure why this one keeps popping up, are we disabling es-link at some point in this component?

packages/bpk-component-banner-alert/stories.js
```diff
@@ -39,8 +40,14 @@ storiesOf('bpk-component-banner-alert', module)
   .add('Success (long message)', () => (
     <BpkBannerAlert message={longMessage} type={ALERT_TYPES.SUCCESS} />
   ))
+  .add('Success (dismissable)', () => (
+    <BpkBannerAlert dismissable message={message} type={ALERT_TYPES.SUCCESS} />
+  ))
   .add('Success with React rendered message', () => (
-    <BpkBannerAlert message={<span style={{ fontWeight: 700 }}>{message}</span>} type={ALERT_TYPES.SUCCESS} />
+    <BpkBannerAlert
+      message={<span style={{ fontWeight: fontWeightBold }}>{message}</span>}
```
matteo-hertel: super awesome!! 👍

packages/bpk-docs/src/pages/BannerAlertsPage/bpk-banner-alers-page.scss
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-banner-alers-page {
```
shaundon: Typo - this should be `alerts`. Likewise for the filename and the import in the docs page.

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
georgegillams: Yup - that was already done here before - https://github.com/Skyscanner/backpack/blob/fd973903db3066106917c45660133537ae36ab7f/packages/bpk-component-banner-alert/src/BpkBannerAlert.js#L182

It was disabled as es-lint is picking up on accessibility issues with the touchable highlight, but the touchable highlight is "inaccessible" by design as the chevron button provides the necessary accessibility 😸

packages/bpk-component-banner-alert/readme.md
```diff
@@ -34,5 +34,9 @@ export default () => (
 | message           | node                 | true     | -             |
 | ariaLive          | ARIA_LIVE (one of)   | false    | 'assertive'   |
 | children          | node                 | false    | null          |
-| toggleButtonLabel | string               | false    | null          |
 | className         | string               | no       | null          |
+| dismissable       | bool                 | false    | false         |
+| dismissLabel      | string               | false    | null          |
```
matthewdavidson: Can we rename `dismissLabel` to `dismissButtonLabel `?

packages/bpk-component-banner-alert/readme.md
```diff
@@ -34,5 +34,9 @@ export default () => (
 | message           | node                 | true     | -             |
 | ariaLive          | ARIA_LIVE (one of)   | false    | 'assertive'   |
 | children          | node                 | false    | null          |
-| toggleButtonLabel | string               | false    | null          |
 | className         | string               | no       | null          |
+| dismissable       | bool                 | false    | false         |
```
matthewdavidson: This `dismissable` prop is a little problematic i.e. what happens if consumers provide `children` and `dismissable` - does the toggle button also show?

Perhaps they should be different components - I'm not sure what the best approach is here.

matthewdavidson: I think we also need to contemplate the API here (also in terms of aligning it to the RN component when we do the work).

Can we provide consumers with an `onDismissed` and a `dismissed` prop and allow them to handle the show/hide state

matthewdavidson: Have a look at some of the examples out in the wild https://js.coach/react/?search=alert

A common approach seems to be to have a container component which renders nothing by default. Consumers add / remove alerts to the container by calling a static `add` function (an example exists in https://js.coach/react/react-alert?search=alert).

PR #230, opened 2017-10-30
[BPK-971] Upgrade enzyme dependencies

matthewdavidson:
This is the last of a few PR's aimed at upgrading dependencies in the backpack monorepo. I focus on updating enzyme in this one.

Please have a read of their upgrade guide: https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md

It details the significant changes of version 3 vs 2. We've had to update some of our tests as a result.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,15 +69,16 @@
     "css-loader": "^0.28.4",
     "danger": "^1.2.0",
     "del": "^3.0.0",
-    "enzyme": "^2.8.2",
-    "enzyme-to-json": "^1.5.1",
+    "enzyme": "^3.1.0",
+    "enzyme-adapter-react-15": "^1.0.3",
+    "enzyme-to-json": "^3.1.4",
     "eslint": "^4.9.0",
     "eslint-config-airbnb": "^16.1.0",
     "eslint-config-skyscanner": "^3.0.0",
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jsx-a11y": "^6.0.2",
     "eslint-plugin-react": "^7.4.0",
-    "extract-text-webpack-plugin": "^3.0.1",
+    "extract-text-webpack-plugin": "^3.0.2",
     "file-loader": "^1.1.5",
     "gulp": "^3.9.1",
     "gulp-chmod": "^2.0.0",
@@ -124,6 +125,9 @@
   },
   "jest": {
     "verbose": true,
+    "setupFiles": [
+      "<rootDir>/scripts/jest/setup.js"
+    ],
     "testRegex": "packages/.*-test\\.js$",
     "testPathIgnorePatterns": [
       "native/"
--- a/packages/bpk-component-calendar/src/BpkCalendarContainer-test.js
+++ b/packages/bpk-component-calendar/src/BpkCalendarContainer-test.js
@@ -68,13 +68,15 @@ describe('BpkCalendarContainer', () => {
       selectedDate={new Date(2010, 1, 15)}
     />);
 
-    const grid = calendar.find('BpkCalendarGridTransition');
+    let grid = calendar.find('BpkCalendarGridTransition');
     const nav = calendar.find('BpkCalendarNav');
     const eventStub = { persist: jest.fn() };
 
     expect(grid.prop('month')).toEqual(new Date(2010, 1, 1));
 
     nav.prop('onMonthChange')(eventStub, { month: new Date(2010, 2, 1) });
+    calendar.update();
+    grid = calendar.find('BpkCalendarGridTransition');
     expect(grid.prop('month')).toEqual(new Date(2010, 2, 1));
   });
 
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -124,6 +125,9 @@
   },
   "jest": {
     "verbose": true,
+    "setupFiles": [
+      "<rootDir>/scripts/jest/setup.js"
```
matthewdavidson: We now require a setup file for enzyme to configure the react 15 adaptor.

packages/bpk-component-calendar/src/BpkCalendarContainer-test.js
```diff
@@ -68,13 +68,15 @@ describe('BpkCalendarContainer', () => {
...
 
-    const grid = calendar.find('BpkCalendarGridTransition');
+    let grid = calendar.find('BpkCalendarGridTransition');
     const nav = calendar.find('BpkCalendarNav');
     const eventStub = { persist: jest.fn() };
 
     expect(grid.prop('month')).toEqual(new Date(2010, 1, 1));
 
     nav.prop('onMonthChange')(eventStub, { month: new Date(2010, 2, 1) });
+    calendar.update();
```
matthewdavidson: See [this item](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#for-mount-updates-are-sometimes-required-when-they-werent-before) to get an idea of why we now have to call .update()` on the root wrapper.

packages/bpk-component-calendar/src/BpkCalendarGridTransition-test.js
```diff
@@ -47,11 +47,6 @@ describe('BpkCalendar', () => {
 
     // Next month
     calendar.setProps({ month: new Date(Date.UTC(2010, 2)) });
-    expect(calendar.state('currentMonth')).toEqual(new Date(Date.UTC(2010, 1)));
```
matthewdavidson: Enzyme appears to be calling `onMonthTransitionEnd` automatically now.

packages/bpk-component-image/src/withLazyLoading-test.js
```diff
@@ -20,30 +20,37 @@ import React from 'react';
...
   });
 
   it('should add inView prop', () => {
-    const documentMock = jest.fn();
+    const documentMock = createDocumentMock();
     const MockImageComponent = () => (<div>Fake Image</div>);
     const LazyLoadedImage = withLazyLoading(MockImageComponent, documentMock);
 
-    const lazyLoadedImage = shallow(<LazyLoadedImage />);
+    const lazyLoadedImage = shallow(<LazyLoadedImage />, { disableLifecycleMethods: true });
```
matthewdavidson: Enzyme previously didn't call lifecycle methods such as `componentDidMount` and as such these tests relied on this behaviour. See [this](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#lifecycle-methods)

packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -19,7 +19,7 @@
 import React from 'react';
 import { wrapDisplayName } from 'bpk-react-utils';
 
-export default function withLazyLoading(Component, document) {
+export default function withLazyLoading(Component, doc) {
```
k0nserv: Why change it from document?

matthewdavidson: Ah right yeah good question - so when myself and Shaun were debugging this we initially thought that document was being used from the global scope instead of the scope of the HOC (it wasn't). To make it very explicit we opted to rename it so as that confusion can be avoided going forward.

shaundon: When was this? I don't remember debugging it with you 😂  but I have a memory like a sieve, so..

matthewdavidson: 😂  maybe I am lying

k0nserv: I feel the opposite `doc` is unclear and makes me wonder what the variable actually is. `document` is very clear and it's being shadowed so still works as expected

packages/bpk-component-calendar/src/BpkCalendarContainer-test.js
shaundon: I'm getting flashbacks of using Angular and having to put `$scope.$digest()` everywhere! But yeah - I see why this is necessary and the explanation you linked to makes sense.

packages/bpk-component-image/src/withLazyLoading.js
matthewdavidson: Ok ill revert!

matthewdavidson:
@k0nserv can i have approval? 😇  🙏

matthewdavidson:
@shaundon yeah I reckon enzymes shallo renderer is now including `key` props in it's output :thumbsup:

PR #235, opened 2017-11-01
WIP: [BPK-1087] Fix popover "on close" element focus behaviour

matthewdavidson:
We use a utility called `a11y-focus-store` to help with managing focus when displaying popovers and modals to the user. Before a popover/modal is triggered, we need to remember the users current focus at that point so that we can refocus that element when the popover/modal is closed.

Consider the following scenario:
1. Two popovers on a page:
two popovers on page

2. User triggers first popover:
trigger first popover

3. Whilst first popover is still open, user triggers second popover:
trigger second popover whilst first is still open

4. User clicks outside to dismiss popover, but wrong target element is focused:
on click outside wrong popover target is focused

5. This PR ensures that the correct target element is focused:
correct behaviour

### DIFF
--- a/packages/bpk-component-popover/stories.js
+++ b/packages/bpk-component-popover/stories.js
@@ -79,7 +79,12 @@ class PopoverContainer extends Component {
   }
 
   render() {
-    const { targetFunction, changeProps, ...rest } = this.props;
+    const {
+      targetFunction,
+      changeProps,
+      id,
+      ...rest
+    } = this.props;
     let target = null;
     let openButton = <BpkButton onClick={this.openPopover}> Open </BpkButton>;
 
@@ -95,7 +100,7 @@ class PopoverContainer extends Component {
       <div>
         {openButton}
         <BpkPopover
-          id="my-popover"
+          id={`my-popover-${id}`}
           target={target}
           onClose={this.closePopover}
           isOpen={this.state.isOpen}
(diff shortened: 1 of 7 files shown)


packages/bpk-component-popover/stories.js
```diff
@@ -141,25 +147,34 @@ PopoverContainer.defaultProps = {
...
+);
+
 storiesOf('bpk-component-popover', module)
   .add('Default', () => (
-    <div style={{ height: '1000px', margin: '30px', textAlign: 'center' }}>
-      <PopoverContainer />
-    </div>
+    <Spacer>
+      <PopoverContainer id="my-popover-1" />
+      <PopoverContainer id="my-popover-2" />
```
shaundon: Is the addition of `id` something that the new `eslint` brought in?

packages/bpk-component-popover/stories.scss
```diff
@@ -3,3 +3,11 @@
 .bpk-popover-paragraph {
   margin-bottom: $bpk-spacing-xs;
 }
+
+.bpk-popover-spacer {
+  display: flex;
+  height: 500 $bpk-one-pixel-rem;
```
shaundon: Should there be a multiplication symbol here?

packages/bpk-component-popover/stories.js
matthewdavidson: Nah I did that because there are two on the page and I wanted them to have different `id`'s

packages/bpk-component-popover/stories.scss
matthewdavidson: Good spot!

matthewdavidson:
I've published a pre-release version of these changes to test on internal projects - don't mind the version number changes.

PR #237, opened 2017-11-01
[WPT-1763] Rename to bpk-scrim with a view to move into own package

danrr:
I'm splitting up the work into a bunch of stand-alone-ish MRs for easier review

This just renames the scrim and all the variables it depends on to eliminate references to the modal component as it won't be the only one to use this. There are a lot of files here but most are generated from variables in json files.


packages/bpk-tokens/src/web/base/aliases.json
```diff
@@ -33,7 +33,7 @@
     "ZINDEX_AUTOSUGGEST": "900",
     "ZINDEX_POPOVER": "900",
     "ZINDEX_TOOLTIP": "900",
-    "ZINDEX_MODAL_SCRIM": "1000",
+    "ZINDEX_SCRIM": "1000",
```
matthewdavidson: Unfortunately I dont think it's worth renaming tokens as part of this change - tokens are a public API in `bpk-tokens` & `bpk-mixins` and we cannot be sure of who is using what. Ok to keep them as is? Either that or we deprecate the existing names and duplicate them with new names.

danrr: Right, I see

I'm happy to go down the duplicate and deprecate route, unless you think it's too much of a pain?

matthewdavidson: Nah I think thats OK - we haven't done it before, but we need some sort of policy for deprecating older tokens. There is a meta property on tokens that we could assign a `deprecated: true,` property to.

danrr: I've updated the files, so most variables are back. This alias though gets output directly in some of the files (eg. packages/bpk-tokens/tokens/travelpro.raw.json) and if I add it back I don't think it can be properly marked as deprecated. Should I leave it as is?

changelog.md
```diff
@@ -5,6 +5,17 @@
 **Added:**
 - New slider component, see http://backpack.prod.aws.skyscnr.com/components/web/sliders
 
+**Changed:**
+- bpk-component-modal:
+    - Renamed `bpk-modal-scrim` to `bpk-scrim`, and `BpkModalScrim` to `BpkScrim`
```
matthewdavidson: Typically we wouldn't list internal changes like this in the changelog - only public API changes or behaviour that consumers are interested in.

changelog.md
```diff
@@ -5,6 +5,17 @@
...
+- bpk-tokens and bpk-mixins:
```
matthewdavidson: We haven't yet deprecated these yet - we've simply marked them for deprecation. At some point in the future we'll do a sweep of all the deprecated tokens and release a MAJOR version of `bpk-tokens` etc.

Essentially I dont think you need a changelog entry for this PR at all 😇 .

packages/bpk-tokens/src/web/base/aliases.json
matthewdavidson: Hmm good point - let me get back to you...

matthewdavidson: We've agreed that it's not part of the public interface, so this is fine!

danrr: cool

changelog.md
danrr: Removed

PR #238, opened 2017-11-01
Use node LTS

matthewdavidson:
cc @matteo-hertel 

So the policy I have opted for here is to stick to `lts/*` which means folks will always be on the latest LTS version of node (and opt in to bug and security patches). I use `package.json`'s `engines` field to record the semver range of node / npm versions that backpack is supposed to be developed on - this is in turn used in `check-versions.js` to ensure that folks aren't on really old incompatible versions. I think this is probably a decent compromise as opposed to strictly locking down versions. 

WDYR?


shaundon:
@matteo-hertel as I understand it, that's correct.

screen shot 2017-11-01 at 14 14 13

matteo-hertel:
Awesome, I just wanted to make sure 
`the constraint will be satisfied by any version matching >=8.9.0 <9.0.0`

once the semver import is fixed, it's good to go and make the world a better place

PR #239, opened 2017-11-01
[BPK-1083] New spinner component

shaundon:
Docs and screenshots to be added in a future PR.

Here's some screenshots of the 'all types' story:

simulator screen shot 1 nov 2017 14 06 18

screenshot_1509545233

The themed one doesn't look very different from the dark one right now because it's using `#2d244c` (same as the themed button), but this may change in future based on feedback from @jamesf3rguson.

### DIFF
--- a/native/packages/react-native-bpk-component-spinner/src/BpkSpinner.js
+++ b/native/packages/react-native-bpk-component-spinner/src/BpkSpinner.js
@@ -1,0 +1,103 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  ActivityIndicator,
+  Platform,
+} from 'react-native';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withTheme } from 'react-native-bpk-theming';
+
+const tokens = Platform.select({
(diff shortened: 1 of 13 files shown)


native/packages/react-native-bpk-component-spinner/src/BpkSpinner.js
```diff
@@ -0,0 +1,103 @@
...
+  // supplied. If not, disable theming.
+  if (theme && (type !== 'primary' || !theme[THEMING_ATTRIBUTE])) {
+    theme = null;
+  }
+
+  return (
+    <ActivityIndicator
+      color={getSpinnerColor(theme, type)}
+      size={small ? 'small' : 'large'}
+      {...rest}
```
matteo-hertel: Because rest is passed here perhaps it's worth to add in the readme that this component can accept any pops of ActivityIndicator https://facebook.github.io/react-native/docs/activityindicator.html, what do you think?

shaundon: Yep absolutely. I meant to put this and then forgot..

shaundon: Actually, I just looked at buttons and that information is conveyed in the docs page, not the readme (see [here](http://backpack.prod.aws.skyscnr.com/components/native/button)), so let's stick to that approach here for consistency.

native/packages/react-native-bpk-component-spinner/stories.js
```diff
@@ -0,0 +1,107 @@
...
+    backgroundColor: tokens.colorGray900,
+    padding: tokens.spacingBase,
+  },
+  bottomMargin: {
+    marginBottom: tokens.spacingBase,
+  },
+});
+
+const themeAttributes = {
+  spinnerPrimaryColor: '#2d244c',
```
matteo-hertel: This is really being picky but can we have a different colour here, looks like there is no difference between this one  and the dark variant

matteo-hertel: and maybe a token colour rather than hardcoded 😄

shaundon: Soooo with the colour, this came from James. He's aware that it doesn't look very different but asked me to leave it as is, for now at least.

shaundon: A nice thing might be to have some sort of mock theme file that exposes all the necessary theme props for our storybooks, rather than each story file having to define it itself. I figured that's best left for a future PR though.

PR #242, opened 2017-11-02
Move scrim and scrolling behaviour out bpk-component-modal

danrr:
~Still early days for this branch but I wanted to get some feedback.~

I ~was thinking of creating~ made a higher order component that applies a scrim and the scrolling behaviour that currently lives in BpkModalDialog (componentDidMount, componentWillUnmount) to any component. That way this can be easily applied to another component (eg. overflow menu) that has similar requirements.


danrr:
The package will probably be renamed to bpk-overlay-utils.

packages/bpk-component-modal/src/BpkModalDialog-test.js
```diff
@@ -19,19 +19,30 @@
...
+      onMouseMove: jest.fn(),
+      onMouseUp: jest.fn(),
+    };
+  });
   it('should render correctly', () => {
     const tree = renderer.create(
       <BpkModalDialog
         id="my-modal"
         title="Modal title"
         onClose={() => null}
```
matteo-hertel: `jest.fn()` here will be more in line with the rest of the test file

packages/bpk-scrim-utils/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "bpk-scrim-utils",
+  "version": "1.0.0",
+  "description": "Higher order component that adds an overlay behind components and manages scroll states.",
```
matthewdavidson: This description seems overly focused on the one HOC, whereas the pakcage name is called suffixed `-utils` implying plural. Check out https://www.npmjs.com/package/bpk-react-utils (not that it's a good example, but it deals with the fact that there are multiple).

packages/bpk-scrim-utils/readme.md
```diff
@@ -0,0 +1,32 @@
...
+    {...props.closeEvents}
+  >
+    <BpkButton onClick={props.onClose}>Close</BpkButton>
+    Hello
+  </div>
+);
+
+const BoxWithScrim = withScrim(Box);
+```
+
```
matthewdavidson: I think there needs to be more docs here detailing the props that the HOC exposes to the wrapped component. `isIphone` is missing for example.

packages/bpk-component-modal/src/BpkModalDialog-test.js
danrr: done

matthewdavidson:
Have a look at the difference behaviour on mobile safari landscape:

Before:
img_0907

After:
img_0906

I suspect that the before isn't quite correct either (the whole screen should be white).

matthewdavidson:
I really need to stop pressing that button.

matthewdavidson:
Other than the issue above and the issue demonstrated offline (I was unable to share the video clip on github), this is good to go. I tested on chrome, FF, safari on OSX, chrome, IE11 & edge on windows, safari on iOS and chrome on android.

packages/bpk-scrim-utils/readme.md
danrr: done

PR #250, opened 2017-11-06
Exposing BpkInput underlying DOM node ref to parent components

marccerrato:
When using the component `BpkInput`, I would want to have access to the underlying `input` DOM node so that I can manage things like `focus` from a parent component.

The recommended way to do so in the React docs is by exposing a prop and attach it to the DOM node as a `ref` attribute: https://reactjs.org/docs/refs-and-the-dom.html#exposing-dom-refs-to-parent-components

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,9 @@
 - react-native-bpk-component-banner-alert:
   - Fix icon color of success alert.
   - User style now takes precedence over encapsualted style.
+  
+- bpk-component-input:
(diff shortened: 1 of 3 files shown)


matthewdavidson:
Hey @marccerrato ! Nice PR, can you add a changelog entry too please!

cc @ksiwei you were asking for this!

marccerrato:
Changelog updated! 👍

PR #254, opened 2017-11-09
[BPK-1039] Add native switch component

shaundon:
**iOS**
ios

**Android**
android

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,9 +5,13 @@
 **Added:**
 - bpk-scrim-utils:
   - New higher order component `withScrim`
-  
+
 - react-native-bpk-component-horizontal-nav:
   - New component `BpkHorizontalNav`. See http://backpack.prod.aws.skyscnr.com/components/native/horizontal-nav
+
+- react-native-bpk-component-switch
+  - New component `BpkSwitch`. See http://backpack.prod.aws.skyscnr.com/components/native/switch
+
 
 ## 2017-11-07 - New slider component
 
--- a/native/packages/react-native-bpk-component-switch/package.json
+++ b/native/packages/react-native-bpk-component-switch/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "react-native-bpk-component-switch",
+  "version": "0.0.0",
+  "main": "index.js",
+  "description": "Backpack React Native switch component.",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^16.0.0-alpha.1",
+    "react-native": ">= 0.47.0"
+  },
+  "dependencies": {
+    "bpk-tokens": "^26.4.5",
+    "prop-types": "^15.5.8",
+    "react-native-bpk-theming": "^1.0.13"
+  }
+}
--- a/native/packages/react-native-bpk-component-switch/readme.md
+++ b/native/packages/react-native-bpk-component-switch/readme.md
@@ -1,0 +1,48 @@
+# react-native-bpk-component-switch
+
+> Backpack React Native switch component.
+
+## Installation
+
+```sh
+npm install react-native-bpk-component-switch --save-dev
+```
+
+## Usage
+
+```js
+import React, { Component } from 'react';
+import { View, StyleSheet, Text } from 'react-native';
+import BpkSwitch from 'react-native-bpk-component-switch';
+import * as TOKENS from 'bpk-tokens/tokens/ios/base.react.native.es6';
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    justifyContent: 'center',
+    padding: TOKENS.spacingBase,
+  }
+});
+
+export default class App extends Component {
+  render() {
+    return (
+      <View style={styles.container}>
+        <BpkSwitch />
+        <BpkSwitch value />
+      </View >
+    );
+  }
+}
+```
+
+## Props
+
+| Property            | PropType                              | Required | Default Value |
+| -----------         | ------------------------------------- | -------- | ------------- |
+| value               | bool                                  | false    | false         |
+| theme               | See [Theme Props](#theme-props) below | false    | null          |
+
+## Theme Props
+
+* `switchPrimaryColor`
--- a/native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
+++ b/native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
@@ -1,0 +1,95 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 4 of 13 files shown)


native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
```diff
@@ -0,0 +1,94 @@
...
+
+const getColors = (theme, value) => {
+  const primaryColor = theme ? theme[THEMING_ATTRIBUTE] : tokens.colorBlue500;
+  const secondaryColor = tokens.colorGray100;
+
+  // The color props mean different things based on the platform.
+  const colors = Platform.select({
+    ios: {
+      tintColor: secondaryColor, // Border around the switch when off.
+      thumbTintColor: null, // Must be unset or the the handles loses its drop shadow.
```
shaundon: https://github.com/facebook/react-native/issues/16764

native/packages/react-native-bpk-component-switch/readme.md
```diff
@@ -0,0 +1,48 @@
...
+npm install react-native-bpk-component-switch --save-dev
+```
+
+## Usage
+
+```js
+import React, { Component } from 'react';
+import { View, StyleSheet, Text } from 'react-native';
+import BpkSwitch from 'react-native-bpk-component-switch';
+import * as TOKENS from 'bpk-tokens/tokens/ios/base.react.native.es6';
```
k0nserv: Can we destruct only the tokens we need instead?

native/packages/react-native-bpk-component-switch/readme.md
```diff
@@ -0,0 +1,48 @@
...
+    padding: TOKENS.spacingBase,
+  }
+});
+
+export default class App extends Component {
+  render() {
+    return (
+      <View style={styles.container}>
+        <BpkSwitch />
+        <BpkSwitch value />
```
k0nserv: I don't like `value` for the prop. Should be `on` IMO

native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
k0nserv: Can we include a reference to this issue in the source too?

native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
```diff
@@ -0,0 +1,94 @@
...
+
+  return <Switch {...getColors(theme, value)} value={value} {...rest} />;
+};
+
+const propTypes = {
+  theme: themePropType,
+  value: PropTypes.bool,
+};
+
+BpkSwitch.propTypes = propTypes;
```
k0nserv: Why are we diverging from just assigning the object directly here?

native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
```diff
@@ -0,0 +1,94 @@
...
+
+BpkSwitch.defaultProps = {
+  theme: null,
+  value: false,
+};
+
+export default withTheme(BpkSwitch);
+export {
+  propTypes,
```
k0nserv: Ah I see... Is this something we aim to always do for the components?

native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
shaundon: Sure

native/packages/react-native-bpk-component-switch/readme.md
shaundon: Even in the readme? We're not doing that in any other readmes.

native/packages/react-native-bpk-component-switch/readme.md
shaundon: Neither do I, but that's the same prop the underlying component uses and I figured it's nicer to use the same as them. Happy to change though if enough people disagree. Any thoughts @matthewdavidson @matteo-hertel?

native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
shaundon: Not always, but in the test we call one of our custom `propTypes` directly to test its functionality. 

Previously we could use `BpkSwitch.propTypes` to do this, but the theme provider that wraps the component strips off extra properties. So we decided to use this method to access them. We don't expose this on the external facing API, just internally.

native/packages/react-native-bpk-component-switch/readme.md
matteo-hertel: If the underlying component is using `value` I'd stick with it, we'll not add much value to the end user by changing it and this component will become an opinionated wrapper of the native switch

native/packages/react-native-bpk-component-switch/readme.md
k0nserv: Seems like a good thing to encourage imo

native/packages/react-native-bpk-component-switch/readme.md
k0nserv: Hmm. It just looks extremely strange when using the shorthand with this version, but maybe that's not common

native/packages/react-native-bpk-component-switch/src/BpkSwitch.js
k0nserv: Okay fair enough

native/packages/react-native-bpk-component-switch/readme.md
shaundon: I know what you mean, it does look weird! If it were `<BpkSwitch value={true} />` it might look less weird, but `eslint` corrects it to just `value`.

native/packages/react-native-bpk-component-switch/readme.md
shaundon: Fair enough, but if we're gonna do that we should probably change all the readmes at once in another PR.

matthewdavidson: Agreed 👍  can we create a task for this please? 😇

native/packages/react-native-bpk-component-switch/readme.md
matthewdavidson: Stick with `value` 👍

changelog.md
```diff
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+**Added:**
+- react-native-bpk-component-switch
+ - New component `BpkSwitch`. See http://backpack.prod.aws.skyscnr.com/components/native/switch
```
matthewdavidson: Can we have a newline here please 😇

native/packages/react-native-bpk-component-switch/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "react-native-bpk-component-switch",
+  "version": "1.0.1",
```
matthewdavidson: This should be `0.0.0`

shaundon: Argh! I copypasta'd this from the spinner component and I was so careful to scrub all mentions of it, then I missed this!

native/packages/react-native-bpk-component-switch/readme.md
shaundon: https://gojira.skyscanner.net/browse/BPK-1111 🙌

PR #256, opened 2017-11-09
[BPK-1025] Introduce react-native-bpk-component-ticket

matthewdavidson:
Added the react native ticket component. Note that theming props are left out for now - we'll decide later on whether we wanna make this a first class citizen or leave it to userland.

We use a 3rd party component for the dashed line - I've checked performance between ticket and card and it seems negligible (see the perf examples in the stories and the official docs on analysing frame rate https://facebook.github.io/react-native/docs/performance.html).


shaundon:
In the screenshots for the ticket, the content and stub are both the same lorem ipsum. I'm wondering if it would be clearer if the stub part was truncated, same as the examples on the web ticket are.

matthewdavidson:
Fixed the inconsistency in this PR and also raised https://github.com/Skyscanner/backpack/pull/261 👍

PR #260, opened 2017-11-10
[no jira] Make theming common amongst native stories

shaundon:
Removes duplication across our stories.

This is a precursor to a future move to make use of the new branding API.


native/storybook/ThemeStyles.js
```diff
@@ -0,0 +1,48 @@
+/*
```
matthewdavidson: This should be named `themeAttributes` - Upper camel case is reserved for components

PR #262, opened 2017-11-10
 [BPK-1084] Enhance RN BpkAnimateHeight component

georgegillams:
This PR contains the changes to BpkAnimateHeight which are required to address BPK-1084 (banner alert animations)

This presents the same issue as web animations when elements are removed from the DOM / Tree where the padding suddenly disappears causing a small jump. I think it best to address this later on. For now, user styling (such as padding) will be applied to the outer container until a better solution can be found.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -96,6 +96,10 @@
 - react-native-bpk-component-card: 1.1.4 => 1.1.5
   - Add support for android
 
+- react-native-bpk-component-animate-height:
+  - Added props for `expandDelay` and `collapseDelay`.
+  - Now removes itself from the tree when collapsed to `height: 0`.
+
 ## 2017-11-10 - New `withScrim` HOC, plus native horizontal nav and switch components
 
 **Added:**
--- a/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
+++ b/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
@@ -20,85 +20,104 @@ import {
   View,
   Platform,
   Animated,
-  ViewPropTypes,
 } from 'react-native';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { animationDurationSm } from 'bpk-tokens/tokens/base.react.native';
 
+const COLLAPSED_HEIGHT = 0.01;
+
 class BpkAnimateHeight extends React.Component {
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
 
     this.state = {
-      expanded: null,
-      expandedHeight: null,
-      collapsedHeight: 0,
-      height: new Animated.Value(1),
-      heightSet: false,
+      expanded: this.props.expanded,
+      height: new Animated.Value(COLLAPSED_HEIGHT),
+      inTree: this.props.expanded,
+      expandAnimationInProgress: false,
     };
+
+    // overflow: hidden is not properly supported on Android,
+    // which causes animation to look shoddy.
+    // See https://facebook.github.io/react-native/releases/0.49/docs/layout-props.html#overflow
+    this.animationDuration = Platform.OS === 'ios' ? props.animationDuration : 0;
 
     this.componentWillReceiveProps = this.componentWillReceiveProps.bind(this);
     this.setExpandedHeight = this.setExpandedHeight.bind(this);
-    this.resize = this.resize.bind(this);
+    this.onHeightAnimationComplete = this.onHeightAnimationComplete.bind(this);
+    this.performAnimation = this.performAnimation.bind(this);
   }
 
   componentWillReceiveProps(nextProps) {
     if (nextProps.expanded !== this.state.expanded) {
-      this.state.expanded = nextProps.expanded;
-      this.resize();
+      if (this.state.expanded) {
+        this.setState({ expanded: false });
+        this.performAnimation(COLLAPSED_HEIGHT, this.props.collapseDelay);
+      } else {
+        this.setState({ inTree: true, expandAnimationInProgress: true });
+      }
     }
   }
 
-  setExpandedHeight(event, expandedImmediately) {
-    const { height } = event.nativeEvent.layout;
-    this.setState({
-      // if the component has started expanded, we should now set the height of the container
-      // why on earth does a value of 1 instead of 0 make this break!
-      height: new Animated.Value(expandedImmediately ? height : 0),
-      heightSet: true,
-      expandedHeight: height,
-    });
+  onHeightAnimationComplete() {
+    if (typeof this.props.onAnimationComplete === 'function') {
+      this.props.onAnimationComplete();
+    }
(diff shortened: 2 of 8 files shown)


native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -34,92 +33,114 @@ const {
...
-      // why on earth does a value of 1 instead of 0 make this break!
-      height: new Animated.Value(expandedImmediately ? height : 0),
-      heightSet: true,
-      expandedHeight: height,
-    });
+  onHeightAnimationComplete() {
+    if (typeof this.props.onAnimationComplete === 'function') {
+      this.props.onAnimationComplete();
+    }
+    if (this.state.height._value - 0.01 < 0.00001) { // eslint-disable-line no-underscore-dangle
```
k0nserv: Is this the best way to get the value?

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -34,92 +33,114 @@ const {
...
-    Animated.timing(
-      this.state.height,
-      {
-        toValue: finalValue,
-        duration: animationDuration,
-      },
-    ).start();
+  setExpandedHeight(event) {
+    const { height } = event.nativeEvent.layout;
+    if (height !== 0) {
```
k0nserv: Is `heigh` a float? Can we invert the condition as `height === 0` and bail out?

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -34,92 +33,114 @@ const {
...
+            <View
+              onLayout={event => (this.setExpandedHeight(event))}
+            >
+              {this.props.children}
+            </View>
+          </Animated.View>
         </View>
-      </Animated.View>
-    );
+      ) : null;
```
k0nserv: So per the CHANGELOG this should be removing itself from the tree, but this actually removes the children not it self

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
georgegillams: It does seem to be. An issue was raised requesting that `getValue()` be made public, but it was not deemed necessary

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
georgegillams: Not sure what you mean here. The render function literally returns null.
The View-count shown below always reverts to 11 after the animation has completed. It also shows 11 views if I delete the `AnimateHeight` component from the story completely.

null

k0nserv: The component still renders right? It just renders no children

georgegillams: So the component shown above is a `View` containing a `BpkAnimateHeight` and a `BpkButton`. Once the `BpkAnimateHeight` has collapsed it becomes null, so all you're seeing is a `Viiew` containing the `BpkButton`

changelog.md
```diff
@@ -6,6 +6,10 @@
 - react-native-bpk-component-ticket:
   - New component, see http://backpack.prod.aws.skyscnr.com/components/native/ticket
 
+- react-native-bpk-component-animate-height:
+  - Added props for `expandDelay` and `collapseDelay`.
+  - Now removes itself from tree when collapsed to height 0.
```
shaundon: Sorry to be pedantic, but I think we can make this more meaningful for users, something like "when collapsed to `height: 0`, or "when collapsed and height is zero". What do you think?

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -34,92 +33,115 @@ const {
   animationDurationSm,
 } = tokens;
 
+const collapsedHeight = 0.01;
```
shaundon: Should probably go with `COLLAPSED_HEIGHT` for a constant like this.

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -34,92 +33,115 @@ const {
...
+    super(props);
 
     this.state = {
-      expanded: null,
-      expandedHeight: null,
-      collapsedHeight: 0,
-      height: new Animated.Value(1),
-      heightSet: false,
+      expanded: this.props.expanded,
+      height: new Animated.Value(0.01),
```
shaundon: Can we make this use the constant defined above, seeing as the value is the same?

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -34,92 +33,115 @@ const {
...
-      },
-    ).start();
+  setExpandedHeight(event) {
+    const { height } = event.nativeEvent.layout;
+    if (height === 0) {
+      return;
+    }
+    if (!this.state.expandAnimationInProgress) {
+      this.setState({
+        height: this.state.expanded ? new Animated.Value(height) : new Animated.Value(collapsedHeight),
```
shaundon: It would be DRYer to do:

```
height: new Animated.Value(this.state.expanded ? height : collapsedHeight)
```

..but not a huge deal

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -34,92 +33,115 @@ const {
...
+      });
+    } else {
+      this.setState({ expanded: true, expandAnimationInProgress: false });
+      Animated.timing(this.state.height, {
```
shaundon: As this is called twice I'm wondering if it should be abstracted into a function that takes a couple of params. Might be premature optimisation though, I'll leave it up to you!

changelog.md
georgegillams: Yeah sure. Also, never apologise for being pedantic 😆

native/packages/react-native-bpk-component-animate-height/stories.js
```diff
@@ -91,10 +94,16 @@ AnimateHeightDemo.propTypes = {
...
+  expandDelay: PropTypes.number,
+  collapseDelay: PropTypes.number,
+  animationDuration: PropTypes.number,
 };
 
 AnimateHeightDemo.defaultProps = {
   children: null,
+  expandDelay: 0,
+  collapseDelay: 0,
+  animationDuration: 400,
```
matteo-hertel: we have tokens for this now!

matteo-hertel:
Just fix the merge conflicts 😄

PR #263, opened 2017-11-10
[BPK-1084] Move state inside Banner Alert component

georgegillams:


matthewdavidson:
Hey @georgegillams! With this change how do consumers show a previously dismissed banner alert?

It looks like this PR hides the dismiss logic inside the component meaning that from a consumer point of view they'd have to re-render higher up the component tree for it to show again.

I reckon  we should expose an `onDismiss` and `dimissed` prop and allow the consumer to control the state themselves.

PR #268, opened 2017-11-14
[BPK-1092] horizontal nav animation

shaundon:
**What does this PR do?**

Previously, consumers would handle the `selected` prop on `BpkHorizontalNavItem`. If a NavItem had this prop, the NavItem would get blue text and a blue indicator underneath.

Their code would look like this:

```
<BpkHorizontalNav>
    <BpkHorizontalNavItem title="One" />
    <BpkHorizontalNavItem title="Two" selected />
    <BpkHorizontalNavItem title="Three" />
</BpkHorizontalNav>
```

With this PR, consumers should add a `selectedId` prop to the container and `id` props to the children instead, like this:

```
<BpkHorizontalNav selectedId="1">
    <BpkHorizontalNavItem title="One" id="0" />
    <BpkHorizontalNavItem title="Two" id="1" />
    <BpkHorizontalNavItem title="Three" id="2" />
</BpkHorizontalNav>
```

**How does it do this?**
The indicator line has been pulled out of the individual NavItems and made part of the parent. To place the line correctly, the parent needs to know the x-position and the width of each NavItem, so it can give the line the correct width and position.

To get this information, each child element is given an [`onLayout`](https://facebook.github.io/react-native/docs/view.html#onlayout) prop, which adds its layout information to an array contained in the parent.

`onLayout` fires after `componentDidMount`, making it hard to hook into a time when all the children have been laid out (and we have all the layout data we need to position the indicator). 

To overcome this, when `onLayout` fires for each NavItem, we check if the NavItem in question is the initially selected one. If so, we know we have enough layout information to render the initial position of the indicator, and trigger a re-render with `this.setState`. In the background, the other NavItem layouts can come in asynchronously, ready for when `selected` changes.

The indicator can now be positioned correctly because our array of positions looks like this:

```
{
  '0': {width: 100, x: 0},
  '1': {width: 80, x: 100},
  '2': {width: 90, x: 180}
}
```

So if `selected={1}`, the indicator's style would include `{ translateX: 100, width: 80 }`.

**Advantages of this approach**
* Very minimal changes to the NavItem code (just removing the indicator).
* Works with NavItems of variable width and with the `spaceAround` property, unlike [`react-native-tab-view`](https://github.com/react-native-community/react-native-tab-view/tree/master/src).
* Uses the least amount of stateful code possible. No need for Redux etc.

**Disadvantages of this approach**
* We're manually triggering a re-render, although in our testing this isn't noticeable.

### DIFF
--- a/native/packages/react-native-bpk-component-horizontal-nav/readme.md
+++ b/native/packages/react-native-bpk-component-horizontal-nav/readme.md
@@ -28,10 +28,10 @@ export default class App extends Component {
   render() {
     return (
       <View style={styles.container}>
-        <BpkHorizontalNav>
-          <BpkHorizontalNavItem selected title="Item One" onPress={() => {}} />
-          <BpkHorizontalNavItem title="Item Two" onPress={() => {}} />
-          <BpkHorizontalNavItem title="Item Three" onPress={() => {}} />
+        <BpkHorizontalNav selectedId="nav-1">
+          <BpkHorizontalNavItem id="nav-0" title="Item One" onPress={() => {}} />
+          <BpkHorizontalNavItem id="nav-1" title="Item Two" onPress={() => {}} />
+          <BpkHorizontalNavItem id="nav-2" title="Item Three" onPress={() => {}} />
         </BpkHorizontalNav>
       </View >
     );
@@ -43,20 +43,21 @@ export default class App extends Component {
 
 ### BpkHorizontalNav
 
-| Property            | PropType                              | Required | Default Value |
-| -----------         | ------------------------------------- | -------- | ------------- |
-| children            | node                                  | true     | -             |
-| spaceAround         | bool                                  | false    | false         |
+| Property            | PropType                                                    | Required | Default Value |
+| -----------         | ----------------------------------------------------------- | -------- | ------------- |
+| children            | node                                                        | true     | -             |
+| selectedId          | string (matching `id` prop of `BpkHorizontalNavItem` child) | true     | -             |
+| spaceAround         | bool                                                        | false    | false         |
 
 ### BpkHorizontalNavItem
 
 | Property            | PropType                              | Required | Default Value |
 | -----------         | ------------------------------------- | -------- | ------------- |
+| id                  | string                                | true     | -             |
 | onPress             | func                                  | true     | -             |
 | title               | string                                | true     | -             |
 | accessibilityLabel  | string                                | false    | props.title   |
 | disabled            | bool                                  | false    | false         |
-| selected            | bool                                  | false    | false         |
 | theme               | See [Theme Props](#theme-props) below | false    | null          |
 
 
--- a/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav-test.common.js
+++ b/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav-test.common.js
@@ -18,14 +18,28 @@
 
 import React from 'react';
 import renderer from 'react-test-renderer';
+
 import BpkHorizontalNav from './BpkHorizontalNav';
+import BpkHorizontalNavItem from './BpkHorizontalNavItem';
+
+jest.mock('./BpkHorizontalNavItem', () => 'BpkHorizontalNavItem');
 
 const commonTests = () => {
   describe('BpkHorizontalNav', () => {
     it('should render correctly', () => {
       const tree = renderer.create(
-        <BpkHorizontalNav>
+        <BpkHorizontalNav selectedId="0">
           My nav content.
+        </BpkHorizontalNav>,
+      ).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly with the "selected" prop', () => {
+      const tree = renderer.create(
+        <BpkHorizontalNav selectedId="0">
+          <BpkHorizontalNavItem id="0" />
+          <BpkHorizontalNavItem id="1" />
         </BpkHorizontalNav>,
       ).toJSON();
       expect(tree).toMatchSnapshot();
@@ -33,7 +47,7 @@ const commonTests = () => {
 
     it('should render correctly with custom "style" prop', () => {
       const tree = renderer.create(
-        <BpkHorizontalNav style={{ marginBottom: 10 }}>
+        <BpkHorizontalNav selectedId="0" style={{ marginBottom: 10 }}>
           My nav content.
         </BpkHorizontalNav>,
       ).toJSON();
@@ -42,7 +56,7 @@ const commonTests = () => {
 
     it('should render correctly with arbitrary props', () => {
       const tree = renderer.create(
-        <BpkHorizontalNav custom="custom-prop">
+        <BpkHorizontalNav selectedId="0" custom="custom-prop">
           My nav content.
         </BpkHorizontalNav>,
       ).toJSON();
--- a/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
+++ b/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
@@ -20,10 +20,16 @@ import {
   ScrollView,
   Platform,
   StyleSheet,
+  View,
   ViewPropTypes,
 } from 'react-native';
 import React from 'react';
 import PropTypes from 'prop-types';
+
+import BpkHorizontalNavSelectedIndicator from './BpkHorizontalNavSelectedIndicator';
+import withAnimatedProps from './withAnimatedProps';
+
+const AnimatedIndicator = withAnimatedProps(BpkHorizontalNavSelectedIndicator, ['xOffset', 'width']);
 
 const tokens = Platform.select({
   ios: () => require('bpk-tokens/tokens/ios/base.react.native.common.js'), // eslint-disable-line global-require
@@ -35,6 +41,8 @@ const styles = StyleSheet.create({
     borderColor: 'transparent',
     borderBottomColor: tokens.colorGray100,
     borderWidth: tokens.borderSizeSm,
+  },
+  horizontalNavInner: {
     flexDirection: 'row',
   },
   spaceAround: {
@@ -44,32 +52,92 @@ const styles = StyleSheet.create({
   },
 });
 
-const BpkHorizontalNav = (props) => {
-  const {
-    children,
-    spaceAround,
-    style,
-    ...rest
-  } = props;
-  const navStyle = [styles.horizontalNav];
-  if (spaceAround) {
-    navStyle.push(styles.spaceAround);
+class BpkHorizontalNav extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      indicatorOffsetX: null,
+      indicatorWidth: null,
+    };
+    this.onChildLayout = this.onChildLayout.bind(this);
   }
-  return (
-    <ScrollView
-      alwaysBounceHorizontal={false}
-      contentContainerStyle={[navStyle, style]}
-      horizontal
-      showsHorizontalScrollIndicator={false}
-      {...rest}
-    >
-      { children }
-    </ScrollView>
-  );
-};
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.selectedId && this.childrenPositions[nextProps.selectedId]) {
+      const nextLayoutProps = this.childrenPositions[nextProps.selectedId];
+      this.setState({
+        indicatorOffsetX: nextLayoutProps.x,
+        indicatorWidth: nextLayoutProps.width,
+      });
+    }
+  }
+
+  onChildLayout(event, id) {
+    const { width, x } = event.nativeEvent.layout;
+    this.childrenPositions[id] = { width, x };
+
+    // If the child in question is the initially selected one, the indicator can now be positioned.
+    if (this.props.selectedId === id) {
+      this.setState({
+        indicatorOffsetX: x,
+        indicatorWidth: width,
+      });
+    }
+  }
+
+  childrenPositions = {};
+
+  render() {
+    const {
+      children,
+      selectedId,
+      spaceAround,
+      style,
+      ...rest
+    } = this.props;
+
+    const navStyle = [styles.horizontalNav];
+    if (spaceAround) { navStyle.push(styles.spaceAround); }
+
+    const enhancedChildren = React.Children.map(children, child => (
+      React.isValidElement(child) && React.cloneElement(child, {
+        key: child.props.id,
+        selected: selectedId === child.props.id,
+
+        // Have children report their layout details after being laid out.
+        // This allows the selected indicator to be correctly positioned.
+        onLayout: event => this.onChildLayout(event, child.props.id),
+      })));
+
+    const renderIndicator = () => {
+      if (this.state.indicatorOffsetX === null) {
+        return null;
+      }
+      return <AnimatedIndicator xOffset={this.state.indicatorOffsetX} width={this.state.indicatorWidth} />;
+    };
+
+    return (
+      <ScrollView
+        alwaysBounceHorizontal={false}
+        contentContainerStyle={[navStyle, style]}
+        horizontal
+        showsHorizontalScrollIndicator={false}
+        {...rest}
+      >
+        <View>
+          <View style={styles.horizontalNavInner}>
+            {enhancedChildren}
+          </View>
+          { renderIndicator() }
+        </View>
+      </ScrollView>
+    );
+  }
+}
 
 BpkHorizontalNav.propTypes = {
   children: PropTypes.node.isRequired,
+  selectedId: PropTypes.string.isRequired,
   spaceAround: PropTypes.bool,
   style: ViewPropTypes.style,
 };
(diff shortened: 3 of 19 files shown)


native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -42,39 +43,90 @@ const styles = StyleSheet.create({
...
-    style,
-    ...rest
-  } = props;
-  const navStyle = [styles.horizontalNav];
-  if (spaceAround) {
-    navStyle.push(styles.spaceAround);
+class BpkHorizontalNav extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { isReady: false };
```
matteo-hertel: this is my poor choice of names, I think there might be a better name for this but I can't think of one

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -42,39 +43,90 @@ const styles = StyleSheet.create({
...
-};
+  onChildLayout = (event) => {
+    const { width, x } = event.nativeEvent.layout;
+    this.childrenPositions.push({ width, x });
+    this.checkChildrenPositions(React.Children.count(this.props.children), this.childrenPositions.length);
+  };
+  checkChildrenPositions(childrenLength, renderedChildren) {
+    if (childrenLength === renderedChildren) {
+      // onLayout is async, so sort the children to ensure correct order.
+      this.childrenPositions = this.childrenPositions.sort((a, b) => a.x > b.x);
```
matteo-hertel: probably  different name rather than `a` and `b` will make it easier to understand

shaundon: Hard to say - we could use `obj1` and `obj2` but they might not be any more descriptive. I'm struggling to think of something that's more descriptive than `a` and `b` for such a trivial function.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
-      alwaysBounceHorizontal={false}
-      contentContainerStyle={[navStyle, style]}
-      horizontal
-      showsHorizontalScrollIndicator={false}
-      {...rest}
-    >
-      { children }
-    </ScrollView>
-  );
+  if (selectedIndex < 0 || selectedIndex + 1 > childrenCount) {
```
k0nserv: `selectedIndex >= childrenCount` reads better than `selectedIndex + 1 > childrenCount` imo.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
+    };
+    this.onChildLayout = this.onChildLayout.bind(this);
+    this.animateIndicator = this.animateIndicator.bind(this);
+  }
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.selected !== null) {
+      this.animateIndicator(this.childrenPositions[nextProps.selected]);
+    }
+  }
+  onChildLayout = (event, index) => {
```
k0nserv: We could put the index as the first argument here and then partially apply the function before passing it to the `BpkHorizontalNavItem`. It's slightly nicer than `onLayout: event => this.onChildLayout(event, index)` since it instead could just be `onLayout: boundOnChildLayout`

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
+
+    const navStyle = [styles.horizontalNav];
+    if (spaceAround) {
+      navStyle.push(styles.spaceAround);
+    }
+
+    const enhancedChildren = (selected !== null) ?
+      React.Children.map(children, (child, index) => (
+        React.isValidElement(child) && React.cloneElement(child, {
+          key: index, // eslint-disable-line react/no-array-index-key
```
k0nserv: Could we use a better key than index?

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
+        toValue: x,
+        duration,
+      }),
+      Animated.timing(this.state.indicatorWidth, {
+        toValue: width,
+        duration,
+      }),
+    ]).start();
+  }
+  childrenPositions = {};
```
k0nserv: Could we introduce more vertical spacing in this file. One line between each member of the component for example?

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
+    // the indicator can now be positioned.
+    if (this.props.selected === index) {
+      this.setState({
+        indicatorOffsetX: new Animated.Value(x),
+        indicatorWidth: new Animated.Value(width),
+      });
+    }
+  }
+  animateIndicator = ({ x, width }) => {
+    const duration = 200;
```
k0nserv: This looks like a contant so it should either be in snake case at the top of the file or possibly a token

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
+      children,
+      selected,
+      spaceAround,
+      style,
+      theme,
+      ...rest
+    } = this.props;
+
+    if (selected && !validateSelectedIndex(selected, React.Children.count(children))) {
+      throw new Error(`Invalid \`selected\` prop. Must be an integer between 0 and ${React.Children.count(children) - 1}.`); // eslint-disable-line max-len
```
k0nserv: Shouldn't this be done in the construtor or possibly as a prop validation?

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
+        React.isValidElement(child) && React.cloneElement(child, {
+          key: index, // eslint-disable-line react/no-array-index-key
+          selected: child.props.selected || selected === index,
+
+          // Have children report their layout details after being laid out.
+          // This allows the selected indicator to be correctly positioned.
+          onLayout: event => this.onChildLayout(event, index),
+        }))) : children;
+
+    const renderSelectedIndicator = (selectedIndex) => {
```
k0nserv: Could this be a component?

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
+      }
+      return null;
+    };
+
+    return (
+      <ScrollView
+        alwaysBounceHorizontal={false}
+        contentContainerStyle={[navStyle, style]}
+        horizontal
+        showsHorizontalScrollIndicator={false}
```
k0nserv: Do we have a method to indicate that this is scrollable or are we planning one? On web we use a gradient that fades out on either side. cc @jamesf3rguson

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -35,48 +40,157 @@ const styles = StyleSheet.create({
...
+          </View>
+          {renderSelectedIndicator(selected)}
+        </View>
+      </ScrollView>
+    );
+  }
+}
+
+const propTypes = {
+  children: PropTypes.arrayOf(PropTypes.element).isRequired,
```
k0nserv: Doesn't `PropTypes.element` already support one or more?

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: Agreed, will change.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: Yes, it should be in the constructor.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: @matthewdavidson and I spoke about it, we do want to do that at some point in the near term.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: When I tried it, I got proptypes errors from having more than one. Will double-check though.

k0nserv: Fair, I might be thinking of node

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: I'm not sure what else we have that's more applicable here. The React docs say to definitely avoid keys for data that can be re-ordered, and the data here will never be re-ordered, so I think we're probably ok.

shaundon:
I've moved the indicator to its own component and file, which decouples the animation and theming logic from `BpkHorizontalNav`, which is now only concerned with laying out the children and listening for changes to the `selected` prop. It also means the test coverage is improved, as we can test the indicator in isolation.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav-test.common.js
```diff
@@ -18,11 +18,16 @@
...
+
 import BpkHorizontalNav from './BpkHorizontalNav';
+import BpkHorizontalNavItem from './BpkHorizontalNavItem';
+
+jest.mock('./BpkHorizontalNavItem', () => 'BpkHorizontalNavItem');
 
 const commonTests = () => {
   describe('BpkHorizontalNav', () => {
     it('should render correctly', () => {
+      jest.spyOn(console, 'error').mockImplementation(() => jest.fn());
```
matthewdavidson: 😈

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav-test.common.js
```diff
@@ -31,6 +36,43 @@ const commonTests = () => {
...
+        <BpkHorizontalNav selected={-1}>
+          <BpkHorizontalNavItem />
+          <BpkHorizontalNavItem />
+        </BpkHorizontalNav>,
+      )).toThrow('Invalid `selected` prop. Must be an integer between 0 and 1.');
+    });
+
+    it('should throw when "selected" prop is greater than the number of children', () => {
+      expect(() => renderer.create(
+        <BpkHorizontalNav selected="2">
```
matthewdavidson: This should be a number.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -44,37 +49,117 @@ const styles = StyleSheet.create({
...
-      { children }
-    </ScrollView>
-  );
+  return true;
 };
 
+class BpkHorizontalNav extends React.Component {
+  constructor(props) {
+    super(props);
+    if (props.selected && !validateSelectedIndex(props.selected, React.Children.count(props.children))) {
```
matthewdavidson: this shouldn't be in ctor, it wil only fire once and won't capture dynamic updates to children

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavSelectedIndicator.js
```diff
@@ -0,0 +1,100 @@
...
+  Animated,
+  Platform,
+  StyleSheet,
+} from 'react-native';
+import PropTypes from 'prop-types';
+import { withTheme } from 'react-native-bpk-theming';
+
+import { THEMING_ATTRIBUTE, themePropType } from './theming';
+
+const ANIMATION_DURATION = 200; // milliseconds.
```
matthewdavidson: Use http://backpack.prod.aws.skyscnr.com/tokens/animation

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavSelectedIndicator.js
```diff
@@ -0,0 +1,100 @@
...
+      width: new Animated.Value(props.width),
+    };
+    this.animateIndicator = this.animateIndicator.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.animateIndicator(nextProps.xOffset, nextProps.width);
+  }
+
+  animateIndicator = (xOffset, width) => {
```
matthewdavidson: This is still proposal syntax - plus the arrow function is redundant.

native/packages/react-native-bpk-component-horizontal-nav/readme.md
```diff
@@ -43,10 +43,11 @@ export default class App extends Component {
...
 ### BpkHorizontalNav
 
-| Property            | PropType                              | Required | Default Value |
-| -----------         | ------------------------------------- | -------- | ------------- |
-| children            | node                                  | true     | -             |
-| spaceAround         | bool                                  | false    | false         |
+| Property            | PropType                                  | Required | Default Value |
+| -----------         | ----------------------------------------- | -------- | ------------- |
+| children            | node                                      | true     | -             |
+| selected            | number (array index of the selected item) | false    | null          |
```
matteo-hertel: if nothing is passed will default to `null` will null be cast a 0 so the first item will be selected or nothing will be selected?

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav-test.common.js
```diff
@@ -18,11 +18,16 @@
 
 import React from 'react';
 import renderer from 'react-test-renderer';
+
 import BpkHorizontalNav from './BpkHorizontalNav';
+import BpkHorizontalNavItem from './BpkHorizontalNavItem';
+
+jest.mock('./BpkHorizontalNavItem', () => 'BpkHorizontalNavItem');
```
matteo-hertel: Not sure I understand this, you're mocking `BpkHorizontalNavItem` with itself ?!

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavSelectedIndicator.js
shaundon: Just tried this and it's not possible with the current tokens as they're strings -- `"200ms"` as opposed to `200`.

I'll do another PR in the future to change the formatting of the tokens and then update this, I think it's out of scope for this PR.

native/packages/react-native-bpk-component-horizontal-nav/readme.md
shaundon: Nothing will be selected. I'm tightening it up a bit though so that `selected` is required.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav-test.common.js
shaundon: I think it's essentially returning a blank component.

matteo-hertel: I don't get we need it, you don't want and actual `BpkHorizontalNavItem` in the test?

shaundon:
@matthewdavidson I've made the requested changes. The main one is to use an `id` field instead of array indexes, so the component is like this now:

```
<Nav selectedId="item-1">
    <NavItem id="item-0" ...etc />
    <NavItem id="item-1" ...etc />
</Nav>
```

I tried to use `key` but I ran into some warnings, and [the docs recommended this approach](https://reactjs.org/warnings/special-props.html). Plus I feel that `id` and `selectedId` is very explicit from an API point of view.

One big benefit of this approach is I was able to remove a lot of defensive code around index out of bounds exceptions.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav-test.common.js
shaundon: (Discussed in person)

shaundon:
Pulled out all the animation code into its own HOC.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav-test.common.js
matteo-hertel: The outcome of the discussion was: we don't want the full `BpkHorizontalNavItem` there because it's testing the `BpkHorizontalNav` component

PR #269, opened 2017-11-14
add TouchableNativeFeedback for android

matteo-hertel:
added `TouchableNativeFeedback` for android, we'll now  ship two BPK cards: iOS and Android, the reason behind it is `TouchableNativeFeedback` and `TouchableHighlight` are very different so having them behaving in one file became a nightmare.

The main difference between the two (`TouchableNativeFeedback` and `TouchableHighlight`) is the `style` `TouchableNativeFeedback`does not accepts any styles therefore the Android version return a `View` wrapping it, in which userland style will be applied

Base Card
screenshot_1510678069

Ripple animation start
screenshot_1510678078

Ripple animation end
screenshot_1510678084

cc @jamesf3rguson  for the visual validation

iOS is unaffected by this changes

### DIFF
--- a/native/packages/react-native-bpk-component-card/src/BpkCard.android.js
+++ b/native/packages/react-native-bpk-component-card/src/BpkCard.android.js
@@ -20,8 +20,7 @@ import {
   View,
   Platform,
   StyleSheet,
-  TouchableHighlight,
-  PixelRatio,
+  TouchableNativeFeedback,
   ViewPropTypes,
 } from 'react-native';
 import React from 'react';
@@ -33,22 +32,11 @@ const tokens = Platform.select({
 })();
 
 const {
-  colorGray50,
   colorWhite,
   elevationXs,
   elevationLg,
   borderRadiusSm,
   spacingBase,
-  shadowSmColor,
-  shadowSmOffsetWidth,
-  shadowSmOffsetHeight,
-  shadowSmOpacity,
-  shadowSmRadius,
-  shadowXlColor,
-  shadowXlOffsetWidth,
-  shadowXlOffsetHeight,
-  shadowXlOpacity,
-  shadowXlRadius,
 } = tokens;
 
 /**
@@ -59,23 +47,12 @@ const styles = StyleSheet.create({
     backgroundColor: colorWhite,
     borderRadius: borderRadiusSm,
     elevation: elevationXs,
-    shadowColor: shadowSmColor,
-    shadowOffset: { width: shadowSmOffsetWidth, height: shadowSmOffsetHeight / PixelRatio.get() },
-    shadowOpacity: shadowSmOpacity,
-    shadowRadius: shadowSmRadius / PixelRatio.get(),
   },
   cardPadded: {
     padding: spacingBase,
   },
   cardFocused: {
     elevation: elevationLg,
-    shadowColor: shadowXlColor,
-    shadowOffset: { width: shadowXlOffsetWidth, height: shadowXlOffsetHeight / PixelRatio.get() },
-    shadowOpacity: shadowXlOpacity,
-    shadowRadius: shadowXlRadius / PixelRatio.get(),
-  },
-  cardInner: {
-    backgroundColor: 'transparent', // otherwise this view's corners would bleed outwith the outer container
   },
 });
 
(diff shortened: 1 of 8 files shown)


shaundon:
I think this is a good approach - there's much less platform specific code than `BpkButton` so it makes sense to keep it together. 

React Native's built-in Button component does the same thing: https://github.com/facebook/react-native/blob/master/Libraries/Components/Button.js#L126. In our own codebase, `BpkHorizontalNavItem` does the same thing: https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavItem.js#L115

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -91,17 +91,23 @@ const BpkCard = (props) => {
...
-    >
-      <View style={styles.cardInner}>{children}</View>
-    </TouchableHighlight>
+    <View style={userStyle}>
+      <Touchable
+        style={styles.outerContainer}
+        useForeground
+        underlayColor={colorGray50}
+        accessibilityComponentType="button"
+        background={Platform.OS === 'android' ? TouchableNativeFeedback.SelectableBackgroundBorderless() : null}
```
matteo-hertel: Android's `TouchableNativeFeedback` does not accepts styles, so the background white needs to be added to the inner view on Android

native/packages/react-native-bpk-component-card/src/BpkCard.js
```diff
@@ -91,17 +91,23 @@ const BpkCard = (props) => {
...
   return (
-    <TouchableHighlight
-      accessibilityComponentType="button"
-      underlayColor={colorGray50}
-      style={style}
-      {...rest}
-    >
-      <View style={styles.cardInner}>{children}</View>
-    </TouchableHighlight>
+    <View style={userStyle}>
```
k0nserv: The `userStyle` have moved is this a regression?

k0nserv: Also might it be confusing that we spread `rest` on the `Touchable`, but add the style to the outer view?

k0nserv: Based on all this can you expand on why we need to outer view at all?

matteo-hertel: During standup we agreed to have a container component like the button component does, once that is done I'll update this and explain the problem

native/packages/react-native-bpk-component-card/index.js
```diff
@@ -15,7 +15,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+/* eslint-disable import/no-unresolved */
+/* eslint-disable import/extensions */
```
matthewdavidson: it's be good if we could add a rule to eslint to allow `.android.js` & `.ios.js`

native/packages/react-native-bpk-component-card/src/BpkCard.android.js
```diff
@@ -0,0 +1,107 @@
...
+    elevation: elevationXs,
+  },
+  cardPadded: {
+    padding: spacingBase,
+  },
+  cardFocused: {
+    elevation: elevationLg,
+  },
+  cardOuter: {
+    backgroundColor: colorWhite,
```
matthewdavidson: This `backgroundColor` be added to `card` and `cardOuter ` can be removed.

native/packages/react-native-bpk-component-card/index.js
matteo-hertel: I'm trying this https://github.com/benmosher/eslint-plugin-import/issues/279#issuecomment-215052176
happy for me to add this to the `.eslintrc`?

native/packages/react-native-bpk-component-card/src/__snapshots__/BpkCard-test.android.js.snap
```diff
@@ -57,73 +27,37 @@ exports[`Android BpkCard should render correctly 1`] = `
...
-          Object {},
-        ]
+        Object {
+          "color": "#333333",
+          "margin": 20,
+        }
       }
     >
-      Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
+      TouchableNativeFeedback is not supported on this platform!
```
matteo-hertel: This does not look correct 😕

native/packages/react-native-bpk-component-card/src/BpkCard-test.common.js
```diff
@@ -15,7 +15,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
```
matthewdavidson: Why is this newline removed?

native/packages/react-native-bpk-component-card/src/__snapshots__/BpkCard-test.android.js.snap
matthewdavidson: Hmm it sure does not 🤔

native/packages/react-native-bpk-component-card/src/BpkCard-test.common.js
matteo-hertel: ah! I've removed the eslint disable and I forgot to put back the new line 😅

matthewdavidson:
@matteo-hertel  This is good to go once the build passes dude.

PR #270, opened 2017-11-15
[NOJIRA] Generate platform specific extension token files

matthewdavidson:
This PR should allow us to stop doing this:
```
const tokens = Platform.select({
  ios: () => require('bpk-tokens/tokens/ios/base.react.native.common.js'), // eslint-disable-line global-require
  android: () => require('bpk-tokens/tokens/android/base.react.native.common.js'), // eslint-disable-line global-require
})();
```

And start doing this:
```
import { spacingBase } from 'bpk-tokens/tokens/base.react.native';
```


matthewdavidson:
Yeah that was intentional and no it's not part of the public interface.

matteo-hertel:
Will this cause problems during testing? e.g. Jest not loading the correct tokens and loading always iOS

matthewdavidson:
@matteo-hertel yes ... yes it will 😂

matteo-hertel:
raining_david_tennant

PR #271, opened 2017-11-15
[BPK-848]update deployment target

matteo-hertel:


matteo-hertel:
yes indeed!

shaundon:
Well that's convenient! ✋ :five:

matteo-hertel:
This is all done and tested but we'll wait for the new domain before merging it

packages/bpk-docs/src/Readme.md
```diff
@@ -0,0 +1,3 @@
+# Backpack Design System Documentation
```
matthewdavidson: This should be `README.md`.

webpack.config.babel.js
```diff
@@ -203,6 +204,10 @@ if (isProduction) {
         warnings: false,
       },
     }),
+    new CopyWebpackPlugin([
+      { from: 'packages/bpk-docs/src/Readme.md', to: 'Readme.md' },
```
matthewdavidson: Does this get copied to the `dist` directory?

matteo-hertel: yes it does!

webpack.config.babel.js
```diff
@@ -203,6 +204,10 @@ if (isProduction) {
         warnings: false,
       },
     }),
+    new CopyWebpackPlugin([
+      { from: 'packages/bpk-docs/src/README.md', to: 'README.md' },
```
k0nserv: Couldn't we just create the `README.md` in https://github.com/backpack/backpack.github.io once and then leave it?

matteo-hertel: Travis will force push to the target branch so it'll be nuked

k0nserv: Really? So it will always just have a single commit?

matteo-hertel: according to the docs yes
https://docs.travis-ci.com/user/deployment/pages/

k0nserv: Cool

PR #272, opened 2017-11-16
[BPK-1094] BpkFieldset validation tweak

shaundon:
If BpkFieldset doesn't have a label, it will no longer include an empty BpkLabel component.


matteo-hertel:
This is still good to go IMHO, I would probably wait for @matthewdavidson to give a 👍

PR #276, opened 2017-11-17
[no jira] Add duration type to token formatter and consume it

shaundon:
React Native animations take an integer in milliseconds as their arguments, so the previous values we were using, such as `"200ms"` do not work.

This PR adds a formatter so that `duration` tokens are converted, and then consumes it in the horizontal nav.

### DIFF
--- a/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
+++ b/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
@@ -118,7 +118,7 @@ BpkAnimateHeight.propTypes = {
 };
 
 BpkAnimateHeight.defaultProps = {
-  animationDuration: parseInt(animationDurationSm, 10),
+  animationDuration: animationDurationSm,
   style: null,
 };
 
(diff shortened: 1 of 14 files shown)


shaundon:
@matteo-hertel it shouldn't do - the only file I can see that happening with is `native/packages/react-native-bpk-component-horizontal-nav/src/withAnimatedProps.js` and I can fix that easily if so.

matteo-hertel:
Awesome, I remember there was another component that was using animation, forgot which one, I guess if CI passes nothing is broken and this is good to go

shaundon:
Actually you're right, I just looked and native `BpkAnimateHeight` uses `animationDurationSm`:

```
BpkAnimateHeight.defaultProps = {
  animationDuration: parseInt(animationDurationSm, 10),
  style: null,
};
```

This PR won't break that functionality, because `parseInt` still works on existing integers, but I'll update my PR to change it there too.

shaundon:
@matteo-hertel now updated 😄

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -118,7 +118,7 @@ BpkAnimateHeight.propTypes = {
 };
 
 BpkAnimateHeight.defaultProps = {
-  animationDuration: parseInt(animationDurationSm, 10),
+  animationDuration: animationDurationSm,
```
matteo-hertel: 🎖

PR #277, opened 2017-11-17
[no jira] Changed the way Danger reports new components

shaundon:
Danger uses this code to advise people introducing new components:

```
if (nativeComponentIntroduced) {
  message('It looks like you are introducing a native component.');
  warn('Ensure the component style is extensible via `style`.');
}
```

However, when this is posted to GitHub, they're placed away from each other, which removes a lot of the context:

screen shot 2017-11-17 at 12 01 18

This PR changes the code to use `warn` for all of the message, which will make more sense:

```
if (nativeComponentIntroduced) {
  warn('It looks like you are introducing a native component. Ensure the component style is extensible via `style`.');
}
```

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -37,8 +37,7 @@ const webComponentIntroduced = createdFiles.some(filePath => (
 ));
 
 if (webComponentIntroduced) {
-  message('It looks like you are introducing a web component.');
-  warn('Ensure the component style is extensible via `className`.');
+  warn('It looks like you are introducing a web component. Ensure the component style is extensible via `className`.');
 }
 
 // Ensure new native components are extensible by consumers.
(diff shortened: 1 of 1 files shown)


dangerfile.js
```diff
@@ -47,8 +46,7 @@ const nativeComponentIntroduced = createdFiles.some(filePath => (
 ));
 
 if (nativeComponentIntroduced) {
-  message('It looks like you are introducing a native component.');
-  warn('Ensure the component style is extensible via `style`.');
+  warn('It looks like you are introducing a native component. Ensure the component style is extensible via `style`.');
```
matteo-hertel: maybe "`style` prop" will make more sense?
It's good anyway!

PR #279, opened 2017-11-17
[BPK-509] Add dismissable configuration

georgegillams:


packages/bpk-component-banner-alert/stories.js
```diff
@@ -39,8 +40,14 @@ storiesOf('bpk-component-banner-alert', module)
   .add('Success (long message)', () => (
     <BpkBannerAlert message={longMessage} type={ALERT_TYPES.SUCCESS} />
   ))
+  .add('Success (dismissable)', () => (
+    <BpkBannerAlert dismissable message={message} type={ALERT_TYPES.SUCCESS} />
```
matteo-hertel: can we add the  `onDismiss` to log in the action logger console?

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
```diff
@@ -101,9 +120,26 @@ class BpkBannerAlert extends Component {
     }));
   }
 
+  onDismiss() {
+    if (this.props.onDismiss) {
```
matteo-hertel: you should probably check if it's a function with `typeof`

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
```diff
@@ -127,7 +163,7 @@ class BpkBannerAlert extends Component {
     */
     // Disabling 'click-events-have-key-events and interactive-supports-focus' because header element is not focusable.
     // ToggleButton is focusable and works for this.
-    return (
+    return show ? (
```
matteo-hertel: this nested ternary operators are really hard to read, can we refactor the logic a bit?
1. if `!show` return null straightaway 
2.  create a const for the expandible and dismissable first set them to null if no needed
3. return the rendered component adding inside the const created before

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
georgegillams: Yeah I also had a chat to Matt about that, and he said that the type is enforced by the proptypes anyway. If it's not a function, react will blow up

shaundon:
This is failing right on local and CI:

```ModuleNotFoundError: Module not found: Error: Can't resolve './AnimateAndFade' in '/Users/shaundonnelly/projects/backpack/packages/bpk-component-banner-alert/src'```

Just posting this here to remind myself to follow up on it next week.

File: packages/bpk-component-banner-alert/src/BpkBannerAlert.js [hunk 4]
```diff
@@ -62,7 +63,7 @@ const getIconForType = (type) => {
 
 const ToggleButton = (props) => {
   const classNames = [getClassName('bpk-banner-alert__expand-icon')];
-
+  console.log(props.expanded);
```
shaundon: This should probably be removed before merging.

PR #281, opened 2017-11-19
Set svg fill colour in BpkButton

sogame:
`currentColor` can be used as `fill` colour for svg's in a buttons, allowing the users not having to manually set it when adding icons inside buttons (see the changes in `BpkLoadingButton` as an example).

`currentColor` is supported even by IE9: https://caniuse.com/#feat=currentcolor


changelog.md
```diff
@@ -2,7 +2,9 @@
 
 ## UNRELEASED
 
-_Nothing yet_
+**Added:**
+- bpk-component-button:
+  - Svg's fill colour is automatically set to text colour.
```
k0nserv: This is a patch change so should be under Fixed imo

sogame:
Thanks Hugo!
Changed to "Fixed" :-)

k0nserv:
Tested on IE10 and it looks good

PR #282, opened 2017-11-20
Replace padded prop on withScrim with more generic contentClassName

danrr:
This allows the extraction of some modal specific code and makes the component more versatile.
Only the modal needs display: flex and padding on the container, it now injects the styles it requires.

### DIFF
--- a/packages/bpk-scrim-utils/src/bpk-scrim-content.scss
+++ b/packages/bpk-scrim-utils/src/bpk-scrim-content.scss
@@ -25,25 +25,9 @@
   bottom: 0;
   left: 0;
   z-index: $bpk-zindex-scrim;
-  overflow: auto;
   overflow-x: hidden;
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
-
-  @include bpk-breakpoint-above-mobile {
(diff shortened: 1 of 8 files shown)


packages/bpk-scrim-utils/src/bpk-scrim-content.scss
```diff
@@ -25,26 +25,10 @@
   bottom: 0;
   left: 0;
   z-index: $bpk-zindex-scrim;
-  overflow: auto;
```
matthewdavidson: Why is this removed?

danrr: overflow-x and overflow-y on the next two lines overwrote it.

danrr:
I had look on browser stack for iPhone and Android and locally in a bunch of browsers.

packages/bpk-scrim-utils/src/bpk-scrim-content.scss
matthewdavidson: Can we keep it in for now? I remember explicitly adding this for x-browser compat - I should have added a comment explaining why :(

PR #283, opened 2017-11-20
[no jira] Replace S3 bucket links with GH pages ones

shaundon:


matteo-hertel:
Awesome, let's just wait for the pipeline and redirect to be ready and this is good to go

PR #285, opened 2017-11-20
Bpk 1109 rethink bpk image positoning

k0nserv:


packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -103,19 +104,32 @@ export default function withLazyLoading(Component, document) {
...
             inView={this.state.inView}
-            {...this.props}
+            {...rest}
           />
         </div>
       );
     }
   }
   WithLazyLoading.displayName = wrapDisplayName(Component, 'withLazyLoading');
+  WithLazyLoading.propTypes = {
```
matthewdavidson: My eyes 😭  (new lines)

changelog.md
```diff
@@ -6,11 +6,17 @@
 - bpk-scrim-utils:
   - Removed `padded` prop in favour of more generic `contentClassName`
   - Fixed modal white background on iPhone not showing
+- bpk-component-image:
+  - `withLazyLoading` now applys styling options to the wrapper div it creates rather than the lazy loaded component.
```
matthewdavidson: *applies

k0nserv: That's what I get for not reading what I write

georgegillams:
When I apply a className with `margin: 50px` it causes some bizarre behaviour:
screen shot 2017-11-20 at 14 39 24

And for lazy loaded ones: 
screen shot 2017-11-20 at 14 40 25

georgegillams:
I think the first is due to the paddingBottom: 66.66% thing, as the image is now not using the full width of the container?

k0nserv:
@georgegillams You applied it to the aspect ratio container right? Consumers can't/shouldn't change the style on that one. All though it seems like we add any className to that div. https://github.com/Skyscanner/backpack/blob/8ebc56deba530024a7df54e526c96709ac26ba89/packages/bpk-component-image/src/BpkImage.js#L104

k0nserv:
We should move that to be added to the outer div

k0nserv:
See 1cad452

georgegillams:
Yeah I applied it directly to BpkImage where it is used on the images page.

k0nserv:
I think the change you saw was reasonable George. At least when I tested locally it made sense.

PR #288, opened 2017-11-20
[No JIRA] Improve image docs

k0nserv:
Documents the combination of the two HOCs and clarifies SSR support.


packages/bpk-component-image/readme.md
```diff
@@ -91,6 +95,34 @@ export default () => (
 );
 ```
 
+### All together now
+
+Combining `withLazyLoading` and `withLoadingBehavior` gives use a lazily loaded image that will show a spinner while the image loads.
```
georgegillams: *us

PR #292, opened 2017-11-20
Make BpkBannerAlert dismissable option self-contained

sogame:
Hi,
I've noticed that when using the "dismissable" feature from `BpkBannerAlert`, the state must be handled by the parent, providing the `show` prop with the right value. Is there a reason for that? Wouldn't this feature be easier to use if it was self-contained?

This PR includes the changes in the component to avoid having to handle the "dismissed" state from the parent.

### DIFF
--- a/packages/bpk-component-banner-alert/readme.md
+++ b/packages/bpk-component-banner-alert/readme.md
@@ -34,53 +34,26 @@ export default () => (
 import React, { Component } from 'react';
 import BpkBannerAlert, { ALERT_TYPES } from 'bpk-component-banner-alert';
 
-class DismissableBpkBannerAlert extends Component {
-  constructor() {
-    super();
-
-    this.setDismissed = this.setDismissed.bind(this);
-
-    this.state = {
-      show: true,
-    };
-  }
-
-  setDismissed() {
-    this.setState({
-      show: false,
-    });
-  }
-
-  render() {
-    return (
-      <BpkBannerAlert
-        message="Successful alert with more information."
-        type={ALERT_TYPES.SUCCESS}
-        dismissable
-        onDismiss={this.setDismissed}
-        show={this.state.show}
-        dismissButtonLabel="Dismiss"
-      />
-    );
-  }
-}
-
 export default () => (
-  <DismissableBpkBannerAlert />
+  <BpkBannerAlert
+    message="Successful alert with dismiss option."
+    type={ALERT_TYPES.SUCCESS}
+    dismissable
+    dismissButtonLabel="Dismiss"
+  />
 );
 ```
 
 ## Props
 
-| Property          | PropType             | Required | Default Value |
-| ----------------- | -------------------- | -------- | ------------- |
-| type              | ALERT_TYPES (one of) | true     | -             |
-| message           | node                 | true     | -             |
-| ariaLive          | ARIA_LIVE (one of)   | false    | 'assertive'   |
-| children          | node                 | false    | null          |
-| dismissable       | bool                 | false    | false         |
-| dismissLabel      | string               | false    | null          |
-| onDismiss         | func                 | false    | null          |
-| show              | bool                 | false    | true          |
-| toggleButtonLabel | string               | false    | null          |
-| className         | string               | false    | null          |
+| Property           | PropType             | Required | Default Value |
+| ------------------ | -------------------- | -------- | ------------- |
+| type               | ALERT_TYPES (one of) | true     | -             |
+| message            | node                 | true     | -             |
+| ariaLive           | ARIA_LIVE (one of)   | false    | 'assertive'   |
+| children           | node                 | false    | null          |
+| dismissable        | bool                 | false    | false         |
+| dismissButtonLabel | string               | false    | null          |
+| onDismiss          | func                 | false    | null          |
+| toggleButtonLabel  | string               | false    | null          |
+| className          | string               | false    | null          |
--- a/packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
+++ b/packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
@@ -18,6 +18,8 @@
 
 import React from 'react';
 import renderer from 'react-test-renderer';
+import { shallow } from 'enzyme';
+import { shallowToJson } from 'enzyme-to-json';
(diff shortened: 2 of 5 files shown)


packages/bpk-component-banner-alert/readme.md
```diff
@@ -34,53 +34,26 @@ export default () => (
...
 ## Props
 
-| Property          | PropType             | Required | Default Value |
-| ----------------- | -------------------- | -------- | ------------- |
-| type              | ALERT_TYPES (one of) | true     | -             |
-| message           | node                 | true     | -             |
-| ariaLive          | ARIA_LIVE (one of)   | false    | 'assertive'   |
-| children          | node                 | false    | null          |
-| dismissable       | bool                 | false    | false         |
-| dismissLabel      | string               | false    | null          |
```
sogame: Just changed `dismissLabel` to `dismissButtonLabel` (the prop was wrong here), but I had to change the width of the whole table 😅

packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
```diff
@@ -78,15 +80,30 @@ describe('BpkBannerAlert', () => {
 
   it('should render correctly with dismissable option', () => {
     const tree = renderer.create(
-      <BpkBannerAlert type={ALERT_TYPES.WARN} message={message} dismissable />,
+      <BpkBannerAlert type={ALERT_TYPES.WARN} message={message} dismissable dismissButtonLabel="Dismiss" />,
```
sogame: I think it makes sense to include the label in the snapshot

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -38,47 +37,18 @@ sapien, et dapibus mi aliquet non. Pellentesque auctor sagittis lectus vitae rho
...
-BpkBannerDismissable.propTypes = {
-  message: PropTypes.string,
-  type: PropTypes.string,
-};
-
-BpkBannerDismissable.defaultProps = {
-  message: null,
-  type: null,
-};
+const BannerAlert = withDefaultProps(BpkBannerAlert, {
```
sogame: Small documentation refactor, I'm using `withDefaultProps` to create three components so there's less repeated code :-)

sogame:
@backpackbot I haven't updated `changelog.md` because the "dismissable" feature has not been released yet, and the current message is valid if this is merged before releasing it 😝

packages/bpk-component-banner-alert/readme.md
georgegillams: Good catch!

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
```diff
@@ -107,6 +107,7 @@ class BpkBannerAlert extends Component {
 
     this.state = {
       expanded: false,
+      dismissed: false,
```
georgegillams: The reason we don't do this is to maintain state as high up as possible in the tree. It may make sense for us to provide a HOC which applies the default show / hide behaviour for the normal dismissable case, but we should still keep state outside of the component so that consumers of the components have full control.
If we hold `dismissed` state inside the component, for instance, it would not be possible to automatically dismiss a banner after a certain period of time has elapsed.

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
georgegillams: 👍

packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
georgegillams: Sure! Good catch 😸

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
sogame: But `expanded` state is inside the component, shouldn't both be in the same place, for consistency?

Having the `onDismiss` callback already allows a parent component to do some action once the component has been dismissed (maybe an `onToggle` callback should also be implemented to be able to notify parent components?)

Regarding automatically dismissing a banner after some period of time, it's funny you mention this functionality, as it's something I asked in the Slack channel and implemented but I hold the PR until the `dismissible` functionality was merged 😂  (this is the thread: https://skyscanner.slack.com/archives/C0JHPDSSU/p1509748373000119)

It's true that this functionality could be handled by a parent component, but I think it makes more sense to add a prop to this component with the period of time, and the component itself to handle it. This is my initial implementation (now I would refactor it to use `onDismiss`): https://github.com/sogame/backpack/commit/fa1dc46a7df9ab614fa956d034bcb8d69157dfc8

Maybe the best option would be to keep the `show` prop (so parent components can have more control on the visibility, if they need to), but at the same time keep the `dismissed` state in the component (as I think that most of the time, users will not want to control it). Oh, and later also add a prop to allow the component to be automatically hidden after a period of time. What do you think about this approach?

georgegillams: We have a task in our backlog to move `expanded` state outside too - so ultimately they will be consistent.

I personally think that a HOC to handle state would be a better approach. Maybe something like `withNormalDismissBehaviour`. @matthewdavidson  and @k0nserv  may have some other thoughts

matthewdavidson:
Hey @sogame! Cheers for this. However we are very much on the side of leaving this stuff up to userland. Whilst it's true that the majority of consumers will probably not want to control the dismiss behaviour, there will be some who do. If we were to encapsulate the dismiss functionality inside the component it would render it unusable for those consumers who need full control.

Like @georgegillams suggests we can provide a HOC to make it an easy one liner for the majority. 

Note that achieving the functionality of auto hiding after a timeout now becomes a userland concern.

matthewdavidson:
@sogame it looks like you have caught some other things as part of this change - wanna raise a separate PR to get them in? :)

I'll close this one for now.

sogame:
Ok, I understand your point. Creating the HOC would be a helpful addition :-)

I'll create a new PR with the other small changes 👍

PR #307, opened 2017-11-22
[BPK-956] Theme Toggle component

matteo-hertel:


packages/bpk-component-theme-switcher/src/theming.js
```diff
@@ -0,0 +1,17 @@
+import {
+  colorGreen100,
+  colorGreen500,
+} from 'bpk-tokens/tokens/base.es6';
+
+const bpkCustomTheme = {
```
matteo-hertel: this is where the hardcoded (for now) custom theme lives, we need to decide colours for the different components

packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
```diff
@@ -61,7 +67,7 @@ class DefaultLayout extends Component {
     const { children } = this.props;
 
     return (
-      <div>
+      <EnhancedThemeProvider themeAttributes={[...themeAttributes]}>
```
matteo-hertel: So here's where it gets interesting: I'm wrapping the whole Layout with the theme provider, the upside is that this will work across the board, the downside is that we have to provide a value for each themable attribute, there are two ways we can go: 
1. we leave like this and we'll provide all the themable attributes
2. we wrap each component with the theme provider

Thoughts?

packages/bpk-docs/package.json
```diff
@@ -13,53 +13,55 @@
     "src"
   ],
   "devDependencies": {
-    "bpk-animate-height": "^1.1.3",
```
matteo-hertel: this does not look like, hang on

changelog.md
```diff
@@ -5,6 +5,9 @@
 **Added:**
 - bpk-theming:
   - New web theming utility, see https://backpack.github.io/components/utilities/theming
+- bpk-component-theme-switcher:
+  - New component to help switching themes on the fly
+
```
shaundon: We might not need to mention this, if it's not particular user-facing.

.storybook/themableAttributes.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable global-require */
```
shaundon: Don't forget the licence :)

.storybook/config.js
```diff
@@ -5,17 +5,26 @@ import '../packages/bpk-stylesheets';
...
-    {story()}
+    <EnhancedThemeProvider themeAttributes={themeAttributes[meta.kind]}>
+      {story()}
+    </EnhancedThemeProvider>
     <br />
     <BpkGridToggle />
     <br />
     <BpkRtlToggle />
+    <br />
+    <BpkThemeSwitcher />
```
matthewdavidson: Why `ThemeSwitcher` as opposed to `ThemeToggle`?

.storybook/config.js
```diff
@@ -5,17 +5,26 @@ import '../packages/bpk-stylesheets';
...
   </div>
-));
-
+),
```
matthewdavidson: This newline seems redundant

.storybook/themableAttributes.js
```diff
@@ -0,0 +1,55 @@
...
+export default {
+  'bpk-animate-height': require('./../packages/bpk-animate-height').themeAttributes,
```
matthewdavidson: Not everything has theme attributes?? This should only contain the packages that have themeAttributes.

changelog.md
```diff
@@ -5,6 +5,9 @@
 **Added:**
 - bpk-theming:
   - New web theming utility, see https://backpack.github.io/components/utilities/theming
+- bpk-component-theme-switcher:
+  - New component to help switching themes on the fly
```
k0nserv: Is this component public?

packages/bpk-component-text/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "bpk-mixins": "^17.0.15",
     "bpk-react-utils": "^2.3.8",
+    "bpk-component-theme-switcher": "^0.0.0",
```
matthewdavidson: Why is the text component getting a dependency on theme switcher?

packages/bpk-component-theme-switcher/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "bpk-component-theme-switcher",
+  "version": "0.0.0",
+  "description": "Backpack RTL toggle component.",
```
matthewdavidson: copypasta

packages/bpk-component-theme-switcher/readme.md
```diff
@@ -0,0 +1,20 @@
...
+
+```sh
+npm install bpk-component-theme-switcher --save-dev
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkThemeSwitcher from 'bpk-component-theme-switcher';
```
matthewdavidson: This should have the docs for the HOC too.

packages/bpk-component-theme-switcher/src/BpkThemeSwitcher.js
```diff
@@ -0,0 +1,60 @@
...
+  const htmlElement = getHtmlElement();
+  htmlElement.dispatchEvent(new CustomEvent(THEME_CHANGE_EVENT, { detail: { theme } }));
+};
+
+class BpkThemeSwitcher extends React.Component {
+  constructor(props) {
+    super(props);
+    this.switchTheme = this.switchTheme.bind(this);
+    this.state = {
+      customTheme: false,
```
matthewdavidson: Can we rename to make it obvious that this is a `bool`. Like `enableCustomTheme` or something.

packages/bpk-component-theme-switcher/src/updateOnThemeChange-test.js
```diff
@@ -0,0 +1,44 @@
...
+
+const EnhancedComponent = updateOnThemeChange('div');
+
+describe('BpkThemeSwitcher', () => {
+  it('should render correctly', () => {
+    const tree = renderer.create(<EnhancedComponent />).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('should force an update when receiving a direction change event', () => {
```
matthewdavidson: copypasta

packages/bpk-component-theme-switcher/src/updateOnThemeChange.js
```diff
@@ -0,0 +1,76 @@
...
+      return !this.state.theme ?
+        (
+          <div>
+            {this.props.children}
+          </div>
+        )
+        :
+        (
+          <div>
+            <EnhancedComponent
```
matthewdavidson: Wouldn't we always render the enhanced component and just let the theme prop pass through?

packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
matthewdavidson: Well yeah, this is what consumer have to do...

We import the `themeAttributes` directly from the component packages that we wanna theme and spread them to theme provider.

packages/bpk-docs/src/themableAttributes.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable global-require */
+
+export default [
+  require('bpk-component-blockquote').themeAttributes,
```
matthewdavidson: Why are we using require here? just import them like normal

packages/bpk-component-theme-switcher/src/theming.js
shaundon: It might be better to do something like we have in native, with a `theme` object that all the specific attributes are derived from. Just makes things DRYer for if we decide to use a different colour.

https://github.com/Skyscanner/backpack/blob/master/native/storybook/themeAttributes.js#L19

Not the end of the world though as this might go away in favour of something using the branding API.

packages/bpk-component-theme-switcher/src/updateOnThemeChange.js
```diff
@@ -0,0 +1,76 @@
...
+    render() {
+      return !this.state.theme ?
+        (
+          <div>
+            {this.props.children}
+          </div>
+        )
+        :
+        (
+          <div>
```
k0nserv: Do we really need the extra div here?

packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
k0nserv: Since it's shared by the docs and the stories having all themeable attributes everywhere is fine imo

.storybook/config.js
matteo-hertel: I thought the end plan was to have multiple themes or theme on the fly applied to storybooks and docs, so it'll switch to a new theme, if it make more sense to change it to toggle I'll do it

packages/bpk-docs/src/themableAttributes.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable global-require */
```
k0nserv: This file should be called `themeableAttributes.js`

.storybook/themableAttributes.js
k0nserv: This file should be called `themeableAttributes`

packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
shaundon: I'm leaning towards the first option - as I feel that's the most optimum use of the theme provider, as one big wrapper around everything you want themed. It feels a bit redundant to wrap everything with it.

.storybook/themableAttributes.js
matteo-hertel: that's true, the thinking here is, because the storybook will load the themes on a story basis, I've added all the packages there so it'll work out of the box in the future, I could easily change it if it makes more sense

.storybook/config.js
k0nserv: I like toggle better :)

packages/bpk-component-text/package.json
k0nserv: Good catch, missed that

matteo-hertel: that's a very good question!

packages/bpk-component-theme-switcher/src/updateOnThemeChange.js
k0nserv: If you render it without providing the theme props you get prop type warnings, which we might be okay with.

packages/bpk-component-theme-switcher/readme.md
matteo-hertel: Will add, thanks

does RTL needs that as well? I copied that component and looks like it does not have it

packages/bpk-component-theme-switcher/src/BpkThemeSwitcher.js
```diff
@@ -0,0 +1,60 @@
...
+
+import React from 'react';
+import { BpkButtonLink } from 'bpk-component-link';
+
+import { getHtmlElement, THEME_CHANGE_EVENT } from './utils';
+import bpkCustomTheme from './theming';
+
+const setTheme = (theme) => {
+  const htmlElement = getHtmlElement();
+  htmlElement.dispatchEvent(new CustomEvent(THEME_CHANGE_EVENT, { detail: { theme } }));
```
k0nserv: If this component is public we should give users the ability to change the EVENT dispatched and the element we dispatch it on.

packages/bpk-component-theme-switcher/src/updateOnThemeChange.js
matteo-hertel: Not with the current set up of the theme provider, if the theme is invalid will throw errors, that's why the conditional render

packages/bpk-component-theme-switcher/src/updateOnThemeChange.js
matteo-hertel: No, I'll get rid of it

packages/bpk-docs/src/themableAttributes.js
matteo-hertel: well doing it like this I can set them into an array and filter straight off, importing them I have to import them as alias, create the array and filter, I think it's much more readable this way, wdyt?

.storybook/config.js
matteo-hertel: Toggle it is!

packages/bpk-component-theme-switcher/src/BpkThemeSwitcher.js
```diff
@@ -0,0 +1,60 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { BpkButtonLink } from 'bpk-component-link';
+
+import { getHtmlElement, THEME_CHANGE_EVENT } from './utils';
+import bpkCustomTheme from './theming';
```
k0nserv: This should also be configurable if we aim for this to be a public component

packages/bpk-component-theme-switcher/src/BpkThemeSwitcher-test.js
```diff
@@ -0,0 +1,36 @@
...
+import BpkThemeSwitcher from './BpkThemeSwitcher';
+
+describe('BpkThemeSwitcher', () => {
+  it('should render correctly', () => {
+    const tree = renderer.create(<BpkThemeSwitcher />).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('should render correctly when clicked', () => {
+    const component = renderer.create(<BpkThemeSwitcher />);
```
k0nserv: Shouldn't this test use enzyme instead?

packages/bpk-component-theme-switcher/src/updateOnThemeChange-test.js
```diff
@@ -0,0 +1,44 @@
...
+
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { mount } from 'enzyme';
+import updateOnThemeChange from './updateOnThemeChange';
+import { THEME_CHANGE_EVENT, getHtmlElement } from './utils';
+
+const EnhancedComponent = updateOnThemeChange('div');
+
+describe('BpkThemeSwitcher', () => {
```
k0nserv: Copypasta error

changelog.md
matteo-hertel: I guess so, RTL is. should it not be public?

changelog.md
matteo-hertel: Shall I remove this from the changelog? Again I've copied over what has been done for RTL
cc @matthewdavidson @k0nserv

packages/bpk-component-theme-switcher/src/BpkThemeSwitcher.js
matteo-hertel: That's the end goal, but walk before you run, we agreed on having an "hardcoded" theme for the moment and we'll decide what to do next

packages/bpk-component-theme-switcher/src/theming.js
matteo-hertel: I sure I'm missing something but I'm importing tokens and and use those values rather than hardcoded one like your example, how is that different? could be something obvious that I'm missing 😄

packages/bpk-component-theme-switcher/src/updateOnThemeChange-test.js
matteo-hertel: I'll have to fix all these error in RTL as well!

packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
matteo-hertel: so we all agree on option 1?

changelog.md
k0nserv: My take is that if we want it to be public we should make it more configurable and generic

packages/bpk-component-theme-switcher/src/theming.js
shaundon: So I was thinking it could be like this:

```
const theme = {
  primaryColor: colorGreen500,
  ...etc
};

const bpkCustomTheme = {
  blockquoteBarColor: primaryColor,
  horizontalNavBarSelectedColor: primaryColor,
  horizontalNavLinkActiveColor: primaryColor,
  horizontalNavLinkColor: primaryColor,
  ...etc
};
```

So then you get two benefits:

1 - DRYer - if we decide that the custom theme should use a different token in future, you only need to change the value of `primaryColor`.

2 - It's closer to how it'll work in real life, as the branding API will return values like `brandPrimaryColor`, `brandAccentColor` etc and then those will be used to create theme attributes.

But it's also not a big deal at all, it's just a minor code style thing after all!

packages/bpk-component-theme-toggle/src/updateOnThemeChange.js
```diff
@@ -0,0 +1,73 @@
...
+    render() {
+      return !this.state.theme ?
+        (
+          <div>
+            {this.props.children}
+          </div>
+        )
+        :
+        (<EnhancedComponent
+          theme={this.state.theme}
```
matthewdavidson: we should be able to pass null here :).

packages/bpk-theming/src/BpkThemeProvider.js
```diff
@@ -31,6 +31,9 @@ const uniq = (arr) => {
 };
 
 const createStyle = (theme, themeAttributes) => {
+  if (!theme) {
```
k0nserv: You need to change `themeAttributesPropType` to take this into account as well. Otherwise it'll still throw a prop type warning

packages/bpk-component-theme-toggle/readme.md
```diff
@@ -0,0 +1,29 @@
...
+export default () => (
+  <BpkThemeToggle />
+);
+```
+
+## HoC usage
+
+```js
+import React from 'react';
+import { updateOnThemeChange } from './../packages/bpk-component-theme-toggle';
```
matthewdavidson: Should be:
```
import { updateOnThemeChange } from 'bpk-component-theme-toggle';
```

packages/bpk-component-theme-toggle/readme.md
```diff
@@ -0,0 +1,29 @@
...
+import BpkThemeProvider from './../packages/bpk-theming';
```
matthewdavidson: Should be:
```
import BpkThemeProvider from 'bpk-theming';
```

packages/bpk-theming/readme.md
```diff
@@ -41,5 +41,5 @@ export default class App extends Component {
 | Property            | PropType        | Required | Default Value |
 | -----------         | --------------- | -------- | ------------- |
 | children            | node            | true     | -             |
-| theme               | object          | true     | -             |
+| theme               | object          | false    | {}            |
```
matthewdavidson: default value is actually `null`

matteo-hertel: ah yeah! it was `{}` yesterday

PR #314, opened 2017-11-23
[BPK-975] Make accordions themeable

k0nserv:
I changed the variables from what we initially had in mind. The same variable apply to both the title and icon in normal, active, and hover states.

Theming on the docs

screen shot 2017-11-24 at 2 15 37 pm

cc @jamesf3rguson

### DIFF
--- a/packages/bpk-component-accordion/src/bpk-accordion-item.scss
+++ b/packages/bpk-component-accordion/src/bpk-accordion-item.scss
@@ -28,21 +28,22 @@
     padding: $bpk-spacing-xs 0;
     border: 0;
     background-color: transparent;
-    color: $bpk-link-color;
     text-align: left;
     cursor: pointer;
     appearance: none;
+
+    @include bpk-themeable-property(color, --bpk-accordion-color, $bpk-link-color);
 
     @include bpk-rtl {
       text-align: right;
     }
 
     @include bpk-hover {
-      color: $bpk-link-hover-color;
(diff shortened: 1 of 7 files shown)


packages/bpk-component-accordion/src/bpk-accordion-item.scss
```diff
@@ -28,21 +28,22 @@
     padding: $bpk-spacing-xs 0;
     border: 0;
     background-color: transparent;
-    color: $bpk-link-color;
     text-align: left;
     cursor: pointer;
     appearance: none;
 
+    @include bpk-themeable-property(color, --bpk-accordion-color, $bpk-link-color);
```
matthewdavidson: This is interesting ... part of me thinks that these should be covered by `themeAttributes` from `bpk-component-link`.

matteo-hertel: but the end interface is so much better and consistent if the provide the accordion colour IMHO

matthewdavidson: From a design system perspective though, if we want to theme links, we should theme all links.

matteo-hertel: Not sure where to draw the line here, if accordion is using a link under the hood should be theme explicitly as part of the accordion and not having some external factor (e.g. change the theme for link) theming part of this component

Shall I change it?

matthewdavidson: As discussed offline, lets just leave as is.

PR #315, opened 2017-11-23
[BPK-978] Add theming to web buttons

matthewdavidson:
Note to reviewers: I am intentionally leaving out `link` buttons as I actually wanna deprecate this api in favor of `BpkLinkButton` which has already been made themable (see https://backpack.github.io/components/web/typography#links).


packages/bpk-component-button/index.js
```diff
@@ -17,5 +17,15 @@
  */
 
 import BpkButton from './src/BpkButton';
+import {
+  primaryThemeAttributes,
+  secondaryThemeAttributes,
+  linkThemeAttributes,
```
shaundon: I can't see this referred to anywhere else - is it meant to be here?

packages/bpk-mixins/src/mixins/_buttons.scss
```diff
@@ -100,23 +104,31 @@
...
 @mixin bpk-button--secondary {
-  background-color: $bpk-button-secondary-background-color;
   background-image: $bpk-button-secondary-background-image;
-  color: $bpk-button-secondary-color;
-  box-shadow: $bpk-button-secondary-box-shadow;
+
+  @include bpk-themeable-property(color, --bpk-button-secondary-text-color, $bpk-button-secondary-color);
+  @include bpk-themeable-property(background-color, --bpk-button-secondary-background-color, $bpk-button-secondary-background-color);
+  @include bpk-border-lg($bpk-button-secondary-border-color);
+  @include bpk-border-lg(var(--bpk-button-secondary-border-color, $bpk-button-secondary-border-color));
```
shaundon: I notice it's changed from `box-shadow` to using a mixin - is this just because it should have been using the mixin in the first place?

packages/bpk-component-button/index.js
matthewdavidson: Ah nice catch! I removed the link theme stuff so that is an orphan.

packages/bpk-mixins/src/mixins/_buttons.scss
matthewdavidson: Yeah, this mixin still uses box-shadow under the hood!

I think the buttons pre-date the mixin.

shaundon: Thought that might be the case :)

matthewdavidson:
@shaundon this needs reviewing again :)

packages/bpk-component-theme-toggle/src/theming.js
```diff
@@ -20,12 +20,33 @@ import {
...
   blockquoteBarColor: '#461962',
+
   horizontalNavBarSelectedColor: '#461962',
   horizontalNavLinkActiveColor: '#2d244c',
   horizontalNavLinkColor: colorBlue500,
   horizontalNavLinkHoverColor: colorBlue500,
   horizontalNavLinkSelectedColor: '#491962',
+
   spinnerPrimaryColor: '#461962',
   barchartBarBackgroundColorNormal: '#865f9e',
```
matteo-hertel: no new line here?

matthewdavidson: yeah should be, i rebased and got more props it seems. :P

PR #321, opened 2017-11-24
[no jira] Abstract out some theme props for DRYness

shaundon:
This replaces the hard-coded values with a mapping derived from the image @jamesf3rguson provided. It should mean for most uses we won't need to refer back to the image.

@matteo-hertel how come `colorBlue500` is in there - can I replace this with one of the themed colours?


matthewdavidson:
This is blocked until #319

matteo-hertel:
colorBlue is there because the nav for web will change so as part of the change I'll update that as well

packages/bpk-component-theme-toggle/src/theming.js
```diff
@@ -19,14 +19,27 @@ import {
   colorBlue500,
 } from 'bpk-tokens/tokens/base.es6';
 
+/* eslint-disable no-unused-vars */
```
matthewdavidson: Whats the unused var here?

shaundon: The secondary ones are all unused - but I think it's good to have them in there for future usage. This is essentially acting as a mini version of `bpk-tokens` for theming variables (which we may not want inside `bpk-tokens` as they're just for usage here).

matthewdavidson: Ah in that case it makes sense to wait until #315 because MY PR uses them. :)

shaundon: Damn, you got me!

PR #325, opened 2017-11-24
[BPK-985] Theme support for BpkNudger

shaundon:


packages/bpk-component-nudger/readme.md
```diff
@@ -61,3 +61,6 @@ class App extends Component {
 | value                 | number                        | true     | -             |
 | className             | string                        | false    | null          |
 
+## Theme Props
+
+Same as [secondary button](/components/web/buttons#theme-props)
```
shaundon: Figured it made sense to just put a link here and keep a single source of truth.

packages/bpk-component-nudger/src/bpk-nudger.scss
```diff
@@ -35,12 +35,4 @@
...
     vertical-align: middle;
   }
-
-  &__icon {
-    fill: $bpk-button-secondary-color;
-
-    &--disabled {
-      fill: $bpk-button-disabled-color;
-    }
-  }
```
shaundon: These seem to be redundant as the styles inherited from button are overriding them. I tested in storybook and still seems to work fine.

packages/bpk-component-nudger/src/themeAttributes-test.js
```diff
@@ -0,0 +1,35 @@
...
+      'buttonSecondaryActiveTextColor',
+      'buttonSecondaryBorderColor',
+      'buttonSecondaryHoverBorderColor',
+      'buttonSecondaryActiveBorderColor',
+      'buttonSecondaryBackgroundColor',
+      'buttonSecondaryHoverBackgroundColor',
+      'buttonSecondaryActiveBackgroundColor',
+    ]);
+  });
+});
```
shaundon: Could argue that this test is redundant as we're importing theme attributes from `bpk-component-button`. I included it to err on the side of caution, but happy for it to go.

PR #328, opened 2017-11-27
Bpk 1129 update horizontal nav colors

k0nserv:


packages/bpk-component-theme-toggle/src/theming.js
```diff
@@ -68,10 +70,10 @@ const bpkCustomTheme = {
   linkVisitedColor: theme.primaryColor700,
 
   horizontalNavBarSelectedColor: theme.primaryColor500,
-  horizontalNavLinkActiveColor: theme.primaryColor600,
```
matteo-hertel: is this correct? I think we should keep the primary color from the theme for `horizontalNavLinkActiveColor`

k0nserv: No I don't think so, we don't do in the new design of our own. It just goes darker and darker grays. Remember active is when you click an unselected item.

matteo-hertel: right, in which case all good 👍

PR #332, opened 2017-11-28
Add a HOC to BpkBannerAlert to wrap default dismissable/expandable behaviour

sogame:
Hi, as a follow up of https://github.com/Skyscanner/backpack/pull/292 I've created this initial version of the HoC to wrap the default behaviour for `dismissable` and `expandable` (I think it makes more sense and it is easier to use just one HoC wrapping both features instead of two different HoC's, one per feature).

* Remove the `expanded` state from BpkBannerAlert (it was mentioned that this was planned, to be consistent with `dismissable`)
* Create HoC with default behaviour for `dismissable` and `expandable`
* Added the prop `hideAfter` to the HoC, so it can also automatically hide the alert after a period of seconds

What to you think about this approach? If you're fine with this, I'll update the tests, create new tests for the HoC, move BpkBannerAlert into a dumb component (now that it doesn't have state) and update the documentation.

### DIFF
--- a/packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
+++ b/packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
@@ -85,21 +85,39 @@ describe('BpkBannerAlert', () => {
 
   it('should render correctly with show set false', () => {
     const tree = renderer.create(
-      <BpkBannerAlert type={ALERT_TYPES.WARN} message={message} dismissable show={false} />,
+      <BpkBannerAlert
+        type={ALERT_TYPES.WARN}
+        message={message}
+        dismissable
+        dismissButtonLabel="Dismiss"
+        show={false}
+      />,
     ).toJSON();
     expect(tree).toMatchSnapshot();
   });
 
   it('should render correctly with animateOnEnter', () => {
     const tree = renderer.create(
-      <BpkBannerAlert type={ALERT_TYPES.WARN} message={message} dismissable animateOnEnter />,
+      <BpkBannerAlert
+        type={ALERT_TYPES.WARN}
+        message={message}
+        dismissable
+        dismissButtonLabel="Dismiss"
+        animateOnEnter
+      />,
     ).toJSON();
     expect(tree).toMatchSnapshot();
   });
 
   it('should render correctly with animateOnLeave', () => {
     const tree = renderer.create(
-      <BpkBannerAlert type={ALERT_TYPES.WARN} message={message} dismissable animateOnLeave />,
+      <BpkBannerAlert
+        type={ALERT_TYPES.WARN}
+        message={message}
+        dismissable
+        dismissButtonLabel="Dismiss"
+        animateOnLeave
+      />,
     ).toJSON();
     expect(tree).toMatchSnapshot();
   });
--- a/packages/bpk-component-banner-alert/src/withBannerAlertState.js
+++ b/packages/bpk-component-banner-alert/src/withBannerAlertState.js
@@ -1,0 +1,116 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import { wrapDisplayName } from 'bpk-react-utils';
+
+const withBannerAlertState = (WrappedComponent) => {
+  class component extends Component {
+    constructor(props) {
+      super(props);
+
+      this.state = {
+        expanded: props.expanded,
+        show: true,
+      };
+
+      this.onExpandToggle = this.onExpandToggle.bind(this);
+      this.onDismiss = this.onDismiss.bind(this);
+    }
+
+    componentWillMount() {
+      const { hideAfter } = this.props;
+
+      if (hideAfter && hideAfter > 0) {
+        this.hideIntervalId = setTimeout(() => {
+          this.onDismiss();
+        }, hideAfter * 1000);
+      }
+    }
+
+    componentWillUnmount() {
+      if (this.hideIntervalId) {
+        clearTimeout(this.hideIntervalId);
+      }
+    }
+
+    onExpandToggle() {
+      const expanded = !this.state.expanded;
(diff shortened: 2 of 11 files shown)


packages/bpk-component-banner-alert/src/withNormalBehaviour.js
```diff
@@ -0,0 +1,96 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+const withNormalBehaviour = (WrappedComponent) => {
```
k0nserv: This name sounds a bit weird to me. Can we call it `withBannerAlertState` instead?

packages/bpk-component-banner-alert/src/withNormalBehaviour.js
```diff
@@ -0,0 +1,96 @@
...
+        <WrappedComponent
+          expanded={this.state.expanded}
+          onExpandToggle={this.onExpandToggle}
+          onDismiss={this.onDismiss}
+          show
+          {...rest}
+        />
+      );
+    }
+  }
```
k0nserv: We should set the componentName for this. See https://github.com/Skyscanner/backpack/blob/73b56b9840139fccad7b6f65dd66f0c964c189de/packages/bpk-scrim-utils/src/withScrim.js#L158

packages/bpk-component-banner-alert/src/withNormalBehaviour.js
```diff
@@ -0,0 +1,96 @@
...
+
+  component.propTypes = {
+    onDismiss: PropTypes.string,
```
k0nserv: This prop type is incorrect. It should be a function

k0nserv:
This is a bit strange, do we really wanna support both of these at the same time?

image


cc @jamesf3rguson

packages/bpk-component-banner-alert/src/withNormalBehaviour.js
sogame: Name changed

packages/bpk-component-banner-alert/src/withNormalBehaviour.js
sogame: I didn't knew about this, nice! I'll add it to our squad apps :-P

Added here

packages/bpk-component-banner-alert/src/withNormalBehaviour.js
sogame: True 😓  Fixed, thanks! :-)

packages/bpk-component-banner-alert/src/withBannerAlertState.js
```diff
@@ -0,0 +1,109 @@
...
+      if (this.hideIntervalId) {
+        clearTimeout(this.hideIntervalId);
+      }
+    }
+
+    onExpandToggle() {
+      const expanded = !this.state.expanded;
+      this.setState({ expanded });
+
+      if (this.props.onExpandToggle) {
```
sogame: I've added the option to bind callbacks when expanded/collapsed/dismissed, in case the user of the component needs to be notified to perform some other action on their side.

sogame:
I agree with @k0nserv, when both "expandable" and "dismissible" are set it looks a bit weird (I forgot to mention it when I opened the pull request), most likely the way this is presented to the user could be reviewed.

Not sure if using both options was expected, but I think it would make sense (in fact, in the application we're building, we would need it).

georgegillams:
@sogame could you speak to @jamesf3rguson about your application's need for dismissable, expandable banner-alerts. We haven't expected consumers to use such a configuration, so we would need to put in some work to support it 😸

sogame:
Sure, I'll talk to him on Monday (today is bank holiday in Barcelona).

sogame:
@k0nserv @georgegillams I've discussed this with @jamesf3rguson, and for now using "expandable" and "dismissible" at the same time is not supported, therefore I'll remove this option from the storybook.

Other than that, does this look good to you? If so, I'll fix the conflicts and finish the remaining work: create new tests for the HoC, move BpkBannerAlert into a dumb component (now that it doesn't have state) and update the documentation.

georgegillams:
Would it be possible to include an example in the Readme for the new HOC, otherwise people may miss it. Probably something similar to how we've documented the HOCs here https://backpack.github.io/components/web/accordions/

sogame:
Yes, sure, this is what I meant with "update the documentation" :-)

Thanks for the example link, that will make it easier for me :-)

packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
```diff
@@ -85,21 +85,39 @@ describe('BpkBannerAlert', () => {
 
   it('should render correctly with show set false', () => {
     const tree = renderer.create(
-      <BpkBannerAlert type={ALERT_TYPES.WARN} message={message} dismissable show={false} />,
+      <BpkBannerAlert
+        type={ALERT_TYPES.WARN}
+        message={message}
+        dismissable
+        dismissButtonLabel="Dismiss"
```
sogame: `dismissButtonLabel` was missing, a warning was displayed in the console (and due to line length, I had to break it)

sogame:
@georgegillams @k0nserv, removed `WIP` from the PR, I think that all changes are there, now :-)

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -258,6 +260,36 @@ const components = [
...
+    blurb: [
+      <Paragraph>
+        Most common use of &quot;expandable&quot; and &quot;dismissable&quot; can be easily achieved
+        using the &quot;withBannerAlertState&quot; HOC.
+      </Paragraph>,
+      <Paragraph>
+        It also adds the option to automatically dismiss a banner after a certain period of time has elapsed.
+      </Paragraph>,
+    ],
+    examples: [
```
georgegillams: I'm not sure there's much point in showing these examples - I think it's probably enough to just show how to use the HOC in the readme.
But it's a minor point and doesn't really matter. The HOC works really well 👍

sogame: Up to you, it's your documentation 😂 

If you want, let's see what other Backpack members think, and I'll update it accordingly, no problem :-)

I've just pushed a rebased version with latest master changes (conflicts in `changelog.md`)

georgegillams: The consensus seems to be to remove the new examples, as we already have examples of the functionality the HOC provides.
But the documentation of the HOC usage is good and should stay!

sogame: Done! I've just removed it :-)

Anything else, just let me know :-)

georgegillams: All LGTM 👍 
Happy to merge!

PR #334, opened 2017-11-28
[BPK-979] Add theming for calendar

k0nserv:
This adds support for theming the calendar. On top of the specified themeable attributes I had to add `calendarDateTextFocusColor` and `calendarDateFocusedBorderColor` which were not accounted for.

**Note:** The Datepicker suffers from the same issues as all components that use the portal, namely it doesn't work.

cc @jamesf3rguson, @matthewdavidson

### DIFF
--- a/packages/bpk-component-calendar/src/bpk-calendar-date.scss
+++ b/packages/bpk-component-calendar/src/bpk-calendar-date.scss
@@ -25,23 +25,26 @@
   border: none;
   border-radius: $bpk-border-radius-pill;
   background-color: $bpk-calendar-day-background-color;
-  color: $bpk-calendar-day-color;
   font-size: $bpk-font-size-base;
   text-align: center;
   cursor: pointer;
   appearance: none;
   box-sizing: border-box;
 
+  @include bpk-themeable-property(color, --bpk-calendar-date-text-color, $bpk-calendar-day-color);
+
   @include bpk-hover {
     &:not(.bpk-calendar-date--selected) {
       background-color: $bpk-calendar-day-hover-background-color;
-      color: $bpk-calendar-day-hover-color;
+
+      @include bpk-themeable-property(color, --bpk-calendar-date-text-hover-color, $bpk-calendar-day-hover-color);
     }
   }
 
   &:not(.bpk-calendar-date--selected):active {
     background-color: $bpk-calendar-day-active-background-color;
-    color: $bpk-calendar-day-active-color;
+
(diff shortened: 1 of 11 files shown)


packages/bpk-component-calendar/src/bpk-calendar-date.scss
```diff
@@ -54,15 +57,18 @@
...
     cursor: default;
+
+    @include bpk-themeable-property(background-color, --bpk-calendar-date-selected-background-color, $bpk-state-selected-background-color);
+    @include bpk-themeable-property(color, --bpk-calendar-date-text-selected-color, $bpk-calendar-day-selected-color);
   }
 
   &--focused:not(:disabled):not(.bpk-calendar-date--selected) {
-    color: $bpk-state-selected-background-color;
     box-shadow: 0 0 0 2px $bpk-state-selected-background-color inset;
+    box-shadow: 0 0 0 2px var(--bpk-calendar-date-focused-border-color, $bpk-state-selected-background-color) inset;
```
matteo-hertel: this looks like a copypasta error, is the box shadow themable?

packages/bpk-component-calendar/src/bpk-calendar-nav.scss
```diff
@@ -47,22 +47,22 @@
     cursor: pointer;
     appearance: none;
 
+    @include bpk-themeable-property(color, --bpk-calendar-nudger-icon-color, $bpk-calendar-nav-icon-fill);
+
     @include bpk-hover {
-      .bpk-calendar-nav__icon {
-        fill: $bpk-calendar-nav-icon-hover-fill;
-      }
+      @include bpk-themeable-property(color, --bpk-calendar-nudger-icon-hover-color, $bpk-calendar-nav-icon-hover-fill);
```
matteo-hertel: Not sure I understand here, you're not applying the fill to the icon but to the hover of the button?

packages/bpk-component-calendar/src/bpk-calendar-date.scss
k0nserv: It's intentional. `box-shadow` is too advanced to use `bpk-themeable-property`. The above line is a fallback for browsers that don't support CSS variables.

packages/bpk-component-calendar/src/bpk-calendar-nav.scss
k0nserv: I'm using `fill: currentColor` which means the icon picks it up from the button's color. Then I'm just setting the button color for the different states.

PR #335, opened 2017-11-28
[BPK-984] themable modal

matteo-hertel:
This will enable modal to be themed, there is a non breaking change to the `Portal` component.

The problems is, by default the portal creates and empty `div`, appends it to the `body` and then the content is rendered, I've introduced a new prop called `renderTarget`, to specify which element the portal needs to be rendered into, this allow us to have a div in the scope of the theme provider, render the modal inside and it'll have access to the css vars set by the theme provider making the modal themable, if no  `renderTarget` is provided, the `Portal` will fall back to the body, making this a backward compatible change


packages/bpk-component-close-button/src/bpk-close-button.scss
```diff
@@ -26,10 +26,10 @@
   appearance: none;
 
   &__icon {
-    fill: $bpk-modal-close-icon-fill;
+    @include bpk-themeable-property(fill, --bpk-modal-close-button-color, $bpk-modal-close-icon-fill);
```
matteo-hertel: the name is to be consistent with the Jira card, can be changed

packages/bpk-component-close-button/src/bpk-close-button.scss
```diff
@@ -26,10 +26,10 @@
...
 
     @include bpk-hover {
-      fill: $bpk-modal-close-icon-hover-fill;
+      @include bpk-themeable-property(fill, --bpk-modal-close-button--hover-color, $bpk-modal-close-icon-hover-fill);
```
matteo-hertel: same as above, name can be changed

packages/bpk-component-modal/src/BpkModal.js
```diff
@@ -44,10 +48,12 @@ const BpkModal = (props) => {
 BpkModal.propTypes = {
   isOpen: PropTypes.bool.isRequired,
   onClose: PropTypes.func.isRequired,
-  target: PropTypes.element,
+  renderTarget: PropTypes.instanceOf(Element),
```
matteo-hertel: `.instanceOf(Element)` will accept any DOM element

packages/bpk-component-modal/src/BpkModalDialog-test.js
```diff
@@ -70,6 +70,45 @@ describe('BpkModalDialog', () => {
...
+    const tree = renderer.create(
+      <BpkModalDialog
+        id="my-modal"
+        className="my-classname"
+        wide
+        title="Modal title"
+        onClose={jest.fn()}
+        closeLabel="Close"
+        closeEvents={closeEvents}
+        getDialogRef={jest.fn()}
```
k0nserv: This prop is changed on master `dialogRef` instead of `getDialogRef`

packages/bpk-component-modal/src/BpkModalDialog-test.js
```diff
@@ -70,6 +70,45 @@ describe('BpkModalDialog', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('should render correctly with wide prop', () => {
```
matteo-hertel: this branch of the code was not covered so I've added the test to bring code coverage to 100%

packages/bpk-component-modal/src/BpkModal.js
```diff
@@ -29,13 +29,17 @@ const ScrimBpkModalDialog = withScrim(BpkModalDialog);
...
+    target,
+    renderTarget,
+    ...rest
   } = props;
 
   delete rest.onClose;
 
   return (
-    <Portal isOpen={isOpen} onClose={onClose} target={target}>
+    <Portal isOpen={isOpen} onClose={onClose} target={target} renderTarget={renderTarget}>
```
k0nserv: So to support theming the users has to supply `renderTarget` otherwise it will not work?

packages/bpk-react-utils/src/Portal.js
```diff
@@ -224,25 +231,27 @@ class Portal extends Component {
 Portal.propTypes = {
   children: PropTypes.node.isRequired,
   isOpen: PropTypes.bool.isRequired,
-  target: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
```
matteo-hertel: I've sorted the props that's why the changes

packages/bpk-component-modal/src/BpkModalDialog-test.js
matteo-hertel: I'll rebase then 😄

matteo-hertel: Sorry I missed the point, I'll update the test

k0nserv:
I'm not crazy about the user having to provide a `renderTarget`, but maybe there's no way around that...

matteo-hertel:
@shaundon Nice catch on the test, I've actually added one in, got removed because of a rebase 😅 

the `target` is the element the Portal uses to get all the events (click, keystrokes etc), it falls back on the element itself maybe that's why is not documented

I'll re-add the test for the `renderTarget`, do you agree with the approach here?

packages/bpk-component-close-button/src/bpk-close-button.scss
matthewdavidson: should just be `--bpk-modal-close-button-hover-color` :)

packages/bpk-component-modal/src/themeAttributes-test.js
```diff
@@ -0,0 +1,32 @@
...
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+  */
+
+import themeAttributes from './themeAttributes';
+
+describe('themeAttributes', () => {
+  it('should export the correct theme attributes', () => {
+    expect(themeAttributes).toEqual([
+      'linkColor',
```
matthewdavidson: This looks like copypasta - link stuff shouldn't be in here right?

matthewdavidson: Or if it should, then it needs to be added to the readme?

matteo-hertel: is not an error, the `themeAttributes`from the `bpk-component-link` have been added to theme the type of modal with the closing button as link

packages/bpk-component-modal/src/BpkModal-test.js
```diff
@@ -0,0 +1,59 @@
...
+    closeEvents = {
+      onTouchStart: jest.fn(),
+      onTouchMove: jest.fn(),
+      onTouchEnd: jest.fn(),
+      onMouseDown: jest.fn(),
+      onMouseMove: jest.fn(),
+      onMouseUp: jest.fn(),
+    };
+  });
+  it('should render correctly in the given target if renderTarget is supplied', () => {
```
matteo-hertel: this is new as I need to test the `renderTarget` the expected behaviour is to see the modal content in the div

packages/bpk-component-modal/src/__snapshots__/BpkModal-test.js.snap
```diff
@@ -0,0 +1,69 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`BpkModal should render correctly in the given target if renderTarget is supplied 1`] = `null`;
+
+exports[`BpkModal should render correctly in the given target if renderTarget is supplied 2`] = `
```
matteo-hertel: 🎉

matteo-hertel:
I've added the tests in for the render target, this PR is not complete as readme and docs are still missing, I think it's time to take a decision, do we go with this approach or not? if so I'll finish this card and move onto the other component that are using portals

webpack.config.babel.js
```diff
@@ -190,6 +190,9 @@ if (isProduction) {
       entry: 'docs',
       paths: staticSiteGeneratorConfig.paths,
       locals: staticSiteGeneratorConfig,
+      globals: {
+        Element: {},
```
matteo-hertel: this is to allow the `PropTypes.instanceOf(Element)` rule, while building the docs `Element` is undefined and breaks the whole thing 😕

k0nserv:
Was testing it locally, but it's currently broken. 
image

k0nserv:
Fix

```
diff --git a/packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js b/packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
index 2265740d..4fc8fce7 100644
--- a/packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
+++ b/packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
@@ -71,7 +71,7 @@ class ModalContainer extends Component {
           isOpen={this.state.isOpen}
           onClose={this.onClose}
           wide={this.state.wide}
-          renderTarget={() => document.getElementById('theamableModalTarget')}
+          renderTarget={document.getElementById('theamableModalTarget')}
           {...rest}
         >
           <div>{children}</div>
```

matteo-hertel:
That's a bit of a problem, the idea is to have `document.getElementById('theamableModalTarget')` not `() => document.getElementById('theamableModalTarget')` but while building the docs site `document` is not available, SSR does not have a `document`, I'll have to change the API to allow both

k0nserv:
Isn't it better to use refs over `document.getElementById`anyway?

packages/bpk-component-modal/src/BpkModal.js
```diff
@@ -44,10 +48,12 @@ const BpkModal = (props) => {
 BpkModal.propTypes = {
   isOpen: PropTypes.bool.isRequired,
   onClose: PropTypes.func.isRequired,
-  target: PropTypes.element,
+  renderTarget: PropTypes.instanceOf(Element),
+  target: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
```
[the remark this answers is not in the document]
matteo-hertel: that's not correct, let me check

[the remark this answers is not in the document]
matteo-hertel: actually now that I remember the `target` is passed straight through to the `Portal` shouldn't the validation be the same?

packages/bpk-react-utils/readme.md
```diff
@@ -63,15 +63,16 @@ class MyComponent extends Component {
...
-| beforeClose | func     | false    | null          |
+| Property     | PropType | Required | Default Value |
+| ------------ | -------- | -------- | ------------- |
+| children     | node     | true     | -             |
+| isOpen       | bool     | true     | -             |
+| beforeClose  | func     | false    | null          |
+| onClose      | func     | false    | noop          |
+| onOpen       | func     | false    | noop          |
+| onRender     | func     | false    | noop          |
+| renderTarget | node     | false    | null          |
```
k0nserv: `renderTarget` is not a `node`. It should be an instance of `Element`

packages/bpk-component-modal/readme.md
```diff
@@ -79,4 +84,12 @@ class App extends Component {
 | getApplicationElement | func                 | true     | -             |
 | closeLabel            | string               | false    | null          |
 | closeText             | string               | false    | null          |
+| renderTarget          | node                 | false    | null          |
```
k0nserv: `renderTarget` wrong here too

changelog.md
```diff
@@ -2,7 +2,12 @@
 
 ## UNRELEASED
 
-_Nothing yet..._
```
k0nserv: This is not under a section

matteo-hertel: I've added `**Added:** `

PR #338, opened 2017-11-29
Create pagination component

ksiwei:


packages/bpk-component-pagination/src/BpkPaginationList.js
```diff
@@ -0,0 +1,83 @@
...
+    currentPage, pageCount, onPageChange, visibleRange,
+  } = props;
+
+  const children = [];
+  const shoulderRange = Math.ceil(visibleRange / 2);
+  const isFirstPageSelected = currentPage === 1;
+  const isLastPageSelected = currentPage === pageCount;
+  const displayRange = isFirstPageSelected || isLastPageSelected ? shoulderRange + 1 : shoulderRange;
+
+  for (let i = 1; i <= pageCount; i += 1) {
```
k0nserv: This could be written as a map expression. `(new Array(pageCount)).map(_, i) => ...`. However this would work better if the whole component used zero-indexing which I also suggest we do

packages/bpk-component-pagination/src/BpkPaginationNudger.js
```diff
@@ -0,0 +1,68 @@
...
+import { cssModules } from 'bpk-react-utils';
+
+import ArrowLeftIcon from 'bpk-component-icon/lg/arrow-left';
+import ArrowRightIcon from 'bpk-component-icon/lg/arrow-right';
+
+import STYLES from './bpk-pagination-nudger.scss';
+
+const getClassName = cssModules(STYLES);
+
+const nudgerIcon = forward =>
```
k0nserv: Consider RTL support for this too.

packages/bpk-component-pagination/src/bpk-pagination-nudger.scss
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-pagination-nudger {
+  text-align: center;
+  vertical-align: middle;
+  box-sizing: border-box;
```
k0nserv: Backpack components expect border-box to be part of the normalization so this is redundant.

packages/bpk-component-pagination/src/bpk-pagination-nudger.scss
```diff
@@ -0,0 +1,57 @@
...
+    padding: $bpk-spacing-xs;
+    border: none;
+    background: none;
+    line-height: 100%; /* stylelint-disable-line scale-unlimited/declaration-strict-value */
+    cursor: pointer;
+    appearance: none;
+
+    @include bpk-hover {
+      .bpk-pagination-nudger__icon {
+        fill: $bpk-color-blue-700;
```
k0nserv: Can we use `currentColor` and setting the color of the button instead?

k0nserv:
Should we have a HOC for this component that encapsulates rendering different content per page and takes care of state? Something like

```javascript
import BpkPagination, { withPaginationState } from 'bpk-component-pagination';

const pageData = [{
  label: "Page 1",
}, {
  label: "Page 2",
}, {
  label: "Page 3",
}];

const PaginatedView = withPaginationState(BpkPagination, pagesData,
  (pageDate, pageIndex) => <PageImplementation data={pageData} />
);

() => (
  <PaginatedView />
);
```

k0nserv:
Style breaks when theming is applied. We are using the selected style here, which AFAIK we are deprecating and hence it's not themed. @jamesf3rguson thoughts?

image

ksiwei:
@k0nserv I like the HOC idea but it will have to be an opt-in option so that consumer who aren't using React for page content (e.g. dayview) can still use the pagination component.

packages/bpk-component-pagination/src/BpkPagination-test.js
```diff
@@ -0,0 +1,100 @@
...
+      currentPage={2}
+      previousLabel="previous"
+      nextLabel="next"
+    />);
+
+    expect(pagination.find({ label: 'next' }).length).toBe(0);
+    expect(pagination.find({ label: 'previous' }).length).toBe(1);
+  });
+
+  it('should call the onPageChange callback when select page', () => {
```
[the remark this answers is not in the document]
ksiwei: it is for testing whether the callback is triggered when page button(with the number as label) is clicked. Similar to the nudger test below.

packages/bpk-component-pagination/src/BpkPaginationList.js
ksiwei: This isn't 1-to-1 mapping, I thought about map expression but then I will have to filter out the null ones. Good point with 0-indexing though.

k0nserv: Is that really problematic though? Just go

```javascript
const children = (new Array(pageCount)).map((_, i) => ...).filter(x => !!x)
```

packages/bpk-component-pagination/src/BpkPagination-test.js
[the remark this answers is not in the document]
k0nserv: But you aren't actually triggering it via interactions in the DOM since you just call it yourself?

packages/bpk-component-pagination/src/BpkPaginationList.js
ksiwei: yeah, but even for mapping an empty array, you will still have to handle it like this https://stackoverflow.com/a/35086350, a bit of overkill, don't you think?

k0nserv: Dunno I personally think it still reads a lot easier than the current imperative version with a for loop.

changelog.md
```diff
@@ -20,6 +20,8 @@ _Nothing yet..._
   - Calendar now supports theming.
 - bpk-component-drawer: 1.0.0
   - A new component which creates a drawer that slides into view from the side of the screen
+- bpk-component-pagination
```
k0nserv: This needs to be moved to the `UNRELEASED` section

packages/bpk-component-pagination/src/BpkPagination.js
```diff
@@ -0,0 +1,95 @@
...
+    visibleRange,
+    className,
+    ...rest
+  } = props;
+
+  if (className) { classNames.push(getClassName(className)); }
+
+  const handlePageChange = (nextPageIndex) => {
+    if (onPageChange &&
+      nextPageIndex <= pageCount - 1 &&
```
k0nserv: Can be simplified to `nextPageIndex < pageCount`

packages/bpk-component-pagination/src/BpkPagination.js
```diff
@@ -0,0 +1,95 @@
...
+    onPageChange,
+    selectedPageIndex,
+    visibleRange,
+    className,
+    ...rest
+  } = props;
+
+  if (className) { classNames.push(getClassName(className)); }
+
+  const handlePageChange = (nextPageIndex) => {
```
k0nserv: Minor: We end up creating this function every time `render` is called. Maybe we should instead move it out and change the signature to 

```javascript
const handlePageChange = onPageChange => nextPageIndex => {
  // ...
}
```

This way it can be partially applied in render as `handelPageChange(onPageChange)` and on line 65 we can skip creating an anonymous function extra

packages/bpk-component-pagination/src/BpkPagination.js
```diff
@@ -0,0 +1,95 @@
...
+          forward
+        />
+      }
+    </section>
+  );
+};
+
+BpkPagination.propTypes = {
+  selectedPageIndex: PropTypes.number.isRequired,
+  onPageChange: PropTypes.func,
```
k0nserv: This should be after the required props

packages/bpk-component-pagination/src/BpkPaginationList-test.js
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { shallow } from 'enzyme';
+import BpkPaginationList from './BpkPaginationList';
+import BpkPaginationBreak from './BpkPaginationBreak';
+
+describe('BpkPaginationList', () => {
```
k0nserv: No snapshot tests for this?

packages/bpk-component-pagination/src/BpkPaginationList.js
```diff
@@ -0,0 +1,85 @@
...
+  const {
+    selectedPageIndex, pageCount, onPageChange, visibleRange,
+  } = props;
+
+  const shoulderRange = Math.ceil(visibleRange / 2);
+  const isFirstPageSelected = selectedPageIndex === 0;
+  const isLastPageSelected = selectedPageIndex === pageCount - 1;
+  const displayRange = isFirstPageSelected || isLastPageSelected ? shoulderRange + 1 : shoulderRange;
+
+  const children = new Array(pageCount).fill(undefined).map((_, i) => {
```
k0nserv: 👍 👍 👍

packages/bpk-component-pagination/src/BpkPaginationList-test.js
ksiwei: don't think it is very useful.

matthewdavidson:
Hey @ksiwei! Thanks for this - looks awesome. 

@jamesf3rguson has had a look from a design POV too.

I have one request: Instead of not rendering the next/prev button, can we still render them and disable them similar to the calendar component:

image

Regardless, can you make the component private so we can merge this particular PR asap?

🎉 🎉 🎉 🎉

PR #341, opened 2017-12-01
[BPK-1134] Add shortcut for switching theme

georgegillams:


packages/bpk-component-rtl-toggle/src/BpkRtlToggle.js
```diff
@@ -41,7 +41,6 @@ class BpkRtlToggle extends React.Component {
     };
   }
 
-
```
matteo-hertel: thanks 😄

georgegillams: Well I couldn't just leave an ~extrean~ ~exteraeneo~ ~unessessary~ not needed blank line, could I 😆

matteo-hertel: @matthewdavidson would be proud of you!

PR #345, opened 2017-12-02
Add className to card props

mush159:


packages/bpk-component-card/readme.md
```diff
@@ -28,5 +28,6 @@ export default () => (
 | Property  | PropType | Required | Default Value |
 | --------- | -------- | -------- | ------------- |
 | children  | -        | true     | -             |
+| className | string   | false    | null          |
```
georgegillams: Personally, I feel that certain props are unnecessary if they exist on all components, or exist on the main underlying component. However, I know that we include `className` in other readme's. This is probably something that we need to standardize on.
@matthewdavidson  @k0nserv  @shaundon  @matteo-hertel  what do we think?

mush159: Yeh thats a good point. It's a bit annoying at the moment when seeing a component without it(knowing that others have it) as I then finding myself looking through the code to make sure it is being passed down.
Personally, I'd rather they be in the readme's as it'd be less confusing for those completely new to backpack. Particularly if coming from other systems were it is explicitly stated
http://www.material-ui.com/#/components/flat-button
http://blueprintjs.com/docs/#core/components/button

shaundon: I see your point there @mush159. Maybe we should be more explicit about this sort of thing, it'll make things easier for consumers after all, even if it is more overhead to add it.

shaundon:
Thanks again for the PR @mush159 😄

packages/bpk-component-card/readme.md
georgegillams: I think we should actually automate the props table, as at the moment we have to define everything twice. If we automated it, we could decide what to automatically include etc 😸

matthewdavidson: @georgegillams sounds like a christmas project ;)

georgegillams: Could be a song: George's 12 scripts of Christmas 😆  🤣

matteo-hertel: something like http://www.episodeyang.com/react-component-props-table/ ?

shaundon: @matteo-hertel that looks awesome!

PR #346, opened 2017-12-04
[BPK-987] themeable popover

matteo-hertel:
Themeable popover with docs and storybook

### DIFF
--- a/packages/bpk-component-popover/readme.md
+++ b/packages/bpk-component-popover/readme.md
@@ -42,22 +42,30 @@ class App extends Component {
 
   render() {
     return (
-      <BpkPopover
-        id="my-popover"
-        target={
-          <BpkButton onClick={this.openPopover}>Open</BpkButton>
-        }
-        onClose={this.closePopover}
-        isOpen={this.state.isOpen}
-        label="My popover"
-        closeButtonText="Close"
-      >
-        <BpkText>My popover content</BpkText>
-      </BpkPopover>
+      <div id="popover-container">
+        <BpkPopover
+          id="my-popover"
+          target={
+            <BpkButton onClick={this.openPopover}>Open</BpkButton>
+          }
+          onClose={this.closePopover}
+          isOpen={this.state.isOpen}
+          label="My popover"
+          closeButtonText="Close"
+          renderTarget={() =>
+            document.getElementById('popover-container')
+          }
+        >
+          <BpkText>My popover content</BpkText>
+        </BpkPopover>
+      </div>
     );
   }
 }
 ```
+
+> **Theming:** In order to theme the modal, a `renderTarget` needs to be supplied as a function which returns a DOM node
+  in the scope of a `BpkThemeProvider`.
 
 ## Props
 
@@ -76,6 +84,7 @@ class App extends Component {
 | closeButtonIcon       | bool                          | false    | true          |
 | portalStyle           | object                        | false    | null          |
 | portalClassName       | string                        | false    | null          |
+| renderTarget          | func                          | false    | null          |
 
 In order to attach the popover to a regular DOM element, provide a function which returns it to `target`:
 
@@ -101,3 +110,13 @@ const onClose = (event, {
   ...
 }
 ```
+
+## Theme Props
+
+* `linkColor`
+* `linkHoverColor`
+* `linkActiveColor`
+* `linkVisitedColor`
+* `closeButtonColor`
+* `closeButtonHoverColor`
+* `closeButtonActiveColor`
--- a/packages/bpk-component-popover/stories.js
+++ b/packages/bpk-component-popover/stories.js
@@ -97,22 +97,25 @@ class PopoverContainer extends Component {
       openButton = null;
     }
     return (
-      <div>
+      <div id="popover-container">
         {openButton}
         <BpkPopover
+          closeButtonText="Close"
           id={`my-popover-${id}`}
-          target={target}
-          onClose={this.closePopover}
           isOpen={this.state.isOpen}
           label="My popover"
-          closeButtonText="Close"
+          onClose={this.closePopover}
+          renderTarget={() =>
+            document.getElementById('popover-container')
+          }
+          target={target}
           {...rest}
         >
           <BpkContentContainer>
             <Paragraph>My popover content.</Paragraph>
             <Paragraph>Some more popover content.</Paragraph>
-            { changeProps
-              ?
+            {
+              changeProps ?
                 <Paragraph>
                   <BpkButton onClick={this.changeContent}>
                     Change content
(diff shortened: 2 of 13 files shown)


packages/bpk-component-popover/stories.scss
```diff
@@ -6,7 +6,7 @@
 
 .bpk-popover-spacer {
   display: flex;
-  height: 500 * $bpk-one-pixel-rem;
+  height: 100 * $bpk-one-pixel-rem;
```
matteo-hertel: this change is because the "switch theme" button was outside the viewport in the storybook, 500 look too big anyway so I changed it

packages/bpk-component-popover/readme.md
```diff
@@ -42,23 +42,30 @@ class App extends Component {
...
+          }
+          onClose={this.closePopover}
+          isOpen={this.state.isOpen}
+          label="My popover"
+          closeButtonText="Close"
+          renderTarget={this.renderTarget}
+        >
+          <BpkText>My popover content</BpkText>
+        </BpkPopover>
+        <div ref={(target) => { this.renderTarget = target; }} />
```
matthewdavidson: This extra `div` is extraneous surely? You can use the one defined on line 45 no?

packages/bpk-component-popover/readme.md
```diff
@@ -42,23 +42,30 @@ class App extends Component {
...
+      </div>
     );
   }
 }
 ```
 
+**Theming:** In order to theme the modal a `renderTarget` needs to be supplied as a DOM reference of a node in
```
matthewdavidson: I have this wording and api changed locally - will submit the PR shortly.

packages/bpk-component-popover/readme.md
```diff
@@ -101,3 +109,10 @@ const onClose = (event, {
   ...
 }
 ```
+
+## Theme Props
+
+Both all the props from [link](/components/web/typography#links) and
+
+* `modalCloseButtonColor`
```
matthewdavidson: Similarly I have fixed the names of these in my PR (should be `closeButtonColor`, `closeButtonHoverColor` and
`closeButtonActiveColor`)

packages/bpk-docs/src/pages/PopoversPage/PopoversPage.js
```diff
@@ -62,27 +62,31 @@ class PopoverContainer extends Component {
...
+              <BpkButton onClick={this.openPopover}>Open</BpkButton>
+          }
+          onClose={this.closePopover}
+          isOpen={this.state.isOpen}
+          label="My popover"
+          closeButtonText="Close"
+          renderTarget={this.theamablePopoverTarget}
+          {...rest}
+        />
+        <div ref={(target) => { this.theamablePopoverTarget = target; }} />
```
matthewdavidson: Have you tested the implications of this? i.e. the popover is no longer appended to the body thereforr zindex issues may occur.

packages/bpk-component-popover/readme.md
matteo-hertel: could do, but that would give the wrong idea of wrapping a `BpkPopover` with a div for the `renderTarget`, the outer `div` is a container, `BpkPopover` and the `renderTarget` are siblings, happy to change it tough!

packages/bpk-component-popover/readme.md
matteo-hertel: Cool, I'll update once you submit so we don't need a new PR

packages/bpk-component-popover/readme.md
matteo-hertel: I guess your PR should go in first and I'll update mine later on

packages/bpk-docs/src/pages/PopoversPage/PopoversPage.js
matteo-hertel: I have tested a few scenarios and it's all holding up 🤞

packages/bpk-component-popover/readme.md
matthewdavidson: Yeah the PR is #347

packages/bpk-component-popover/readme.md
matthewdavidson: See the example https://github.com/Skyscanner/backpack/pull/347/files#diff-0ef12b7e9b89d736d7a6cb0987c78837R70

packages/bpk-component-popover/readme.md
```diff
@@ -42,23 +42,31 @@ class App extends Component {
...
+        >
+          <BpkText>My popover content</BpkText>
+        </BpkPopover>
+      </div>
     );
   }
 }
 ```
 
+> **Theming:** In order to theme the popover, a `renderTarget` needs to be supplied as a function which returns a DOM node
```
matteo-hertel: @matthewdavidson should we point out the support for refs? technically we support both 🤔

matthewdavidson: Erm we could ... I guess hugo probably wants that too. I guess also we'd need to update the modal docs too 🙈

packages/bpk-component-popover/readme.md
```diff
@@ -76,6 +84,7 @@ class App extends Component {
 | closeButtonIcon       | bool                          | false    | true          |
 | portalStyle           | object                        | false    | null          |
 | portalClassName       | string                        | false    | null          |
+| renderTarget          | func                          | false    | null          |
```
matthewdavidson: Also this should be `oneOf([func, node])`

packages/bpk-component-popover/stories.js
```diff
@@ -121,16 +124,19 @@ class PopoverContainer extends Component {
...
-                  This is some long content.
+                This is some long content.
                   This is some long content.
                   This is some long content.
                 </Paragraph>
-              : null }
+              : null}
           </BpkContentContainer>
         </BpkPopover>
+        <div ref={(renderTarget) => { this.renderTarget = renderTarget; }} />
```
matthewdavidson: this is extraneous right?

matteo-hertel: yep! nice catch!

packages/bpk-component-popover/readme.md
matteo-hertel: I can put it in my PR, is not a major issue, something like
```
> **Theming:** In order to theme the popover, a `renderTarget` needs to be supplied as either a function which returns a DOM node or a [ref](https://reactjs.org/docs/refs-and-the-dom.html) in the scope of a `BpkThemeProvider`.
```

matthewdavidson: ya!

packages/bpk-component-popover/readme.md
matteo-hertel: right, I copy pasted from modal 😇

packages/bpk-react-utils/src/Portal.js
```diff
@@ -147,7 +147,7 @@ class Portal extends Component {
       return this.props.renderTarget();
     }
 
-    return this.props.renderTarget || document.body;
```
matteo-hertel: @matthewdavidson we're now accepting a function as render target so this bit can be removed right?

matthewdavidson: Sure 👍

PR #348, opened 2017-12-04
[BPK-981] Supporting theming checkboxes and radio buttons

k0nserv:


packages/bpk-component-checkbox/src/BpkCheckbox.js
```diff
@@ -49,6 +52,10 @@ const BpkCheckbox = (props) => {
         disabled={disabled}
         {...rest}
       />
+      <Tick
+        className={getClassName('bpk-checkbox__icon')}
+        style={{ width: bpkSpacingSm, height: bpkSpacingSm }}
```
matthewdavidson: Why set width/height via inline style vs scss?

packages/bpk-component-checkbox/src/bpk-checkbox.scss
```diff
@@ -41,6 +41,10 @@
     @include bpk-checkbox__input;
   }
 
+  &__icon {
+    @include bpk-checkbox__icon;
```
matthewdavidson: Do we think folk will use this mixin?

packages/bpk-tokens/tokens/base.common.js
```diff
@@ -204,6 +204,9 @@ module.exports = {
   colorYellow800: "rgb(194, 142, 0)",
   colorYellow900: "rgb(156, 114, 0)",
   primaryGradient: "linear-gradient(135deg, #00b2d6 0%, #02DDD8 100%)",
+  checkboxCheckedTickColor: "rgb(0, 178, 214)",
+  checkboxDisabledTickBackgroundColor: "rgb(230, 228, 235)",
```
matthewdavidson: Where does this come from?

k0nserv: I should remove those I reverted the approach

packages/bpk-component-checkbox/src/BpkCheckbox.js
k0nserv: Because the component itself sets it via style and that overrides any values set from scss

packages/bpk-component-checkbox/src/bpk-checkbox.scss
k0nserv: Dunno, it's something we should discuss for sure. Should we just remove all these mixins at this point?

packages/bpk-component-checkbox/src/BpkCheckbox.js
matthewdavidson: Cool ok - just acknowledging the hackety hack,

packages/bpk-component-checkbox/src/__snapshots__/BpkCheckbox-test.js.snap
```diff
@@ -10,6 +10,23 @@ exports[`BpkCheckbox should render correctly 1`] = `
     name="checkbox"
     type="checkbox"
   />
+  <svg
+    className="bpk-checkbox__icon"
+    height="18"
+    style={
+      Object {
+        "height": undefined,
+        "width": undefined,
```
matthewdavidson: Why are these undefined?

packages/bpk-component-checkbox/src/__snapshots__/BpkCheckbox-test.js.snap
```diff
@@ -28,6 +45,23 @@ exports[`BpkCheckbox should render correctly with a small label 1`] = `
...
```
matthewdavidson: Same as above ^

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -468,11 +468,32 @@
     background: none !important; /* stylelint-disable-line declaration-no-important */
   }
 
+  + .bpk-checkbox__icon {
```
matthewdavidson: This class shouldn't exist anywhere in `bpk-mixins` package - can it not be in the components scss file?

packages/bpk-component-checkbox/src/__snapshots__/BpkCheckbox-test.js.snap
k0nserv: Good question

packages/bpk-mixins/src/mixins/_forms.scss
k0nserv: Yeah fair it should be possible to move it

PR #350, opened 2017-12-05
[BPK-1133] Add theming to datepickers

matthewdavidson:
Ronseal.


packages/bpk-component-datepicker/src/themeAttributes.js
```diff
@@ -0,0 +1,27 @@
...
+
+import { themeAttributes as calendarAttributes } from 'bpk-component-calendar';
+import { themeAttributes as modalAttributes } from 'bpk-component-modal';
+import { themeAttributes as popoverAttributes } from 'bpk-component-popover';
+
+export default [
+  ...calendarAttributes,
+  ...popoverAttributes,
+  ...modalAttributes,
+].filter((attribute, index, attributes) => attributes.indexOf(attribute) === index);
```
matteo-hertel: I'm just curious here, what this filter is supposed to catch? `attributes.indexOf(attribute) === index` will always evaluate to `true` right?

matthewdavidson: This is a `uniq` filter. Any items that match but are not at the same index are tossed.

matthewdavidson: In this case, modal and popover have similar (if not identical) theme attributes .. this accounts for any instance in which their theme attributes might change.

packages/bpk-component-datepicker/stories.js
```diff
@@ -57,15 +57,21 @@ class CalendarContainer extends Component {
...
+        <div id="application-element">
+          <BpkDatepicker
+            {...this.props}
+            date={this.state.date}
+            onDateSelect={(date) => {
+              this.setState({ date });
+              action('Selected date')(date);
+            }}
+            onMonthChange={action('Changed month')}
+            getApplicationElement={() => document.getElementById('application-element')}
```
k0nserv: nit: Shouldn't we prefer the ref method instead?

packages/bpk-component-datepicker/src/themeAttributes.js
matteo-hertel: ah! nice!

packages/bpk-component-datepicker/stories.js
matthewdavidson: Could do, but a lot of usage in Skyscanner (particularly with OC components) depends on containers defined outwith the components react tree (e.g. the header culture selector needs to get application element defined in microsite, not in header).

k0nserv: Good point, but shouldn't we use the React-way when we can?

matthewdavidson: Myself and hugo caught up offline and discussed that the most common usage in Skyscanner is preferred.

PR #351, opened 2017-12-05
[Decision] Add scss filename decision

k0nserv:


shaundon:
Could you add this to the table of contents in readme.md too please?

shaundon:
Also I like how you've labelled the PR with `[Decision]`, makes it easy to parse in Outlook etc.

PR #357, opened 2017-12-05
[BPK-1136] Add theme support to docs sidenav

shaundon:
localhost-8080-components-web-chips

### DIFF
--- a/packages/bpk-docs/src/layouts/SideNavLayout/nav-list.scss
+++ b/packages/bpk-docs/src/layouts/SideNavLayout/nav-list.scss
@@ -51,8 +51,9 @@
   }
 
   &__link {
-    color: $bpk-link-color;
     text-decoration: $bpk-link-text-decoration;
+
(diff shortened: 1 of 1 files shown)


packages/bpk-docs/src/layouts/SideNavLayout/nav-list.scss
```diff
@@ -65,23 +66,26 @@
...
 
+      @include bpk-themeable-property(background-color, --bpk-link-active-color, $bpk-color-blue-500);
+
       @include bpk-hover {
-        background-color: $bpk-color-blue-500;
         color: $bpk-color-white;
         font-weight: bold;
         text-decoration: none;
+
+        @include bpk-themeable-property(background-color, --bpk-link-hover-color, $bpk-color-blue-500);
```
k0nserv: This shouldn't change color so it should be the active color too.

shaundon: Good point - now updated.

PR #361, opened 2017-12-06
[BPK-1120] Add react-native-bpk-styles package and apply it

shaundon:


native/packages/react-native-bpk-styles/src/shadows-test.js
```diff
@@ -0,0 +1,26 @@
...
+  */
+
+import shadows from './shadows';
+
+describe('shadows', () => {
+  it('should export the correct shadow styles', () => {
+    expect(shadows.base).toBeDefined();
+    expect(shadows.large).toBeDefined();
+  });
+});
```
shaundon: Could argue that this test is pointless - I was going to make it check the actual values, but then that's already covered by snapshot tests. I've left it in because it does no harm to have it, but I can remove it.

shaundon:
I haven't updated changelog as this isn't really user facing.

native/packages/react-native-bpk-styles/src/shadows-test.js
matthewdavidson: Where is it covered by snapshot tests? You mean in the components that use those styles?

matthewdavidson:
I'm just wondering if we default to functions for shared styles. The thinking here is to emulate scss mixins so that some styles can accept and argument to say ... customise the color applied etc. Im thinking borders etc. 

```
StyleSheet.create({
  container: {
    ...shadows.small(),
    ...borders.small(colorBlue500),
  }
});
```

Thoughts?

shaundon:
@matthewdavidson 

> Where is it covered by snapshot tests? You mean in the components that use those styles?

Yeah exactly - I figured I could import the tokens into the tests and check that `shadowSmColor === tokens.shadowSmColor` but it felt a bit redundant and longwinded for something that snapshots are going to cover every time these styles are used in a component.

shaundon:
I like the idea of using functions - it covers us in future if we do wish to make them take arguments, and it makes it a bit more explicit in the code that it's a mixin-like thing. I'll update it now.

native/packages/react-native-bpk-styles/src/shadows-test.js
```diff
@@ -0,0 +1,26 @@
...
+  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+  */
+
+import shadows from './shadows';
+
+describe('shadows', () => {
+  it('should export the correct functions', () => {
+    expect(shadows.base()).toBeDefined();
```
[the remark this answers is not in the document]
shaundon: Done

native/packages/react-native-bpk-styles/src/shadows.js
```diff
@@ -0,0 +1,65 @@
...
+    width: tokens[type].offsetWidth,
+  },
+  shadowOpacity: tokens[type].opacity,
+  shadowRadius: tokens[type].radius / PixelRatio.get(),
+});
+
+const shadows = {
+  base: () => getShadowStyles('base'),
+  large: () => getShadowStyles('large'),
+};
```
matthewdavidson: Why not just:
```
export const base = () => ({
  shadowColor: shadowSmColor,
  shadowOffset: {
    height: shadowSmOffsetHeight / PixelRatio.get(),
    width: shadowSmOffsetWidth,
  },
  shadowOpacity: shadowSmOpacity,
  shadowRadius: shadowSmRadius / PixelRatio.get(),
});

export const large = () => ({
  shadowColor: shadowXlColor,
  shadowOffset: {
    height: shadowXlOffsetHeight / PixelRatio.get(),
    width: shadowXlOffsetWidth,
  },
  shadowOpacity: shadowXlOpacity,
  shadowRadius: shadowXlRadius / PixelRatio.get(),
});
```

shaundon: Just over-optimisation I guess 😅 . Updated now anyway, I'll merge it.

PR #364, opened 2017-12-06
[BPK-1140] pagination theming

matteo-hertel:
This makes the pagination themable, two main changes:
`paginationBackgroundColorSelected` get applied to the background and border of a selected button and the cursor `pointer` has been removed

in the storybook the secondary button had a different behaviour than the docs site, now the default border is `Gray100`

both changes discussed with @jamesf3rguson 

Screenshots time!

screen shot 2017-12-06 at 3 04 16 pm
screen shot 2017-12-06 at 3 04 23 pm
screen shot 2017-12-06 at 3 04 44 pm

### DIFF
--- a/packages/bpk-component-pagination/src/BpkPaginationPage.js
+++ b/packages/bpk-component-pagination/src/BpkPaginationPage.js
@@ -26,19 +26,20 @@ import STYLES from './bpk-pagination-page.scss';
 const getClassName = cssModules(STYLES);
 
 const BpkPaginationPage = (props) => {
+  const classNames = [getClassName('bpk-pagination-page')];
   const {
     label, onSelect, isSelected,
   } = props;
 
+  if (isSelected) { classNames.push(getClassName('bpk-pagination-page--selected')); }
+
   return (
     <BpkButton
-      iconOnly
       secondary
-      selected={isSelected}
       onClick={onSelect}
-      className={getClassName('bpk-pagination-page')}
+      className={classNames.join(' ')}
     >
-      <span>{ label }</span>
+      <span>{label}</span>
     </BpkButton>
   );
 };
--- a/packages/bpk-component-pagination/src/bpk-pagination-page.scss
+++ b/packages/bpk-component-pagination/src/bpk-pagination-page.scss
@@ -23,3 +23,38 @@
   height: $bpk-spacing-xl;
   padding: 0;
 }
+
+.bpk-pagination-page--selected {
+  background-image: $bpk-button-secondary-background-image;
+  color: $bpk-button-selected-color;
+  box-shadow: $bpk-button-selected-box-shadow;
+  cursor: auto;
+
+  @include bpk-themeable-property(background-color, --bpk-pagination-selected-background-color, $bpk-color-blue-700);
+  @include bpk-border-lg($bpk-color-blue-700);
+  @include bpk-border-lg(var(--bpk-pagination-selected-background-color, $bpk-color-blue-700));
+
+  @include bpk-hover {
+    background-image: $bpk-button-selected-hover-background-image;
+    color: $bpk-button-selected-hover-color;
+    box-shadow: $bpk-button-selected-hover-box-shadow;
+
+    @include bpk-themeable-property(background-color, --bpk-pagination-selected-background-color, $bpk-color-blue-700);
+    @include bpk-border-lg($bpk-color-blue-700);
+    @include bpk-border-lg(var(--bpk-pagination-selected-background-color, $bpk-color-blue-700));
+  }
+
+  &:active {
+    background-image: $bpk-button-selected-active-background-image;
+    color: $bpk-button-selected-active-color;
+    box-shadow: $bpk-button-selected-active-box-shadow;
+
+    @include bpk-themeable-property(background-color, --bpk-pagination-selected-background-color, $bpk-color-blue-700);
+    @include bpk-border-lg($bpk-color-blue-700);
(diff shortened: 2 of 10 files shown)


packages/bpk-component-pagination/src/themeAttributes.js
```diff
@@ -0,0 +1,29 @@
...
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+  */
+import {
+  secondaryThemeAttributes,
+} from 'bpk-component-button';
+
+export default [
+  ...secondaryThemeAttributes,
+  'paginationBackgroundColorSelected',
```
matthewdavidson: These names do not follow our naming convention, see https://github.com/Skyscanner/backpack/blob/master/decisions/theme-props-naming.md

matteo-hertel: I know, I'm a rebel! 😄 

I've used the name on the Jira ticket, I won't do that anymore

packages/bpk-component-pagination/src/bpk-pagination-page.scss
```diff
@@ -23,3 +23,35 @@
   height: $bpk-spacing-xl;
   padding: 0;
 }
+
+.bpk-pagination-page--selected {
+  cursor: auto;
```
matthewdavidson: this code needs room to breath 😅  - stylelint will complain soon I suspect.

packages/bpk-component-pagination/src/themeAttributes.js
matteo-hertel: just to confirm, is
```
  'paginationSelectedBackgroundColor',
  'paginationNudgerActiveColor',
  'paginationNudgerHoverColor',
  'paginationNudgerNormalColor',
```
correct? 
cc @shaundon

packages/bpk-component-pagination/src/bpk-pagination-page.scss
matteo-hertel: I've been betrayed by VSCode, the inline linting did not complain so I thought I could get away with it </3

packages/bpk-component-pagination/src/BpkPaginationPage.js
```diff
@@ -25,20 +25,23 @@ import STYLES from './bpk-pagination-page.scss';
 
 const getClassName = cssModules(STYLES);
 
+const computeClassName = isSelected => [
```
k0nserv: This is a very different approach from what we usually do. Why?

packages/bpk-component-theme-toggle/src/theming.js
```diff
@@ -59,7 +60,7 @@ const bpkCustomTheme = {
   buttonSecondaryTextColor: theme.primaryColor500,
   buttonSecondaryHoverTextColor: theme.primaryColor600,
   buttonSecondaryActiveTextColor: theme.primaryColor700,
-  buttonSecondaryBorderColor: theme.primaryColor500,
+  buttonSecondaryBorderColor: colorGray100,
```
k0nserv: Hmm I wonder if this was intentional. @jamesf3rguson?

jamesf3rguson: Hmm, I guess its because we don't need to theme it... i.e. gray should always be gray. I guess its a bit redundant though?

packages/bpk-component-pagination/src/themeAttributes.js
shaundon: Sounds good to me!

packages/bpk-component-pagination/src/BpkPaginationPage.js
matteo-hertel: No particular reason, I like having a function that returns a ready to use string e.g. I don't like 
https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-accordion/src/BpkAccordion.js#L34
with the join call in the JSX block

Do you want me to change it?

packages/bpk-component-theme-toggle/src/theming.js
matteo-hertel: I've mentioned it in the PR description, I get it might look redundant but it's explicit, all the theme props that the button is using are there, ready to use a quick reference without relying in fallback, however if it's a deal breaker I'll remove it

packages/bpk-component-pagination/src/BpkPaginationPage.js
shaundon: I've always preferred it the other way, with the join call in the JSX block. 

I like that you start with an array that is a useful data structure that *could* be used elsewhere at some point, versus the string which is only useful in the JSX block. And then it's the render function's job to convert that to the format it needs.

matthewdavidson: Consistency matters above all else 👍

matteo-hertel: https://media.giphy.com/media/sJPJTSbXSevxS/giphy.gif

packages/bpk-component-pagination/src/themeAttributes.js
shaundon: Update: Sorry, what I *meant* to say was that the 'normal' bit is redundant and can be omitted. 😅

shaundon:
I think this needs to be rebased due to https://github.com/Skyscanner/backpack/pull/356 - I'm running this PR locally and the disabled nudgers aren't rendering.

matteo-hertel:
I've merged with master, @shaundon ready for your 👍

PR #367, opened 2017-12-07
[BPK-1153] Downgrade @storybook/react-native to fix issue

shaundon:
**The issue**
When using storybook actions (or indeed any addons), we're getting an issue where triggering the action causes an error to appear.

The error message itself links to [this page](https://storybook.js.org/basics/faq/#why-is-there-no-addons-channel), which in turn links to [this issue](https://github.com/storybooks/storybook/issues/1192).

**The solution**
I tried all sorts of different changes to the code but I couldn't get it to work. In the end, through trial and error I figured out that downgrading `@storybook/react-native` by one patch version (`3.2.17` :arrow_right: `3.2.16`) sorts it out.

It's not an ideal long term solution, but it does fix things for now.

`3.3.0` of Storybook is currently in alpha, I intend to next find out if that fixes the issue, and also find out what exactly changed from `3.2.16` to `3.2.17` to break things.

### DIFF
--- a/native/package-lock.json
+++ b/native/package-lock.json
@@ -10,6 +10,62 @@
       "integrity": "sha1-6pn2EhtKjwZbTHH4VZXbJxRJiAc=",
       "dev": true
     },
+    "@storybook/addon-actions": {
+      "version": "3.2.17",
+      "resolved": "https://registry.npmjs.org/@storybook/addon-actions/-/addon-actions-3.2.17.tgz",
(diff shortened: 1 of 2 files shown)


shaundon:
Just upgraded `@storybook/react-native` to `3.3.0-alpha.4` and it works there, although it does regress Android to the 'attempting to change value of a readonly property' bug we saw on `3.2.16`.

So, assuming that the Android fix that was in `3.2.17` makes it to the final version of `3.3.0`, upgrading upon release should hopefully fix our issue.

PR #368, opened 2017-12-08
[BPK-1108] Initial implementation of floating text label inputs

matthewdavidson:
Note: This PR isn't merging into master, instead it's going into `bpk-1108-floating-label-text-inputs`. This is still a WIP, I just want to get an early review. Follow up PR's will focus on refactoring and tests.


matthewdavidson:
@shaundon oops - you are correct. I'll add that in a follow up PR.

PR #370, opened 2017-12-09
[BPK-434] Add datatable component from react-virtualized

janeklb:
_Originally opened up by @jslvtr here: #182_

Added datatable with a subset of the suggested features:

- Sortable columns.
- RTL and accessibility support.
- Support for many rows via react-virtualized.
- Row click callback

Also includes a readme and example in the storybook.

BpkDataTable

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -369,6 +369,8 @@
 - react-native-bpk-component-banner-alert:
   - New component `BpkBannerAlert`, see https://backpack.github.io/components/native/banner-alerts/
 
+- New component `bpk-component-datatable`, from `react-virtualized`
+
 **Fixed:**
 - bpk-tokens: 26.3.0 => 26.4.0
   - Fixed package meta data to point to correct entry file
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,9 @@
 - react-native-bpk-component-banner-alert:
   - New animations available when `show` is toggled. See https://backpack.github.io/components/native/banner-alerts
 
+- bpk-component-datatable:
+  - New data table component, see https://backpack.github.io/components/web/datatables
+
 ## 2017-12-06 - More theming support
 
 **Added:**
@@ -368,8 +371,6 @@
 
 - react-native-bpk-component-banner-alert:
   - New component `BpkBannerAlert`, see https://backpack.github.io/components/native/banner-alerts/
-
-- New component `bpk-component-datatable`, from `react-virtualized`
 
 **Fixed:**
 - bpk-tokens: 26.3.0 => 26.4.0
--- a/packages/bpk-component-datatable/index.js
+++ b/packages/bpk-component-datatable/index.js
@@ -15,9 +15,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-import BpkDataTable from './src/BpkDataTable';
 import { Column } from 'react-virtualized';
+import BpkDataTable from './src/BpkDataTable';
 
 const BpkColumn = Column;
 
--- a/packages/bpk-component-datatable/index.js
+++ b/packages/bpk-component-datatable/index.js
@@ -15,9 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { Column } from 'react-virtualized';
 import BpkDataTable from './src/BpkDataTable';
+import BpkDataTableColumn from './src/BpkDataTableColumn';
 
-const BpkColumn = Column;
-
-export { BpkDataTable, BpkColumn };
+export { BpkDataTable, BpkDataTableColumn };
--- a/packages/bpk-component-datatable/index.js
+++ b/packages/bpk-component-datatable/index.js
@@ -1,0 +1,24 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import BpkDataTable from './src/BpkDataTable';
+import { Column } from 'react-virtualized';
+
+const BpkColumn = Column;
+
+export { BpkDataTable, BpkColumn };
--- a/packages/bpk-component-datatable/package.json
+++ b/packages/bpk-component-datatable/package.json
@@ -13,8 +13,8 @@
     "react": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
-    "bpk-mixins": "^17.0.5",
-    "bpk-react-utils": "^2.3.6",
+    "bpk-mixins": "^17.2.0",
+    "bpk-react-utils": "^2.4.0",
     "prop-types": "^15.5.8",
     "react-table": "^6.6.0",
     "react-virtualized": "^9.10.1"
--- a/packages/bpk-component-datatable/package.json
+++ b/packages/bpk-component-datatable/package.json
@@ -15,6 +15,7 @@
   "dependencies": {
     "bpk-mixins": "^17.2.0",
     "bpk-react-utils": "^2.4.0",
+    "lodash": "^4.17.4",
     "prop-types": "^15.5.8",
     "react-virtualized": "^9.10.1"
   }
--- a/packages/bpk-component-datatable/package.json
+++ b/packages/bpk-component-datatable/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "bpk-component-datatable",
+  "version": "0.0.0",
+  "description": "Backpack datatable component.",
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^15.0.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "bpk-mixins": "^17.0.1",
+    "bpk-react-utils": "^2.3.4",
+    "prop-types": "^15.5.8",
+    "react-table": "^6.6.0",
+    "react-virtualized": "^9.10.1"
+  }
+}
--- a/packages/bpk-component-datatable/package.json
+++ b/packages/bpk-component-datatable/package.json
@@ -13,8 +13,8 @@
     "react": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
-    "bpk-mixins": "^17.0.1",
-    "bpk-react-utils": "^2.3.4",
+    "bpk-mixins": "^17.0.5",
+    "bpk-react-utils": "^2.3.6",
     "prop-types": "^15.5.8",
     "react-table": "^6.6.0",
     "react-virtualized": "^9.10.1"
--- a/packages/bpk-component-datatable/package.json
+++ b/packages/bpk-component-datatable/package.json
@@ -16,7 +16,6 @@
     "bpk-mixins": "^17.2.0",
     "bpk-react-utils": "^2.4.0",
     "prop-types": "^15.5.8",
-    "react-table": "^6.6.0",
     "react-virtualized": "^9.10.1"
   }
 }
--- a/packages/bpk-component-datatable/readme.md
+++ b/packages/bpk-component-datatable/readme.md
@@ -12,7 +12,7 @@ npm install bpk-component-datatable --save-dev
 
 ```js
 import React from 'react';
-import BpkDataTable from 'bpk-component-datatable';
+import { BpkDataTable, BpkDataTableColumn } from 'bpk-component-datatable';
 
 const rows = [
   { name: 'Jose', description: 'Software Engineer' },
@@ -23,12 +23,12 @@ const onRowClick = row => alert(JSON.stringify(row));
 
 export default () => (
   <BpkDataTable rows={rows} height={200} onRowClick={onRowClick}>
-    <BpkColumn
+    <BpkDataTableColumn
       label={'Name'}
       dataKey={'name'}
       width={100}
     />
-    <BpkColumn
+    <BpkDataTableColumn
       label={'Description'}
       dataKey={'description'}
       width={100}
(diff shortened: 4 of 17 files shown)


changelog.md
```diff
@@ -369,6 +369,8 @@
 - react-native-bpk-component-banner-alert:
   - New component `BpkBannerAlert`, see https://backpack.github.io/components/native/banner-alerts/
 
+- New component `bpk-component-datatable`, from `react-virtualized`
```
matthewdavidson: format should be:
```
- bpk-component-datatable:
  - New data table component, see https://backpack.github.io/components/web/datatables
```

matthewdavidson: We'll follow up with the docs to be displayed at that url in a separate url.

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+
+  componentWillReceiveProps({ rows }) {
+    if (rows !== this.state.rows) {
+      const { sortBy, sortDirection } = this.state;
+      const sortedList = sortList({ sortBy, sortDirection, list: rows });
+      this.setState({ rows, sortedList, rowSelected: undefined });
+    }
+  }
+
+  rowClicked({ index }) {
```
matthewdavidson: can we rename this to `onRowClicked`?

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+      this.setState({ rowSelected: undefined });
+    } else {
+      this.setState({ rowSelected: index });
+    }
+    if (this.props.onRowClick !== undefined) {
+      this.props.onRowClick(this.state.rows[index]);
+    }
+  }
+
+  rowStyle({ index }) {
```
matthewdavidson: Can we rename this to `rowClassName`?

packages/bpk-component-datatable/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^15.0.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "bpk-mixins": "^17.2.0",
+    "bpk-react-utils": "^2.4.0",
+    "prop-types": "^15.5.8",
+    "react-table": "^6.6.0",
```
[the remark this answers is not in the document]
janeklb: indeed it is 👍

packages/bpk-component-datatable/index.js
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Column } from 'react-virtualized';
+import BpkDataTable from './src/BpkDataTable';
+
+const BpkColumn = Column;
```
matthewdavidson: Oh so we aren't using the column component defined in `BpkColumn.js`, we are using the one from `react-virtualized` directly? is this intentional?

packages/bpk-component-datatable/src/BpkColumn.js
```diff
@@ -0,0 +1,29 @@
...
+
+import React from 'react';
+import { Column } from 'react-virtualized';
+import { cssModules } from 'bpk-react-utils';
+
+import STYLES from './bpk-column.scss';
+
+const getClassName = cssModules(STYLES);
+
+const BpkColumn = props => <Column className={getClassName('bpk-column')} {...props} />;
```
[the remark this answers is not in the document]
matthewdavidson: This component should be renamed to `BpkDataTableColumn`. Also we need snapshot test(s).

packages/bpk-component-datatable/src/bpk-data-table.scss
```diff
@@ -0,0 +1,69 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins';
+
+.bpk-data-table {
+  width: 100%;
+
+  &__headerColumn {
```
matthewdavidson: For consistency, this should be `&__header-column`.

packages/bpk-component-datatable/src/bpk-data-table.scss
```diff
@@ -0,0 +1,69 @@
...
+    border: $bpk-border-size-sm solid $bpk-color-gray-50;
+    border-bottom: none;
+    color: $bpk-color-gray-700;
+    font-weight: normal;
+
+    &:last-child {
+      border-bottom: $bpk-border-size-sm solid $bpk-color-gray-50;
+    }
+
+    &:not(.bpk-data-table__headerRow):hover {
```
matthewdavidson: This should be `.bpk-data-table__header-row.

Just noticed... there are no styles associated with this class - do you mean `.bpk-data-table__row` (defined on [line 43](https://github.com/Skyscanner/backpack/pull/370/files#diff-55ba15298ffc9780198b9a01011f4642R43))

packages/bpk-component-datatable/index.js
janeklb: Yeah i noticed this last night - just getting my head around it. The `BpkDataTable` code doesn't render them as JSX elements, but calls them as functions (https://github.com/Skyscanner/backpack/pull/370/files#diff-2bda430a0beb77956f2b5269a49be0e0R129) since `react-virtualized`'s `Table` *requires* that it's children are of type `Column`. 

So, we export `react-virtualized` column, but we actually use our own `BpkColumn` in the tests. Going to re-work this stuff to make it a bit more consistent.

packages/bpk-component-datatable/stories.js
```diff
@@ -0,0 +1,58 @@
...
+        flexGrow={1}
+      />
+      <BpkColumn
+        label="Bla"
+        dataKey="bla"
+        width={100}
+        disableSort
+      />
+    </BpkDataTable>
+  ));
```
matthewdavidson: We need some more example stories which demonstrates the different functionality we are supporting with this component. For example...there is a `width` prop for the table and the columns as well as `flexGrow` - anything else worth providing an example for? This helps future maintainers to know what to test when they are making changes.

packages/bpk-component-datatable/index.js
janeklb: @jslvtr can you shine some light as to why it's set up this way? Do i have the right idea?

matthewdavidson:
So I've had a play locally @janeklb, i have a few comments on the behaviour / design aspect:

data table column header behaviour

- Notice how the column header text jumps to accommodate the position of the icon - ideally we wouldn't want this jarring behaviour
- The outline of clickable elements are clipped by the table column and row containers (blue border behaviour in chrome - it's different in other browsers). react virtualised is setting overflow hidden explicitly so you might have to use something like `outline-offset: -(3 * $bpk-one-pixel-rem;)` and adjust padding as necessary to achieve a better result.

I suggest checking in with Kartikey for a more thorough design review 👍

packages/bpk-component-datatable/readme.md
```diff
@@ -0,0 +1,61 @@
...
+
+const rows = [
+  { name: 'Jose', description: 'Software Engineer' },
+  { name: 'Rolf', description: 'Manager' }
+]
+
+const onRowClick = row => alert(JSON.stringify(row));
+
+export default () => (
+  <BpkDataTable rows={rows} height={200} dir={'rtl'} onRowClick={onRowClick}>
```
k0nserv: This requires intervention by the user to handle RTL, most components just do this automatically so this might trip people up

packages/bpk-component-datatable/index.js
k0nserv: Doesn't this conflict with what is below? We aren't using the version we define ourselves

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Table, AutoSizer } from 'react-virtualized';
+import { cssModules } from 'bpk-react-utils';
+import _sortBy from 'lodash/sortBy';
```
k0nserv: Lodash isn't listed as a dependency of this package

janeklb:
@matthewdavidson I noticed that from the previous MR  #182 -- it seemed like it was hardcoded in `react-virtualized` but can have another go at it

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+    sorted.reverse();
+  }
+  return sorted;
+};
+
+class BpkDataTable extends Component {
+  constructor(props) {
+    super(props);
+
+    const sortBy = 'name';
```
k0nserv: Is `name` here referring to a column? This seems awefuly specific to our examples

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+      columns.reverse();
+    }
+
+    this.state = {
+      sortedList,
+      sortBy,
+      sortDirection,
+      columns,
+      rows: props.rows,
+      rowSelected: undefined,
```
k0nserv: Can we call this `selectedRow` instead?

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+      columns,
+    } = this.state;
+    const { height, dir } = this.props;
+
+    return (
+      <Table
+        className={getClassName('bpk-data-table')}
+        width={width}
+        height={height}
+        headerHeight={60}
```
k0nserv: Do we foresee consumers wanting to change the header height?

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+        rowHeight={60}
```
k0nserv: Same

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+    );
+  }
+
+  render() {
+    if (this.props.width !== null) {
+      return this.renderTable(this.props.width);
+    }
+
+    return (
+      <AutoSizer disableHeight>{({ width }) => this.renderTable(width)}</AutoSizer>
```
k0nserv: We can bind `this.renderTable` to `this` and avoid creating this redundant anonymous function

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -0,0 +1,162 @@
...
+    );
+  }
+}
+
+BpkDataTable.propTypes = {
+  rows: PropTypes.arrayOf(Object).isRequired,
+  children: PropTypes.arrayOf(PropTypes.element).isRequired,
```
k0nserv: Can we make this stricter with `PropTypes.instaceOf`?

packages/bpk-component-datatable/src/bpk-data-table.scss
janeklb: yeah - no styles associated, but the class is programatically set on header rows to give it access to this rule

packages/bpk-component-datatable/index.js
janeklb: Yeah, we need some way of declaring Columns, but we need to use something that'll automatically add `bpk-column` (or whatever the class name is). What about something like this:
```
<BpkDataTable ...props>
  {
    [
      { dataKey: 'name', label: 'Name', width: 100 },
      { dataKey: 'description', label: 'Description', width: 200, flexGrow: 1 },
      { dataKey: 'bla', label: 'Bla', width: 100, disableSort: true },
    ]
  }
</BpkDataTable>
```

with 
```
const columnShape = PropTypes.shape({
  dataKey: PropTypes.string.isRequired,
  label: PropTypes.string.isRequired,
  width: PropTypes.number.isRequired,
  flexGrow: PropTypes.number,
  disableSort: PropTypes.boolean,
});

BpkDataTable.propTypes = {
  ...,
  children: PropTypes.arrayOf(columnShape).isRequired,
  ...,
};
```
?

janeklb: Or would you guys prefer these intermediate data components that define how `react-virtualized` `Column`s should be created (kinda like what we have now, but a bit better organised)?

packages/bpk-component-datatable/src/BpkDataTable.js
janeklb: you're missing the destructured `{ width }` (function signatures don't match) :/

packages/bpk-component-datatable/src/BpkDataTable.js
janeklb: Good catch. I'll update that.

packages/bpk-component-datatable/stories.js
janeklb: The `width` prop one makes sense, will add that in. However, the `flexGrow` one less so, since it's a property on `react-virtualized`'s [`Column`](https://github.com/bvaughn/react-virtualized/blob/master/docs/Column.md) - ok to skip it?

packages/bpk-component-datatable/readme.md
```diff
@@ -0,0 +1,58 @@
...
+| height       | number                  | yes      | -                    |
+| onRowClick   | function                | no       | -                    |
+| width        | number                  | no       | full width of parent |
+| headerHeight | number                  | no       | 60                   |
+| rowHeight    | number                  | no       | 60                   |
+| className    | number                  | no       | -                    |
+
+### BpkDataTableColumn
+
+Supports all properties defined in [`Column`](https://github.com/bvaughn/react-virtualized/blob/master/docs/Column.md) (from `react-virtualized`)
```
janeklb: Rather than explicitly listing all props, is this ok?

packages/bpk-component-datatable/src/BpkDataTableColumn.js
```diff
@@ -0,0 +1,41 @@
...
+
+import React from 'react';
+import { Column } from 'react-virtualized';
+import { cssModules } from 'bpk-react-utils';
+
+import STYLES from './bpk-column.scss';
+
+const getClassName = cssModules(STYLES);
+
+const BpkDataTableColumn = () => null;
```
matthewdavidson: This is interesting - what's the workaround here? I'm assuming it's to get round the prop type issues we discussed in a previous comment, but I think it needs a comment for other contributors so that they understand whats going on.

janeklb:
Alright - tested this locally with our project and it's works great, and so I'm happy to ship this. Is there anything else you guys want to address?

PR #372, opened 2017-12-11
[BPK-1108] floating label text inputs

shaundon:
screen shot 2017-12-12 at 13 20 55

screen shot 2017-12-12 at 14 19 05


k0nserv:
@shaundon for the clear button. Look at `clearButtonMode`

shaundon:
@k0nserv thanks, didn't know about that prop!

Just spoke to @jamesf3rguson and showed him the clear button in action, the button appears when the component is focused and `value.length > 0`.

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -16,185 +16,143 @@
...
+
+    return (
+      <View style={[styles.container, userStyle]}>
+        <Animated.Text style={animatedLabelStyle}>{label}</Animated.Text>
+        <Animated.View style={animatedInputStyle}>
+          <TextInput
+            editable={editable}
+            value={value || ''}
+            style={styles.input}
+            onFocus={() => this.setState(() => ({ isFocused: true }), onFocus)}
```
matteo-hertel: can we have this as it's own function somewhere? it'll look much better IMHO

native/packages/react-native-bpk-component-text-input/src/BpkTextInputIcons-test.android.js
```diff
@@ -0,0 +1,39 @@
...
+  const reactNative = require.requireActual('react-native');
+  jest
+    .spyOn(reactNative.Platform, 'select')
+    .mockImplementation(obj => obj.android || obj.default);
+  reactNative.Platform.OS = 'android';
+
+  return reactNative;
+});
+
+jest.mock('./../node_modules/react-native-bpk-component-text/node_modules/bpk-tokens/tokens/base.react.native',
```
matteo-hertel: Do we have to mock this? so if we have a component that brings in other components we have to mock the tokens?

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
```diff
@@ -0,0 +1,126 @@
...
+
+  fontFamily,
+
+  textSmLineHeight,
+  textSmFontSize,
+  textSmFontWeight,
+  textBaseLineHeight,
+  textBaseFontSize,
+  textBaseFontWeight,
+} from 'bpk-tokens/tokens/base.react.native';
```
matteo-hertel: this is beautiful, but alphabetically ordered would be absolutely amazing! 😄

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
```diff
@@ -0,0 +1,126 @@
...
+  container: {
+    paddingTop: textSmLineHeight,
+  },
+  input: {
+    flex: 1,
+    paddingTop: spacingSm,
+    paddingRight: 0,
+    paddingBottom: spacingSm,
+    paddingLeft: 0,
+    minHeight: textBaseLineHeight + (spacingSm * 2),
```
matteo-hertel: isn't `spacingSm * 2` `spacingMd`? we could use that

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
shaundon: Done

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
shaundon: Good plan!

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
shaundon: Yep I believe it is, good catch

native/packages/react-native-bpk-component-text-input/src/BpkTextInputIcons-test.android.js
matteo-hertel: Summon @matthewdavidson for this one 😄

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
shaundon: Changed

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
shaundon: Done

native/packages/react-native-bpk-component-text-input/src/BpkTextInputIcons-test.android.js
shaundon: I removed the line and the tests still passed, so YOLO

shaundon:
So, I removed the clear button logic we had in favour of the native iOS one provided by `clearButtonMode`. This is awesome (thanks Hugo!) because it means we can delete a load of code and let the OS handle it for us.

However there's an issue - when you quickly change focus onto a field with a validation icon on the right, you can sometimes see a bit of jank. It doesn't happen every time, but when it does happen it looks pretty bad. There's a gif below.

I'd like to find a way to fix this as I prefer using `clearButtonMode` - any thoughts @k0nserv @matthewdavidson @matteo-hertel @georgegillams?

kapture 2017-12-12 at 17 21 05

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
matteo-hertel: image

k0nserv:
How about animating the width of the tick rather than just outright removing it? From whatever width it is by default to 0.

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -16,185 +16,140 @@
...
   value: PropTypes.string.isRequired,
+  clearButtonMode: TextInput.propTypes.clearButtonMode,
   inputRef: PropTypes.func,
+  validationMessage: PropTypes.string,
+  valid: PropTypes.oneOf(true, false, null),
+  editable: PropTypes.bool,
+  style: ViewPropTypes.style,
+  onFocus: PropTypes.func,
+  onBlur: PropTypes.func,
+  onChangeText: PropTypes.func,
```
k0nserv: Why is this documented? We aren't handling this directly as it's part of the spread

native/packages/react-native-bpk-component-text-input/src/BpkTextInputIcons.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+import React from 'react';
+import { StyleSheet } from 'react-native';
+import BpkIcon from 'react-native-bpk-component-icon';
+import { colorRed500, colorGreen500, spacingSm } from 'bpk-tokens/tokens/base.react.native';
+
+const styles = StyleSheet.create({
+  icon: {
+    marginRight: spacingSm * 1.5,
```
k0nserv: What's the status of RTL here?

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
```diff
@@ -0,0 +1,128 @@
...
+const INPUT_RANGE = [0, 1];
+
+const styles = StyleSheet.create({
+  container: {
+    paddingTop: textSmLineHeight,
+  },
+  input: {
+    flex: 1,
+    paddingTop: spacingSm,
+    paddingRight: 0,
```
k0nserv: 0 is the default. Specifying these is redundant right?

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
```diff
@@ -0,0 +1,128 @@
...
+    paddingBottom: spacingSm,
+    paddingLeft: 0,
+    minHeight: textBaseLineHeight + spacingMd,
```
k0nserv: Why plus `spacingMd`? Can we make it clearer by renaming the `spacingMd` value here?

shaundon:
`position: absolute` did the trick. Thanks all :)

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
shaundon: Actually not sure - have removed.

k0nserv:
So the absolute version doesn't work if you use `clearButtonMode="always"`

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
shaundon: @matthewdavidson what was the reason for this?

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
shaundon: Also just noticed that `paddingTop` and `paddingBottom` can become `paddingVertical`.

matthewdavidson: Erm I actually don;t think these are redundant - android inputs come with a default value for these props.

shaundon: Just tried it and Matt's right. I've changed it to `paddingHorizontal: 0` to save a line of code at least

shaundon:
@k0nserv just tried it with `clearButtonMode="always"` and it works mostly fine, it looks like this:

screen shot 2017-12-13 at 09 57 24

When you click one of the ones with a validation icon, the clear button appears as normal. So I think they are covering it, but is that a big issue? Or if it is an issue, something we can patch later once this is out?

native/packages/react-native-bpk-component-text-input/src/BpkTextInputStyles.js
matthewdavidson: `spacingMd` is equivalent to `(spacingSm * 2)`. As for the reason for setting min height ... im sorry i cant remember 😇  try removing and observing the impact android & ios (check the "box model" in inspector too!)

matthewdavidson:
@shaundon can we rename `BpkTextInputStyles.js` to just `styles.js` - it's not exporting a component so it doesn;t need to follow the naming convention.

k0nserv:
@shaundon I should clarify. With `clearButtonMode="always"` I'd expect to see both the clear and the tick, but I guess this is one for @jamesf3rguson

changelog.md
```diff
@@ -2,6 +2,15 @@
 
 ## UNRELEASED
 
+**Breaking:**
+- react-native-bpk-component-text-input:
+  - Now supports Android.
+  - A new design, see https://backpack.github.io/components/native/text-input.
```
matthewdavidson: Can we change this to "A new floating text label design" with no link? What happens when we change the style again in the future? Hypothetical I know, but the content at that url is dynamic so it shouldnt be in the changelog.

changelog.md
```diff
@@ -2,6 +2,15 @@
...
+  - The `small` prop has been removed. Existing components using it will now get the standard size.
+  - The `disabled` prop has been replaced with `editable`, to follow React Native's built in [TextInput](https://facebook.github.io/react-native/docs/textinput.html) component.
+  - A new optional prop, `validationMessage`, allows a message to be shown when `valid={false}`.
+  - A new required `label` prop is required.
```
matthewdavidson: How about: "A new required `label` prop."

matthewdavidson:
For a follow up PR, but we'll need to update the screenshots and modify the docs accordingly.

changelog.md
shaundon: D'oh!

shaundon:
Spoke to @jamesf3rguson about behaviour when `clearButtonMode="always"` and we agreed that it behaves relatively sensibly for now, so we can patch it in future if needed.

shaundon:
https://github.com/Skyscanner/backpack/pull/372#issuecomment-351344413

Thanks for clarifying it's a native issue, I was thinking how the hell are we gonna fix that!

native/packages/react-native-bpk-component-text-input/src/BpkTextInputIcons.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import React from 'react';
+import { StyleSheet, I18nManager } from 'react-native';
+import BpkIcon from 'react-native-bpk-component-icon';
+import { colorRed500, colorGreen500, spacingSm } from 'bpk-tokens/tokens/base.react.native';
+
+// Required so that it appears in the same place as the clear button.
+// TODO put these into tokens because this isn't so nice.
+const iconPlacement = I18nManager.isRTL ? 9 : spacingSm * 1.5;
```
shaundon: I feel the need to point out how awful this is. We have to hard set this so that the icon overlaps iOS's close icon just right. In RTL it's placed differently for some reason (they couldn't even make it an even number!).

As @k0nserv rightly points out, this is not a good long term solution as if iOS changes, it breaks our layout. We should soon look at rendering our icons outside of the text field to prevent this.

PR #373, opened 2017-12-11
[BPK-880] add star rating docs

matteo-hertel:
Docs for star rating component

screen shot 2017-12-11 at 16 40 41-fullpage


packages/bpk-docs/src/pages/NativeStarRatingPage/NativeStarRatingPage.js
```diff
@@ -0,0 +1,66 @@
...
+const NativeStarRatingPage = () => (<DocsPageBuilder
+  title="Star Ratings"
+  blurb={[
+    <Paragraph>
+      Star ratings are used for displaying a score within a given range.
+    </Paragraph>,
+  ]}
+  components={components}
+  readme={readme}
+  showMenu
```
shaundon: Change `showMenu` to `false`, as the menu isn't needed when there's only the default configuration.

matteo-hertel: really good shout!

PR #376, opened 2017-12-12
[No JIRA] Rename pageLabelFormatter to pageLabel

k0nserv:
The word `formatter` in `pageLabelFormatter` is redundant as this
is capture by the prop type being `func`.


shaundon:
Looks good - will merge once CI passes. Do snapshots need to update?

k0nserv:
There are no snapshots for this component. @ksiwei never created any. I think we should probably have some, but let's do that later

PR #381, opened 2017-12-13
[BPK-1108] Updated docs

shaundon:
localhost_8080_components_native_text-input 1


matthewdavidson:
@shaundon the dimensions of the images shouldn't have changed, should they?

shaundon:
@matthewdavidson no they shouldn't have, no idea why that's happened. I'll fix anyway

PR #384, opened 2017-12-13
[NOJIRA] Get description lists over the line

matthewdavidson:
**Note:** I have renamed `Definition` -> `Description` as per Mozilla's terminology as this is the source of truth we use (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl).

I also added LTR support and modified the indent margin for `<dd>`'s.

Here is what it looks like:

image


matthewdavidson:
cc @janeklb - i hope you don't mind. I wanna get this component released :)

janeklb:
not at all, got plenty of things to do :)

.storybook/config.js
```diff
@@ -58,7 +58,7 @@ configure(() => {
   require('./../packages/bpk-component-chip/stories');
   require('./../packages/bpk-component-close-button/stories');
   require('./../packages/bpk-component-code/stories');
-  require('./../packages/bpk-component-definition-list/stories');
+  require('./../packages/bpk-component-description-list/stories');
```
shaundon: Just needs to be one line lower, as right now it's appearing before `bpk-component-content-container`.

matthewdavidson: Good catch - should actually be 2 lines lower 😉  😂

shaundon: Close enough :P

PR #386, opened 2017-12-14
[no jira] Reinstate placeholder support to native text input

shaundon:
Feel silly we didn't think of this approach before!

kapture 2017-12-14 at 10 13 10

CC @tadam313

### DIFF
--- a/native/packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
+++ b/native/packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
@@ -138,7 +138,7 @@ const commonTests = () => {
       expect(testRenderer.toJSON()).toMatchSnapshot();
     });
 
-    it('should ignore when placeholder is provided', () => {
+    it('should ignore when placeholder is provided, as element is not focused', () => {
       const testRenderer = TestRenderer.create(
         <BpkTextInput
           label="Name"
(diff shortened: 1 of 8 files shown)


native/packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
```diff
@@ -137,18 +137,6 @@ const commonTests = () => {
 
       expect(testRenderer.toJSON()).toMatchSnapshot();
     });
-
-    it('should ignore when placeholder is provided', () => {
```
matthewdavidson: Is there any way we can test the placeholder support?

jamesf3rguson: I'm happy

jamesf3rguson:
I'm happy.

native/packages/react-native-bpk-component-text-input/src/BpkTextInput-test.common.js
shaundon: We could test that it's null by default (because the element isn't focused). Is there a way you know of to trigger focus in a test and then snapshot that?

shaundon: I've added a test for it being null when not focused (it's just the same test I removed, reinstated). Not sure about testing when focused though.

PR #387, opened 2017-12-14
[BPK-1152] Add prettier

k0nserv:
Introduces prettier to the project. Prettier is ran via eslint and eslint fix. 

New dependeices

+ prettier 
+ eslint-config-prettier - Changes ESLint rules to be compatible with prettier
+ eslint-plugin-prettier - Runs prettier as part of ESLint
+ eslint_d - To ensure `eslint_d` is available to Ale which I use for Vim linting and fixing.
+ husky - Git commit hooks
+ lint-staged - Make commit hook faster

Currently we are running `npm run lint:js` in precommit and preventing the commit on any violations. Style is also verified on CI as before

## Todo

- [x] Make it faster(precommit takes 20s)
- [x] Document prettier in CONTRIBUTING.md #


k0nserv:
@matthewdavidson mentioned https://github.com/okonet/lint-staged which can lint staged files only. This should cut down the 20s it currently takes to run the precommit hook. Their example runs fix and then adds the files I'd prefer to just lint and fail instead

dangerfile.js
```diff
@@ -131,5 +156,9 @@ const androidSnapshotsWithIosTokens = fileChanges.filter((filePath) => {
 
 if (androidSnapshotsWithIosTokens.length > 0) {
   // eslint-disable-next-line max-len
-  fail(`iOS tokens have been found in the following Android snapshots:\n  - ${androidSnapshotsWithIosTokens.join('\n  - ')}`);
+  fail(
+    `iOS tokens have been found in the following Android snapshots:\n  - ${androidSnapshotsWithIosTokens.join(
```
matthewdavidson: ^^^ I reckon some of our console formatting is going to be screwed up with this change.

package.json
```diff
@@ -44,7 +45,11 @@
     "android": "(cd ./native && npm run android)",
     "check-owners": "npm whoami && grep -E $(npm whoami) NPM_OWNERS && node scripts/npm/check-owners.js",
     "publish": "npm run check-owners && npm run build && git pull --rebase && npm run test && lerna publish",
-    "danger": "danger"
+    "danger": "danger",
+    "prettier": "prettier --config .prettierrc --write \"{packages,native,scripts}/**/*.js\" dangerfile.js webpack.config.babel.js"
```
matthewdavidson: Surely `prettier --config .prettierrc --write \"**/*.js\"` is enough here?

package.json
```diff
@@ -44,7 +45,11 @@
...
+  },
+  "lint-staged": {
+    "*.js": ["eslint"]
```
matthewdavidson: How does this relate to our `npm run lint:js`script? I.e. we have some custom stuff in there (`eslint . .storybook --ext .js,.jsx`). It feels like its going down a different path.

k0nserv: Shouldn't be a problem. `--ext` is covered by the `*.js` filter and the other args are which files to lint and we don't want to explicitly lint everything here

package.json
k0nserv: Yeah you are probably right

dangerfile.js
k0nserv: That case seems fine though

package.json
k0nserv: We don't have any jsx files as far as I can see

janeklb:
This is awesome. How much work would it be to add to mshell-node-react?

PR #388, opened 2017-12-14
[BPK-1158] Theme selector

shaundon:
screen shot 2017-12-14 at 16 29 06

So we now have three themes; London, Hong Kong and Doha. The names and the colours came directly from @jamesf3rguson. London is the one we had before.

The keyboard shortcut is still present, except it now cycles between the three.


.storybook/config.js
```diff
@@ -38,7 +38,9 @@ addDecorator(story => (
     <br />
     <BpkRtlToggle />
     <br />
-    <BpkThemeToggle />
+    <div style={{ width: '10rem' }}>
```
shaundon: Unsure if there's a better way to do this, seeing as we don't have stylesheets for the storybook config (which also rules out using `TOKENS.onePixelRem * 10` because `string * 10 == NaN`

shaundon:
@matteo-hertel @matthewdavidson @k0nserv @georgegillams 

What do you think about renaming this to `bpk-component-theme-switcher`? It'll be a lot of hassle for minimal gain, as we'll be effectively orphaning a package and starting a new one. And it's still a toggle when you think about it

matthewdavidson:
@shaundon Agreed - keep the name as is

.storybook/config.js
georgegillams: `\`${parseInt(TOKENS.onePixelRem, 10) * 10}${TOKENS.onePixelRem.replace(/[0-9]/g, '')}\``

georgegillams: 🤣

georgegillams:
💯

PR #389, opened 2017-12-14
[BPK-1156] Fix Web banner alert animation jank

georgegillams:
Think this is good except for changelog entry and maybe a more in-depth explanation in the readme?

### DIFF
--- a/packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
+++ b/packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
@@ -69,6 +69,17 @@ describe('BpkBannerAlert', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('should render correctly with a custom banner-alert class name', () => {
+    const tree = renderer.create(
+      <BpkBannerAlert
+        type={ALERT_TYPES.WARN}
+        message={message}
+        bannerClassName="custom-class"
+      />,
+    ).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
   it('should render correctly with arbitrary props', () => {
     const tree = renderer.create(
       <BpkBannerAlert type={ALERT_TYPES.WARN} message={message} id="custom-id" hidden="hidden" />,
--- a/packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
+++ b/packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
@@ -17,10 +17,11 @@
  */
 
 import React, { Component } from 'react';
-import { fontWeightBold } from 'bpk-tokens/tokens/base.es6';
+import { durationSm, fontWeightBold } from 'bpk-tokens/tokens/base.es6';
 import PropTypes from 'prop-types';
 import { cssModules, withDefaultProps } from 'bpk-react-utils';
-import BpkBannerAlert, { ALERT_TYPES } from 'bpk-component-banner-alert';
+import BpkAnimateHeight from 'bpk-animate-height';
+import BpkBannerAlert, { ALERT_TYPES, withBannerAlertState } from 'bpk-component-banner-alert';
 import BpkButton from 'bpk-component-button';
 import bannerAlertReadme from 'bpk-component-banner-alert/readme.md';
 import DocsPageBuilder from './../../components/DocsPageBuilder';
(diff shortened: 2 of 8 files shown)


packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
```diff
@@ -55,7 +55,11 @@ describe('BpkBannerAlert', () => {
```
shaundon: We should get a test for the new `bannerClassName` prop

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -84,6 +84,81 @@ BpkBannerDismissable.defaultProps = {
   type: null,
 };
 
+// eslint-disable-next-line react/no-multi-comp
+class BpkBannerAlertDismissDemo extends Component {
+  constructor() {
+    super();
+
+    this.newBannerAlerts = [
```
shaundon: Why the `new` part?

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -84,6 +84,81 @@ BpkBannerDismissable.defaultProps = {
...
+    const updatedBannerAlerts = JSON.parse(JSON.stringify(this.state.bannerAlerts));
+    updatedBannerAlerts[index].show = false;
+    this.setState({
+      bannerAlerts: updatedBannerAlerts,
+    });
+  }
+
+  reset() {
+    this.setState({
+      bannerAlerts: JSON.parse(JSON.stringify(this.newBannerAlerts)),
```
shaundon: How come you're stringifying it and then parsing it?

packages/bpk-docs/src/pages/BannerAlertsPage/bpk-banner-alerts-page.scss
```diff
@@ -19,7 +19,7 @@
 @import '~bpk-mixins/index';
 
 .bpk-banner-alerts-page {
-  &__component:not(:last-child) {
-    margin-bottom: $bpk-spacing-base;
+  &__component {
+    margin-top: $bpk-spacing-base;
```
shaundon: How come this has changed?

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -84,6 +84,81 @@ BpkBannerDismissable.defaultProps = {
...
+  }
+
+  render() {
+    return (
+      <div>
+        <BpkButton
+          onClick={this.reset}
+        >
+          Reset
+        </BpkButton>
```
shaundon: I feel that we should make this only appear when they've removed one or more of the alerts, as otherwise it won't have a function. We can just have a state variable called `objectsEdited` which starts `false`, becomes true when `setDismissed` is called and false again when `reset` is called. 

What do you think?

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
georgegillams: Javascript doesn't have a built in clone option. This is the most performant way to achieve the same without relying on 3rd party stuff!

shaundon: Yeah fair enough - I googled it and came to the same thing! https://stackoverflow.com/a/23481096/1011161

packages/bpk-docs/src/pages/BannerAlertsPage/bpk-banner-alerts-page.scss
georgegillams: because I added the reset button, this is now unessessary as we want the top one to be paded too. Will possible revert when I make the reset button appear / disapear etc

packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
georgegillams: tru dat!

packages/bpk-docs/src/pages/BannerAlertsPage/bpk-banner-alerts-page.scss
shaundon: Ahh ok!

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
shaundon: I guess what I'm getting at is it reads a bit funny - you can probably just call it `bannerAlerts`

shaundon:
There are some syntax errors in `BpkBannerAlert.js` that's breaking CI. It looks like something went really wrong with fixing a merge conflict?

georgegillams:
Yeah I'm partway through rebasing 🤣

PR #392, opened 2017-12-17
[BPK-434] use bpk-component-icon for sort direction icons

janeklb:
- https://github.com/Skyscanner/backpack/commit/bf00596c8a0c941e8e6af80c90b24357f3f1aadf: add `hasChildrenOfType` util (probably a good candidate for `bpk-react-utils`) to ensure children of `BpkDataTable` are `BpkDataTableColumn`s.
- https://github.com/Skyscanner/backpack/commit/6f019c38fb25d9d9034416daa8965d54ac9b5d35: use `bpk-component-icon` for sort direction icons

### DIFF
--- a/packages/bpk-component-datatable/src/hasChildrenOfType.js
+++ b/packages/bpk-component-datatable/src/hasChildrenOfType.js
@@ -1,0 +1,59 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Children } from 'react';
+
+export function getDisplayName(Component) {
+  return (
+    Component.displayName ||
+    Component.name ||
+    (typeof Component === 'string' && Component.length > 0
+      ? Component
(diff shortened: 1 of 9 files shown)


packages/bpk-component-datatable/src/hasChildrenOfType.js
```diff
@@ -0,0 +1,53 @@
...
+const hasChildrenOfType = (type, atLeast = 1) => (props, propType, componentName) => {
+  if (Children.count(props[propType]) < atLeast) {
+    const inflectedNoun = atLeast === 1 ? 'child' : 'children';
+    return Error(
+      `${componentName} requires at least ${atLeast} '${getDisplayName(type)}' ${inflectedNoun}.`,
+    );
+  }
+
+  const children = Children.toArray(props[propType]);
+  for (let i = 0; i < children.length; i += 1) {
```
shaundon: You can use `React.Children.forEach` here: https://reactjs.org/docs/react-api.html#reactchildrenforeach

janeklb:
hey @shaundon - luckily the component hasn't been announced yet (was just merged and published last week https://github.com/Skyscanner/backpack/pull/370) so I think it's safe to update the prop types. Also, there are no docs yet :/

shaundon:
@janeklb my bad - I confused this with `bpk-component-table`!

packages/bpk-component-datatable/src/hasChildrenOfType.js
janeklb: The reason for this approach is to be able to return directly out of the iteration block -- can you do `return React.Children.forEach(...)`?

shaundon: Ah yeah, good point. I'm surprised eslint didn't object to that, it usually shouts at me when I use a native `for` loop. But as long as it works for you 😄

shaundon:
Updated my earlier comment to address me being silly before and talking about the wrong component

packages/bpk-component-datatable/src/hasChildrenOfType.js
janeklb: Yeah i was waiting to hear from it :)

PR #393, opened 2017-12-18
[BPK-1161] Added Do's/Don'ts for Popovers and Modals

jamesf3rguson:


packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
```diff
@@ -146,6 +146,14 @@ const ModalsPage = () => (
     ]}
     components={components}
     readme={modalReadme}
+    usageTable={{
+      dos: ['Use for rich / complex content related to the current page.'],
```
matthewdavidson: What is "rich / complex content"? Do we have an example of the opposite? These do's and don't should be really simple and obvious.

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
```diff
@@ -146,6 +146,14 @@ const ModalsPage = () => (
...
+      donts: [
+        "Don't use for alerts, errors and confirmations - use a dialog instead.",
+        `Don't overuse modals as interrupts as they will get in the way of the
+      user's main tasks, and may disrupt context.`,
```
matthewdavidson: This sentence is grammatically incorrect. Perhaps: "Don't overuse modals. They are an intrusive UI pattern and should be used sparingly." Im not sure though...

shaundon: I heard "grammar" and came as fast as I could.

I think it's probably grammatically correct, but it's a bit ambiguous as we're using 'interrupt' as a noun. I like what Matt suggests as it's less ambiguous.

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
shaundon: Maybe something along the lines of "complex content, e.g. forms, multiple controls"?

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
```diff
@@ -146,6 +146,17 @@ const ModalsPage = () => (
     ]}
     components={components}
     readme={modalReadme}
+    usageTable={{
+      dos: [
+        'Use when users need to regain context of where they are in the UI',
+        'Use when users need to focus on a separate task/information in relation to their current goal.',
```
shaundon: I think this would be better with the `/information` part removed, as it reads a bit funny. What do you think?

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
```diff
@@ -146,6 +146,17 @@ const ModalsPage = () => (
     ]}
     components={components}
     readme={modalReadme}
+    usageTable={{
+      dos: [
+        'Use when users need to regain context of where they are in the UI',
```
shaundon: Should end in a full stop.

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
```diff
@@ -146,6 +146,17 @@ const ModalsPage = () => (
...
+        'Use when users need to focus on a separate task/information in relation to their current goal.',
+      ],
+      donts: [
+        "Don't use for alerts, errors and confirmations - use a dialog instead.",
+        `Don't overuse modals as interrupts as they will get in the way of the
```
shaundon: I still think `interrupts` ➡️  `interruptions` 😂

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
jamesf3rguson: We want to keep this as users won't always have a task to complete

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
jamesf3rguson: Fixed.

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
jamesf3rguson: Fixed :)

PR #398, opened 2017-12-19
Bpk 1059 android banner alert

matteo-hertel:
This is now splitted in two commits, the fist commit had a bunch of feedback (https://github.com/Skyscanner/backpack/pull/382) the second commit is the latest version

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,9 @@
 ## UNRELEASED
 
 **Added:**
+- react-native-bpk-component-banner-alert:
+  - Android support
+
 - bpk-component-banner-alert:
   - `withBannerAlertState` HOC, see https://backpack.github.io/components/web/banner-alerts/#withBannerAlertState
   - `bannerClassdName` prop added to allow stlying to be applied to the internal banner rather than the outer container. See https://backpack.github.io/components/web/banner-alerts/#props
--- a/native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
+++ b/native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
@@ -1,0 +1,272 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import {
+  StyleSheet,
+  TouchableNativeFeedback,
+  TouchableWithoutFeedback,
+  View,
+  ViewPropTypes,
+} from 'react-native';
+
+import {
+  borderRadiusSm,
+  colorBlue500,
+  colorGray50,
+  colorGray500,
+  colorGreen500,
+  colorRed500,
+  colorYellow500,
+  spacingBase,
+  spacingLg,
+  spacingMd,
+  spacingSm,
+  spacingXxl,
+} from 'bpk-tokens/tokens/base.react.native';
+
+import BpkAnimateHeight from 'react-native-bpk-component-animate-height';
+import BpkIcon from 'react-native-bpk-component-icon';
+import BpkText from 'react-native-bpk-component-text';
+
+import AnimateAndFade from './AnimateAndFade';
+import {
+  dismissablePropType,
+  toggleExpandedButtonLabelPropType,
+  dismissableLabelPropType,
+} from './customPropTypes';
+
+export const ALERT_TYPES = {
+  SUCCESS: 'success',
+  WARN: 'warn',
+  ERROR: 'error',
+  NEUTRAL: 'neutral',
+};
+
+const styles = StyleSheet.create({
+  mainContainer: {
+    paddingVertical: spacingBase,
+    paddingHorizontal: spacingLg,
+    backgroundColor: colorGray50,
+    minHeight: spacingXxl + spacingBase,
+    flex: 1,
+  },
+  dismissableContainer: {
+    paddingHorizontal: spacingLg - spacingMd,
+    backgroundColor: colorGray50,
+    justifyContent: 'center',
+  },
+  childrenContainer: {
+    paddingVertical: spacingSm,
+  },
+  flexRow: {
+    display: 'flex',
+    flexDirection: 'row',
+  },
+  text: {
+    flex: 1,
+  },
+  icon: {
+    paddingRight: spacingSm,
+    paddingTop: spacingSm - 1,
+  },
+  dismissLabel: {
+    color: colorBlue500,
+  },
+  dismissButtonContainer: {
+    padding: spacingMd,
+    borderRadius: borderRadiusSm,
+  },
+  dismissButtonContainerRadius: {
+    borderRadius: borderRadiusSm,
+  },
+});
+
+const ALERT_TYPE_STYLES = {
+  [ALERT_TYPES.SUCCESS]: {
+    iconSource: 'tick-circle',
+    iconStyle: {
+      color: colorGreen500,
+    },
+  },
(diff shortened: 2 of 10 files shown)


native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -0,0 +1,255 @@
...
+  flexRow: {
+    display: 'flex',
+    flexDirection: 'row',
+  },
+  text: {
+    flex: 1,
+  },
+  icon: {
+    paddingRight: spacingSm,
+    paddingTop: spacingSm - spacingSm / spacingSm,
```
k0nserv: Huh? This is just `spacingSm - 1` right?

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -0,0 +1,255 @@
...
+    style: userStyle,
+    animateOnEnter,
+    animateOnLeave,
+    children,
+    ...rest
+  } = props;
+  const expandable = children !== null;
+  const { iconSource, iconStyle } = ALERT_TYPE_STYLES[type] || {};
+
+  const dismissableButton = () => {
```
k0nserv: It's probably not great to create this component inside render every time. I would suggest moving it out

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -0,0 +1,255 @@
...
+    if (!dismissable) {
+      return null;
+    }
+    return (
+      <View style={styles.dismissableContainer}>
+        <View style={styles.dismissButtonContainerRadius}>
+          <TouchableNativeFeedback
+            accessibilityComponentType="button"
+            accessibilityLabel="dismiss"
```
k0nserv: This looks like an error. This should be coming from props rights?

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -0,0 +1,255 @@
...
+      ) : (
+        mainContent
+      )}
+    </AnimateAndFade>
+  );
+};
+
+BpkBannerAlert.propTypes = {
+  message: PropTypes.string.isRequired,
+  type: PropTypes.oneOf(Object.keys(ALERT_TYPES).map(key => ALERT_TYPES[key]))
```
k0nserv: This would be simpler with `Object.values` instead

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
shaundon: I think it also doesn't need to be a function - the only place it's called is on line 197, that looks like this:

``` {dismissable && dismissableButton()}```

Therefore you don't need to do `if (!dismissable) { return null; }` as that's already been checked. 

I recommend changing it to be a variable, same as you have for `expandableContent` and `expandableIcon`.

File: native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js (hunk 1 above)
Comment (matteo-hertel) [reply native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js]: yep, reason being, the icon renders with some spacing up top, not sure why but this does it 😄

File: native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js (hunk 3 above)
Comment (matteo-hertel) [reply native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js]: Woah, nice catch!! 🦅

**Comment by shaundon** [comment]
I like how it's split into components now, but the components are nested inside `BpkBannerAlert`, which makes the component quite large and hard to read.

I feel that it would benefit from a flatter structure, so we'd go from this:

```
const BpkBannerAlert = props => {
   ...
  const expandableContent = ( ... );
  const expandableIcon = ( ... );
  const mainContent = ( ... );
  
  return (
    <AnimateAndFade>
      ...
      {mainContent}
    </AnimateAndFade>
  );
};
```

..to this:

```
const BpkBannerAlert = props => {
   ...  
  return (
    <AnimateAndFade>
      <MainContent />
    </AnimateAndFade>
  );
};

const ExpandableContent = () => ( ... )
const ExpandableIcon = () => ( ... )
const MainContent = (props) => ( ... )
```

I do get that this is a prime example of what we discussed at stand-up this morning and that others might disagree with how to lay it out - I feel that this way is a lot more readable though.

File: native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js (hunk 1 above)
Comment (shaundon) [reply native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js]: I'd be tempted to just make that `spacingSm - 1` in that case, it feels a bit like we're going out of our way to use a token when `1` would make more sense (if it were a number other than 1 I'd say use a token though)

Comment (k0nserv) [reply native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js]: But why not write it as `spacingSm - 1` though? All though we should investigate why the spacing is there and get rid of it

File: native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js [hunk 5]
```diff
@@ -0,0 +1,255 @@
...
+      color: colorRed500,
+    },
+  },
+  [ALERT_TYPES.NEUTRAL]: {
+    iconSource: 'information-circle',
+    iconStyle: {
+      color: colorGray500,
+    },
+  },
+};
```
shaundon: Let's leave it out of this PR, but I think this should be moved to its own file as it's duplicated between platforms. I'll make a card for it.

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
matteo-hertel: Fair, will do

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
matteo-hertel: I thought magic arbitrary numbers were banned, I'll amend to `1`

matteo-hertel:
I've moved some bits and bobs outside the render function, the main content will say there as move to a function/component will generate much more code than it does now, other than that it has been all actioned, ready for a second pass

changelog.md
```diff
@@ -3,6 +3,9 @@
 ## UNRELEASED
 
 **Added:**
+- react-native-bpk-component-banner-alert:
+  - android support
```
shaundon: Teeny thing but `Android` should be capitalised 📚

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -0,0 +1,257 @@
...
+
+const dismissableButton = (dismissable, dismissButtonLabel, onDismiss) => {
+  if (!dismissable) {
+    return null;
+  }
```
shaundon: I think you can remove this `if` statement, because before this function is called you're checking if `dismissable == true`

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -0,0 +1,257 @@
...
+  },
+  [ALERT_TYPES.NEUTRAL]: {
+    iconSource: 'information-circle',
+    iconStyle: {
+      color: colorGray500,
+    },
+  },
+};
+
+const dismissableButton = (dismissable, dismissButtonLabel, onDismiss) => {
```
shaundon: My gut feeling says this would be better written as a React component and not a function, e.g.:

```
const DismissableButton = ({ dismissButtonLabel, onDismiss }) => (
  <View style={styles.dismissableContainer}>
     ...
  </View>
);
```

So then it could be called from `BpkBannerAlert` like this:

```
{dismissable && <DismissableButton dismissButtonLabel={dismissButtonLabel} onDismiss={onDismiss} />}
```
as opposed to:
```
{dismissable && dismissableButton(dismissable, dismissButtonLabel, onDismiss)}
```

But in fairness, I'm unsure if there are any specific advantages we get from changing it to this. (any thoughts @matthewdavidson @k0nserv?)

(same applies to `expandableContent` and `expandableIcon` too)

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
matteo-hertel: Yes, this is non needed anymore!

PR #402, opened 2017-12-21
[nojira] Remove import * as React

matthewdavidson:
This fixes the following warnings caused by `import * as React from 'react';` change:

```
  console.warn node_modules/react/lib/lowPriorityWarning.js:38
    Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs

  console.warn node_modules/react/lib/lowPriorityWarning.js:38
    Warning: Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
```

### DIFF
--- a/packages/bpk-docs/src/pages/ButtonsPage/ButtonsPage.js
+++ b/packages/bpk-docs/src/pages/ButtonsPage/ButtonsPage.js
@@ -18,7 +18,7 @@
 
 /* @flow */
 
-import * as React from 'react';
+import React, { Component } from 'react';
 import BpkButton from 'bpk-component-button';
 import { colors, buttons } from 'bpk-tokens/tokens/base.es6';
 import { alignToButton, alignToLargeButton } from 'bpk-component-icon';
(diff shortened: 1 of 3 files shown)


packages/bpk-docs/src/pages/ButtonsPage/ButtonsPage.js
```diff
@@ -51,7 +51,7 @@ const AlignedBpkLargeHelpIcon = alignToLargeButton(TestBpkLargeHelpIcon);
 const AlignedBpkSmallSearchIcon = alignToButton(TestBpkSmallSearchIcon);
 const AlignedBpkLargeSearchIcon = alignToLargeButton(TestBpkLargeSearchIcon);
 
-class LoadingButtonContainer extends React.Component<{}, { loading: boolean }> {
+class LoadingButtonContainer extends Component<{}, { loading: boolean }> {
   constructor() {
     super();
```
shaundon: From what I can see, you've not changed anything in this file to make it work with the `import React, { Component } from 'react';` syntax. Doesn't Flow complain?

PR #403, opened 2017-12-21
[NoJira] update check-owners

matteo-hertel:
I got a bit carried away but this fixes the issue with `.` files we have during release, and I've added some async/await cause why not? 😄

### DIFF
--- a/scripts/npm/check-owners.js
+++ b/scripts/npm/check-owners.js
@@ -89,8 +89,9 @@ console.log(`Maintainers are:\n  ${owners.join('\n  ')}\n`);
 
 Promise.all([readdir('packages/'), readdir('native/packages/')])
   .then(packages =>
-    Promise.all([...packages[0], ...packages[1]].map(getPackageMaintainers)),
+    [...packages[0], ...packages[1]].filter(i => !i.startsWith('.')),
   )
+  .then(packages => Promise.all(packages.map(getPackageMaintainers)))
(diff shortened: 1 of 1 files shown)


scripts/npm/check-owners.js
```diff
@@ -19,23 +19,80 @@
 /* @flow */
 
 /* eslint-disable no-console */
+/* eslint-disable no-use-before-define */
```
matteo-hertel: I do want hoisting in my JS!

matthewdavidson: Unfortunately the rule is there for a reason Matteo :)

matteo-hertel: In this particular instance I want to allow hoisting, it dramatically improve readability having the uniform function on top, opening the file will give a full overview of the business logic, implementation details are further down

georgegillams: I'm with @matthewdavidson  on this one. Maybe C# / Java devs would find this more readable.
If the business logic is really much lower-level than the overall business logic, it should probably be split into multiple files.

matteo-hertel: I completely disagree with that, that's not C# / Java at all, splitting the logic in multiple file doesn't really make sense here as won't be re-used by other scripts, at least not for now

matteo-hertel:
I've decided to revert most of the changes I've done and just add the fix for the core issue of having the script picking up the dotfiles

PR #404, opened 2017-12-21
[BPK-1177] Adding horizontal nav to docs site header

jamesf3rguson:


k0nserv:
This approach unfortunately does not work in Firefox and Edge. Even in Chrome and Safari where it does work you can only click the text rather than the whole item which is meh at best

packages/bpk-docs/src/components/Header/Header.js
```diff
@@ -56,68 +56,90 @@ const toNavList = (links, hideOnTablet) => {
...
+    };
+    itemProps.className = getClassName(
+      'bpkdocs-header__horizontal-nav__nav-list-link',
+    );
+  } else {
+    itemProps.href = link.href;
+    itemProps.target = link.blank ? '_blank' : '_self';
+    itemProps.className = getClassName(
+      'bpkdocs-header__horizontal-nav__nav-list-link',
+    );
```
shaundon: (sorry for the repeat comment, I added it in the wrong place first)

This looks like it's repeated in both the `if` and `else`, can we move it up?

k0nserv: Yeah this code can definitely be refactored anyway, will do if we agree with the approach

shaundon: Ok cool - in that case I definitely agree with the approach :)

k0nserv:
Updated :) @jamesf3rguson can you try this out and make sure you like the look/feel?

PR #406, opened 2017-12-22
[BPK-1078] Flow support for native BpkIcon

shaundon:
Also added a `.flowconfig` to `/native` to make Flow play nicely with React Native. Comments about the code are inline.

### DIFF
--- a/native/.flowconfig
+++ b/native/.flowconfig
@@ -1,0 +1,53 @@
+[ignore]
+; Ignore templates for 'react-native init'
+<PROJECT_ROOT>/node_modules/react-native/local-cli/templates/.*
+
+<PROJECT_ROOT>/react-native-bpk-*/node_modules/.*
+
+; Ignore RN jest
+<PROJECT_ROOT>/node_modules/react-native/jest/.*
+
+; Ignore RNTester
+<PROJECT_ROOT>/node_modules/react-native/RNTester/.*
+
+; Ignore the website subdir
+<PROJECT_ROOT>/node_modules/react-native/website/.*
+
+; Ignore the Dangerfile
+<PROJECT_ROOT>/node_modules/react-native/danger/dangerfile.js
+
+; Ignore Fbemitter
+<PROJECT_ROOT>/node_modules/fbemitter/.*
+
+; Ignore "BUCK" generated dirs
+<PROJECT_ROOT>/node_modules/react-native/\.buckd/
+
+; Ignore unexpected extra "@providesModule"
+.*/node_modules/.*/node_modules/fbjs/.*
+
+; Ignore polyfills
+<PROJECT_ROOT>/node_modules/react-native/Libraries/polyfills/.*
+
+; Ignore various node_modules
+<PROJECT_ROOT>/node_modules/react-native-gesture-handler/.*
+<PROJECT_ROOT>/node_modules/expo/.*
+<PROJECT_ROOT>/node_modules/react-navigation/.*
+<PROJECT_ROOT>/node_modules/xdl/.*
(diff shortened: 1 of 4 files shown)


native/.flowconfig
```diff
@@ -0,0 +1,53 @@
+[ignore]
```
shaundon: This file is largely copied from a vanilla `create-react-native-app` app.

shaundon: We may be able to remove bits but for now I've left it mostly intact.

native/.flowconfig
```diff
@@ -0,0 +1,53 @@
...
+
+[options]
+emoji=true
+
+unsafe.enable_getters_and_setters=true
+
+module.file_ext=.js
+module.file_ext=.json
+module.file_ext=.ios.js
+module.file_ext=.android.js
```
shaundon: Without these lines, we get errors from RN code that implicitly required `.ios` and `.android`, e.g.:

```
import { colorGray700 } from 'bpk-tokens/tokens/base.react.native';
```

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -42,7 +44,14 @@ const styles = StyleSheet.create({
 const mapCharacterCode = characterCode =>
   String.fromCharCode(parseInt(characterCode, 16));
 
-const BpkIcon = props => {
+type IconName = $Keys<typeof iconMappings>;
+type Props = {
+  icon: IconName,
+  small: ?boolean,
+  style: ?{} | Array<{}>,
```
shaundon: This is probably going to become a common pattern, so we could abstract it out. Alternatively, we could pull a type out of `react-native` itself. For now, this works well.

native/packages/react-native-bpk-component-icon/stories.js
```diff
@@ -16,6 +16,8 @@
  * limitations under the License.
```
shaundon: I tested that Flow works here by changing the `small` prop on an icon to a string (should be a boolean) and it does display an error.

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -42,7 +44,14 @@ const styles = StyleSheet.create({
 const mapCharacterCode = characterCode =>
   String.fromCharCode(parseInt(characterCode, 16));
 
-const BpkIcon = props => {
+type IconName = $Keys<typeof iconMappings>;
+type Props = {
+  icon: IconName,
+  small: ?boolean,
```
k0nserv: This shouldn't be optional, we always expect it to have a valid value (true or false).

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
k0nserv: How does React Native itself do these?

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
shaundon: But it *is* an optional proptype - that's what this is analogous to, right?

k0nserv: No that’s not true. The prop type is external to the component, but flow is internal i.e while a consumer can omit `small`, inside the component we expect it to always have a value. In this case React and the default props ensure that that is the case

shaundon: Ok, makes sense. I'll change it

PR #407, opened 2017-12-26
[NO JIRA] add 'hasChildrenOfType' to bpk-react-utils

janeklb:
Currently lives in bpk-component-datatable (added here: https://github.com/Skyscanner/backpack/commit/bf00596c8a0c941e8e6af80c90b24357f3f1aadf) but makes sense to move to bpk-react-utils.

This PR is just to add the functionality to bpk-react-utils, then will submit a separate one for bpk-component-datatable.

### DIFF
--- a/packages/bpk-react-utils/readme.md
+++ b/packages/bpk-react-utils/readme.md
@@ -164,3 +164,48 @@ const MyComponent = (props) => (
 | appearClassName       | string   | true     | -             |
 | appearActiveClassName | string   | true     | -             |
 | transitionTimeout     | number   | true     | -             |
+
+## hasChildrenOfType.js
+
+A prop type checker that allows you to restrict a component's children to a specific type.
+
+### Usage
+
+```js
+import React from 'react';
+import { hasChildrenOfType } from 'bpk-react-utils';
+
+const Child = () => <span>Child</span>;
+
+const Parent = ({ children }) => <div>{ children }</div>;
+Parent.propTypes = {
+  children: hasChildrenOfType(Child),
+};
+
+// Valid:
+const WithValidChildren = () => (
+  <Parent>
+    <Child />
+    <Child />
+  </Parent>
+);
+
+// Invalid (will get prop type validation errors):
+const WithInvalidChildren = () => (
+  <Parent>
+    <Child />
+    <div>something else</div>
+  </Parent>
+);
+```
+
+### Parameters
+
+```js
+function hasChildrenOfType(Type, atLeast = 1)
+```
+
+| Parameter  | Description                     | Required | Default Value |
+| ---------- | ------------------------------- | -------- | ------------- |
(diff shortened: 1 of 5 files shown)


packages/bpk-react-utils/readme.md
```diff
@@ -164,3 +164,48 @@ const MyComponent = (props) => (
 | appearClassName       | string   | true     | -             |
 | appearActiveClassName | string   | true     | -             |
 | transitionTimeout     | number   | true     | -             |
+
+## hasChildrenOfType.js
+
+A prop type checker that allows you to restrict a component's children to a specific type.
```
k0nserv: Can we explain how this is different to `PropTypes.instanceOf`? That was my first question upon seeing this.

janeklb: Would adding 'Component' help? Totally open to suggestions if you have any ideas. 
eg.
> A prop type checker that allows you to restrict a component's children to a specific (Component) type.

matthewdavidson: @janeklb I second @k0nserv's question, what's the difference between using `hasChildrenOfType(MyComponent)` & `PropTypes.instanceOf(MyComponent)`? I.e. i think we are questioning whether this new prop type is redundant or not.

janeklb: `PropTypes.instanceOf(...)` uses js's `instanceof` operator to see if the prop value is an instance of the supplied class/type. It doesn't work with components (ie. jsx or `React.createElement` etc.) since those can't be operated on with `instanceof`.

This type checker validates that the _children_ of a component are all "instances" of a certain component type.

janeklb: See the tests - they should give some more clarity :)

janeklb: tbh im not sure why this isn't a part of `prop-types` itself -- seems like a common enough requirement

janeklb: And airbnb's prop types seem to already support something like this https://www.npmjs.com/package/airbnb-prop-types

matthewdavidson:
@janeklb where do you see this being used in backpack? Im tempted to close this as it can safely exist in userland and will most likely be available in `prop-types` at some point as you point out.

janeklb:
This is being used in `bpk-component-datatable` :)

It was inspired by `react-virtualized`'s `Table` [children validator](https://github.com/bvaughn/react-virtualized/blob/master/source/Table/Table.js#L30)

matthewdavidson:
@janeklb Lets keep it inlined within the data table component package then dude - we don't think `bpk-react-utils` is necessarily the best place for it. We can revisit it as per rule of three if the times comes - hopefully it'll be in `prop-types` by then.

janeklb:
sounds good! cheers

PR #409, opened 2017-12-27
[BPK-1122] Flow type web banner alerts

k0nserv:
I did not do the test because of how complex getting `flow-typed` and `jest` working is. I have concerns around CI etc with that approach so I figured let's discuss it in the new year.

### DIFF
--- a/packages/bpk-component-banner-alert/src/BpkBannerAlert.js
+++ b/packages/bpk-component-banner-alert/src/BpkBannerAlert.js
@@ -15,9 +15,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* @flow */
 
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { type Node } from 'react';
 import { withButtonAlignment } from 'bpk-component-icon';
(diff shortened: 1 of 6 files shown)


packages/bpk-component-banner-alert/src/BpkBannerAlert.js
```diff
@@ -45,14 +46,16 @@ export const ALERT_TYPES = {
   NEUTRAL: 'neutral',
 };
 
+type AlertTypeValue = $Values<typeof ALERT_TYPES>;
```
k0nserv: Should we export this and similar types? It makes it easier to reuse them in the docs for example.

packages/bpk-component-banner-alert/src/BpkBannerAlert.js
```diff
@@ -70,7 +73,12 @@ const getIconForType = type => {
   return map[type];
 };
 
-const ToggleButton = props => {
+type ToggleButtonProps = {
+  label: ?string,
```
k0nserv: This was previously required, but the prop in the outer component wasn't so this is a clarification of what should happen here. i.e this was a bug. We should consider if we want to handle this differently

PR #410, opened 2017-12-28
New web banner alert API 

k0nserv:
This adds three new components 

+ `BpkBannerAlert`
+ `BpkBannerAlertDismissable`
+ `BpkBannerAlertExpandable`

The previous `BpkBannerAlert` component is renamed `BpkBannerAlertInner` and contains most of the implementation. The three new components delegate to this component for all behaviour. The idea is that this will produce a clearer API since prop types can be more specific and expressive for each of these components. For example `dismissButtonLabel` was previously optional on `BpkBannerAlert` because it's only used when the banner alert is dismissable in `BpkBannerAlertDismissable` it's required since we always want the consumer to specify it for accessibility reasons


.eslintrc
```diff
@@ -59,6 +59,20 @@
         },
         "allowChildren": false
       }
-    ]
+    ],
+    "react/sort-comp": ["error", {
```
matteo-hertel: any particular reason why this is part of this PR?
if it was a group decision, then ignore this comment, if it wasn't I think it should be discussed and add it to the decision folder

k0nserv: No I just didn't consider the option of putting the flow attributes below render since that just seemed silly. Happy to discuss it further though

k0nserv:
Two very good points and I agree with both. Let's discuss when everyone is back

k0nserv:
So for the default config should we go with `CONFIGURATION.DEFAULT` then?

changelog.md
```diff
@@ -10,6 +10,18 @@ _Nothing yet.._
...
  - `icons` object is now exposed, allowing users to refer to icon names using a constant.
 
+**Breaking:**
+- bpk-component-banner-alert:
+  - Removed `dismissable` property in favour of `configuration`.
+  - A banner alert with `children` will no longer be auotmatically expandable. Use the newly introduce `configuration` prop instead.
+
+**Added:**
+- bpk-component-banner-alert:
+  - Introduced new component for the different banner alert configurations.
```
k0nserv: The grammar is shit here. Need to rewrite

packages/bpk-component-banner-alert/index.js
```diff
@@ -17,8 +17,16 @@
...
 import withBannerAlertState from './src/withBannerAlertState';
 
-export { ALERT_TYPES, withBannerAlertState };
+export {
+  ALERT_TYPES,
+  BpkBannerAlertDismissable,
+  BpkBannerAlertExpandable,
+  withBannerAlertState,
+};
 export default BpkBannerAlert;
```
k0nserv: What should be the default export here? Or should we just not have one so the import looks like

```javascript
import { BpkBannerAlert, BpkBannerAlertDismissable, BpkBannerAlertExpandable, ALERT_TYPES } from 'bpk-component-banner-alert';
```

packages/bpk-component-banner-alert/readme.md
```diff
@@ -14,12 +14,13 @@ npm install bpk-component-banner-alert --save-dev
 
 ```js
 import React from 'react';
-import BpkBannerAlert, { ALERT_TYPES } from 'bpk-component-banner-alert';
+import BpkBannerAlert, { ALERT_TYPES, CONFIGURATION } from 'bpk-component-banner-alert';
```
k0nserv: This still needs to be updated

packages/bpk-component-banner-alert/src/BpkBannerAlert-test.js
```diff
@@ -18,163 +18,41 @@
...
-          type={ALERT_TYPES.WARN}
-          message={message}
-          dismissable
-          dismissButtonLabel="Dismiss"
-          animateOnLeave
-        />,
-      )
-      .toJSON();
-    expect(tree).toMatchSnapshot();
+  it('should render correctly', () => {
```
k0nserv: I opted for shallow rendering since `BpkBannerAlertInner` is already well tested and these components just delegate to it

packages/bpk-component-banner-alert/src/BpkBannerAlertExpandable-test.js
```diff
@@ -0,0 +1,43 @@
...
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJson from 'enzyme-to-json';
+
+import BpkBannerAlertExpandable from './BpkBannerAlertExpandable';
+import { ALERT_TYPES } from './common-types';
+
+const message = 'Lorem ipsum dolor sit amet, consectetuer adipiscing elit.';
+
+describe('BpkBannerAlertExpandable', () => {
```
k0nserv: Should test expanded vs not here

packages/bpk-component-banner-alert/src/BpkBannerAlertInner-test.js
```diff
@@ -0,0 +1,187 @@
...
+/* @flow */
+
+import React from 'react';
+import renderer from 'react-test-renderer';
+import BpkBannerAlertInner, { CONFIGURATION } from './BpkBannerAlertInner';
+import { ALERT_TYPES } from './common-types';
+
+const message = 'Lorem ipsum dolor sit amet, consectetuer adipiscing elit.';
+
+describe('BpkBannerAlertInner', () => {
```
k0nserv: These are just the old tests from `BpkBannerAlert`

changelog.md
matteo-hertel: @shaundon is the man for the job!

packages/bpk-component-banner-alert/index.js
matteo-hertel: I think what you have now make sense, you're exporting by default the main component and the different variants

k0nserv:
> First pass looks good, one thing I noticed is inconsistent is the use of shallow vs renderer, we should probably make a group decision on what to use

I'm only using Shallow for the outer components and HOC. I could change it back for the HOC, but for the outer components they are only using `BpkBannerAlertInner` which is heavily tested itself so I felt it made sense that way

packages/bpk-component-banner-alert/src/BpkBannerAlertInner.js
```diff
@@ -0,0 +1,257 @@
+/*
```
matthewdavidson: Is this file mostly just a rename?

k0nserv: Yeah and the addition of the `configuration` prop

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
```diff
@@ -15,6 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* @flow */
```
matthewdavidson: You'll need to add the flow annotation to the index file too right?

changelog.md
```diff
@@ -10,6 +10,18 @@ _Nothing yet.._
 - react-native-bpk-component-icon: 1.1.15 => 1.2.0
  - `icons` object is now exposed, allowing users to refer to icon names using a constant.
 
+**Breaking:**
+- bpk-component-banner-alert:
+  - Removed `dismissable` property in favour of `configuration`.
+  - A banner alert with `children` will no longer be auotmatically expandable. Use the newly introduce `configuration` prop instead.
+
```
matthewdavidson: You've also removed `ARIA_LIVE` stuff...

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
k0nserv: Yeah true, haven't done the docs yet though

changelog.md
k0nserv: Good eye

packages/bpk-docs/src/pages/BannerAlertsPage/BannerAlertsPage.js
matthewdavidson: You touched this page tho :P

changelog.md
```diff
@@ -33,6 +33,20 @@
 - react-native-bpk-component-icon: 1.1.15 => 1.2.0
   - `icons` object is now exposed, allowing users to refer to icon names using a constant.
 
+**Breaking:**
+- bpk-component-banner-alert:
+  - Removed `dismissable` property in favour of `configuration`.
+  - A banner alert with `children` will no longer be auotmatically expandable. Use the newly introduce `configuration` prop instead.
```
matthewdavidson: > Removed `dismissable` property in favour of `configuration`.
> Use the newly introduce `configuration` prop instead.

These statements about the `configuration` prop don't make sense from a consumer POV. I.e. that prop isn't exposed to consumers. Instead it should focus on using one of the three new components (i.e. the contents of your migration guide).

packages/bpk-component-banner-alert/readme.md
```diff
@@ -10,29 +10,25 @@ npm install bpk-component-banner-alert --save-dev
 
 ## Usage
 
-### Non-dismissable
+### Simple
```
matthewdavidson: Call this Default

changelog.md
k0nserv: Yeah that's just me messing up

k0nserv:
Per my chat with @jamesf3rguson I've removed the ability to use `hideAfter` with expandable banner alerts. Please review 5ba4862

packages/bpk-component-banner-alert/docs/migrating-from-v1-to-v2.md
```diff
@@ -0,0 +1,120 @@
...
+    type={ALERT_TYPES.SUCCESS}
+    message="Everything went well"
+    onDismiss={() => undefined}
+  />
+);
+```
+
+## The expandable case
+
+In version 1 the expandable behaviour was enabled if the component had any children. With version 2 there is now a dedicated component, `BpkBannertAlertExpandable`, for this bevaiour.
```
shaundon: `BpkBannertAlertExpandable` ➡️  `BpkBannerAlertExpandable`, and `bevaiour` ➡️  `behaviour`.

packages/bpk-component-banner-alert/docs/migrating-from-v1-to-v2.md
```diff
@@ -0,0 +1,120 @@
...
+
+**V1:**
+
+```javascript
+import BpkBannerAlert, { ALERT_TYPES } from "bpk-component-banner-alert";
+
+const render = () => (
+  <BpkBannerAlert
+    type={ALERT_TYPES.SUCCESS}
+    message="Everything went horribly wrong"
```
shaundon: Haha love the message

packages/bpk-component-banner-alert/docs/migrating-from-v1-to-v2.md
```diff
@@ -0,0 +1,120 @@
...
+    onExpandToggle={() => undefined}
+  >
+    An even longer more detailed message
+  </BpkBannerAlertExpandable>
+);
+```
+
+## `withBannerAlertState`
+
+`withBannerAlertState` continues to work with both `BpkBannerAlertDismissable` and `BpkBannerAlertExpandable` as before. However when using `hideAfter` `onDimiss` is no longer called when the banner alert is removed, instead a new prop `onHide` has been introduced for this purposeas before. However when using `hideAfter` `onDimiss` is no longer called when the banner alert is removed, instead a new prop `onHide` has been introduced for this purpose.
```
shaundon: Looks like the same sentence (beginning `However when using...`) is repeated here.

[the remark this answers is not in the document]
k0nserv: Yes I definitely agree. This is why I started talking about the ability to render these as part of the docs page

PR #411, opened 2018-01-02
[BPK-1122] Use property initializer callbacks

k0nserv:


matthewdavidson:
Although now wouldn't you remove the eslint config changes? You might say that we'll need them in the future anyway but I'd rather we be explicit.

k0nserv:
No as they are still required by `hideIntervalId`

PR #412, opened 2018-01-02
[BPK-1122] Adds flow-typed and flow in tests

k0nserv:
This adds `flow-typed` as a dev dependency and `flow-typed` definitions
for jest so that we can use flow in tests.


packages/bpk-component-banner-alert/src/AnimateAndFade-test.js
```diff
@@ -27,7 +28,7 @@ const message = (
 describe('AnimateAndFade', () => {
   it('should render correctly', () => {
     const tree = renderer
-      .create(<AnimateAndFade>{message}</AnimateAndFade>)
+      .create(<AnimateAndFade show>{message}</AnimateAndFade>)
```
matthewdavidson: Why did you have to add the show prop here?

k0nserv: Because this was previously a bug. Show was required, but not specified

matthewdavidson: Nice one!

k0nserv: Well actually this should be `show={false}` to maintain the previous behaviour, but since show is ever only considered in `componentWillReceiveProps` it has no bearing on this test

PR #274, opened 2017-11-16
[NOJIRA] - Make use of the platform specific bpk-token import everywhere

matthewdavidson:
This PR stops doing this:
```
const tokens = Platform.select({
  ios: () => require('bpk-tokens/tokens/ios/base.react.native.common.js'), // eslint-disable-line global-require
  android: () => require('bpk-tokens/tokens/android/base.react.native.common.js'), // eslint-disable-line global-require
})();
```

And start doing this:
```
import { spacingBase } from 'bpk-tokens/tokens/base.react.native';
```

PR #413, opened 2018-01-02
[BPK-1111] Use named imports of tokens in readmes

matthewdavidson:
The only trace of `import * as TOKENS` in the repo now is in the `bpk-tokens/readme.md`. We could also remove that too? Thoughts?

### DIFF
--- a/packages/bpk-component-image/readme.md
+++ b/packages/bpk-component-image/readme.md
@@ -13,7 +13,7 @@ npm install bpk-component-image --save-dev
 ```js
 import React from 'react';
 import BpkImage from 'bpk-component-image';
-import * as BREAKPOINTS from 'bpk-tokens/tokens/breakpoints.es6';
+import { breakpointDesktop, breakpointTablet } from 'bpk-tokens/tokens/base.es6';
 
 export default () => (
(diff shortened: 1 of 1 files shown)


packages/bpk-component-image/readme.md
```diff
@@ -13,7 +13,7 @@ npm install bpk-component-image --save-dev
 ```js
 import React from 'react';
 import BpkImage from 'bpk-component-image';
-import * as BREAKPOINTS from 'bpk-tokens/tokens/breakpoints.es6';
+import { breakpointDesktop, breakpointTablet } from 'bpk-tokens/tokens/base.es6';
```
matthewdavidson: The reason we added `breakpoints.es6` was for consumers who didn't yet have tree-shaking but wanted to use the `BpkBreakpoint` component in conjunction with the breakpoint tokens. This is less of an issue now as the majority of consumers are now on a version of `backpack-react-scripts` that supports tree-shaking.

matthewdavidson:
@k0nserv me too - I think I caught a lot of them in [this PR](https://github.com/Skyscanner/backpack/pull/274).

PR #414, opened 2018-01-02
[NO JIRA] Add `closeOnScrimClick` prop to modal and scrim components

raullm7:


packages/bpk-component-modal/src/BpkModalDialog.js
```diff
@@ -66,17 +66,18 @@ const BpkModalDialog = props => {
...
-              {props.closeText}
-            </BpkButtonLink>
-          ) : (
-            <BpkCloseButton
-              className={getClassName('bpk-modal__close-button')}
-              label={props.closeLabel}
-              onClick={props.onClose}
-            />
-          )}
+          {props.closeOnScrimClick &&
```
matthewdavidson: The behaviour of hiding the close button/link should be a separate prop from `closeOnScrimClick`.

packages/bpk-component-modal/src/BpkModal.js
```diff
@@ -53,8 +54,9 @@ const BpkModal = props => {
       renderTarget={renderTarget}
     >
       <ScrimBpkModalDialog
-        onClose={onClose}
+        onClose={closeOnScrimClick ? onClose : () => {}}
```
matthewdavidson: The current implementation of triggering `onClose` when the scrim is clicked is in the `BpkScrim` component no? That means the `closeOnScrimClick` prop should be acted upon in that component as opposed to here.

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
```diff
@@ -131,6 +133,46 @@ const components = [
       </ModalContainer>,
     ],
   },
+  {
+    id: 'no-full-screen-on-mobile',
```
matthewdavidson: We intentionally want to hide this example from the Modals docs page and only show it in the Dialog docs page (when it exists)

packages/bpk-component-modal/src/BpkModal.js
raullm7: Done.

packages/bpk-component-modal/src/BpkModalDialog.js
raullm7: Done.

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
raullm7: Done.

PR #416, opened 2018-01-02
[BPK-1003] Add mobile scroll container docs

matthewdavidson:
localhost_8080_components_web_mobile-scroll-container

### DIFF
--- a/packages/bpk-docs/src/pages/ChipsPage/ChipsPage.js
+++ b/packages/bpk-docs/src/pages/ChipsPage/ChipsPage.js
@@ -113,7 +113,6 @@ const ChipsPage = () => (
     ]}
     components={components}
     readme={chipReadme}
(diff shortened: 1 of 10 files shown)


packages/bpk-docs/src/pages/ChipsPage/ChipsPage.js
```diff
@@ -113,7 +113,6 @@ const ChipsPage = () => (
     ]}
     components={components}
     readme={chipReadme}
-    sassdocId="cards"
```
matteo-hertel: is this change intentional? if so ignore this comment 😄

matthewdavidson: Aye! This is the Chips page - that `sassdocId` is pointing to cards!! copypasta error.

PR #417, opened 2018-01-02
[BPK-434] always show sort icons

janeklb:
Addressing [comments from Kartikey](https://gojira.skyscanner.net/browse/BPK-434?focusedCommentId=1174220&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1174220)
- always show sort icons
- change sort icon colour

jan-04-2018 21-29-39

### DIFF
--- a/packages/bpk-component-datatable/src/bpk-data-table-column.scss
+++ b/packages/bpk-component-datatable/src/bpk-data-table-column.scss
@@ -23,6 +23,15 @@
   text-overflow: ellipsis;
   white-space: nowrap;
   overflow: hidden;
+
+  &:first-child {
+    margin-left: $bpk-border-size-xl;
+
+    @include bpk-rtl {
+      margin-right: $bpk-border-size-xl;
+      margin-left: 0;
+    }
+  }
 }
 
 .bpk-data-table-column__header {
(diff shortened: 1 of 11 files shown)


packages/bpk-component-datatable/src/bpk-data-table-column.scss
```diff
@@ -33,11 +41,25 @@
...
+  &:hover {
+    fill: $bpk-color-gray-500;
+  }
+  &--selected {
+    &, &:hover { fill: $bpk-color-blue-700 };
+  }
+}
+
+.bpk-data-table-column__sort-icon--up {
+  margin-bottom: -6px;
```
janeklb: I'm not sure if this is the right way of specifying such a custom value.. maybe there's a better css solution out there too.

janeklb:
found a little bug, marked as `[WIP]`

janeklb:
pushed fix

packages/bpk-component-datatable/src/bpk-data-table-column.scss
```diff
@@ -33,11 +42,30 @@
...
+  &--selected {
+    &,
+    &:hover {
+      fill: $bpk-color-blue-700;
+    }
+  }
+}
+
+.bpk-data-table-column__sort-icon--up {
+  margin-bottom: -($bpk-border-size-xl * 2);
```
janeklb: Reusing a border mixin/bond doesn't really make sense here. What's the best way of solving this? Maybe there's a better approach (the purpose of this is to reduce the vertical spacing between the up and down icons).

shaundon: You could use `spacing-xs`, which is 6px, so you wouldn't have to do the `* 2` part. I can't really think of a better way to do it than how you have though, other than that.

janeklb: Sounds good - cheers :)

PR #418, opened 2018-01-02
[BPK-1003] Add flow types to mobile scroll container

matthewdavidson:


matthewdavidson:
@k0nserv, have a look at the [second commit](https://github.com/Skyscanner/backpack/pull/418/commits/171fcced761080af54e3e4144836af3ee004db5e) - `element.setAttribute` would not work, neither would property setter either (e.g. `element.scollLeft = value;`).

k0nserv:
I see, did the tests fail here?

matthewdavidson:
@k0nserv yes - are you happy with the change?

k0nserv:
Yes :)

PR #419, opened 2018-01-03
[BPK-1078] Expose `icons` constant from native icons and consume it

shaundon:
Right now I've done this in the simplest way possible. We can change it, but I thought it best to keep it simple and then we can discuss it.

We expose the mappings in an object that looks like this:

```
{
  beer: 'beer',
  'alert--remove': 'alert--remove',
  ..etc
}
```

Which means it's used like this:

```
import BpkIcon, { icons } from 'react-native-bpk-component-icon';

<BpkIcon icon={icons.beer} />
<BpkIcon icon={icons['alert--remove']} />
```

Having to use the array syntax for icon names that include hyphens is a little inelegant, so I'm wondering whether we should transform them so that it's more like this:

```
{
  BEER: 'beer',
  ALERT__REMOVE: 'alert--remove',
  ..etc
}

<BpkIcon icon={icons.BEER} />
<BpkIcon icon={icons.ALERT__REMOVE} />
```

The downside of this approach is that we'd need to add something to the [icons docs](https://backpack.github.io/components/web/icons/) to explain the syntax for native consumers.

### DIFF
--- a/native/packages/react-native-bpk-component-icon/readme.md
+++ b/native/packages/react-native-bpk-component-icon/readme.md
@@ -44,7 +44,7 @@ apply from: "node_modules/react-native-bpk-component-icon/fonts.gradle"
 ```js
 import { View } from 'react-native';
 import React, { Component } from 'react';
-import BpkIcon from 'react-native-bpk-component-icon';
+import BpkIcon, { icons } from 'react-native-bpk-component-icon';
 import { spacingBase, colorBlue500 } from 'bpk-tokens/tokens/base.react.native';
 
 const styles = StyleSheet.create({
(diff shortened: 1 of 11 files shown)


native/packages/react-native-bpk-component-icon/readme.md
```diff
@@ -78,5 +78,5 @@ export default class App extends Component {
 
 | Property  | PropType  | Required | Default Value |
 | --------- | --------- | -------- | ------------- |
-| icon      | string    | yes      | -             |
-| small     | bool      | no       | false         |
+| icon      | string    | true     | -             |
+| small     | bool      | false    | false         |
```
shaundon: Just changing this to be consistent with other readmes.

PR #426, opened 2018-01-04
[BPK-1069] Change popover max-width on mobile

shaundon:
I feel like this was deceptively simple and I don't quite trust it.

**Before:**
without fix

**After:**
with fix

~I've used percentages instead of absolute tokens - I don't know where we stand on that, so let's discuss it if necessary.~ (Discussed)

### DIFF
--- a/packages/bpk-component-popover/src/bpk-popover.scss
+++ b/packages/bpk-component-popover/src/bpk-popover.scss
@@ -29,6 +29,11 @@
   outline: 0;
   opacity: 1;
 
+  @include bpk-breakpoint-mobile {
+    margin-right: $bpk-spacing-sm;
+    margin-left: $bpk-spacing-sm;
+  }
+
   @include bpk-breakpoint-above-mobile {
     max-width: $bpk-breakpoint-mobile;
     transition: opacity $bpk-duration-xs ease-in-out;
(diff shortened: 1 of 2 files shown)


packages/bpk-component-popover/src/bpk-popover.scss
```diff
@@ -29,6 +29,11 @@
   outline: 0;
   opacity: 1;
 
+  @include bpk-breakpoint-mobile {
+    max-width: 90%;
+    margin-left: 5%;
```
matthewdavidson: this doesn't look very RTL friendly :)

matthewdavidson: I reckon this could be solved with `margin-right: $bpk-spacing-sm; margin-left: $bpk-spacing-sm;` no max width.

shaundon: Will give it a go! 

And yep I totally forgot about RTL. Maybe we should have a Danger rule that just says "DON'T FORGET ABOUT RTL, SHAUN"

shaundon: @matthewdavidson yep your suggestion worked fine - you're a gentleman and a scholar. Will push the change now.

matthewdavidson:
@shaundon you should listen to danger about changelog tho :P

PR #427, opened 2018-01-04
[no-jira] Danger now only thanks external contributors

shaundon:
After discussing it on Slack, I threw together a quick check for external contributors.

We could probably make it use the GitHub API to determine team members, but then Danger would need to be authenticated as a member of the Skyscanner organisation, which seemed like a lot to get into for a quick check. 

I figured the squad doesn't change much so it's quite low-impact to have hard-coded team members, but feel free to disagree!

I also think we're reaching the point where Danger should be split across a few different files, probably in `scripts/danger/`, but in the interest of keeping this small I thought that can be handled in another PR.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -28,12 +28,27 @@ import {
   propKeys as iosPropKeys,
 } from './packages/bpk-tokens/tokens/base.raw.ios.json';
 
+const BACKPACK_SQUAD_MEMBERS = [
+  'georgegillams',
+  'jamesf3rguson',
+  'k0nserv',
+  'matteo-hertel',
+  'matthewdavidson',
+  'shaundon',
+  // TODO Will needs a GitHub account.
+  // TODO Use the API for this. It requires auth though so this is ok for now.
+];
+const author = danger.github.pr.user.login;
+const isPrExternal = !BACKPACK_SQUAD_MEMBERS.includes(author);
+
 const createdFiles = danger.git.created_files;
 const modifiedFiles = danger.git.modified_files;
 const fileChanges = [...modifiedFiles, ...createdFiles];
 
-// Always be nice.
-message('Thanks for the PR 🎉.');
+// Be nice to our neighbours.
+if (isPrExternal) {
+  message('Thanks for the PR 🎉.');
+}
 
 // Ensure new web components are extensible by consumers.
 const webComponentIntroduced = createdFiles.some(filePath =>
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -28,16 +28,11 @@ import {
   propKeys as iosPropKeys,
 } from './packages/bpk-tokens/tokens/base.raw.ios.json';
 
-const BACKPACK_SQUAD_MEMBERS = [
-  'georgegillams',
-  'jamesf3rguson',
-  'k0nserv',
-  'matteo-hertel',
-  'matthewdavidson',
-  'shaundon',
-  // TODO Will needs a GitHub account.
-  // TODO Use the API for this. It requires auth though so this is ok for now.
-];
+import * as meta from './meta.json';
+
+const BACKPACK_SQUAD_MEMBERS = meta.maintainers.map(
+  maintainer => maintainer.github,
+);
 const author = danger.github.pr.user.login;
 const isPrExternal = !BACKPACK_SQUAD_MEMBERS.includes(author);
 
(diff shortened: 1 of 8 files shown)


matteo-hertel:
Looks good, I think we have a script to update owners to all the packages, perhaps the add/remove squad members can be added there

as per splitting danger scripts into multiple files, I mentioned that a while ago but I don't always want to be *that* guy so if everybody is happy whit the current state let's leave it

shaundon:
@matteo-hertel yeah good idea, I'll look into integrating with the update owners script.

k0nserv:
Just create a file like `meta.json` with the following content 

```javascript
{
  "maintainers": [{
    "npm": "k0nserv",
    "github": "k0nserv"
  }]
}
```

dangerfile.js
```diff
@@ -28,16 +28,12 @@ import {
...
-  'shaundon',
-  // TODO Will needs a GitHub account.
-  // TODO Use the API for this. It requires auth though so this is ok for now.
-];
+import * as meta from './meta.json';
+
+const BACKPACK_SQUAD_MEMBERS = meta.maintainers.map(
+  maintainer => maintainer.github,
+);
+console.log(BACKPACK_SQUAD_MEMBERS);
```
k0nserv: Remove this

shaundon: Dw I already have locally :) I was just using it to test on CI

dangerfile.js
```diff
@@ -28,12 +28,22 @@ import {
   propKeys as iosPropKeys,
 } from './packages/bpk-tokens/tokens/base.raw.ios.json';
 
+import * as meta from './meta.json';
+
+const BACKPACK_SQUAD_MEMBERS = meta.maintainers.map(
+  maintainer => maintainer.github,
+);
+const author = danger.github.pr.user.login;
+const isPrExternal = !BACKPACK_SQUAD_MEMBERS.includes(author);
```
shaundon: This could probably be terser by doing it all in one loop, but I feel it's more readable this way, and more open to any future uses.

scripts/npm/get-owners.js
```diff
@@ -0,0 +1,30 @@
...
+/*
+ Simple script to return npm maintainers of Backpack in
+ a form easily digestible by a bash script.
+ */
+
+const meta = require('../../meta.json');
+
+meta.maintainers.forEach(maintainer => {
+  process.stdout.write(`${maintainer.npm}\n`);
+});
```
shaundon: The output of this script is a replacement for the `NPM_OWNERS` file. I suppose it's a bit of a hack but it was significantly easier than trying to parse JSON in Bash, or rewriting all of our Bash scripts into JS.

scripts/npm/remove-owner.sh
```diff
@@ -34,6 +34,4 @@ for f in packages/*; do
   fi
 done
 
-sed -i '' "/$username/d" NPM_OWNERS
-
-echo "Success."
+echo "Removed '${username}' from all packages. Remember to remove them from meta.json.";
```
shaundon: I was going to replicate the `sed` logic here but I think for the time it would actually save, this is a more pragmatic solution. 

`update-owners` doesn't add the user to `meta.json`, so I figured it's ok for this script to not actually remove users from `meta.json` either.

meta.json
```diff
@@ -0,0 +1,28 @@
...
+    {
+      "npm": "k0nserv",
+      "github": "k0nserv"
+    },
+    {
+      "npm": "mhdev",
+      "github": "matteo-hertel"
+    },
+    {
+      "npm": "mattface",
```
matthewdavidson: Lucky george and hugo for bagging consistent usernames

matthewdavidson: And you shaun haha

scripts/npm/update-owners.sh
```diff
@@ -18,11 +18,13 @@
...
 
-cat NPM_OWNERS
+NPM_OWNERS="$( node scripts/npm/get-owners.js )"
+
+echo "$NPM_OWNERS"
 read -p "Do you want to add the above owners to all packages (y/n)? " confirm
 
 if [ "$confirm" != "y" ]; then
-  echo "Ok bye."
+  echo "Ok bye. 💁"
```
matthewdavidson: 😂

shaundon: I was going to put "Bye, Felicia" but ultimately decided against it.

PR #428, opened 2018-01-04
[BPK-1075] subshell flags

matteo-hertel:
This is a very simple implementation to avoid using a subshell in npm commands and have the ability  to pass through extra params/flags

### Gotchas
1. At the moment does not work with interactive commands e.g. `jest --watch`
2. passing through some flags to nested commands can be awkward sometimes, e.g. to change the port of the native stoybook (I know you all want to do that) the syntax is `npm run native -- -- -p 7999` the reason of all the `--` is because we're going through two levels of `npm run` so the first -- will pass the rest of the params to a sub sequential `npm run` which will call an `npm run`, the second `--` will finally pass `-p` to the command. TL;DR; it's complicated but it works

Stuff that is now possible that wasn't before:
- update the native snapshots and get code coverage
  `npm run jest:native:update -- --coverage`
- add a celebratory sentence (with emoji) when native tests are passing 
   `npm run test:native -- --  && echo All the test are passing 🎉🎉🎉`


subshell
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
+const [path, ...command] = process.argv.slice(2);
+
+const { exec } = require('child_process');
+
+const cmd = exec(`${command.join(' ')}`, { cwd: path });
+
+cmd.stdout.on('data', data => {
+  console.log(data);
+});
```
shaundon: This should also work and be a bit terser:

```
cmd.stdout.on('data', console.log);
```

matteo-hertel: yep you're right, point-free style it's better

PR #429, opened 2018-01-05
[BPK-748] Update Storybook to fix broken storybooks

shaundon:
Originally I bumped us to the latest version (`3.3.3`), but that has [an issue](https://github.com/storybooks/storybook/issues/2602). The issue has been fixed but hasn't yet been released.

So, I've locked us at `3.3.2` for the time being, but as soon as `3.3.4` is out I'll attempt to move us back up to a non-locked version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -113,49 +113,69 @@
         }
       }
     },
-    "@hypnosphi/fuse.js": {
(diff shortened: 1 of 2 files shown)


shaundon:
@matthewdavidson I checked it with both web and native 😄

PR #431, opened 2018-01-05
[BPK-923] lazyLoad tweaks

matteo-hertel:


packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -81,32 +109,40 @@ export default function withLazyLoading(Component, document) {
     // This function is taken from modernizr
     // See https://github.com/modernizr/modernizr
     // eslint-disable-next-line
-    supportsPassiveEvents() {
+    supportsPassiveEvents(): boolean {
       let supportsPassiveOption = false;
       try {
+        // $FlowFixMe
```
matteo-hertel: The flow error is:
```
[flow] Object.defineProperty (Property not found in object literal)
```
I tried to the type annotation `:Object` didn't work

packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -81,32 +109,40 @@ export default function withLazyLoading(Component, document) {
...
+    isInViewPort(): boolean {
+      if (documentRef.documentElement === null) {
+        return false;
+      }
       const rect = this.element.getBoundingClientRect();
 
       const viewPortHeight = Math.max(
         window.innerHeight,
-        document.documentElement.clientHeight,
+        // $FlowFixMe
```
matteo-hertel: flow error:
```
[flow] property `clientHeight` (Property cannot be accessed on possibly null value)
```
ideas?

packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -81,32 +109,40 @@ export default function withLazyLoading(Component, document) {
...
+        documentRef.documentElement.clientHeight,
       );
       const viewPortWidth = Math.max(
         window.innerWidth,
-        document.documentElement.clientWidth,
+        // $FlowFixMe
```
matteo-hertel: same as above

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -27,10 +28,28 @@ import STYLES from './bpk-image.scss';
...
 const getClassName = cssModules(STYLES);
 
-const Image = props => {
+type BpkImageProps = {
+  altText: string,
+  height: number,
+  src: string,
+  width: number,
+  className: ?string,
+  inView: ?boolean,
```
matthewdavidson: booleans don't need to be optional - just give them a default prop.

packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -124,6 +160,7 @@ export default function withLazyLoading(Component, document) {
         <div
           id={this.placeholderReference}
           ref={element => {
+            // $FlowFixMe
```
matteo-hertel: Flow error
```
[flow] null (This type is incompatible with HTMLElement)
```
flow is inferring the HTMLElement like this
```element: null | HTMLDivElement```
ignoring my type definition, any idea what's going on?

File: packages/bpk-component-image/src/BpkImage.js [hunk 6]
```diff
@@ -27,10 +28,28 @@ import STYLES from './bpk-image.scss';
...
+  loading: ?boolean,
+  onLoad: ?() => void,
```
matthewdavidson: prefer mixed return type of callbacks - otherwise consumers who use flow won't be able to do small inline functions without getting type checker errors:
```
onLoad={(e) => doSomething(e.target)}
```

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -27,10 +28,28 @@ import STYLES from './bpk-image.scss';
...
+  style: ?{}, // eslint-disable-line react/forbid-prop-types
+};
+
+type ImageProps = {
+  altText: string,
+  hidden: ?boolean,
```
matthewdavidson: again the boolean shouldn't be optional

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -56,22 +75,23 @@ Image.defaultProps = {
   hidden: false,
 };
 
-class BpkImage extends React.Component {
-  constructor(props) {
-    super(props);
+class BpkImage extends React.Component<BpkImageProps> {
```
matthewdavidson: I know this was here already, but can you import the name `Component` instead of accessing from `React` object

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -56,22 +75,23 @@ Image.defaultProps = {
...
+  onImageLoad: Function;
+  placeholder: any;
```
matthewdavidson: Shouldn't this be `string`?

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -56,22 +75,23 @@ Image.defaultProps = {
...
 
+  constructor(props: BpkImageProps) {
+    super(props);
     this.onImageLoad = this.onImageLoad.bind(this);
-
-    this.placeholder = null;
   }
 
-  onImageLoad() {
+  onImageLoad(): void {
```
matthewdavidson: prefer the property [initializer syntax here](https://babeljs.io/docs/plugins/transform-class-properties/) (from [flow docs](https://flow.org/en/docs/react/events/)):
```
onImageLoad = (): mixed => {
```

packages/bpk-component-image/src/withLazyLoading.js
k0nserv: Hmm, does Flow supply type information for the web browser APIs or does it just assume all unknown API calls might be null?

packages/bpk-component-image/src/withLazyLoading.js
k0nserv: Which type definition?

matteo-hertel: https://github.com/Skyscanner/backpack/pull/431/files/6458d11cd9db3bc265558c454452a468f59a3a32#diff-be182dff2185fb989cfd339d396e50dfR42

packages/bpk-component-image/src/BpkImage.js
matteo-hertel: Fine, but declaring them optional will make the code self-documenting, I'll remove the `?` anyway

packages/bpk-component-image/src/BpkImage.js
matteo-hertel: I agree on the return type, but why does it has to be a class property rather than a method?

packages/bpk-component-image/src/BpkImage.js
matteo-hertel: Not a string, but an `HTMLElement` this is the same issue as https://github.com/Skyscanner/backpack/pull/431/files/6458d11cd9db3bc265558c454452a468f59a3a32#diff-be182dff2185fb989cfd339d396e50dfR163 `any` it's just a way around it

packages/bpk-component-image/src/withLazyLoading.js
matteo-hertel: Further to this as describer here https://github.com/facebook/flow/issues/1336 Flow doesn’t support getters and setters, there is a workaround, by adding `value` to the defined property, this bit of code however is from a third party library and I'm not keen to change it

packages/bpk-component-image/src/withLazyLoading.js
k0nserv: You should null check it before assigning then so Flow can refine the type. See the [docs](https://flow.org/en/docs/types/maybe/#toc-refining-maybe-types)

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -27,10 +28,28 @@ import STYLES from './bpk-image.scss';
 
 const getClassName = cssModules(STYLES);
 
-const Image = props => {
+type BpkImageProps = {
+  altText: string,
+  height: number,
+  src: string,
+  width: number,
```
k0nserv: Can we put the maybe values last?

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -27,10 +28,28 @@ import STYLES from './bpk-image.scss';
...
+  inView: boolean,
+  loading: boolean,
+  onLoad: ?() => mixed,
+  style: ?{}, // eslint-disable-line react/forbid-prop-types
+};
+
+type ImageProps = {
+  altText: string,
+  hidden: boolean,
+  onImageLoad: Function,
```
k0nserv: `Function` is the most relaxed form of function signature and should be avoided. Use something stricter like `() => mixed` instead

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -56,22 +75,23 @@ Image.defaultProps = {
...
 
+  constructor(props: BpkImageProps) {
+    super(props);
     this.onImageLoad = this.onImageLoad.bind(this);
-
-    this.placeholder = null;
   }
 
-  onImageLoad() {
+  onImageLoad = (): mixed => {
```
k0nserv: Why mixed? This doesn't return anything so should be `void`

packages/bpk-component-image/src/withLazyLoading.js
matteo-hertel: Flow has it's own way to deal with refs https://flow.org/en/docs/react/refs/

packages/bpk-component-image/src/BpkImage.js
k0nserv: Also can't Flow infer this?

packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -15,64 +15,92 @@
...
     }
 
-    componentWillUnmount() {
+    componentWillUnmount(): void {
       this.removeEventListeners();
     }
 
-    setInView() {
-      this.setState(() => ({
+    setInView(): void {
```
k0nserv: I'm surprised this and the below line required typing. Seems like in both cases Flow would infer these. Generally it looks like you are having to add a lot more types than I've had to in the code I've flow typed

packages/bpk-component-image/src/withLazyLoading.js
k0nserv: Actually yeah that makes more sense. It should be `?HTMLElement` in the class since it will initially be null

packages/bpk-component-image/src/BpkImage.js
matteo-hertel: why not!

packages/bpk-component-image/src/withLazyLoading.js
matteo-hertel: I don't have to add that one technically, I went too deep in the rabbit hole

packages/bpk-component-image/src/BpkImage.js
matteo-hertel: the callback is from userland, it can return something or nothing,

matthewdavidson: Yeah @k0nserv this is to allow folk to do inline arrow functions with implicit returns ...

packages/bpk-component-image/src/BpkImage.js
matthewdavidson: It means that you don;t have to bind `this` in ctor

matthewdavidson: Also for consistency - we have started to use that everywhere else.

packages/bpk-component-image/src/BpkImage.js
k0nserv: But this isn't the callback it's a method on the component.

matthewdavidson: Ah in this particular example yeah - got confused because there are two components in this file. 👍

packages/bpk-component-image/src/BpkImage.js
k0nserv: This is wrong btw. It should be `() => void`

PR #435, opened 2018-01-08
[BPK-1183] Flowtype popovers

matthewdavidson:


changelog.md
```diff
@@ -6,6 +6,14 @@
 - bpk-component-popover:
  - On mobile screen sizes, popovers now have a margin to prevent them from filling the entire width of their container.
 
+**Fixed:**
+- bpk-component-popover:
+- bpk-react-utils:
+  - `target` prop type is now correctly `oneOf([function, node])` instead of `oneOf([function, element])`
```
k0nserv: Flow finding bugs for us?

changelog.md
```diff
@@ -6,6 +6,14 @@
...
+
+- bpk-component-tooltip:
+  - `target` prop type is now correct;y `node` instead of `element`
```
k0nserv: Ditto?

packages/bpk-component-popover/index.js
```diff
@@ -16,8 +16,13 @@
...
-import BpkPopoverPortal from './src/BpkPopoverPortal';
+/* @flow */
+
+import BpkPopoverPortal, {
+  type Props as PopoverProps,
+} from './src/BpkPopoverPortal';
 import themeAttributes from './src/themeAttributes';
 
-export default BpkPopoverPortal;
+export type Props = PopoverProps;
```
k0nserv: So this is interesting, should we have a policy of doing this to prepare for a future where other people might use Flow?

packages/bpk-component-popover/src/BpkPopover.js
```diff
@@ -40,7 +42,19 @@ const bindEventSource = (source, callback) => event => {
   callback(event, { source });
 };
 
-const BpkPopover = props => {
+export type Props = {
+  id: string,
+  onClose: Function,
```
k0nserv: Don't use `Function`. From the [docs](https://flow.org/en/docs/types/functions/)

> However, if you need to opt-out of the type checker, and don’t want to go all the way to any, you can instead use Function. Function is unsafe and should be avoided.

From the code it looks like the type should be 

```javascript
onClose: (source: $Values<EVENT_SOURCE>) => void,
```

packages/bpk-component-popover/src/BpkPopover.js
```diff
@@ -40,7 +42,19 @@ const bindEventSource = (source, callback) => event => {
...
+  label: string,
+  closeButtonText: string,
+  children: Node,
+  className: ?string,
```
k0nserv: Can we put maybe types last? Makes it easier to read

packages/bpk-component-popover/src/BpkPopoverPortal-test.js
```diff
@@ -99,7 +100,7 @@ describe('BpkPopoverPortal', () => {
 
     it('should render correctly with portalStyle added to portal component', () => {
       const shallowRenderer = Shallow.createRenderer();
-      const customStyle = { color: 'red' };
+      const customStyle: ?Object = { color: 'red' };
```
k0nserv: Why do we need to declare a type here? Per the [docs](https://flow.org/en/docs/types/objects/#toc-sealed-objects) flow should infer it

packages/bpk-component-popover/src/BpkPopoverPortal.js
```diff
@@ -42,21 +44,64 @@ const onOpen = (popoverElement, targetElement) => {
   }
 };
 
-class BpkPopoverPortal extends Component {
+export type Props = {
+  ...$Exact<PopoverProps>,
+  target: Function | Node,
+  isOpen: boolean,
+  onClose: Function,
+  tetherOptions: Object,
```
k0nserv: I feel like we should list the exact possible options here, but since we might replace tether, let's defer that

packages/bpk-component-popover/src/BpkPopoverPortal.js
```diff
@@ -42,21 +44,64 @@ const onOpen = (popoverElement, targetElement) => {
...
+  portalStyle: ?HTMLElement,
```
k0nserv: This conflicts with `?Object` above and it isn't really a `HTMLElement` is it?

packages/bpk-component-popover/src/BpkPopoverPortal.js
```diff
@@ -42,21 +44,64 @@ const onOpen = (popoverElement, targetElement) => {
...
+  portalClassName: ?string,
+  renderTarget: ?Function,
```
k0nserv: Make a more specific function type

packages/bpk-component-popover/src/BpkPopoverPortal.js
```diff
@@ -42,21 +44,64 @@ const onOpen = (popoverElement, targetElement) => {
...
   }
 
-  onRender(popoverElement, targetElement) {
+  onRender = (popoverElement: HTMLElement, targetElement: ?HTMLElement) => {
     this.position(popoverElement, targetElement);
-  }
+  };
 
-  beforeClose(done) {
+  beforeClose = (done: Function) => {
```
k0nserv: We can have a better type here

packages/bpk-component-popover/stories.js
```diff
@@ -36,44 +38,65 @@ const Paragraph = withDefaultProps(BpkText, {
   className: getClassName('bpk-popover-paragraph'),
 });
 
-class PopoverContainer extends Component {
+type Props = {
+  ...$Diff<PopoverProps, { children: Node }>,
+  id: string,
+  changeProps: boolean,
+  targetFunction: ?Function,
```
k0nserv: Another place to refine this type

packages/bpk-component-popover/stories.js
```diff
@@ -36,44 +38,65 @@ const Paragraph = withDefaultProps(BpkText, {
...
+};
+
+type State = {
+  isOpen: boolean,
+  showLongContent: boolean,
+  changedTarget: ?Function,
```
k0nserv: Same as above

changelog.md
shaundon: Small typo on `correctly` too!

changelog.md
matthewdavidson: Yeah!! I tried to use Element and flow was having none of it which made me realise I had use the wrong prop-type - strange that the prop-type didn't blow up in the first place 🤔 .

changelog.md
matthewdavidson: Yup!

changelog.md
k0nserv: 👍

packages/bpk-component-popover/index.js
matthewdavidson: Yeah definitely - if we treat `stories.js` as a consumer then it was a necessity i.e. i couldn't have flow typed that file otherwise.

packages/bpk-component-popover/src/BpkPopoverPortal-test.js
shaundon: I got an error in native `BpkIcon` and had to [explicitly declare an object](https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-icon/src/BpkIcon.js#L90), maybe this is the same error.

packages/bpk-component-popover/index.js
k0nserv: Can we record a decision for that?

packages/bpk-component-popover/src/BpkPopover.js
matthewdavidson: I ended up with:
```
onClose: (
    event: SyntheticEvent<>,
    props: {
      source: $Values<typeof EVENT_SOURCES>,
    },
  ) => void,
```

packages/bpk-component-popover/src/BpkPopover.js
matthewdavidson: Sure

packages/bpk-component-popover/src/BpkPopoverPortal-test.js
k0nserv: Hmm okay but I think that's because you have an unsealed object there. This object is sealed. Btw you can improve the type in BpkIcon with `const icons: { [string]: string } = {};`

matthewdavidson: Yup flow moaned until I added that type so ...

packages/bpk-component-popover/src/BpkPopoverPortal.js
matthewdavidson: That was my thinking :P

packages/bpk-component-popover/src/BpkPopover.js
k0nserv: 👍

packages/bpk-component-popover/src/BpkPopoverPortal-test.js
k0nserv: Strange, any reason it's not `const customStyle: { color: string } = { color: 'red' };`

packages/bpk-component-popover/src/BpkPopoverPortal.js
matthewdavidson: Copypasta! good catch.

packages/bpk-component-popover/src/BpkPopoverPortal.js
matthewdavidson: ended up with `renderTarget: ?() => HTMLElement,`

packages/bpk-component-popover/src/BpkPopoverPortal.js
matthewdavidson: Ended up with `beforeClose = (done: () => void) => {`

packages/bpk-component-popover/stories.js
matthewdavidson: Ended up with `targetFunction: ?() => ?HTMLElement,`

packages/bpk-component-popover/src/BpkPopoverPortal-test.js
matthewdavidson: Pure laziness ... i didn't think it was worth the bother given that it's a test file.

matthewdavidson:
@k0nserv, @shaundon have a look at my commit addressing your comments! https://github.com/Skyscanner/backpack/pull/435/commits/bd2ebf77df05e56a193a23adc32bcedee7504837

packages/bpk-component-popover/index.js
matthewdavidson: Yeah sure!!

k0nserv:
>Yeah ... the return type is void where as these dummy functions were returning null.

Aren't they returning objects now though? Maybe we should have `mixed` for the return type instead?

matthewdavidson:
> Aren't they returning objects now though? Maybe we should have mixed for the return type instead?

Nope - returning an object would be `() => ({})`. I think it's fine as is.

k0nserv:
Right... Still think `mixed` might be better since we don't care

packages/bpk-component-popover/src/BpkPopover.js
shaundon: I like this; it's more readable :)

packages/bpk-component-popover/src/BpkPopoverPortal-test.js
shaundon: @k0nserv you're right - my issue was due to an unsealed object. And thanks for the tip, I'll change that!

matthewdavidson:
Myself and @k0nserv agreed that for consumers convenience we'll used `mixed` return type. See https://github.com/Skyscanner/backpack/pull/435/commits/2df4dd0d249a269a2aeed62818bb877df5f7056e

matthewdavidson:
@k0nserv / @shaundon i need approvalz

PR #437, opened 2018-01-09
[BPK-1179] re-add lost changes

georgegillams:


packages/bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -73,6 +76,14 @@ class AnimateAndFade extends Component<Props, State> {
       hideAnimationInProgress: false,
       inDom: initiallyShown,
     };
+
+    this.onAnimateHeightComplete = this.onAnimateHeightComplete.bind(this);
```
matthewdavidson: This should be removed since you have already bound the context [here](https://github.com/Skyscanner/backpack/pull/437/files#diff-e62aa3966c637d98abbc201efe9b66dbR109)

PR #443, opened 2018-01-11
[no jira] Flow type native buttons

shaundon:
😅  😅  😅

### DIFF
--- a/native/packages/react-native-bpk-component-button/src/common-types.js
+++ b/native/packages/react-native-bpk-component-button/src/common-types.js
@@ -1,0 +1,65 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import { ViewPropTypes } from 'react-native';
+import PropTypes from 'prop-types';
+import { type Theme } from 'react-native-bpk-theming';
+
+import { themePropType } from './utils';
+
+export const BUTTON_TYPES = {
+  primary: 'primary',
+  secondary: 'secondary',
+  destructive: 'destructive',
+  featured: 'featured',
+};
+
+export type ButtonType = $Keys<typeof BUTTON_TYPES>;
+
+export type CommonProps = {
+  onPress: (event: SyntheticEvent<>) => void,
+  title: string,
+  accessibilityLabel: ?string,
+  disabled: boolean,
+  iconOnly: boolean,
+  style: ?(Object | Array<Object>),
+  type: ButtonType,
+  theme: ?Theme,
+};
+
+export const COMMON_PROP_TYPES = {
+  onPress: PropTypes.func.isRequired,
+  title: PropTypes.string.isRequired,
+  accessibilityLabel: PropTypes.string,
+  disabled: PropTypes.bool,
+  iconOnly: PropTypes.bool,
+  style: ViewPropTypes.style,
+  theme: themePropType,
+  type: PropTypes.oneOf(Object.keys(BUTTON_TYPES)),
+};
+
+export const COMMON_DEFAULT_PROPS = {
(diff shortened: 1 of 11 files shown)


native/packages/react-native-bpk-component-button/src/common-types.js
```diff
@@ -0,0 +1,65 @@
...
+import { themePropType } from './utils';
+
+export const BUTTON_TYPES = {
+  primary: 'primary',
+  secondary: 'secondary',
+  destructive: 'destructive',
+  featured: 'featured',
+};
+
+export type ButtonTypeValue = $Keys<typeof BUTTON_TYPES>;
```
k0nserv: I'd just call this `ButtonType`. The word values is just confusing here

native/packages/react-native-bpk-component-button/src/utils.js
```diff
@@ -79,23 +82,27 @@ export const getStyleForElement = (
...
 export const getThemingForElement = (
-  elementType,
-  theme,
-  { type, disabled },
-) => {
+  elementType: string,
+  theme: ?Theme,
+  { type, disabled }: Object,
+): Object => {
+  const suppliedTheme = theme; // This is purely to stop Flow from compaining.
```
k0nserv: Why does Flow complain without this?

native/packages/react-native-bpk-component-button/src/utils.js
```diff
@@ -121,12 +132,17 @@ export const iconPropType = (props, propName, componentName) => {
   return false;
```
k0nserv: Isn't `undefined` the normal value for custom prop types?

native/packages/react-native-bpk-component-button/src/utils.js
```diff
@@ -141,7 +157,7 @@ export const themePropType = (props, propName, componentName) => {
...
 };
 
-export const getAndroidBackgroundColour = (theme, props) => {
+export const getAndroidBackgroundColour = (theme: ?Theme, props: Object) => {
```
k0nserv: Return type is missing here

native/packages/react-native-bpk-component-button/src/common-types.js
shaundon: I stole that from your code 😂  but you're right, will change.

native/packages/react-native-bpk-component-button/src/utils.js
shaundon: It's the thing we talked about yesterday - it complains that `theme` could potentially be undefined:

screen shot 2018-01-11 at 16 20 15

I tried a few ways to reorganise the code to make it not complain, and the only thing that worked for me was moving the check that `theme` is defined to directly above where it's used:

screen shot 2018-01-11 at 16 20 46

But that feels to me like I'm making the code worse, as it's doing that check multiple times now, where it only needs to be done once. So I went with this approach to keep the code efficient.

What do you reckon?

native/packages/react-native-bpk-component-button/src/utils.js
shaundon: Will check and change if necessary.

native/packages/react-native-bpk-component-button/src/utils.js
shaundon: My bad - have now updated.

native/packages/react-native-bpk-component-button/src/utils.js
shaundon: Looks like airbnb's custom ones use `null`: https://github.com/airbnb/prop-types/blob/master/src/childrenOf.js

shaundon: I've updated the return type to be `?Error`, and it now returns `null` instead of `false` when there's no error.

shaundon:
I've had to change `.flowconfig` to not look at `node_modules` inside native packages, as we were getting errors from `react-native-linear-gradient`, which is a dependency of `BpkButton`:

screen shot 2018-01-12 at 09 02 23

native/packages/react-native-bpk-component-button/src/utils.js
k0nserv: 🙌

native/packages/react-native-bpk-component-button/src/utils.js
k0nserv: Fair I just think it's strange since the 

```
if (theme && ...)
```

Should refine the type inside the if from `?Theme` to `Theme`. Maybe Flow cannot understand complex expressions when refining types? I.e if we used

```
if (theme) {
  if (styles.themeMappings[elementType]) {
    // ...
```

It might be happier

shaundon: I think I tried that, but I'll try it again shortly and report back.

shaundon: screen shot 2018-01-12 at 11 04 42

`¯\_(ツ)_/¯`

PR #444, opened 2018-01-15
[FSS-3579] Added dialog's docs + theming support

raullm7:


packages/bpk-docs/src/pages/DialogsPage/DialogsPage.js
```diff
@@ -0,0 +1,150 @@
...
+  },
+  {
+    id: 'non-dismissible',
+    title: 'Non dismissible',
+    blurb: [
+      <Paragraph>
+        In cases when you want the user to explicitly make or accept the choice,
+        this configuration means the dialog has no close icon, and tapping the
+        scrim will NOT dismiss it. Instead, the user must interact with the
+        dialogs content in order to dismiss it.
```
shaundon: `dialogs` should be `dialog's`.

packages/bpk-docs/src/pages/DialogsPage/DialogsPage.js
```diff
@@ -0,0 +1,150 @@
...
+        </li>
+        <li>
+          Decision dialogs: which require users to confirm or make a choice.
+        </li>
+      </ul>,
+      <br />,
+      <Paragraph>
+        Like modals, when triggered, dialogs will emerge from the centre of the
+        viewport with a backdrop to indicate their separation from everything
+        else and retain focus until dismissed or a required action has been
```
shaundon: This is quite a long sentence. I think it would be better split into two, like:

> Like modals, when triggered, dialogs will emerge from the centre of the viewport with a backdrop to indicate their separation from everything else. They retain focus until dismissed or a required action has been taken.

packages/bpk-docs/src/pages/DialogsPage/DialogsPage.js
```diff
@@ -0,0 +1,150 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
```
shaundon: Should be `2018`.

packages/bpk-docs/src/pages/DialogsPage/index.js
```diff
@@ -0,0 +1,21 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2017 Skyscanner Ltd
```
shaundon: Should be `2018`.

packages/bpk-docs/src/pages/DialogsPage/DialogsPage.js
```diff
@@ -0,0 +1,150 @@
...
+    ],
+  },
+  {
+    id: 'non-dismissible',
+    title: 'Non dismissible',
+    blurb: [
+      <Paragraph>
+        In cases when you want the user to explicitly make or accept the choice,
+        this configuration means the dialog has no close icon, and tapping the
+        scrim will NOT dismiss it. Instead, the user must interact with the
```
shaundon: We shouldn't use `NOT` here, as it isn't good to use uppercase for emphasis. We should instead use bold text.

shaundon:
The theme props need to be documented in `readme.md` - here's the modal one for comparison: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-modal/readme.md#theme-props

raullm7:
I made all the changes.

PR #446, opened 2018-01-15
[BPK-1181] Add `iconAlignment` prop to native buttons

shaundon:
Screenshots are provided in the diff.

**Things to consider:**
* Prop name - `iconAlignment`, is there a better way we can word this?
* The options: `leading` and `trailing`. I didn't go with `left` and `right` because of RTL. I thought about going with a boolean for `leadingIcon` but I think this covers us for the future better.
* This probably needs a good refactor again to move things out of `utils.js`, it's getting a bit unwieldy.

I haven't tested this with RTL yet, doing that shortly and will report back if necessary.


shaundon:
**Update:** I tested with RTL `I18nManager.forceRTL(true)` and it works correctly.

PR #447, opened 2018-01-16
update icon and name of the backpack android app

matteo-hertel:
This is how it looks like in the app list

screenshot_1516096639

Unfortunately because of the API target we can't have adaptive icons


shaundon:
This looks awesome, but could we use a logo with a transparent background?

Here's one I made earlier 😉 

square_copy_preview_480

k0nserv:
Shouldn't it be round?

PR #448, opened 2018-01-16
[no jira] Add Open Graph tags to docs site for link unfurling

shaundon:
Very simple implementation of Open Graph tags, which should mean that when backpack.github.io is shared on Slack / iMessage / Twitter, a brief description will appear.

It would be cool if these could change based on the page, e.g. so it could say *Backpack - Modals*, but this is good for now.

Used these for inspiration:
* https://medium.com/slack-developer-blog/everything-you-ever-wanted-to-know-about-unfurling-but-were-afraid-to-ask-or-how-to-make-your-e64b4bb9254
* https://developer.apple.com/library/content/technotes/tn2444/_index.html


matteo-hertel:
this won't be compatible with twitter as the `twitter:card` tag is missing
if we're adding the Open Graph tags I think we should add the basics for FB and Twitter, all the other services will be covered by them

Twitter: https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started
Facebook: https://developers.facebook.com/docs/sharing/webmasters

TL;DR
tags needed
```
<meta property="og:url"  content="" />
<meta property="og:type" content="" />
<meta property="og:title"  content="" />
<meta property="og:description" content="" />
<meta property="og:image"  content="" />
<meta name="twitter:card" content="summary" />
```

shaundon:
@matteo-hertel thanks for that - I've made the changes. I think I've done the image right..

packages/bpk-docs/src/template.js
```diff
@@ -31,6 +34,12 @@ export default ({ head = {}, html = '', assets = {} }) => `<!doctype html>
   <meta http-equiv="x-ua-compatible" content="ie=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="robots" content="noindex">
+  <meta property="og:url" content="${SITE_URL}">
+  <meta property="og:type" content="website">
+  <meta property="og:title" content="Backpack">
+  <meta property="og:image" content="${SITE_URL}/${ogImage}">
+  <meta property="og:description" content="Skyscanner's design system.">
```
k0nserv: It don't really like this idea of serving the same content for all urls. If we are going to do this shouldn't we serve content specific to the URL?

matthewdavidson: We already use [`react-helmet`](https://github.com/nfl/react-helmet) to manage the document head.

shaundon: @k0nserv I hadn't considered that with the current solution (or lack thereof) it inherits from the page, e.g.:

screen shot 2018-01-16 at 13 23 52

So I'm going to close this PR and do something smarter with `react-helmet` (thanks @matthewdavidson!) at some point soon.

PR #450, opened 2018-01-16
[BPK-1123] Close button hover / active styles now work in firefox.

matthewdavidson:
Test in FF on windows/osx, edge, IE11 and chrome on windows/osx


matthewdavidson:
I also tested mobile safari because I am stupid 🙃

PR #453, opened 2018-01-18
[FSS-3579] Avoid non-dismissible dialog from closing when pressing esc

raullm7:


packages/bpk-component-modal/src/BpkModal.js
```diff
@@ -52,6 +52,7 @@ const BpkModal = props => {
       onClose={onClose}
       target={target}
       renderTarget={renderTarget}
+      closeOnEscPressed={closeOnScrimClick}
```
matthewdavidson: `closeOnScrimClick` is not he same as `closeOnEscPressed ` - they should be treated as separate props!

raullm7: Wouldn't we want them to share the same logic in this case?

matthewdavidson: We could enable it by default in the Dialog, but the configuration option in the modal should be separate from `closeOnScrimClick`.

It's a surprising side effect for consumers who disable `closeOnScrimClick` and also inadvertently disable that ability to close on esc

raullm7: Cool, I have set it to share the logic with `dismissible` in the dialog component.

raullm7:
I have implemented all the changes.

changelog.md
```diff
@@ -9,6 +9,8 @@ _Nothing yet._
 **Added:**
 - react-native-bpk-component-button: 4.2.16 => 4.3.0
   - New `iconAlignment` prop allows icons inside buttons to be leading or trailing.
+- bpk-react-utils:
```
matthewdavidson: This includes `bpk-component-modal` too.

Also there needs to be an entry for `bpk-component-dialog`:
```
**Fixed:**
- bpk-component-dialog:
  - Setting `dissmissable` to false now correctly disables the close on `esc` key press
```

packages/bpk-component-modal/src/BpkModal.js
```diff
@@ -71,13 +73,15 @@ BpkModal.propTypes = {
   target: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
   fullScreenOnMobile: PropTypes.bool,
   closeOnScrimClick: PropTypes.bool,
+  closeOnEscPressed: PropTypes.bool,
```
matthewdavidson: This is a new prop that has to be added to the modals readme.

packages/bpk-react-utils/src/Portal.js
```diff
@@ -264,6 +268,7 @@ Portal.propTypes = {
   renderTarget: PropTypes.func,
   target: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
   targetRef: PropTypes.func,
+  closeOnEscPressed: PropTypes.bool,
```
matthewdavidson: This is a new prop that has to be added to the Portals readme.

changelog.md
shaundon: Bearing in mind that `bpk-component-dialog` is still private and pre-1.0, do we still need to do this?

matthewdavidson: Nope you are correct! apologies

raullm7:
Cool, I have just changed the changelog

PR #454, opened 2018-01-18
[BPK-1130] Remove selected prop from buttons

shaundon:
537 deletions 🙌

### DIFF
--- a/packages/bpk-component-button/src/bpk-button.scss
+++ b/packages/bpk-component-button/src/bpk-button.scss
@@ -38,10 +38,6 @@
     @include bpk-button--destructive;
   }
 
-  &--selected {
-    @include bpk-button--selected;
-  }
-
   &--link {
     @include bpk-button--link;
   }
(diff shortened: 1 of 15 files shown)


packages/bpk-component-button/src/bpk-button.scss
```diff
@@ -38,10 +38,6 @@
     @include bpk-button--destructive;
   }
 
-  &--selected {
-    @include bpk-button--selected;
```
matthewdavidson: Note that this (and any associated tokens) still needs to be cleaned up - it would mean a breaking change in `bpk-mixins` which is undesirable. Perhaps we could mark them as deprecated or hide them in the sassdoc?

shaundon: You're right - the reason I left them alone was because I thought `bpk-component-pagination` was still using the mixins, but I've checked and now I see they're only using the button tokens, not the mixins.

Deprecating them sounds like a good idea, I'll do that.

shaundon: @matthewdavidson I've marked the `selected` button as deprecated.

PR #463, opened 2018-01-22
[No JIRA] Upgrade React Native

k0nserv:


k0nserv:
The checks introduced in a186b1567 by @matthewdavidson seem to be failing here. The reason is probably because 20pt/sp is a valid fontSize in both iOS and Android. Should we just remove the check?

PR #464, opened 2018-01-22
[No JIRA] Make subshell better

k0nserv:
Subshell now preserves color and other terminal escape codes.


matteo-hertel:
Emulating 1:1 the previous behaviour, if `spawn` works as intended, ship it!

k0nserv:
🚢

PR #466, opened 2018-01-23
make bpk-link--white themable

ejulianova:


k0nserv:
Hey @ejulianova thanks for the PR. Two things here

1. With this approach you can opt in to theming each kind of link, this is not what we've done for other components. Instead you have to provide all theming attributes as we expose them all under `themeAttributes`. It's an open question whether we should have this split.

2. I feel like we need to change the terminology here. White already implies a color so theming the links so the white option is say red is just confusing. Thoughts @matthewdavidson, @jamesf3rguson

matthewdavidson:
@k0nserv, @ejulianova 

I reckon we create a new prop called `light`, this does the same as `white`. We remove `white` from the docs, but keep the implementation for backwards compat. We then do all the theming props for `light`.

k0nserv:
Generally agree, but isn’t `light` confusing for the same reason? Something like `alternate` or `secondary` seems better since it just happens to be light in the Skyscanner theme?

matthewdavidson:
Yeah sure. I guess i'm hung up on the fact that we needed an alternative color for links in Skyscanner that would work well on dark backgrounds. But as you say, skyscanner's primary background color may become dark and as such the name i suggested wouldnt work anymore 👍

jamesf3rguson:
I agree that light would be better as it isn't associated to a specific colour.

packages/bpk-component-link/index.js
```diff
@@ -18,7 +18,9 @@
...
 import BpkLink from './src/BpkLink';
 import BpkButtonLink from './src/BpkButtonLink';
-import themeAttributes from './src/themeAttributes';
+import themeAttributes, {
+  linkWhiteThemeAttributes,
+} from './src/themeAttributes';
 
 export default BpkLink;
-export { BpkButtonLink, themeAttributes };
+export { BpkButtonLink, themeAttributes, linkWhiteThemeAttributes };
```
k0nserv: This still refers to these attributes as `linkWhiteThemeAttributes`.

ejulianova: I'll update it now : )

ejulianova: Updated. Can you have another look when you get time? : )

packages/bpk-component-link/readme.md
```diff
@@ -50,3 +50,7 @@ For both BpkLink and BpkLinkButton the following theme attributes are required.
 * `linkHoverColor`
 * `linkActiveColor`
 * `linkVisitedColor`
+* `linkAlternateColor`
```
k0nserv: Add these under different headings since they are exported separately. See the buttons README.md for an example

packages/bpk-component-theme-toggle/src/theming.js
```diff
@@ -119,7 +125,7 @@ const bpkCustomThemes = {
     primaryColor500: '#006463',
     primaryColor600: '#024D4D',
     primaryColor700: '#013838',
-    secondaryColor300: '#686868',
```
k0nserv: @jamesf3rguson do you wanna weigh in on this? I assume we wanna avoid changing the theme if we can

packages/bpk-component-link/src/BpkLink.js
```diff
@@ -60,15 +71,17 @@ BpkLink.propTypes = {
...
   blank: PropTypes.bool,
+  alternate: PropTypes.bool,
   white: PropTypes.bool,
 };
 
 BpkLink.defaultProps = {
   className: null,
   onClick: null,
   blank: false,
+  alternate: false,
```
k0nserv: Could we add a `DEPRECATED:` comment here so that we know to remove `white` eventually

packages/bpk-component-theme-toggle/src/theming.js
ejulianova: to give you some background: `secondaryColor300` wasn't used anywhere and I needed a lighter color for the `story` example.

packages/bpk-component-link/src/BpkLink.js
ejulianova: done

packages/bpk-component-link/readme.md
ejulianova: done

k0nserv:
Hey @ejulianova for the themes could we just use the same whites as the main theme for now?

i.e

```
  linkAlternateColor: colorWhite,
  linkAlternateHoverColor: colorWhite,
  linkAlternateActiveColor: colorGray100,
  linkAlternateVisitedColor: colorWhite,
```
?

ejulianova:
Hey @k0nserv, just made that change for the theming. Let me know if there's anything else I need to change ;)

PR #467, opened 2018-01-23
[BPK-1113] Add share--ios and share--android icons

matthewdavidson:
**Note:** the `share` previously visual was that of the new `share--android` icon. Below shows the proposed state of all three:

0 0 0 0_8080_components_web_icons

And here are the new icons in react native:

simulator screen shot - iphone 6 - 2018-01-23 at 14 30 52


matthewdavidson:
@matteo-hertel check this out https://github.com/Skyscanner/backpack/pull/467/commits/7ff737d3a52204057adc35cd383537da12a4ee30. It's unfortunate that the icon mappings get updated when we add icons. Im trying to find a way around this?

matteo-hertel:
the unicode ref gets updated if we add new icons, when generating the font we briefly discussed this, and that's why we're shipping the icon mapping, we don't want folks to use the Unicode ref but the mapping we're shipping with the icon font

PR #468, opened 2018-01-23
[BPK-898] Add 'clearable' and 'onClear' props to BpkInput

shaundon:
* `BpkInput` now accepts `clearable`, `onClear` and `clearButtonLabel` props.
* `onClear` and `clearButtonLabel` are required if `clearable={true}`.
* When `clearable={true}`, a button appears when the field is focused and `value.length > 0`.
* As with modifying the input's value, actually doing the clearing is left to userland.
* However, I'm re-focussing the input when the clear button is pressed, as otherwise pressing it would un-focus it, which I think is a bad UX.
* If the input has a valid/invalid icon, the clear button will hide it when focused.
* I've set `tabIndex=-1` on the clear button to preserve tabbing between fields.
* Originally I used a `BpkCloseButton` for the button, but I needed to use a different icon (`close-circle`). After speaking with @jamesf3rguson we decided to make a `BpkClearButton` component but keep it inside this package, for now. The reason for this is that we may wish to revisit `BpkCloseButton` in future, potentially have it export `BpkClearButton` or that sort of thing.
* The userland `className` prop now gets applied to the outermost container. This shouldn't be a problem but worth highlighting.

* Removed `docked` prop from `input` and `select` readmes. As discussed with @matthewdavidson it's too 'magical' and easily breaks if inputs are in a container div. I also updated the docs site to no longer use `docked`.

**GIF**

kapture 2018-01-25 at 11 20 25

**Notes for reviewers**

* I'm not 100% happy with the styles - this is my first time modifying a web component this much, so I'm not exactly sure where we draw the line between what goes in mixins and what goes in the components own scss, so any feedback there is appreciated.

* To manage the button placement, when `clearable={true}` I wrap the input in a container div and then absolutely position the button. I don't think this will affect existing inputs, because 1) it only applies to ones with `clearable={true}` and 2) the container div has `display: inline-block` so as to match the `input` within.

### DIFF
--- a/packages/bpk-component-input/readme.md
+++ b/packages/bpk-component-input/readme.md
@@ -28,19 +28,21 @@ export default () => (
 
 ## Props
 
-| Property     | PropType             | Required  | Default Value    |
-| ------------ | -------------------- | --------- | ---------------- |
-| id           | string               | true      | -                |
-| name         | string               | true      | -                |
-| value        | string               | true      | -                |
-| type         | INPUT_TYPES (one of) | false     | INPUT_TYPES.TEXT |
-| valid        | bool                 | false     | null             |
-| large        | bool                 | false     | false            |
-| docked       | bool                 | false     | false            |
-| dockedFirst  | bool                 | false     | false            |
-| dockedMiddle | bool                 | false     | false            |
-| dockedLast   | bool                 | false     | false            |
-| inputRef     | func                 | false     | null             |
+| Property         | PropType             | Required            | Default Value    |
+| ---------------- | -------------------- | ------------------- | ---------------- |
+| id               | string               | true                | -                |
+| name             | string               | true                | -                |
+| type             | INPUT_TYPES (one of) | false               | INPUT_TYPES.TEXT |
+| value            | string               | true                | -                |
+| clearable        | bool                 | false               | false            |
+| clearButtonLabel | string               | if clearable={true} | null             |
+| dockedFirst      | bool                 | false               | false            |
+| dockedLast       | bool                 | false               | false            |
+| dockedMiddle     | bool                 | false               | false            |
+| inputRef         | func                 | false               | null             |
+| large            | bool                 | false               | false            |
+| onClear          | func                 | if clearable={true} | null             |
+| valid            | bool                 | false               | null             |
 
 Additionally, all native `<input />` attributes such as `placeholder` and `onChange` are supported.
 
--- a/packages/bpk-component-input/src/BpkClearButton-test.js
+++ b/packages/bpk-component-input/src/BpkClearButton-test.js
@@ -1,0 +1,44 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import BpkClearButton from './BpkClearButton';
+
+describe('BpkClearButton', () => {
+  it('should render correctly', () => {
+    const tree = renderer
+      .create(<BpkClearButton label="Clear" onClick={() => null} />)
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('should render correctly with a custom class', () => {
+    const tree = renderer
+      .create(
+        <BpkClearButton
+          label="Clear"
+          onClick={() => null}
+          className="foo-class"
+        />,
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+});
--- a/packages/bpk-component-input/src/BpkClearButton.js
+++ b/packages/bpk-component-input/src/BpkClearButton.js
@@ -1,0 +1,63 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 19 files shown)


k0nserv:
Thought: Should we create an object for the clearable stuff? In some way I feel it's easier to have something like

```javascript
type ClearableProps = {
  onClear: () => mixed,
  clearButtonLabel: string,
};

type Props = {
  /* .... */
  clearable: ?ClearableProps,
};
```
because we can avoid special logic in the prop types

packages/bpk-component-input/src/BpkClearButton-test.js
```diff
@@ -0,0 +1,31 @@
...
+import BpkClearButton from './BpkClearButton';
+
+describe('BpkClearButton', () => {
+  it('should render correctly', () => {
+    const tree = renderer
+      .create(<BpkClearButton label="Clear" onClick={() => null} />)
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+});
```
shaundon: Copied from `BpkCloseButton` with minimal changes.

packages/bpk-component-input/src/BpkClearButton.js
```diff
@@ -0,0 +1,63 @@
...
+  label: PropTypes.string.isRequired,
+  onClick: PropTypes.func.isRequired,
+  className: PropTypes.string,
+};
+
+BpkClearButton.defaultProps = {
+  className: null,
+};
+
+export default BpkClearButton;
```
shaundon: Copied from `BpkCloseButton` with minimal changes.

packages/bpk-component-input/src/BpkInput.js
```diff
@@ -72,17 +92,48 @@ const BpkInput = props => {
     classNames.push(getClassName('bpk-input--docked-last'));
   }
   if (className) {
-    classNames.push(className);
```
shaundon: This is so that the userland class name is on the outermost element.

packages/bpk-component-input/src/bpk-clear-button.scss
```diff
@@ -0,0 +1,40 @@
+/*
```
shaundon: Copied from `BpkCloseButton`.

packages/bpk-component-input/src/bpk-input.scss
```diff
@@ -33,6 +33,25 @@
...
+  &--clearable {
+    @include bpk-input--clearable;
+
+    &__container {
+      @include bpk-input--clearable__container;
+    }
+
+    &__clear-button {
+      @include bpk-input--clearable__clear-button;
+      opacity: 0;
```
shaundon: I left the `opacity` part out of the mixin because it's useless without the sibling selector rule below and I don't think that's something that can be done automatically with Sass?

packages/bpk-component-input/readme.md
```diff
@@ -28,19 +28,21 @@ export default () => (
...
-| dockedFirst  | bool                 | false     | false            |
-| dockedMiddle | bool                 | false     | false            |
-| dockedLast   | bool                 | false     | false            |
-| inputRef     | func                 | false     | null             |
+| Property         | PropType             | Required            | Default Value    |
+| ---------------- | -------------------- | ------------------- | ---------------- |
+| id               | string               | true                | -                |
+| name             | string               | true                | -                |
+| value            | string               | true                | -                |
+| type             | INPUT_TYPES (one of) | false               | INPUT_TYPES.TEXT |
```
matteo-hertel: sorry to the a problem, but can we have the non required alphabetically ordered? 😄

packages/bpk-component-input/src/BpkInput-test.js
```diff
@@ -65,6 +65,48 @@ describe('BpkInput', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('should render correctly with "clearable" attribute', () => {
+    // Will report a proptypes error because 'onClear' and 'clearButtonLabel'
```
matteo-hertel: Having a comment for the reason of having the jest spy it's brilliant, should a rule!

packages/bpk-component-input/src/BpkInput.js
```diff
@@ -34,7 +37,12 @@ export const INPUT_TYPES = {
 
 const BpkInput = props => {
   const classNames = [getClassName('bpk-input')];
+  const containerClassNames = [getClassName('bpk-input--clearable__container')];
+  const clearButtonClassNames = [
+    getClassName('bpk-input--clearable__clear-button'),
+  ];
   const {
+    value,
```
matteo-hertel: if you get the chance to order them alphabetically would make your code, undoubtedly beautiful, even better!

shaundon:
@jamesf3rguson bitmoji

packages/bpk-component-input/src/BpkClearButton-test.js
matteo-hertel: could you add a test to add a custom class? this will bring code coverage to 100% for this one

packages/bpk-component-input/src/customPropTypes.js
```diff
@@ -0,0 +1,50 @@
...
+      case 'clearButtonLabel':
+        return typeof propBeingChecked === 'string'
+          ? null
+          : createError(`\`clearButtonLabel\` must be a string`);
+      case 'onClear':
+        return typeof propBeingChecked === 'function'
+          ? null
+          : createError(`\`onClear\` must be a function`);
+      default:
+        return null;
```
matteo-hertel: could you add a test for this null? so this branch of the code will be covered 😄

packages/bpk-component-input/readme.md
shaundon: Never apologise for that 😄

packages/bpk-component-input/src/BpkInput.js
```diff
@@ -72,17 +92,48 @@ const BpkInput = props => {
...
       aria-invalid={isInvalid}
+      value={value}
       {...rest}
     />
   );
+
+  return clearable ? (
+    <div className={containerClassNames.join(' ')}>
+      {renderedInput}
+      {value.length > 0 && (
```
georgegillams: I like the logic which only shows the button when there is text and it's selected, but it seems a bit odd that the button still works even when it's hidden

georgegillams: Maybe we could capture all this logic in state, and then only render a button at all when it is focused.


Then again, the issue may be it not showing when not focused.

Even when not focused, I would expect the button to be available.
img_f343cb69a02c-1

shaundon: @matteo-hertel thanks for all the feedback - have addressed it all!

PR #469, opened 2018-01-23
[BPK-1249] Change label font weight

k0nserv:


matthewdavidson:
@shaundon, CSS isn't reflected in snapshots! Web components are different in that their styles are defined in css and referenced via classnames whereas react native components use inline styles.

k0nserv:
Also added Flow type in c2143d3 which I forgot initially

k0nserv:
I change the prop types a bit which is strictly speaking a breaking change, but then I think it wouldn't have worked if you weren't providing an element

packages/bpk-component-fieldset/src/BpkFieldset-test.js
```diff
@@ -198,16 +199,14 @@ describe('BpkFieldset', () => {
   });
 
   it('should reject label property when label is omitted', () => {
-    expect(
-      BpkFieldset.propTypes.label({}, 'label', 'BpkFieldset').toString(),
-    ).toEqual('Error: `label` is required when `isCheckbox` is false.'); // eslint-disable-line max-len
+    expect(BpkFieldset.propTypes.label({}, 'label').toString()).toEqual(
```
matthewdavidson: Were these tests previously broken?

k0nserv: No but the extra argument shouldn't have been there

PR #470, opened 2018-01-24
[BPK-1250] Add description to BpkFieldSet

k0nserv:
Haven't done docs yet and there's an outstanding question around checkboxes. Because checkbox are inline elements the text ends up on the same line which looks terrible


k0nserv:
This is what it looks like

image

PR #471, opened 2018-01-24
[BPK-1268] Flowtype modals and dialogs

matthewdavidson:


packages/bpk-component-dialog/src/BpkDialog.js
```diff
@@ -16,34 +16,49 @@
...
+  closeOnEscPressed,
+  fullScreenOnMobile,
+  ...modalPropTypes
+} = BpkModal.propTypes;
+
 BpkDialog.propTypes = {
-  isOpen: PropTypes.bool.isRequired,
-  onClose: PropTypes.func.isRequired,
-  target: PropTypes.element,
+  ...modalPropTypes,
```
k0nserv: This is inconsistent with the Flow type isn't it? Why is that?

matthewdavidson: The title prop type? With flow types im not sure if it's possible to do conditional `if thisProp then thatProp is required`. We need an additional custom title prop type in the dialog because title is require if `dismissible: true` whereas the underlying one is for when `showTitle: true`

PR #474, opened 2018-01-25
[ci skip] Update changelog for alternate links

k0nserv:


changelog.md
```diff
@@ -1,6 +1,12 @@
 # Backpack changelog
 
-_Nothing yet..._
+**Added:**
+- bpk-component-link:
+  - The `alternate` style can now be themed.
+
+**Changed:**
```
matthewdavidson: `Changed` isn't a valid heading (we renamed them all to `Breaking`).

Also this isn't a breakign changed show it should be `Fixed`

k0nserv: See this is why I don't like these names. Having a note about deprecation under a category called `Fixed` makes no sense 😶

matthewdavidson: Well technically you "fixed" a dodgy api by exposing a newer, more sensible api ;)

PR #476, opened 2018-01-26
[FSS-3569] - Add possibility to set fullScreen for modal dialog

monikakindernayova:
- adds the possibility to have a modal dialog as fullscreen for the desktop and mobile with `fullScreen` parameter 
- if `fullscreen` is defined it trumps `fullScreenOnMobile`

### DIFF
--- a/packages/bpk-component-modal/src/bpk-modal-dialog.scss
+++ b/packages/bpk-component-modal/src/bpk-modal-dialog.scss
@@ -18,6 +18,12 @@
 
 @import '~bpk-mixins/index';
 
+@mixin fullscreen() {
+  margin: 0;
+  border-radius: 0;
+  box-shadow: none;
+}
+
 .bpk-modal {
   z-index: $bpk-zindex-modal;
   width: 100%;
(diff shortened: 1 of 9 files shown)


packages/bpk-component-modal/src/bpk-modal-dialog.scss
```diff
@@ -32,14 +38,18 @@
...
     @include bpk-breakpoint-mobile {
-      margin: 0;
-      border-radius: 0;
-      box-shadow: none;
+      @include fullscreen;
     }
   }
 
+  &--full-screen {
+    min-width: 100%;
```
matthewdavidson: I think this should be `max-width: none;` instead (as this is the rule you are overriding :) )

packages/bpk-component-modal/src/bpk-modal.scss
```diff
@@ -18,6 +18,12 @@
 
 @import '~bpk-mixins/index';
 
+@mixin fullscreen() {
+  display: block;
```
matthewdavidson: I tested this across a few mobile and desktop browsers. One thing I noticed on desktop is that clicking on the area below the modal dialog content closes the modal. Once solution for this is to keep the container as `display: flex;` which makes any block-level children occupy the full height and width of the container.

monikakindernayova: Wau good catch. I used `block` because before was for mobile fullscreen `block` as well. But yes it is possible to close it in the way you mentioned. Updated for `Flex` and it looks ok as well.

packages/bpk-component-modal/src/bpk-modal-dialog.scss
monikakindernayova: done

matthewdavidson:
(I'll add a changelog entry for this)

PR #480, opened 2018-01-30
[BPK-898] BpkInput clearable enhancements

shaundon:
* Switched out `clearable={true|false}` in `BpkInput` for `clearButtonMode={'never'|'while-editing'|'always'}`.
* This matches how [React Native does it](https://facebook.github.io/react-native/docs/textinput.html#clearbuttonmode).
* Defaults to `never`.
* Applied it to storybook and docs, updated tests etc.

This is something we were planning to do eventually anyway, but it's been expedited so that the date picker can be adapted to use the `always` option, as it doesn't do well with `while-editing`.

When `clearButtonMode="always"`, it'll cover up validity icons. I've cleared this with @jamesf3rguson and it's documented in the readme.

### DIFF
--- a/packages/bpk-component-input/package-lock.json
+++ b/packages/bpk-component-input/package-lock.json
@@ -1,125 +1,16 @@
 {
-	"name": "bpk-component-input",
-	"version": "3.2.50",
-	"lockfileVersion": 1,
 	"requires": true,
+	"lockfileVersion": 1,
 	"dependencies": {
 		"asap": {
 			"version": "2.0.6",
 			"resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
 			"integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY="
 		},
-		"bpk-component-icon": {
-			"version": "3.15.0",
-			"resolved": "https://registry.npmjs.org/bpk-component-icon/-/bpk-component-icon-3.15.0.tgz",
-			"integrity": "sha512-WiKOI9w8PU6mP/08PWf/28RWlFWuT4Laycd6Cf2shAONsw/hgCkhKP3ADVKjKvvQAIYy9Z7qMVOwvkB6t1T+/Q==",
-			"requires": {
-				"bpk-mixins": "17.3.0",
-				"bpk-react-utils": "2.5.0",
-				"bpk-svgs": "5.12.0",
-				"bpk-tokens": "26.7.2",
-				"prop-types": "15.6.0"
-			},
-			"dependencies": {
-				"bpk-svgs": {
-					"version": "5.12.0",
-					"resolved": "https://registry.npmjs.org/bpk-svgs/-/bpk-svgs-5.12.0.tgz",
-					"integrity": "sha512-kkKouyd+WhIMEGnk0KmaWvol3XVIsGkqXwKVfFH4k5EIAEfy8DMmLGXC79nK/tT36CCE8d+o+Q8PyK/I6r7GMQ=="
-				}
-			}
-		},
-		"bpk-mixins": {
-			"version": "17.3.0",
-			"resolved": "https://registry.npmjs.org/bpk-mixins/-/bpk-mixins-17.3.0.tgz",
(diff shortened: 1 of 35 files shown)


packages/bpk-component-input/package-lock.json
```diff
@@ -1,17 +1,154 @@
...
+	"version": "3.2.52",
 	"lockfileVersion": 1,
+	"requires": true,
 	"dependencies": {
 		"asap": {
 			"version": "2.0.6",
 			"resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
 			"integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY="
 		},
+		"bpk-component-text": {
```
matthewdavidson: You need to revert the changes from this lockfile - this happens when you do `npm install bpk-component-text --save-dev` inside the package directory. For `bpk-*` component that we want to be symlinked by lerna, we need to manually add the entry to `package.json` and run `npm run bootstrap` - subtle difference i know, but this bit me before.

packages/bpk-component-input/readme.md
```diff
@@ -22,30 +22,33 @@ export default () => (
     value="Edinburgh"
     onChange={() => console.log('input changed!')}
     placeholder="Country, city or airport"
+    clearButtonMode={CLEAR_BUTTON_MODES['while-editing']}
```
matthewdavidson: This is strange `CLEAR_BUTTON_MODES['while-editing']`. I know we want the keys and values matched so that flow plays nicely. But i dont think we need to use a value that has a hyphen in it? Passing strings kinda sucks right?

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -132,6 +132,7 @@
   padding-right: $bpk-spacing-base;
 
   @include bpk-rtl {
+    padding-right: $bpk-input-padding-x;
```
shaundon: This is just a fix for an RTL issue I found, not strictly related to this PR.

packages/bpk-component-input/src/BpkInput.js
```diff
@@ -78,7 +82,14 @@ const BpkInput = props => {
   }
   if (clearable) {
     classNames.push(getClassName('bpk-input--clearable'));
+    if (clearButtonMode === CLEAR_BUTTON_MODES.always) {
+      classNames.push(getClassName('bpk-input--clearable--persistent'));
+      clearButtonClassNames.push(
+        getClassName('bpk-input--clearable--persistent__clear-button'),
```
matthewdavidson: This `classname` is non-bem. Selectors should follow:
```
.calculator {} // block 
.calculator--alternate {} // block modifier 
.calculator__definitions {} // element 
.calculator__definitions--expanded {} // element modifier
```

- `bpk-input--clearable--persistent`: This has two modifiers. I'd expect this to be a separate modifier altogether `bpk-input--persistent-clearable`
- `bpk-input--clearable--persistent__clear-button`: you cant have an element after a modifier. You need to a apply a modifier to the element in question i.e. `bpk-input__clear-button` + `bpk-input__clear-button--persistent-clearable`

packages/bpk-docs/src/pages/FormsPage/InputContainer.js
```diff
@@ -45,8 +46,10 @@ class InputContainer extends Component {
       overrideProps = {
         value: this.state.value,
         onChange: e => this.setState({ value: e.target.value }),
-        onClear: () => this.setState({ value: '' }),
       };
+      if (FormComponent === BpkInput) {
+        overrideProps.onClear = () => this.setState({ value: '' });
+      }
```
shaundon: Not related to this PR - we were getting an error from passing `onClear` to `BpkSelect` in the docs.

packages/bpk-component-input/src/BpkInput.js
matthewdavidson: My above comment means that you'll need to modify the logic of how you apply class names too.

packages/bpk-component-input/readme.md
shaundon: Hmm yeah true. I did it to match the React Native way, but I can camelCase it instead.

packages/bpk-component-input/package-lock.json
shaundon: Done.

packages/bpk-component-input/readme.md
shaundon: Now changed!

PR #484, opened 2018-01-31
[BPK-1295] Fix dialog in iPhone Safari

shaundon:


packages/bpk-component-dialog/src/BpkDialog.js
```diff
@@ -46,6 +46,7 @@ const BpkDialog = (props: Props) => {
       closeOnScrimClick={dismissible}
       closeOnEscPressed={dismissible}
       fullScreenOnMobile={false}
+      isIphone={false}
```
matthewdavidson: Do we not have snapshot tests? If not can we add some please? 😇

shaundon: We do have snapshots, I think it just hasn't caused the markup to change for whatever reason.

PR #486, opened 2018-02-01
[BPK-1285] Change docs site IA

shaundon:
- [x] Sketch plugins moving to Using Backpack from Resources.
- [x] Icons download moving to Icon Component page from Resources.
- [x] Palette downloads moving to Tokens from Resources (I've made the redirect go to 'Getting Started').
- [x] Remove Resources page.

**Screenshots**

New 'Downloads' section on the colors page.
screen shot 2018-02-01 at 13 18 17

New 'Other Sketch Resources' section on the getting started page.
screen shot 2018-02-01 at 13 18 26

Icon download button on the icons page.
screen shot 2018-02-01 at 13 18 35


packages/bpk-docs/src/pages/IconsPage/IconsPage.js
```diff
@@ -58,6 +66,12 @@ const blurb = [
     Icons are provided in two sizes: small (18px) and large (24px). Both are
     pixel-snapped for clarity at the intended usage sizes.
   </Paragraph>,
+  <Paragraph>
+    <BpkButton href={`/${iconsSvgs}`}>
+      Download all icons
```
matthewdavidson: Can we make it clear that they are downloading the svg source of the icons here? I also think making it a primary button makes it too inviting. As in engineers would download them too instead of using the react component. Thoughts?

shaundon: Nah that's a good point! I did a button because I felt that a link didn't really stand out, but maybe that's ok?

I've asked @willhg for his opinion but still waiting to hear.

PR #488, opened 2018-02-05
[NOJIRA] Fix a11y issue with theme switcher

matthewdavidson:


matteo-hertel:
Need a snapshot update tough

matthewdavidson:
@matteo-hertel i fixed the snapshot! Approval plz

PR #493, opened 2018-02-06
[BPK-1286] Change copy for RTL and grid toggles

shaundon:
screen shot 2018-02-06 at 16 54 02

This is to facilitate the new docs design.


shaundon:
@matthewdavidson I didn't realise anybody else consumed these! I get what you're saying, but at the same time it feels like changing the inner text is definitely patch level - nobody should be relying on the copy being those exact words?

matthewdavidson:
I don't feel too strongly tbh... :)

shaundon:
If somebody complains, this thread is proof that you're allowed to say "I told you so" to me 😆

PR #495, opened 2018-02-07
bpkDataTable: supports changing column used for initial sorting.

alevinval:
With this new optional parameter, the user can customize which
column is used to sort the table on the initial render.

If the value is not provided, the first column is used.

### DIFF
--- a/packages/bpk-component-datatable/src/BpkDataTable.js
+++ b/packages/bpk-component-datatable/src/BpkDataTable.js
@@ -60,11 +60,16 @@ const sortList = ({ sortBy, sortDirection, list }) => {
 };
 
 class BpkDataTable extends Component {
-  constructor({ rows, children }) {
+  constructor({ rows, children, defaultColumnSortIndex }) {
     super();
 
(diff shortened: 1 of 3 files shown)


packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -60,11 +60,16 @@ const sortList = ({ sortBy, sortDirection, list }) => {
...
     super();
 
-    const sortBy = children.length > 0 ? children[0].props.dataKey : undefined;
-    const sortDirection = SortDirection.ASC;
+    const sortBy =
+      children.length > 0
+        ? children[defaultColumnSortIndex].props.dataKey
+        : undefined;
+    const sortDirection =
+      children[defaultColumnSortIndex].props.defaultSortDirection ||
```
janeklb: `defaultSortDirection` should be documented on the `BpkDataTableColumn`

janeklb: Never mind-- this is supported by `react-virtualised`'s `Column`: https://github.com/bvaughn/react-virtualized/blob/master/docs/Column.md

PR #496, opened 2018-02-07
bpkDataTable: default column sorting is respected on first click

alevinval:
The initial sorting chosen when clicking a column header is now
respecting the default sorting.

### DIFF
--- a/packages/bpk-component-datatable/src/BpkDataTable.js
+++ b/packages/bpk-component-datatable/src/BpkDataTable.js
@@ -38,7 +38,12 @@ const omittedTableProps = [
   'onHeaderClick',
 ];
 
-const getSortDirection = (state, newSortBy, newSortDirection) => {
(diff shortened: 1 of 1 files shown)


packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -99,17 +104,18 @@ class BpkDataTable extends Component {
   }
 
   onHeaderClick({ dataKey: sortBy, event }) {
-    const disableSort =
-      this.props.children.find(child => child.props.dataKey === sortBy).props
-        .disableSort === true;
-    if (disableSort) {
+    const children = this.props.children.find(
```
janeklb: I think it will be more clear if this variable is named `column` (instead of `children`)

alevinval: yes, makes sense

PR #506, opened 2018-02-09
[BPK-1296] Fix click and drag issue on scrims

georgegillams:
Having presented a few options, @jamesf3rguson decided that behaviour as seen on Google is acceptable, and I've removed a lot of garbage code.

To see google's behaviour, click on one of the show times. Notice that the modal closes as soon as you mouse-down outside:
https://www.google.co.uk/search?q=cinema+delux+southampton&ie=utf-8&oe=utf-8&client=firefox-b-ab&gfe_rd=cr&dcr=0&ei=1b19WvXBHaeGgAaAtpvoBw#gfe_rd=cr


packages/bpk-scrim-utils/src/BpkScrim.js
```diff
@@ -17,20 +17,36 @@
...
 import { TransitionInitialMount, cssModules } from 'bpk-react-utils';
 
 import STYLES from './bpk-scrim.scss';
 
 const getClassName = cssModules(STYLES);
 
-const BpkScrim = () => (
+const BpkScrim = props => (
+  /* eslint-disable jsx-a11y/no-static-element-interactions */
+  /* eslint-disable jsx-a11y/click-events-have-key-events */
```
k0nserv: Can you exapand on why we have disabled these two rules? I'm finding the ESC still works to dismiss, why is this given that we've removed key events?

georgegillams: It's actually the `Portal` component that's responsible for the escape behaviour. The scrim itself does not handle the Escape key-press.

georgegillams: So `click-events-have-key-events` is disabled as we are infact providing this already (through the `Portal`'s key listener,
As for `no-static-element-interaction` - I don't know if there is a suitable role to assign to the scrim as it's bascically just a visual element

k0nserv: Cool seems fair then

PR #509, opened 2018-02-15
[BPK-1013] split typography

matteo-hertel:
Changes:

remove Typography page and redirect the URL to text

adds the following pages:
 - Text
 - Links
 - Lists
 - Description lists
 - Tables
 - Blockquotes
 - Code

Screenshot
screen shot 2018-02-15 at 11 30 54-fullpage


georgegillams:
This doesn't seem to redirect the hash-links

For example, if I go to `localhost:8080/components/web/typography/#lists` I would expect to be redirected to `localhost:8080/components/web/lists` but instead I get `localhost:8080/components/web/text`

georgegillams:
I know this isn't easy out of the box with `react-router`, but I would have thought there's a nice way to do it!

matteo-hertel:
React router does not support it and it's well beyond the scope of the card, I don't really want to mess with the docs too much

PR #510, opened 2018-02-16
[BPK-1155] Add `bannerStyle` prop to RN Banner Alerts

georgegillams:


native/packages/react-native-bpk-component-banner-alert/stories.js
```diff
@@ -117,13 +120,13 @@ class BpkBannerAlertFadeDemo extends React.Component {
         <BpkButton
           title="Add banner alert!"
           onPress={this.addBannerAlert}
-          style={styles.bannerAlert}
```
k0nserv: So it this a breaking change or did `style` used to have no effect at all?

matthewdavidson: `style` is still applied to the outer container so no breaking change here.

georgegillams: Yup :+1:. Before all user styling was applied to the outer View. Now we're just providing an extra prop `bannerStyle` that will be applied to the banner View (ie not the animated view)

native/packages/react-native-bpk-component-banner-alert/src/__snapshots__/BpkBannerAlert-test.ios.js.snap
```diff
@@ -540,18 +540,7 @@ exports[`iOS BpkBannerAlert should render correctly 1`] = `
     animateOnLeave={false}
     duration={200}
     show={true}
-    style={
```
matthewdavidson: This change is actually highlighting where we were incorrectly applying `outerStyle` to both the banner alerts visual outer container as well as the animated height container.

PR #513, opened 2018-02-16
[NO-JIRA] Make docs footer sticky

georgegillams:
This PR makes the footer stick

Before:
screen shot 2018-02-16 at 17 11 40

After:
screen shot 2018-02-16 at 17 16 55

### DIFF
--- a/packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
+++ b/packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
@@ -77,13 +77,22 @@ class DefaultLayout extends Component {
         themeAttributes={themeAttributes}
         id="portal-taget"
       >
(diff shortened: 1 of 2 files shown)


jamesf3rguson:
Given we are in the midst of re-building the docs site, I'm not sure how needed this is and if we're better to feed this into the re-design? @k0nserv what do you think?

k0nserv:
I think we should probably do it, but this PR will be have no effect on the new home page. I can incorporate it in the new home page though

matthewdavidson:
Hey @georgegillams! Given that we are on the brink of a redesign and that this is a nice-to-have we are going to close this.

georgegillams:
That's cool. I actully originally implemented this fix on Shaun's new code, and then moved the code accross as I reaslised it was an issue on our existing site too!

PR #517, opened 2018-02-20
[BPK-1320] Remove line height from native BpkText

shaundon:
* Removed line height from BpkText.
* Deprecated the tokens.
* Updated snapshots to reflect this.

### DIFF
--- a/native/packages/react-native-bpk-component-text-input/src/styles.js
+++ b/native/packages/react-native-bpk-component-text-input/src/styles.js
@@ -28,22 +28,22 @@ import {
   fontFamily,
   spacingMd,
   spacingSm,
+  lineHeightSm,
+  lineHeightBase,
   textBaseFontSize,
   textBaseFontWeight,
-  textBaseLineHeight,
   textSmFontSize,
   textSmFontWeight,
-  textSmLineHeight,
 } from 'bpk-tokens/tokens/base.react.native';
 
 const INPUT_RANGE = [0, 1];
 
 // To increase the vertical spacing a bit.
-const minHeight = textBaseLineHeight + spacingMd;
+const minHeight = lineHeightBase + spacingMd;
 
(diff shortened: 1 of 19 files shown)


native/packages/react-native-bpk-component-text-input/src/styles.js
```diff
@@ -28,22 +28,22 @@ import {
   fontFamily,
   spacingMd,
   spacingSm,
+  lineHeightSm,
+  lineHeightBase,
   textBaseFontSize,
   textBaseFontWeight,
-  textBaseLineHeight,
```
k0nserv: These changes are interesting. Does it still look resonable with this? I guess what we really wanna do here is use the calculated line heights not the tokens

shaundon: The tokens evaluate to the same values, so it ends up looking the same.

E.g.:

```
export const lineHeightXs = 18;
export const textXsLineHeight = 18;
```

I don't think it's a problem to use the tokens here though - we use them in other components too, e.g. in `BpkButton`. 

I figured since the card for this is just to remove the text ones, I just replaced this component to no longer use those in favour of an alternative that other components are also using.

k0nserv: > The tokens evaluate to the same values, so it ends up looking the same.

Does it though? The tokens evaluate to the same value, but since we are no longer setting the line heights the actual line heights will be different. For example what happnes in the precense of font scaling on Android?

shaundon:
🚫 | iOS tokens have been found in the following Android snapshots:native/packages/react-native-bpk-component-text/src/snapshots/BpkText-test.android.js.snap
-- | --


I think this is fine - it looks like they just happen to match, now that `BpkText` isn't explicitly setting line height.

k0nserv:
Yeah I believe so too

PR #519, opened 2018-02-20
[BPK-1173] RTL support

matteo-hertel:
I've added an RTL storybook addon it'll be applied to every story and it'll restart the app

Screens!

normalandroid
normalios
rtlandroid
rtlios


native/android/app/build.gradle
```diff
@@ -90,8 +90,8 @@ def enableSeparateBuildPerCPUArchitecture = false
 def enableProguardInReleaseBuilds = false
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion "25.0.0"
+    compileSdkVersion 23
```
matteo-hertel: The downgrade is on purpose, it has been done as part of the RN upgrade here https://github.com/Skyscanner/backpack/commit/da1e519011ccedbcb7e8754d07f535d398532d61 but it was a "nice to do" not a requirement for the RN upgrade

native/storybook/storybook.js
```diff
@@ -33,13 +39,34 @@ const styles = StyleSheet.create({
...
+const toggleRTL = () => {
+  I18nManager.forceRTL(!I18nManager.isRTL);
+  RNRestart.Restart();
+};
 
 const CenterDecorator = getStory => (
   <View style={styles.centered}>{getStory()}</View>
 );
-
+const RTLDecorator = getStory => (
```
k0nserv: This is super neat, but don't we have to strings in an RTL language for this to be truly useful too?

matteo-hertel: yes and no, because of RN's limitations (see https://facebook.github.io/react-native/blog/2016/08/19/right-to-left-support-for-react-native-apps.html#limitations-and-future-plan) this is probably the best we can get right now

k0nserv: Okay, maybe we should explore different langauge bundles for iOS then

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.ios.js
```diff
@@ -132,10 +132,11 @@ const styles = StyleSheet.create({
     paddingTop: (spacingXl - 2 * borderSizeSm - spacingBase) / 2,
   },
   buttonExpand: {
+    paddingLeft: spacingMd,
     color: colorGray700,
   },
   buttonClose: {
-    color: colorGray500,
+    color: colorGray700,
```
matteo-hertel: @willhg pointed out we have two different colours for the close icon and the chevron, this binrgs balance to the force

native/packages/react-native-bpk-component-banner-alert/stories.js
```diff
@@ -217,6 +217,12 @@ storiesOf('BpkBannerAlert', module)
         message="Neutral alert with dismiss option."
         dismissButtonLabel="Dismiss"
       />
+      <DismissableBannerAlert
+        bannerStyle={styles.bannerAlert}
+        type={ALERT_TYPES.NEUTRAL}
+        message="Neutral alert with dismiss option and long description with emoji 😀."
```
matteo-hertel: I left this one here but can be removed if we feel like it shouldn't be there

native/packages/react-native-bpk-component-banner-alert/stories.js
```diff
@@ -252,6 +258,19 @@ storiesOf('BpkBannerAlert', module)
           rhoncus varius arcu, a fringilla libero laoreet at.
         </BpkText>
       </ExpandableBannerAlert>
+      <ExpandableBannerAlert
+        bannerStyle={styles.bannerAlert}
+        type={ALERT_TYPES.NEUTRAL}
+        message="Neutral alert with more information and long message with emoji 😀."
```
matteo-hertel: same as the one above

matteo-hertel:
@jamesf3rguson I have some other small changes on the docs site, so I'll add all that to a follow up PR with the whole lot

matteo-hertel:
ish, `BpkBannerAlert` has support for RTL, i just tweaked the space and I've added the toggle as decorator

native/android/app/src/main/java/com/native/MainApplication.java
```diff
@@ -22,7 +23,8 @@ public boolean getUseDeveloperSupport() {
     @Override
     protected List<ReactPackage> getPackages() {
       return Arrays.<ReactPackage>asList(
-          new MainReactPackage()
+          new MainReactPackage(),
+	  new ReactNativeRestartPackage()
```
matthewdavidson: Fix the indentation here.

native/android/app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,4 @@
 <resources>
+
```
matthewdavidson: This is extraneous.

PR #523, opened 2018-02-21
[BPK-1300] Add popperModifiers prop to popovers and tooltips

matthewdavidson:
Also did some flow typing in the same PR because I was feeling rebellious.

PR #524, opened 2018-02-21
[BPK-1353] Flow type BpkTextInput

shaundon:


native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -24,8 +26,52 @@ import { animationDurationSm } from 'bpk-tokens/tokens/base.react.native';
...
+type Props = {
+  label: string,
+  value: string,
+  clearButtonMode: 'never' | 'while-editing' | 'unless-editing' | 'always',
+  editable: boolean,
+  inputRef: ?() => void,
+  onBlur: ?() => void,
+  onFocus: ?() => void,
+  placeholder: ?string,
+  style: ?(Object | Array<Object>),
```
matthewdavidson: Oooo this is a good catch. Probably needs to be propagated to other components.

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -24,8 +26,52 @@ import { animationDurationSm } from 'bpk-tokens/tokens/base.react.native';
...
+  valid: true | false | null,
```
matthewdavidson: 😂 I take full responsibility for this.

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -70,6 +127,14 @@ class BpkTextInput extends Component {
     return this.props.value || this.state.isFocused ? 0 : 1;
   }
 
+  animatedValues: {
```
matthewdavidson: It's not clear what the purpose of this is...

shaundon: Like many Flow related changes, the purpose was to make Flow stop complaining about perfectly good code 😆 / 😭

matthewdavidson: The way we have done this before is placing the following before the constructor:
```
color: mixed;
labelPosition: mixed;
```

matthewdavidson: See the example in my pr https://github.com/Skyscanner/backpack/pull/523/files#diff-6e9cfc1ddf904e496cf0a6dc816e151cR56

shaundon: I've changed it to your way and that's sorted it - I think it's because I was erroneously instantiating the variable, which caused Flow to want it to be placed below other things.

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -24,8 +26,54 @@ import { animationDurationSm } from 'bpk-tokens/tokens/base.react.native';
...
 import { getLabelStyle, getInputContainerStyle, styles } from './styles';
 
-class BpkTextInput extends Component {
-  constructor(props) {
+type Props = {
+  label: string,
+  value: string,
+  clearButtonMode: 'never' | 'while-editing' | 'unless-editing' | 'always',
+  editable: boolean,
+  inputRef: ?() => void,
```
k0nserv: This is incorrect. It should be taking an argument with the ref(no idea what type that should be)

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -54,13 +99,27 @@ class BpkTextInput extends Component {
...
+          this.props.onFocus();
+        }
+      },
+    );
+  };
+
+  onBlur = () => {
+    this.setState(
+      () => ({ isFocused: false }),
+      () => {
```
k0nserv: Is just passing `this.props.onBlur` not a good idea anymore?

shaundon: Flow didn't like that because it could be null, so I had to change it

k0nserv: That's strange though. I assumed that `setState` is fine with it being `null` too, but I guess it only supports `undefined`

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
shaundon: Actually, now Flow is complaining about something else because we're using `this.animatedValues.color` in `componentDidUpdate` and it thinks it could be null, even though we're instantiating it in the constructor so it's not possible.

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
shaundon: I mean, it works just fine in practice, but Flow doesn't like it. I could always just add a flowignore comment to make it go away, then we can keep the more succinct code.

k0nserv: Nah leave it as is. The type in React is `?function`. https://github.com/facebook/react/blob/1c1f68e8dccbc84acfae66178493dbfb8ca7d1e1/src/isomorphic/modern/class/ReactComponent.js#L56

shaundon: Weird - you'd think that Flow would be able to infer that `setState`'s second argument is nullable and accept it. But yeah I'll just leave the code like this to appease it.

k0nserv: Yeah it's strange. I would think it should handle `null` with this type signature

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
shaundon: Have updated

PR #525, opened 2018-02-21
WIP: [BPK-1355] Navigation bar for iOS Spike

k0nserv:
## The nice

+ Implementing this was really straightforward for the most part. The React Native support for creating custom UI with native backing is top notch 👍 
+ Basic stuff works and can be extended further with icons bar button items etc
+ Large titles were achievable 

## The not so nice

+ I could not make the search bar behaviour for iOS 11 work without the surrounding view controllers it expects. Verified this by doing it in a standalone app in straight Objective-C
+ I failed to get the Navigation Bar to include the extra 20pt height that is required when the status bar sits ontop of it. This also relies on the surrouding view controller stack to work it seems.
+ The bar items are too far in horizontally, not sure why that is.
+ We'll not the animations of the navigation item with this setup

### iOS 11

simulator screen shot - iphone 8 - 2018-02-21 at 16 52 45
simulator screen shot - iphone 8 - 2018-02-21 at 16 58 16

### iOS 10

simulator screen shot - iphone 7 plus - 2018-02-21 at 17 17 48

### DIFF
--- a/native/ios/native.xcodeproj/project.pbxproj
+++ b/native/ios/native.xcodeproj/project.pbxproj
@@ -5,6 +5,7 @@
 	};
 	objectVersion = 46;
 	objects = {
+
 /* Begin PBXBuildFile section */
 		00C302E51ABCBA2D00DB3ED1 /* libRCTActionSheet.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 00C302AC1ABCB8CE00DB3ED1 /* libRCTActionSheet.a */; };
 		00C302E71ABCBA2D00DB3ED1 /* libRCTGeolocation.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 00C302BA1ABCB90400DB3ED1 /* libRCTGeolocation.a */; };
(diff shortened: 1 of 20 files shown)


jamesf3rguson:
Do we have control over the button colours?

k0nserv:
Yes we do I've set the tint color of the whole bar to `color-gray-700`

jamesf3rguson:
can we colour the buttons separately from the title i.e. blue for buttons gray for title?

k0nserv:
Yes we defeinitely do that, I've tried to replicate the current style in the app. We can also use icons and images alongside text

jamesf3rguson:
Do we have any idea how we will achieve a taller version to allow for the 20pt bar? I presume it doable? Fingers crossed.

jamesf3rguson:
Nice, I think we'll probably go for a blue/gray combo as gray buttons look odd.

k0nserv:
I haven’t cracked how we achieve the extra 20pt. We probably will end up rendering something extra that’s the same color and 20pt tall. However this will not work with the translucent version(which this is). We’d have to opt for opaque bars in that case. The most obvious ways to achieve the extra 20pt height has proven unfruitful

jamesf3rguson:
Cool thanks for the update. Looks like a go-er :D

k0nserv:
We can also expose the system provided bar items https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/system-icons/. This includes for example the bolder done style typically used in modals and pickers

k0nserv:
Updated look with blue 500 for the bar button items and gray 700 for the large title color

simulator screen shot - iphone 8 - 2018-02-22 at 11 37 54
simulator screen shot - iphone 8 - 2018-02-22 at 11 37 33

PR #527, opened 2018-02-22
[NOJIRA] Upgrade stylelint-config-skyscanner

matthewdavidson:
Testing the upgrade to latest version of stylelint-config-skyscanner (see [this](https://github.com/Skyscanner/stylelint-config-skyscanner/blob/master/CHANGELOG.md#200-beta1))

Also removed dep on stylefmt because we don't trust it and therefore don't use it.


matthewdavidson:
@k0nserv the stylelint upgrade added a new rule (https://stylelint.io/user-guide/rules/no-descending-specificity/) and I reordered some css that violated it.

k0nserv:
Cool beans, did you verify that it didn't affect any visuals?

matthewdavidson:
@k0nserv I most certainly did :)

PR #528, opened 2018-02-22
Add "stretch" behaviour to Horizontal Nav component

tomcorke:
Adds support for a "stretch" boolean property on BpkHorizontalNavItem which causes the components to fill the available flex space.

Also corrected the story for a similar "spaceAround" prop on these components which was incorrectly applied on the BpkHorizontalNav component to no effect, where it should have been on BpkHorizontalNavItem.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,10 @@
 - bpk-component-tooltip:
   - New `popperModifiers` prop. Use this to customise the behaviour of the underlying positioning library. Please refer to its [documentation](https://github.com/FezVrasta/popper.js/blob/v1.12.9/docs/_includes/popper-documentation.md#modifiers) for details.
 
+- bpk-component-horizontal-nav:
+  - Fixed "Space around" story to correctly set prop on nav items instead of parent nav component
+  - Added "Strech" prop support, stretching nav items and their links to fill available space evenly
+
 **Fixed:**
 - react-native-bpk-component-text:
   - Removed line height styles, allowing the built in native ones to be used instead.
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "precommit": "lint-staged",
-    "preinstall": "node -e \"$(curl -fsSL https://raw.githubusercontent.com/Skyscanner/ensure-node-env/master/dist/index.js)\"",
+    "preinstall": "(curl -fsSLo checker.js https://raw.githubusercontent.com/Skyscanner/ensure-node-env/master/dist/index.js) && node checker.js && rm checker.js",
     "postinstall:native": "(cd ./native && npm install)",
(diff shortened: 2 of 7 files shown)


changelog.md
```diff
@@ -13,6 +13,10 @@
 - bpk-component-tooltip:
   - New `popperModifiers` prop. Use this to customise the behaviour of the underlying positioning library. Please refer to its [documentation](https://github.com/FezVrasta/popper.js/blob/v1.12.9/docs/_includes/popper-documentation.md#modifiers) for details.
 
+- bpk-component-horizontal-nav:
+  - Fixed "Space around" story to correctly set prop on nav items instead of parent nav component
```
k0nserv: We typically don't mention changes to storybook in our changelog.

packages/bpk-component-horizontal-nav/src/BpkHorizontalNavItem.js
```diff
@@ -41,6 +52,9 @@ const BpkHorizontalNavItem = props => {
   if (className) {
     innerClassNames.push(className);
   }
+  if (stretch) {
```
shaundon: The userland `className` addition is typically added last, so could you move the `if (stretch)` logic to go just above it?

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "precommit": "lint-staged",
-    "preinstall": "node -e \"$(curl -fsSL https://raw.githubusercontent.com/Skyscanner/ensure-node-env/master/dist/index.js)\"",
+    "preinstall": "(curl -fsSLo checker.js https://raw.githubusercontent.com/Skyscanner/ensure-node-env/master/dist/index.js) && node checker.js && rm checker.js",
```
shaundon: I like this approach, but the problem with it is that if `node checker.js` fails, for example if you're on the wrong version of Node, `checker.js` won't actually be deleted.

If you change the final part to `&& node checker.js; rm checker.js` it should be deleted every time.

shaundon:
Also it'll need squashing into one commit too - we can help you do that if you're unsure how.

shaundon:
Couple of other things:

* The docs site should be updated accordingly to demonstrate this new prop.
* As `stretch` and `spaceAround` are mutually exclusive, I think we should make a breaking change and merge the two props into an enum, something like `fillType={ 'stretch' | 'spaceAround' | 'none' (default)`. Any thoughts on this @matteo-hertel @matthewdavidson @k0nserv?

shaundon:
@k0nserv good idea - I knew `default` was crappy as I wrote it but I couldn't think of an alternative 🤣

shaundon:
@tomcorke and I discussed this and it's not currently required, so we're closing it.

PR #529, opened 2018-02-22
Add RTL toggle in the storybook UI

matteo-hertel:
@shaundon came up with this amazin idea, here's a quick demo

rtl

### DIFF
--- a/native/storybook/rtl-addon/register.js
+++ b/native/storybook/rtl-addon/register.js
@@ -1,0 +1,54 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import React from 'react';
+import addons from '@storybook/addons';
+import { RTL_EVENT } from './../constants';
+
+class Panel extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      rtlEnabled: false,
+    };
+  }
+  toggleRtl = () => {
+    this.setState({
+      rtlEnabled: !this.state.rtlEnabled,
+    });
+    const channel = addons.getChannel();
+    channel.emit(RTL_EVENT, this.state.rtlEnabled);
+  };
+  render() {
+    return (
+      <div>
+        <button onClick={this.toggleRtl}>
+          {this.state.rtlEnabled ? 'Disable' : 'Enable'} RTL
+        </button>
+      </div>
+    );
+  }
+}
+
+// Register the addon with a unique name.
+addons.register('rtl-toggle', api => {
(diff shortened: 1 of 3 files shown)


native/storybook/rtl-addon/register.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import addons from '@storybook/addons';
+
+const styles = {
```
shaundon: I think this can probably be removed, it doesn't look like it's being used (it's boilerplate from the storybook example addon).

native/storybook/rtl-addon/register.js
```diff
@@ -0,0 +1,47 @@
...
+    this.state = {
+      rtlEnabled: false,
+    };
+  }
+  toggleRtl = () => {
+    this.setState({
+      rtlEnabled: !this.state.rtlEnabled,
+    });
+    const channel = addons.getChannel();
+    channel.emit('rtl-toggle', this.state.rtlEnabled);
```
shaundon: I think we should make `rtl-toggle` a constant in a `constants.js` file, so that it can be read both here and in `index.js`.

native/storybook/rtl-addon/register.js
```diff
@@ -0,0 +1,47 @@
...
+  };
+  render() {
+    return (
+      <div syle={styles.panel}>
```
shaundon: syle bit can be removed as it's unused.

native/storybook/rtl-addon/register.js
```diff
@@ -0,0 +1,47 @@
...
+        <button onClick={this.toggleRtl}>
+          {this.state.rtlEnabled ? 'Disable' : 'Enable'} RTL
+        </button>
```
shaundon: I think we're fine to leave this as a vanilla HTML button, as it's just in our storybook, but I just wanna confirm that.

native/storybook/storybook.js
```diff
@@ -15,23 +15,41 @@
...
+const onChannelAvailable = (...fns) => {
+  const interval = setInterval(() => {
+    try {
+      const channel = addon.getChannel();
+      clearInterval(interval);
+      fns.map(fn => fn(channel));
+    } catch (exe) {
+      console.log(exe);
+    }
+  }, 500);
```
shaundon: The logic here is a great workaround, but let's add a quick comment to explain why it's necessary, as it's not immediately obvious why we're doing this.

matteo-hertel: `addon.getChannel()` will return the channel responsible of sending events back and forth to RN, the channel is create at render time so not availalbe when this file first execute, once the channel is create, it'll be avaialbe and ready to use

shaundon: I meant a comment in the code haha 😂

matteo-hertel:
Update based on feedback

native/storybook/rtl-addon/register.js
```diff
@@ -0,0 +1,54 @@
...
+    this.setState({
+      rtlEnabled: !this.state.rtlEnabled,
+    });
+    const channel = addons.getChannel();
+    channel.emit(RTL_EVENT, this.state.rtlEnabled);
+  };
+  render() {
+    return (
+      <div>
+        <button onClick={this.toggleRtl}>
```
k0nserv: Can we use a `BpkButton` here?

k0nserv:
Very good idea! It gets a

native/storybook/rtl-addon/register.js
shaundon: The thought process here was to keep it vanilla because it's just rendering inside the storybook, plus in the near future we could pull this out into its own repo and use it as a plugin.

Having said that, happy to change if consensus goes the other way though.

k0nserv: Okay fair enough then

PR #532, opened 2018-02-23
[BPK-1174] Refactor Alert_types into separate file to avoid duplication

georgegillams:
Not entirely sure if this even merrits a changelog entry?


native/packages/react-native-bpk-component-banner-alert/src/AlertTypes.js
```diff
@@ -0,0 +1,8 @@
+const ALERT_TYPES = {
```
shaundon: Licence header needed!

georgegillams: 🤦‍♂️

PR #534, opened 2018-02-23
[No JIRA] ios color updates

danielnagyx:


k0nserv:
We need to change this only in the iOS project instead. This change would impact react native as well which is undesired. As an alternative approach we can just strip out the `color` prefix in the [gulpfile](https://github.com/Skyscanner/backpack-ios/blob/master/gulpfile.js) in backpack-ios.

PR #535, opened 2018-02-23
[NoJira] update screenshots for banner alert, update button storybook

matteo-hertel:


native/package.json
```diff
@@ -6,7 +6,7 @@
     "link": "react-native link",
     "start": "react-native start",
     "ios": "react-native run-ios",
-    "android:emulator": "$ANDROID_HOME/tools/emulator -avd bpk-avd -dpi-device 420 -skin 1080x1920",
+    "android:emulator": "$ANDROID_HOME/tools/emulator -avd bpk-avd",
```
matteo-hertel: This is change is on purpose, will prevent the Pixel2 emulator to be forced at 1080 and render the content is a weird way

shaundon: Oh awesome - I wrote this quite a while ago and it was more-or-less hacked together, so thanks for optimising it!

native/packages/react-native-bpk-component-button/stories.js
```diff
@@ -46,11 +46,17 @@ const styles = StyleSheet.create({
 const getIconType = type =>
   type === 'destructive' ? icons.trash : icons['long-arrow-right'];
 
-const generateButtonStoryForType = (type: string) => {
+const generateButtonStoryForType = (
+  type: string,
+  storyAsLable: boolean = false,
```
shaundon: Should be `storyAsLabel`

matteo-hertel: 💔

matteo-hertel:
I've fixed the typo and below a screenshot of the docs with the new scresnshots, should be good to go

screen shot 2018-02-26 at 08 45 19-fullpage

PR #538, opened 2018-02-23
[BPK-1368] web badge styles

georgegillams:
TODO:
 - [x] Rename types if @jamesf3rguson says to (see below)
 - [x] Add examples to docs site (ask @jamesf3rguson)
 - [x] Add examples to storybook
 - [x] Changelog

### DIFF
--- a/packages/bpk-component-badge/readme.md
+++ b/packages/bpk-component-badge/readme.md
@@ -12,17 +12,22 @@ npm install bpk-component-badge --save-dev
 
 ```js
 import React from 'react';
-import BpkBadge from 'bpk-component-badge';
+import BpkBadge, { BADGE_TYPES } from 'bpk-component-badge';
 
 export default () => (
-  <BpkBadge>My Badge</BpkBadge>
+  <BpkBadge
+    type={BADGE_TYPES.warning}
+  >
+    My Badge
+  </BpkBadge>
 );
 ```
 
 ## Props
 
-| Property  | PropType        | Required | Default Value |
-| --------- | --------------- | -------- | ------------- |
-| centered  | bool            | false    | null          |
-| docked    | 'left', 'right' | false    | null          |
-| className | string          | false    | null          |
+| Property  | PropType                                                                                                                             | Required | Default Value       |
+| --------- | -------------------------------------------------------------------------------------------------------------------------------------| -------- | ------------------- |
+| centered  | bool                                                                                                                                 | false    | null                |
+| className | string                                                                                                                               | false    | null                |
+| docked    | oneOf('left', 'right')                                                                                                               | false    | null                |
+| type      | oneOf(BADGE_TYPES.warning, BADGE_TYPES.success, BADGE_TYPES.destructive, BADGE_TYPES.light, BADGE_TYPES.inverse, BADGE_TYPES.outline)| false    | BADGE_TYPES.warning |
--- a/packages/bpk-component-badge/src/BpkBadge-test.js
+++ b/packages/bpk-component-badge/src/BpkBadge-test.js
@@ -19,7 +19,7 @@
 import React from 'react';
 import renderer from 'react-test-renderer';
 
-import BpkBadge from './BpkBadge';
+import BpkBadge, { BADGE_TYPES } from './BpkBadge';
 
 describe('BpkBadge', () => {
   it('should render correctly', () => {
@@ -46,5 +46,14 @@ describe('BpkBadge', () => {
       .create(<BpkBadge docked="left">Promociando</BpkBadge>)
       .toJSON();
     expect(tree).toMatchSnapshot();
+  });
+
+  Object.keys(BADGE_TYPES).forEach(badgeType => {
+    it(`should render correctly with type="${badgeType}"`, () => {
+      const tree = renderer
+        .create(<BpkBadge type={badgeType}>Promociando</BpkBadge>)
+        .toJSON();
+      expect(tree).toMatchSnapshot();
+    });
   });
 });
--- a/packages/bpk-component-badge/src/BpkBadge.js
+++ b/packages/bpk-component-badge/src/BpkBadge.js
@@ -22,11 +22,29 @@ import { cssModules } from 'bpk-react-utils';
 
 import STYLES from './bpk-badge.scss';
 
+export const BADGE_TYPES = {
+  warning: 'warning',
+  success: 'success',
+  destructive: 'destructive',
+  light: 'light',
+  inverse: 'inverse',
+  outline: 'outline',
+};
+
 const getClassName = cssModules(STYLES);
 
+const badgeTypeClassNames = {
+  [BADGE_TYPES.warning]: null,
+  [BADGE_TYPES.success]: getClassName('bpk-badge--success'),
+  [BADGE_TYPES.destructive]: getClassName('bpk-badge--destructive'),
+  [BADGE_TYPES.light]: getClassName('bpk-badge--light'),
+  [BADGE_TYPES.inverse]: getClassName('bpk-badge--inverse'),
+  [BADGE_TYPES.outline]: getClassName('bpk-badge--outline'),
+};
+
 const BpkBadge = props => {
-  const { docked, centered, className, ...rest } = props;
-  const classNames = [getClassName('bpk-badge')];
+  const { type, docked, centered, className, ...rest } = props;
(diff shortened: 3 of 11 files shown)


packages/bpk-component-badge/readme.md
```diff
@@ -14,15 +14,14 @@ npm install bpk-component-badge --save-dev
 import React from 'react';
 import BpkBadge from 'bpk-component-badge';
 
-export default () => (
-  <BpkBadge>My Badge</BpkBadge>
-);
+export default () => <BpkBadge>My Badge</BpkBadge>;
```
georgegillams: Should be reverted to previous layout for readability

packages/bpk-docs/src/pages/BadgePage/BadgePage.js
```diff
@@ -43,7 +45,7 @@ const components = [
       <BpkBadge className={badgeClassName}>Apples</BpkBadge>,
       <BpkBadge className={badgeClassName}>Bananas</BpkBadge>,
       <BpkBadge className={badgeClassName}>Strawberries</BpkBadge>,
-      <BpkBadge>Pears</BpkBadge>,
+      <BpkBadge className={badgeClassName}>Pears</BpkBadge>,
```
georgegillams: Shouldn't have removed the classname. Also, the classname should probably be on the other added ones too!

packages/bpk-component-badge/src/BpkBadge-test.js
```diff
@@ -47,4 +47,39 @@ describe('BpkBadge', () => {
       .toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  it('should render correctly when success', () => {
```
shaundon: We could reduce the code here a fair bit by iterating over `BADGE_TYPES` and testing it that way.

See here for an example: https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-button/src/BpkButton-test.common.js#L105

packages/bpk-component-badge/src/BpkBadge.js
```diff
@@ -22,11 +22,29 @@ import { cssModules } from 'bpk-react-utils';
 
 import STYLES from './bpk-badge.scss';
 
+export const BADGE_TYPES = {
+  WARNING: 'warning',
+  SUCCESS: 'success',
+  DESTRUCTIVE: 'destructive',
+  LIGHT: 'light',
+  INVERSE: 'inverse',
+  OUTLINE: 'outline',
```
shaundon: Lately we've started matching the keys and values, because it makes these easier to consume in Flow. E.g. in buttons we have:

```
export const BUTTON_TYPES = {
  primary: 'primary',
  secondary: 'secondary',
  destructive: 'destructive',
  featured: 'featured',
};
```

Then when we Flow type it we can do this:

```
export type ButtonType: $Keys<typeof BUTTON_TYPES>,
```

It's a workaround to get around Flow's current inability to have `$Values`.

Another benefit is that we can use it in PropTypes:

```
 type: PropTypes.oneOf(Object.keys(BUTTON_TYPES))
```

(I know `Object.values` exists but the browser support isn't good enough yet).

packages/bpk-component-badge/src/BpkBadge.js
```diff
@@ -45,12 +63,21 @@ const BpkBadge = props => {
...
 
 BpkBadge.propTypes = {
+  type: PropTypes.oneOf([
+    BADGE_TYPES.WARNING,
+    BADGE_TYPES.SUCCESS,
+    BADGE_TYPES.DESTRUCTIVE,
+    BADGE_TYPES.LIGHT,
+    BADGE_TYPES.INVERSE,
+    BADGE_TYPES.OUTLINE,
+  ]),
```
shaundon: As discussed above, if `BADGE_TYPES` matches keys and values, this can become `oneOf(Object.keys(BADGE_TYPES))`.

packages/bpk-mixins/src/mixins/_badges.scss
```diff
@@ -105,3 +104,76 @@
...
+///
+/// @example scss
+///   .selector {
+///     @include bpk-badge();
+///     @include bpk-badge--outline();
+///   }
+
+@mixin bpk-badge--outline {
+  border: solid $bpk-border-size-sm $bpk-color-white;
+  background-color: rgba(0, 0, 0, .2);
```
shaundon: Can be addressed in a future PR but maybe we should have a token for absolute black, same as we do for white.

packages/bpk-docs/src/pages/BadgePage/BadgePage.js
```diff
@@ -83,6 +85,117 @@ const components = [
...
+        </BpkBadge>
+        <BpkBadge type={BADGE_TYPES.OUTLINE} className={badgeClassName}>
+          Pears
+        </BpkBadge>
+      </BadgeLayout>,
+    ],
+  },
 ];
 
 const BadgePage = () => (
```
shaundon: GH won't let me add a comment on the correct line as it's not part of the diff, but on line 205 with the page blurb, it mentions that badges are only available in one colour. This should be removed.

packages/bpk-component-badge/readme.md
```diff
@@ -14,15 +14,14 @@ npm install bpk-component-badge --save-dev
 import React from 'react';
 import BpkBadge from 'bpk-component-badge';
```
shaundon: As we're using `BADGE_TYPES` in the prop types table, it should be imported and used in the example code too.

packages/bpk-mixins/src/mixins/_badges.scss
georgegillams: Woops this should have been 20% opacity white, not 20% opacity black! (so `rgba(255, 255, 255, .2)`) 

And dear god no! We shouldn't use absolute black anywhere, and definitely shouldn't have a token for it! That'd be criminal!

georgegillams: The reason I didn't use the white token is I'm not sure how to apply opacity to a token value. I had a quick google and it didn't seem to be something you can do without nesting divs

georgegillams: Maybe we should have tokens for transparent versions of our colours, or add a function to `token-helper.js`?

packages/bpk-docs/src/pages/BadgePage/BadgePage.js
georgegillams: Done 👍

packages/bpk-mixins/src/mixins/_badges.scss
shaundon: > The reason I didn't use the white token is I'm not sure how to apply opacity to a token value. I had a quick google and it didn't seem to be something you can do without nesting divs

You can do this natively in Sass:

`rgba(#fff, 0.2)`

packages/bpk-docs/src/pages/BadgePage/BadgePage.js
```diff
@@ -91,8 +204,9 @@ const BadgePage = () => (
     blurb={[
       <Paragraph>
         Badges are labels which hold small amounts of information. They are
-        currently available in one colour. Badges are most often used as
-        counters, such as an indication of unread notifications.
+        available in a number of colours to signify different meanings. Badges
```
shaundon: Sadly, `colours` needs to be changed to the Americanised `colors` to match the rest of the docs (and this page).

packages/bpk-mixins/src/mixins/_badges.scss
```diff
@@ -105,3 +104,76 @@
...
+/// @require {mixin} bpk-badge
+///
+/// @example scss
+///   .selector {
+///     @include bpk-badge();
+///     @include bpk-badge--outline();
+///   }
+
+@mixin bpk-badge--outline {
+  background-color: rgba(255, 255, 255, .2);
```
shaundon: `background-color: rgba($bpk-color-white, .2);` should work here.

packages/bpk-docs/src/pages/BadgePage/BadgePage.js
georgegillams: 😞

shaundon:
I'm happy with this once CI passes - will approve after that 👍

shaundon:
@georgegillams tests are failing now 😭

georgegillams:
Think it was just complaining about some redundant snapshots that I'd left in

PR #542, opened 2018-02-26
[BPK-1357] add native button link component

matteo-hertel:
Behold, the Button Link component!

Docs update will follow as separate PR

It follows mostly what button does and it has more or less the same structure

some screenshots:
defaultandroid
defaultios
themeandroid
themeios


native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.js
```diff
@@ -0,0 +1,120 @@
...
+
+import BpkButtonLinkContainer from './layout/BpkButtonLinkContainer';
+
+type Props = {
+  ...$Exact<CommonProps>,
+  iconAlignment: IconAlignment,
+  large: boolean,
+};
+
+const BpkButtonLink = (props: Props) => {
```
k0nserv: Shouldn't this just be a prop for button? What's the reasoning for it being a separate component. In my head if anything this variant is just the secondary button style

native/packages/react-native-bpk-component-button-link/src/utils.js
```diff
@@ -0,0 +1,88 @@
...
+      }
+    });
+  }
+  return themeForElement;
+};
+
+export const themeAttributesSupplied = (theme: Theme): boolean =>
+  difference(REQUIRED_THEME_ATTRIBUTES, Object.keys(theme)).length === 0;
+
+export const themePropType = (
```
k0nserv: This is duplicated in a lot of places right? When do we start thinking about pulling into a single dependency?

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.js
```diff
@@ -0,0 +1,120 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+/* eslint-disable import/no-unresolved */
```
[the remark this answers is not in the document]
shaundon: We should update this decision to mention this too: https://github.com/Skyscanner/backpack/blob/master/decisions/eslint-ignores.md

changelog.md
```diff
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-_Nothing yet..._
+- react-native-bpk-component-button-link:
+  - Introducing the new react native button link component
```
shaundon: `React Native` should be capitalised, and the sentence should end in a full stop 🙈 .

native/packages/react-native-bpk-component-button-link/readme.md
```diff
@@ -0,0 +1,65 @@
...
+> Backpack React Native button link component.
+
+## Installation
+
+1. Install the NPM package:
+```sh
+npm install react-native-bpk-component-button-link --save-dev
+```
+
+
```
shaundon: y u no step 2?

native/packages/react-native-bpk-component-button-link/readme.md
```diff
@@ -0,0 +1,65 @@
...
+3. `react-native-bpk-component-button-link` depends on [react-native-bpk-component-icon](https://github.com/Skyscanner/backpack/tree/master/native/packages/react-native-bpk-component-icon) make sure to follow the installation guide for the icon component for the different platforms
```
shaundon: Might be better to make this link point to the one on npm? Feel free to disagree though.

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  const renderIcon = (): ?Node => {
+    if (!icon) {
+      return null;
+    }
+    if (typeof icon === 'string') {
+      return (
+        <BpkIcon icon={icon} style={iconStyle(theme, props)} small={!large} />
```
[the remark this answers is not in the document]
shaundon: If we do that, it won't be consistent with `BpkButton`. It's a separate component so it's not a dealbreaker, but worth mentioning.

native/packages/react-native-bpk-component-button-link/src/utils.js
shaundon: Now! How about we add a card to do it, we can take care of it in another PR soon.

native/packages/react-native-bpk-component-button-link/index.js
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import BpkButtonLink from './src/BpkButtonLink';
```
matthewdavidson: We should export the flow types too

changelog.md
matteo-hertel: 😅 sorry, I should know this by now

native/packages/react-native-bpk-component-button-link/readme.md
matteo-hertel: There isn't a step two, I need to remove the `1.`

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink-styles.js
```diff
@@ -0,0 +1,88 @@
+/*
```
matthewdavidson: The name of this file needs to change - just call it `styles.js`. We only ever name files as UpperCamelCase if they export components. Otherwise it's kebab-case.

native/packages/react-native-bpk-component-button-link/readme.md
matteo-hertel: No make sense, it's copied from buttons tough so I might need to update that one as well, wdyt?

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.js
matthewdavidson: Just because it's easier. Look at the nic of our existing button component as well as the web variant. A little bit of duplication goes a long way.

native/packages/react-native-bpk-component-button-link/src/layout/BpkButtonLinkContainer.android.js
```diff
@@ -0,0 +1,75 @@
+/*
```
matthewdavidson: Can we rename this file to not have the word container in it? Containers in react conventions have state.

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.js
```diff
@@ -0,0 +1,120 @@
...
+    </BpkButtonLinkContainer>
+  );
+};
+
+const propTypes = {
+  ...COMMON_PROP_TYPES,
+  iconAlignment: PropTypes.oneOf(Object.keys(ICON_ALIGNMENTS)),
+  large: PropTypes.bool,
+};
+BpkButtonLink.propTypes = propTypes;
```
matthewdavidson: let this code breath 😅  (add new lines). Im surprised prettier doesn't enforce this.

matthewdavidson:
@matteo-hertel I appreciate that a bunch of this stuff is probably copied over from the button component - if thats the case then we should fix it in a separate PR.

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.js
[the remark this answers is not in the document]
k0nserv: I don't think it should be blanket rule, but it should be avoid

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.js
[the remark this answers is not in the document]
k0nserv: Hmm okay yeah then we shouldn't imo

changelog.md
k0nserv: We should have a Danger rule

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink-styles.js
shaundon: This can go in the decisions folder too.

native/packages/react-native-bpk-component-button-link/readme.md
shaundon: Yeah good shout

native/packages/react-native-bpk-component-button-link/index.js
matteo-hertel: I was just looking at all the other packages and we're not exporting flow types anywhere else, shall I start from this package?

matthewdavidson: Here is an example -> https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-chip/index.js

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.js
```diff
@@ -0,0 +1,120 @@
...
+    ...rest
+  } = props;
+  let { theme } = props;
+  if (theme) {
+    if (!themeAttributesSupplied(theme)) {
+      theme = null;
+    }
+  }
+
+  const renderIcon = (): ?Node => {
```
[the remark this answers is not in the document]
matteo-hertel: I will actually not change this, the function signature will look too bloated 
```js
const renderIcon = (
  icon: ?(String | Node),
  theme: ?Object,
  props: Props,
  large: boolean,
): ?Node => {
```
and the usage

```js
{renderIcon(icon, theme, props, large)}
`
Let me know if you think it's a deal breaker

**Comment by matteo-hertel** [comment]
I've action every feedback, most of this is from the button link component so some conventions are from there

**Comment by jamesf3rguson** [comment]
In the trailing icon example: The spacing between the text and icon is missing.

**Comment by jamesf3rguson** [comment]
I'm still not convinced the icon is vertically centered. Wanna chat?

**Comment by k0nserv** [comment]
@matteo-hertel we should be using `TouchableOpacity` on iOS

**Comment by matteo-hertel** [comment]
@k0nserv how so? the one we built does exaclty what we need 😕

File: native/packages/react-native-bpk-component-button-link/src/layout/BpkButtonLinkWrapper.android.js [hunk 13]
```diff
@@ -0,0 +1,75 @@
...
+  COMMON_PROP_TYPES,
+  COMMON_DEFAULT_PROPS,
+} from './../common-types';
+
+type Props = {
+  ...$Exact<CommonProps>,
+  children: Node,
+};
+
+const BpkButtonLinkContainer = (props: Props) => {
```
k0nserv: This still says `Container` should be `Wrapper` per the file name

native/packages/react-native-bpk-component-button-link/src/layout/BpkButtonLinkWrapper.ios.js
```diff
@@ -0,0 +1,83 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React from 'react';
```
k0nserv: Is it really worth splitting these two up instead of just using a single file with one `Platform.select`? Seems like we are ending up with a lot of duplication here

native/packages/react-native-bpk-component-button-link/src/layout/BpkButtonLinkWrapper.android.js
matteo-hertel: Fair enough

native/packages/react-native-bpk-component-button-link/src/layout/BpkButtonLinkWrapper.ios.js
matteo-hertel: the two wrappers are different for instance the andorid one requires a static `TouchableNativeFeedback.SelectableBackgroundBorderless()` and I like the separated strucutre a lot, so at least for now, I would keep it like this

k0nserv: I don't agree, but let's agree to disagree and leave it :)

PR #546, opened 2018-02-27
[BPK-1376] DialingCodeList and DialingCodeListItem components

shaundon:
# Overview
Adds `BpkDialingCodeList` and `BpkDialingCodeListItem` components.

# Example Usage

## `BpkDialingCodeListItem`
```
<BpkDiallngCodeListItem
    id="44"
    dialingCode="44"
    name="United Kingdom"
    onPress={action('BpkDialingCodeListItem pressed.')}
    flag={
      <Image
        source={{
          uri: 'https://upload.wikimedia.org/wikipedia/commons/thumb/6/6f/Flag_of_the_Democratic_Republic_of_the_Congo.svg/800px-Flag_of_the_Democratic_Republic_of_the_Congo.svg.png',
        }}
      />
    }
/>
```
Note that an instance of `Image` is passed in. I did it this way so that consumers are free to choose whichever type of image source they want: https://facebook.github.io/react-native/docs/images.html.

## `BpkDialingCodeList`
```
const CODES = [
  {
    id: '0',
    dialingCode: '0',
    name: 'Zero',
  },
  {
    id: '1',
    dialingCode: '1',
    name: 'One',
  },
  {
    id: '2',
    dialingCode: '2',
    name: 'Two',
  },
];

<BpkDialingCodeList
  codes={CODES}
  onItemPress={(item) => {
    action(`${item.name} selected`);
  }}
  selectedId="0"
/>
```
* `selectedId` is not required.
* consumers can use `onItemPress` to handle state.

* See storybook for a stateful example.

# Other info

## Android radios
The wireframe for Android looks like this:
screen shot 2018-03-01 at 15 54 28

However, React Native doesn't appear to expose these radio buttons (I could be wrong, still making sure). So for now I've used the tick icon, like we have on iOS. We should decide as a squad what to do about this, whether we just add a new icon or what.

## Storybook images
In the storybook, I'm using direct links to Wikipedia for the flags. This probably isn't the best, but I don't love the idea of adding a load of flags to the repo just for this. I'm open to any suggestions!

# GIFs!!

## iOS
kapture 2018-03-01 at 15 57 46

## Android
kapture 2018-03-01 at 15 58 52

# To do
* [x] Boilerplate
* [x] Rendering properly
* [x] Selection rendering
* [x] Render tick (iOS)
* [x] Render radios (Android)
* [x] Flag support
* [x] Flags in storybook example
* [x] Tests
* [x] Changelog
* [x] Fix that one Flow error I can't get rid of!
* [x] RTL
* [x] a11y
* [x] Rebase
* [x] Test on different screen sizes

### DIFF
--- a/native/ios/native.xcodeproj/project.pbxproj
+++ b/native/ios/native.xcodeproj/project.pbxproj
@@ -5,6 +5,7 @@
 	};
 	objectVersion = 46;
 	objects = {
+
 /* Begin PBXBuildFile section */
 		00C302E51ABCBA2D00DB3ED1 /* libRCTActionSheet.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 00C302AC1ABCB8CE00DB3ED1 /* libRCTActionSheet.a */; };
 		00C302E71ABCBA2D00DB3ED1 /* libRCTGeolocation.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 00C302BA1ABCB90400DB3ED1 /* libRCTGeolocation.a */; };
@@ -21,6 +22,7 @@
 		13B07FC11A68108700A75B9A /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB71A68108700A75B9A /* main.m */; };
 		140ED2AC1D01E1AD002B40FF /* libReact.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 146834041AC3E56700842450 /* libReact.a */; };
 		146834051AC3E58100842450 /* libReact.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 146834041AC3E56700842450 /* libReact.a */; };
+		28EA01A797384AA6BA660F6A /* libRCTRestart.a in Frameworks */ = {isa = PBXBuildFile; fileRef = B3896E08C3804FF089E1E50B /* libRCTRestart.a */; };
 		2D02E4BC1E0B4A80006451C7 /* AppDelegate.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB01A68108700A75B9A /* AppDelegate.m */; };
 		2D02E4BD1E0B4A84006451C7 /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 13B07FB51A68108700A75B9A /* Images.xcassets */; };
 		2D02E4BF1E0B4AB3006451C7 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 13B07FB71A68108700A75B9A /* main.m */; };
@@ -37,10 +39,9 @@
 		60D802631F62BDF400FA4879 /* libBVLinearGradient.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 60D802431F62BDCA00FA4879 /* libBVLinearGradient.a */; };
 		6A4A4F4B8F7F4269A13BB2A7 /* BpkIcon.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 349CF328083441A0845E0C8D /* BpkIcon.ttf */; };
 		832341BD1AAA6AB300B99B32 /* libRCTText.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 832341B51AAA6A8300B99B32 /* libRCTText.a */; };
+		8E33D9A4365E4F84826CF0C6 /* iconMapping.json in Resources */ = {isa = PBXBuildFile; fileRef = 3CA50A36B56148BD95AF98C2 /* iconMapping.json */; };
 		C8B831744C5A4256BD1B5010 /* BpkIcon.eot in Resources */ = {isa = PBXBuildFile; fileRef = 8C8F437769B54A038A210159 /* BpkIcon.eot */; };
 		E19140ED6394488C9626FE7A /* BpkIcon.woff in Resources */ = {isa = PBXBuildFile; fileRef = E436C94A65D1410092B05450 /* BpkIcon.woff */; };
-		8E33D9A4365E4F84826CF0C6 /* iconMapping.json in Resources */ = {isa = PBXBuildFile; fileRef = 3CA50A36B56148BD95AF98C2 /* iconMapping.json */; };
-		28EA01A797384AA6BA660F6A /* libRCTRestart.a in Frameworks */ = {isa = PBXBuildFile; fileRef = B3896E08C3804FF089E1E50B /* libRCTRestart.a */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -247,6 +248,41 @@
 			remoteGlobalIDString = 3D383D621EBD27B9005632C8;
 			remoteInfo = "double-conversion-tvOS";
 		};
+		607D1B62204964EB0082F575 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 146833FF1AC3E56700842450 /* React.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = EBF21BDC1FC498900052F4D5;
+			remoteInfo = jsinspector;
+		};
+		607D1B64204964EB0082F575 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 146833FF1AC3E56700842450 /* React.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = EBF21BFA1FC4989A0052F4D5;
+			remoteInfo = "jsinspector-tvOS";
+		};
+		607D1B66204964EB0082F575 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 146833FF1AC3E56700842450 /* React.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = 9936F3131F5F2E4B0010BF04;
+			remoteInfo = privatedata;
+		};
+		607D1B68204964EB0082F575 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 146833FF1AC3E56700842450 /* React.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = 9936F32F1F5F2E5B0010BF04;
+			remoteInfo = "privatedata-tvOS";
+		};
+		607D1B6D204964EC0082F575 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 0334EEA012364B5980516D8A /* RCTRestart.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = 3245CDED1BFEE35C00EABF68;
+			remoteInfo = RCTRestart;
+		};
 		60D802421F62BDCA00FA4879 /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
 			containerPortal = 60D8023D1F62BDCA00FA4879 /* BVLinearGradient.xcodeproj */;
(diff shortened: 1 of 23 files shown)


k0nserv:
A few thoughts:

+ Should we expose `BpkDiallingCodeListItem` at all? My thoughts are no.
+ Should the dialing code not include the `+`?
+ Should we use something more appropriate for lookups as the structure. Maybe
```javsacript
type Id = string;
type CountryData = {| dialingCode: string, name: string, id: Id |};
type Data = { [key: Id]: CountryData };
```

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.ios.js
```diff
@@ -0,0 +1,147 @@
...
+  {name: 'Baz'},
+]
+
+and returns:
+{
+  f: [{name: 'Foo'}],
+  b: [{name: 'Bar'}, {name: 'Baz'}],
+}
+*/
+const createAlphabetisedList = data => {
```
k0nserv: This can be simplified as

```javascript
_.groupBy(data, x => x.name[0].toLowerCase());
```

If we are okay with using lodash

shaundon:
> Should we expose BpkDiallingCodeListItem at all? My thoughts are no.

Yeah perhaps not. Reason I did it here was because I matched how `BpkStarRating` exposes `BpkStar`, but we don't have to match that.

> Should the dialing code not include the +?

Could go either way - I decided to have the component add it in, but very happy to change it.

> Should we use something more appropriate for lookups as the structure?

Maybe - interested to see what others think here. So you're saying from a consumer point of view, they'd pass in data like this?:

```
{
  '44': {dialingCode: '44', name: 'United Kingdom', id: '44' }
}
```

Your spelling of `dialing` also made me look up whether it should be `dialing` or `dialling` - looks like U.S. English, which our codes adheres to, only uses a single `L`. So I'll update the code.

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.android.js
```diff
@@ -0,0 +1,77 @@
...
+    marginHorizontal: spacingMd,
+  },
+});
+
+type Props = {
+  ...$Exact<ListCommonProps>,
+};
+
+const BpkDiallingCodeList = ({ codes, selectedId, onItemPress }: Props) => {
+  codes.sort((a, b) => {
```
k0nserv: This seems redudant to do in every render, can't we just delegate ordering to consumers?

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.ios.js
```diff
@@ -0,0 +1,147 @@
...
+  spacingMd,
+} from 'bpk-tokens/tokens/base.react.native';
+import { BpkDiallingCodeListItem } from '..';
+import {
+  type ListCommonProps,
+  LIST_COMMON_PROP_TYPES,
+  LIST_COMMON_DEFAULT_PROPS,
+} from './common-types';
+
+const spacing = spacingMd * 1.5;
```
k0nserv: This should be spacingBase(16pt) and maybe it could be given a more descriptive name?

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.ios.js
```diff
@@ -0,0 +1,147 @@
...
+  }));
+};
+
+type Props = {
+  ...$Exact<ListCommonProps>,
+};
+
+const BpkDiallingCodeList = ({ codes, selectedId, onItemPress }: Props) => (
+  <SectionList
+    sections={convertCodesIntoSections(codes)}
```
k0nserv: I'd wanna look at the affects of this in render and a long list of say 196 entries.

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeListItem.ios.js
```diff
@@ -0,0 +1,128 @@
...
+import BpkIcon, { icons } from 'react-native-bpk-component-icon';
+import {
+  spacingMd,
+  spacingLg,
+  colorBlue600,
+  colorGray100,
+} from 'bpk-tokens/tokens/base.react.native';
+import { type ListItemProps, LIST_ITEM_PROP_TYPES } from './common-types';
+
+const padding = spacingMd * 1.5;
```
k0nserv: `spacingBase` here too

native/packages/react-native-bpk-component-phone-input/src/common-types.js
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import { Component, type Element } from 'react';
+import PropTypes from 'prop-types';
+
+export type ListItemProps = {
+  id: string,
```
k0nserv: Create a new type for `Id`.

```javascript
export type Id = string;
```

k0nserv:
> Maybe - interested to see what others think here. So you're saying from a consumer point of view, they'd pass in data like this?:
>```
>{
>  '44': {dialingCode: '44', name: 'United Kingdom', id: '44' }
>}

Yes

native/packages/react-native-bpk-component-phone-input/index.js
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import BpkDialingCodeList from './src/BpkDialingCodeList';
+import BpkDialingCodeListItem from './src/BpkDialingCodeListItem';
+
```
matteo-hertel: we should expose the PropsTypes here see https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-button/index.js#L23

native/packages/react-native-bpk-component-phone-input/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "main": "index.js",
+  "description": "Backpack React Native phone input component.",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^16.0.0-alpha.1",
```
matteo-hertel: both react and native should be the same as the main package.json https://github.com/Skyscanner/backpack/blob/master/native/package.json#L22

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.ios.js
shaundon: I've changed `spacingMd * 1.5` to `spacingBase` across the board and it looks too spaced out now:

screen shot 2018-03-02 at 08 48 56

I should also point out that when I spoke to James about this before he left, he said he'd just used values from iOS and not Backpack tokens, which is why we agreed to hack it together with `spacingMd * 1.5`. In Sketch, the vertical padding is 12, hence the calculation.

What do you reckon?

native/packages/react-native-bpk-component-phone-input/src/BpkDialingCodeList.android.js
```diff
@@ -0,0 +1,77 @@
...
+    marginHorizontal: spacingMd,
+  },
+});
+
+type Props = {
+  ...$Exact<ListCommonProps>,
+};
+
+const BpkDialingCodeList = ({ codes, selectedId, onItemPress }: Props) => {
+  codes.sort((a, b) => {
```
matteo-hertel: If you move the sort function implementation outside the stateless component IMHO it'll make more readable and maintainable (feel free to disagree)

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.android.js
shaundon: Yeah let's just do that - I'll remove it.

shaundon: They'll get the side-effect that way of being able to use different orders if they want too (e.g. show the current country at the top or something).

native/storybook/storybook.js
```diff
@@ -56,7 +56,7 @@ const styles = StyleSheet.create({
     flex: 1,
     flexDirection: 'column',
     justifyContent: 'center',
-    padding: spacingBase,
+    // padding: spacingBase,
```
matteo-hertel: this will affect all the stories, is that what we want?

native/packages/react-native-bpk-component-phone-input/src/BpkDialingCodeList.android.js
shaundon: That's a good idea, and I *would* do it, but Hugo's suggested we leave sorting to consumers so I'm just gonna remove it 💅

native/storybook/storybook.js
shaundon: Definitely not! I committed that by mistake 😅

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.ios.js
shaundon: Good tip, thanks :)

native/packages/react-native-bpk-component-phone-input/src/BpkDialingCodeList.android.js
matteo-hertel: Aight, in which case let's make sure we add that to the docs / readme

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.ios.js
k0nserv: Yes agree we should use the iOS defaults, but the iOS default is 16pt AFAIK. That's also what I am seeing in Sketch with the UI library builtin to 49

k0nserv: Also just to be clear I'm only talking about the leading margin being 16pt here not the vertical padding of the cell. Cells should be 44pt high if we align with the default

native/packages/react-native-bpk-component-phone-input/package.json
k0nserv: I believe we need to add a card to update these across the board

native/packages/react-native-bpk-component-phone-input/index.js
matthewdavidson: The flow types you mean?

shaundon:
With regards to performance with long lists, I added a load of countries to the array and the performance on device was shocking (iPhone 7 Plus and Pixel 2). Scrolling was more or less ok, but changing the selected item was taking seconds.

After a bit of googling I came across the solution of using rolling your own `shouldComponentUpdate`, so now the list item components have this:

```
  shouldComponentUpdate(nextProps: Props) {
    return nextProps.selected !== this.props.selected;
  }
```

This sped things up massively. There is still a slight lag when there are >150 items but it's IMO not too bad.

cc @k0nserv @matteo-hertel

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.ios.js
shaundon: Ah yeah I see it a lot more now I'm testing on device. I've updated it accordingly!

native/packages/react-native-bpk-component-phone-input/src/BpkDiallingCodeList.ios.js
shaundon: I've made a comment further down about performance.

native/packages/react-native-bpk-component-phone-input/index.js
shaundon: Have now done so 👍

native/packages/react-native-bpk-component-phone-input/package.json
shaundon: Ok cool, I'll leave this for then.

shaundon:
Update:
* Consumers now need to add the `+`.
* Updated storybook and readme to use ISO country codes for ids.
* Fixed layout issues.
* Confirmed it works on small screen sizes (tested on iPhone SE).
* Set `private: true` in `package,json`.

@k0nserv thanks for the approval, I'll merge once CI passes. I still need to make the Android radio buttons, I'll do that post-merge.

native/packages/react-native-bpk-component-phone-input/package.json
matteo-hertel: Do we have a card or do we need to create one?

native/ios/native.xcodeproj/project.pbxproj
```diff
@@ -5,6 +5,7 @@
 	};
 	objectVersion = 46;
 	objects = {
+
```
matteo-hertel: How come this file changed? are there new native deps?

shaundon: No clue lol

PR #548, opened 2018-02-28
[NOJIRA] Refactor RN buttons

matthewdavidson:
This PR get's rid of the notion of a "Container" for what is a stateless component. We instead rely on platform specific extensions at the top level i.e: 
- `BpkButton.ios.js`
- `BpkButton.android.js`

Instead of:
- `BpkButton.js`
- `BpkButtonContainer.ios.js`
- `BpkButtonContainer.android.js`


native/packages/react-native-bpk-component-button/src/BpkButton.android.js
```diff
@@ -76,66 +72,70 @@ const BpkButton = (props: Props) => {
     );
   }
 
-  // Validate that button is themeable and all theming attributes
-  // have been supplied. If not, disable theming.
-  if (theme) {
-    if (!isTypeThemeable(type) || !themeAttributesSupplied(type, theme)) {
-      theme = null;
-    }
+  const shouldApplyTheme =
```
shaundon: I like this abstraction a lot!

native/packages/react-native-bpk-component-button/src/__snapshots__/BpkButton-test.android.js.snap
```diff
@@ -27,7 +27,7 @@ exports[`Android BpkButton should render correctly 1`] = `
           "margin": 10,
           "width": 300,
         },
-        undefined,
+        null,
```
shaundon: Nothing's jumping out at me that'd cause this - I'm guessing it's the result of a style that was being passed down as undefined to Android before due to the wrapper/container, but now isn't?

native/packages/react-native-bpk-component-button/stories.js
```diff
@@ -26,7 +26,7 @@ import BpkThemeProvider from 'react-native-bpk-theming';
 import { icons } from 'react-native-bpk-component-icon';
 import { spacingMd } from 'bpk-tokens/tokens/base.react.native';
 
-import BpkButton from './src/BpkButton';
+import BpkButton from './index';
```
shaundon: 🙌

native/packages/react-native-bpk-component-button/src/utils.js
```diff
@@ -23,7 +23,7 @@ import difference from 'lodash/difference';
 import { colorWhite } from 'bpk-tokens/tokens/base.react.native';
 import { type Theme } from 'react-native-bpk-theming';
 
-import styles from './BpkButton-styles';
+import styles from './styles';
```
shaundon: 💯

native/packages/react-native-bpk-component-button/src/BpkButton.android.js
```diff
@@ -76,66 +72,70 @@ const BpkButton = (props: Props) => {
...
-      {...rest}
-    >
-      <View style={getStyleForElement('view', props)}>
-        {!iconOnly && (
-          <BpkText
-            textStyle={large ? 'lg' : 'sm'}
-            emphasize
-            style={textStyle(theme, props)}
-          >
-            {Platform.OS === 'android' ? title.toUpperCase() : title}
```
shaundon: The `toUpperCase` call was accidentally removed - Android buttons should be in uppercase.

native/packages/react-native-bpk-component-button/src/BpkButton.android.js
```diff
@@ -76,66 +72,70 @@ const BpkButton = (props: Props) => {
...
+      </TouchableNativeFeedback>
+    </View>
   );
 };
 
-const propTypes = {
+BpkButton.propTypes = {
   ...COMMON_PROP_TYPES,
   iconAlignment: PropTypes.oneOf(Object.keys(ICON_ALIGNMENTS)),
   large: PropTypes.bool,
```
shaundon: `iconAlignment` and `large` are common so can both be abstracted.

shaundon: `large` isn't required on Android (readme should reflect this too).

native/packages/react-native-bpk-component-button/src/BpkButton.android.js
matthewdavidson: Strange how that didn't show up in snapshots...

native/packages/react-native-bpk-component-button/src/BpkButton.android.js
matthewdavidson: Im gonna do these in a follow up PR - I promise 🙏

native/packages/react-native-bpk-component-button/src/__snapshots__/BpkButton-test.android.js.snap
matthewdavidson: Yeah I mean either way it's the same behaviour - I think it could be related to the theme 🤔

native/packages/react-native-bpk-component-button/src/BpkButton.android.js
shaundon: I've figured out why it isn't, something we should really address if possible to prevent this:

```
    <Text
      accessible={true}
      allowFontScaling={true}
      ellipsizeMode="tail"
      style={
        Object {
          "color": "#333333",
          "margin": 20,
        }
      }
    >
      TouchableNativeFeedback is not supported on this platform!
    </Text>
```

matthewdavidson: Ah I can fix this!

matthewdavidson: see https://github.com/Skyscanner/backpack/pull/549 :)

matthewdavidson:
@shaundon can i get approvalz please?

PR #553, opened 2018-03-01
[BPK-1388] Refactor button links

matthewdavidson:
Similar to https://github.com/Skyscanner/backpack/pull/548


native/packages/react-native-bpk-component-button-link/src/BpkButtonLink-test.common.js
```diff
@@ -16,16 +16,20 @@
  * limitations under the License.
  */
 
+/* @flow */
```
matteo-hertel: Are flow typing the tests as well?

matthewdavidson: Did we make a decision not to do this? 🤔 😂

matteo-hertel: I thought it was public API only, I might have missed something 😅

PR #554, opened 2018-03-02
[BPK-1365] native badge component

matteo-hertel:
Native Badge component!

There are a couple of bits that might require discussion, I'll add comments inline 

Demo 
badge


native/packages/react-native-bpk-component-badge/src/styles.js
```diff
@@ -0,0 +1,110 @@
...
+    },
+  }),
+};
+const addStyle = pool => (style, prop) => {
+  if (!pool[prop]) return style;
+  style.push(pool[prop]);
+  return style;
+};
+
+export const getStyle = (docked, type) => ({
```
matteo-hertel: The idea here is to get out from this function an object that has the styles ready to consume and make the render method much more readable, it stacks styles and allows pick and choose, but it's a bit different from what we have elsewhere, so I want to double check it's readable and make sense to everybody

native/packages/react-native-bpk-component-badge/src/styles.js
```diff
@@ -0,0 +1,110 @@
...
+      backgroundColor: colorRed500,
+    },
+    light: {
+      backgroundColor: colorGray50,
+    },
+    inverse: {
+      backgroundColor: colorWhite,
+    },
+    outline: {
+      paddingHorizontal: spacingMd - borderSizeSm,
```
matteo-hertel: when the badge is outlined the border is substracted from the padding in order to have outlined and not outlined badges aligned with each other

native/packages/react-native-bpk-component-badge/src/styles.js
```diff
@@ -0,0 +1,110 @@
...
+      paddingVertical: spacingSm - borderSizeSm,
+      backgroundColor: setOpacity(colorWhite, 0.2),
```
matteo-hertel: we agreed offline to use opacity, I'm not going rogue 😄

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -0,0 +1,91 @@
...
+    outline: outlineStyle,
+  } = getStyle(docked, type);
+
+  const outlineViewStyle =
+    type === BADGE_TYPES.outline ? [outlineStyle, borderStyle] : null;
+  const innerViewStyle =
+    type === BADGE_TYPES.outline ? innerStyle : [innerStyle, borderStyle];
+
+  return (
+    <View style={userStyle}>
```
matthewdavidson: Coming in to this PR I expected to see a tree more resembling this:
```
<View>
  <BpkText style={textStyle} textStyle="xs">
    {message}
  </BpkText>
</View>
```

Home come the two extra wrapper views are required?

native/packages/react-native-bpk-component-badge/src/styles.js
k0nserv: Nifty 👍

matthewdavidson: My immediate reaction is that the styling for this component is not complex - I reckon it could probably be inline with the component without requiring a separate `styles.js` file (let alone some functions which apply reducers)

native/packages/react-native-bpk-component-badge/stories.js
```diff
@@ -0,0 +1,81 @@
...
+    />
+  ));
+  return <View style={containerStyle}>{badges}</View>;
+}
+
+storiesOf('BpkBadge', module)
+  .add('docs:default', () => (
+    <View>
+      <StorySubheading>Default</StorySubheading>
+      {generateBadgeStory(null)}
```
matthewdavidson: This doesn't look great - i know it's just for the story's but it'd be much easier to understand if it looked like:
```
generateBadgeStory({
  dark: false,
})

generateBadgeStory({
  docked: 'left',
  dark: true,
})
```

native/packages/react-native-bpk-component-badge/stories.js
```diff
@@ -0,0 +1,81 @@
...
+    justifyContent: 'flex-start',
+  },
+  badge: {
+    margin: spacingSm,
+  },
+  dark: {
+    backgroundColor: colorGray900,
+  },
+});
+function generateBadgeStory(docks, isDark = false) {
```
matthewdavidson: prefer arrow functions

native/packages/react-native-bpk-component-badge/src/styles.js
matteo-hertel: I disagree with that as having a bunch of styles in the components bloats the whole file, when viewing a component to me make sense to see what's logic is there (if any) and if I want checked the stying I go to the style file.

The whole logic is no different from a component like button (see https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-button/src/utils.js#L59) a base style array and modifiers pushed to it, I appreciate we don't agree on how to structure styles/components, what could be a good compromise that makes us all happy?

native/packages/react-native-bpk-component-badge/src/BpkBadge-test.common.js
```diff
@@ -0,0 +1,62 @@
...
+    it('should render correctly docked to the right', () => {
+      const tree = renderer.create(generateBadgeStory('right')).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+    it('should complain if an extraneous docks is passed', () => {
+      jest.spyOn(console, 'error').mockImplementation(err => {
+        throw err;
+      });
+      expect(() => {
+        renderer.create(generateBadgeStory('underground'));
```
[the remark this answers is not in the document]
matteo-hertel: yes indeed, prehaps it needs a better description 😄

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
matteo-hertel: that was the original implementation, but the outmost `View` is to allow external styling without messing with the component and the one immediately after is for dealing with the outline, and the few problems I had with it, I'll double check if the inner one is stll needed

native/packages/react-native-bpk-component-badge/stories.js
matteo-hertel: fair enough, easy to change

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
matteo-hertel: I checked and because we can use opacity the inner view is not needed anymore, thanks for spotting that, I'll update the PR

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -0,0 +1,179 @@
...
+  [BADGE_TYPES.success]: styles.viewSuccess,
+  [BADGE_TYPES.warning]: styles.viewWarning,
+  [BADGE_TYPES.destructive]: styles.viewDestructive,
+  [BADGE_TYPES.light]: styles.viewLight,
+  [BADGE_TYPES.inverse]: styles.viewInverse,
+  [BADGE_TYPES.outline]: [styles.viewOutline, styles.borderOutline],
+};
+
+const textStyleMap = {
+  [BADGE_TYPES.success]: null,
```
matthewdavidson: Im not sure what people think of this... this was to fix the flow warning. I think it' makes it quite explicit that there are no text styles that are specific to `success`, `warning`, `light` or `inverse`. Alternatively i could check for existence and then push the style on line 143

k0nserv: Why doesn't Flow like it without this? Is it because the value might be undefined and you cannot push undefined? If so it'd be better to write it with a conditional on line 143 imo

matthewdavidson: Fixed with type annotation: `{ [key: string]: Object | Array<Object> }`

matthewdavidson:
@matteo-hertel  Right i've stopped messing about with this PR 😂

native/packages/react-native-bpk-component-badge/readme.md
```diff
@@ -0,0 +1,54 @@
...
+        <BpkBadge message="Badge" type="destructive" />
+        <BpkBadge message="Badge" type="inverse" />
+        <BpkBadge message="Badge" type="light" />
+        <BpkBadge message="Badge" type="outline" />
+        <BpkBadge message="Badge" type="success" />
+        <BpkBadge message="Badge" docked="left" type="warning" />
+        <BpkBadge message="Badge" docked="right" type="destructive" />
+        <BpkBadge message="Badge" type="inverse" />
+        <BpkBadge message="Badge" docked="left" type="light" />
+        <BpkBadge message="Badge" docked="right" type="outline" />
```
shaundon: I'm wondering if it'll be better to make these examples use `BADGE_TYPES` - what do you think?

native/packages/react-native-bpk-component-badge/readme.md
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+```
+## Props
+
+| Property              | PropType                                                                 | Required | Default Value |
+| --------------------- | -------------------------------------------------------------------------| -------- | ------------- |
+| message               | string                                                                   | true     | -             |
+| docked                | oneOf('left', 'right')                                                   | false    | null          |
+| type                  | oneOf('success', 'warning', 'destructive', 'light', 'inverse', 'outline')| false    | warning       |
```
shaundon: This could also be `oneOf(BADGE_TYPES)`.

native/packages/react-native-bpk-component-badge/src/BpkBadge-test.common.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    it('should throw an error is an extraneous docked values is provided', () => {
+      jest.spyOn(console, 'error').mockImplementation(err => {
+        throw err;
+      });
+
+      expect(() => {
+        // $FlowFixMe
+        renderer.create(generateBadgeStory({ docked: 'unknown' }));
```
shaundon: Uh oh - what didn't Flow like here?

native/packages/react-native-bpk-component-badge/src/BpkBadge-test.common.js
```diff
@@ -0,0 +1,85 @@
...
+      }).toThrow();
```
shaundon: I think we usually put the actual error message in `toThrow` tests - is that possible here too? 

E.g. https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-input/src/customPropTypes-test.js#L47

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -0,0 +1,179 @@
...
+
+  const viewStyle = [styles.viewBase, styles.borderBase];
+  const textStyle = [styles.textBase];
+
+  viewStyle.push(viewStyleMap[type]);
+  textStyle.push(textStyleMap[type]);
+
+  if (docked === BADGE_DOCKED_TYPES.left) {
+    viewStyle.push(styles.borderLeft);
+  }
```
shaundon: I like this a lot, it's very readable to me :)

matteo-hertel:
I've discussed with @shaundon offline and some comments will be addressed in the follow up PR for the docs

native/packages/react-native-bpk-component-badge/src/BpkBadge-test.common.js
matthewdavidson: Flow doesn't like that `unknown` isn't a valid `docked` type ... but that's explicitly what we are testing here

native/packages/react-native-bpk-component-badge/readme.md
matthewdavidson: to iterate through them? Nah this code should be as straightforward as possible.

shaundon: Not to iterate, I meant instead of `type="light"`, `type={BADGE_TYPES.light}`.

However I've just looked at our readmes and we don't do that in others anyway, so it doesn't matter so much.

matthewdavidson: Ah in that case I agree :)

native/packages/react-native-bpk-component-badge/readme.md
matteo-hertel: I think for the readme is fine to have the types listed so as a quick ref the used doesn't have to look up what the BADGE_TYPES are?

native/packages/react-native-bpk-component-badge/src/BpkBadge-test.common.js
matteo-hertel: To be honest this test is unnecessary as it's not a custom prop type but it relies on the prop types errors, I've added it on early stages but it's now redundant, I'll probably remove it

native/packages/react-native-bpk-component-badge/readme.md
shaundon: Yeah you're probably right, it is better for consumers that way.

PR #557, opened 2018-03-02
Set svg fill colour in BpkText and BpkLink

sogame:
Some time ago `currentColor` was used as `fill` colour for svg's in `BpkButton` (see https://github.com/Skyscanner/backpack/pull/281/).

I think it makes sense to set it also for `BpkText` and `BpkLink` so if an icon is added as part of the text/link its colour is set by default to the text colour.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,10 @@
  - New `BpkDiallingCodeList` component.
 - react-native-bpk-component-badge:
   - Introducing the new React Native badge component.
+- bpk-component-text:
+  - SVGs inside `BpkText` will now automatically have their fill colour set to match the text colour.
+- bpk-component-link:
+  - SVGs inside `BpkLink` and `BpkButtonLink` will now automatically have their fill colour set to match the text colour.
 
 ## 2018-03-01 - Fix for Android button props
 
--- a/packages/bpk-component-link/src/bpk-link.scss
+++ b/packages/bpk-component-link/src/bpk-link.scss
@@ -28,4 +28,8 @@
   &--alternate {
     @include bpk-link--alternate;
   }
+
+  svg {
+    fill: currentColor;
+  }
 }
--- a/packages/bpk-component-text/src/BpkText.scss
+++ b/packages/bpk-component-text/src/BpkText.scss
@@ -30,4 +30,8 @@
   &--bold {
     @include bpk-text--bold;
   }
+
+  svg {
+    fill: currentColor;
+  }
 }


changelog.md
```diff
@@ -9,6 +9,10 @@
   - Introduced a new `accessoryView` prop that enables rendering any view on the leading side of the text input.
 - react-native-bpk-component-phone-input:
  - New `BpkDiallingCodeList` component.
+- bpk-component-text:
+  - Svg's fill colour is automatically set to text colour.
```
shaundon: I think this could be more descriptive. How about:

```
- SVGs inside `BpkText` will now automatically have their fill colour set to match the text colour.
```

Same for below except for `BpkLink`.

sogame:
Good point! No idea if some client relies on the icon being black, but that would be an error, right? The default text is never black...

Anyway, to be safe, this could be released as a major, then? I could modify the changelog to make it "Breaking" instead of "Added".

What do you think about this?

shaundon:
> but that would be an error, right? The default text is never black...

Yep, good point!

> Anyway, to be safe, this could be released as a major, then? I could modify the changelog to make it "Breaking" instead of "Added".

It's probably fine to be honest, we can probably assume it won't cause an issue. I'll see what the rest of the squad thinks.

sogame:
👍 

I've updated the changelog message for now :-)

k0nserv:
It would end up as pure black which is never desirable since it doens't follow our color scheme hence it should be okay to publish this as a non breaking change

matthewdavidson:
Lets think of this the opposite way round ... when consumers want a different colour icon to the text, how do they achieve it?

It looks like they'd have to do this:
```js
import React from 'react';
import BpkSmallFlightIcon from 'bpk-component-icon/sm/flight';

import STYLES from './icons.scss';

export default () => (
  <BpkText>
    <BpkSmallFlightIcon className={STYLES['my-custom-icon']} />
  </BpkText>
);
```

```scss
@import '~bpk-mixins';

.my-custom-icon {
  svg {
    fill: $bpk-color-yellow-500;
  }
}
```

shaundon:
@matthewdavidson I think you're right, but currently they would still need to do that to make the icon any colour other than the default (which I assume is black), right?

matthewdavidson:
@shaundon no, currently they would apply the class to the icon itself, no the parent element of the icon.

shaundon:
This will be a breaking change for only a small benefit, so we think it'll be better to close this. 

The rule introduced here has a higher specificity than a class name on the icon component, meaning it'll override any colours being set on icons.

PR #558, opened 2018-03-04
Add "withModalState" HOC to BpkModal

sogame:
I think it would be useful to provide a HOC to manage default open/close events in `BpkModal`.

This HOC allows setting `onOpen` and `onClose` functions, so parent component could be still notified about the events (if needed), but at least wouldn't have to handle `isOpen` state.

I would like to get your feedback on the implementation (specially the part about `target` rendering (not sure if adding an `onClick` to a span is the best solution).

If/when you're happy with this, I'll update the documentation and add another HOC like this for BpkDialog.

### DIFF
--- a/packages/bpk-component-modal/index.js
+++ b/packages/bpk-component-modal/index.js
@@ -20,8 +20,9 @@
 
 import BpkModal, { type Props } from './src/BpkModal';
 import themeAttributes from './src/themeAttributes';
+import withModalState from './src/withModalState';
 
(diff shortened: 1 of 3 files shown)


shaundon:
As discussed on Slack, we think state management should be left to consumers, and this can be achieved quite easily using [recompose](https://github.com/acdlite/recompose#lift-state-into-functional-wrappers), so we'll close this PR.

PR #560, opened 2018-03-05
[BPK-1365] native badge docs

matteo-hertel:
I propose this solution for the docs on the native badge component

screen shot 2018-03-06 at 04 15 26-fullpage


shaundon:
I think it would be better for consumers if it had a separate `View` for each instance of the badge, like this:

screen shot 2018-03-05 at 15 28 17

That way, the fourth set of screenshots wouldn't be required either.

k0nserv:
If we need to have three distinct sets of screenshots anyway I think we should go with what Shaun proposed, i.e showcasing all options.

matteo-hertel:
Updated based on feedback please review 😄

PR #564, opened 2018-03-06
[no jira] Rename left/right in native badge to start/end

shaundon:
* Also made the readme example code use the `BADGE_TYPES` and `BADGE_DOCKED_TYPES` enums.
* Redid screenshots so that web and native look much closer (see below).

**Web**
web

**Native**
native


k0nserv:
Does this actually work in RTL? I'm thinking specifically about the left/right borders not being correct in RTL

shaundon:
Forgot to check that, one moment and I'll confirm.

shaundon:
RTL is all good!

k0nserv:
It is? I would think that the outline version when docked would do the wrong thing when removing the borders for example

shaundon:
Android screenshots with RTL enabled:

screenshot_1520328895
screenshot_1520328900

k0nserv:
Intersting, so that means that the left/right wording in the border properties also correct for RTL then

shaundon:
Yeah, RN just intelligently inverts it all - Matteo was extolling its virtues last week 😆

PR #569, opened 2018-03-07
[BPK-1391] disabled prop support for native link buttons

shaundon:


native/packages/react-native-bpk-component-button-link/readme.md
```diff
@@ -35,10 +35,8 @@ export default class App extends Component {
           title="Book flight"
           onPress={() => {}} />
         <BpkButtonLink
-          title="Book flight"
-          onPress={() => {}} />
-        <BpkButtonLink
-          title="Book flight"
```
shaundon: There were three buttons in the example code that were all the same, so I've removed one and made another showcase the `disabled` prop.

native/packages/react-native-bpk-component-button-link/readme.md
```diff
@@ -53,11 +51,6 @@ export default class App extends Component {
           icon="baggage"
           iconAlignment="leading"
           onPress={() => {}} />
-        <BpkButtonLink
-          title="Book flight"
-          icon="baggage"
-          iconOnly
-          onPress={() => {}} />
```
shaundon: Removed this example, as this component doesn't currently support `iconOnly` (was probably copypasta'd from `BpkButton`).

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.android.js
```diff
@@ -75,13 +78,21 @@ const BpkButtonLink = (props: Props) => {
     containerStyle.push(style);
   }
 
+  // Disabled styles must be applied last to ensure they
```
matthewdavidson: Remove this comment, the snapshot tests catch this.

shaundon: Done!

matthewdavidson: Actually I'd argue that userland styles come last. One of the key lessons we've learned in the past is that if consumers want to customise stuff they'll either find a way or just not use the component (that latter of which is entirely undesirable from a design system perspective).

matthewdavidson: Most other components follow this order of precedence...

shaundon: Yeah that's fair, I'll move it.

shaundon: Done!

PR #568, opened 2018-03-06
[BPK-1399] Add iconOnly prop to Android BpkButton

shaundon:

PR #570, opened 2018-03-07
[BPK-1389] Native nudger

shaundon:
## Notes
* API is pretty much the same as the web version.
* I expect we'll need to change the Android version to use `BpkButtonLink` instead, because the design has the disabled button greyed out, but with no grey background. `BpkButtonLink` doesn't support `iconOnly` so if so, that'll need to come first.

### Android
screenshot_1520443176

### iOS
simulator screen shot - iphone se - 2018-03-07 at 17 19 52

## To do
* [x] Rebase once #568 is merged
* [x] Boilerplate
* [x] Implementation
* [x] Readme
* [x] Storybook
* [x] Changelog
* [x] Tests
* [x] Check RTL
* [x] Theming support via the secondary buttons
* [x] Flow

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,9 @@
 - react-native-bpk-component-button:
   - Android now supports the `iconOnly` prop.
 
+- react-native-bpk-component-nudger:
+  - Introducing the React Native nudger component.
+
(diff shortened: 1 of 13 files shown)


shaundon:
Once #575 is merged, the design of this component will be updated to match the wireframe.

PR #563, opened 2018-03-05
Add BpkTouchableNativeFeedback and use it instead of TouchableNativeFeedback

wiyarmir:
As a follow-up from #555 , let's use this safe option everywhere for pre-Lollipop safety.

PR #572, opened 2018-03-07
[No JIRA] Fix regressions with Andriod highlights

k0nserv:
The changes in #563 introduced some regressions in our use of touchable
feedback for Android. This fixes that.

### DIFF
--- a/native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
+++ b/native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
@@ -119,18 +119,18 @@ const ALERT_TYPE_STYLES = {
 };
 
 const DismissableButton = ({ dismissButtonLabel, onDismiss }) => (
-  <View style={styles.dismissableContainer}>
-    <View style={styles.dismissButtonContainerRadius}>
-      <BpkTouchableNativeFeedback
-        accessibilityComponentType="button"
-        accessibilityLabel={dismissButtonLabel}
-        onPress={onDismiss}
-      >
-        <BpkText textStyle="sm" emphasize style={styles.dismissLabel}>
-          {dismissButtonLabel.toUpperCase()}
-        </BpkText>
-      </BpkTouchableNativeFeedback>
-    </View>
+  <View
+    style={[styles.dismissableContainer, styles.dismissButtonContainerRadius]}
+  >
+    <BpkTouchableNativeFeedback
+      accessibilityComponentType="button"
+      accessibilityLabel={dismissButtonLabel}
+      onPress={onDismiss}
+    >
+      <BpkText textStyle="sm" emphasize style={styles.dismissLabel}>
+        {dismissButtonLabel.toUpperCase()}
+      </BpkText>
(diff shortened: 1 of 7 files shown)


native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavItem.js
```diff
@@ -16,7 +16,13 @@
  * limitations under the License.
  */
 
-import { View, Platform, StyleSheet, ViewPropTypes } from 'react-native';
+import {
+  Platform,
+  StyleSheet,
+  TouchableNativeFeedback,
```
shaundon: If we're going to use our `BpkTouchableNativeFeedback` wrapper, maybe we could expose `TouchableNativeFeedback.SelectableBackground()` through there, removing the need to import both that and  `TouchableNativeFeedback`. What do you think?

k0nserv: I don't care that much, but I'm lazy so I lean towards just keeping it like this. cc @matthewdavidson

matthewdavidson: It's strange from a consumer POV to import `BpkTouchableNativeFeedback` as well as `TouchableNativeFeedback`. Can we add a prop?

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -119,18 +119,18 @@ const ALERT_TYPE_STYLES = {
 };
 
 const DismissableButton = ({ dismissButtonLabel, onDismiss }) => (
-  <View style={styles.dismissableContainer}>
-    <View style={styles.dismissButtonContainerRadius}>
```
matthewdavidson: This is you removing a redundant view, right?

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavItem.js
```diff
@@ -81,13 +87,17 @@ const BpkHorizontalNavItem = props => {
     ? BpkTouchableNativeFeedback
     : BpkTouchableOverlay;
   const formattedTitle = isAndroid ? title.toUpperCase() : title;
+  const platformProps = isAndroid
+    ? { background: TouchableNativeFeedback.SelectableBackground() }
```
matthewdavidson: Correct me if I am wrong, but I was under the impression that our usage of `SelectableBackground`/`SelectableBackgroundBorderless` was as follows:
- `(Platform.Version < 21) ? SelectableBackground : SelectableBackgroundBorderless`
- Use `overflow: hidden` if ripple needs to be contained

k0nserv: Or just use `SelectableBackground` without clipping. I'm maintaining previous behaviour here. We should reevulate this later

PR #575, opened 2018-03-08
[BPK-1389] Remove background colour on disabled Android buttons

shaundon:
I've just stuck an `if` statement inside the `getAndroidBackgroundColour` function. This isn't particularly nice, but this component is such a hornet's nest that there was no easy way to add it without changing loads of stuff.

I'm very motivated to finish refactoring this now and will sort it out soon!


native/packages/react-native-bpk-component-button/src/BpkButton-test.android.js
```diff
@@ -60,6 +60,20 @@ jest.mock('./BpkButton', () => jest.requireActual('./BpkButton.android.js'));
 
 describe('Android', () => {
   commonTests();
+
+  it('secondary disabled buttons should have a transparent background', () => {
```
matthewdavidson: this is common to both iOS and androind - consider renaming to "should render correctly with secondary type and disabled"

PR #577, opened 2018-03-08
[BPK-1389] Native nudger docs

shaundon:
* Also removed `private: true` from `package.json`, as this PR means it's good to go.
* Copy came from web nudger.

0 0 0 0_8080_components_native_nudger


packages/bpk-docs/src/layouts/DocsLayout/DocsLayout.js
```diff
@@ -153,8 +154,6 @@ const links = [
         route: routes.NATIVE_TOUCHABLE_OVERLAY,
         children: 'Touchable Overlay',
       },
-      { id: 'NATIVE_BUTTON_LINK', route: null, children: 'Button Link' },
```
shaundon: Removed because it's a duplicate.

packages/bpk-docs/src/pages/NativeNudgerPage/NativeNudgerPage.js
```diff
@@ -0,0 +1,74 @@
...
+
+import Paragraph from './../../components/Paragraph';
+import DocsPageBuilder from './../../components/DocsPageBuilder';
+
+const components = [
+  {
+    id: 'default',
+    title: 'Default',
+    blurb:
+      'Nudgers come with decrease and increase buttons to either side of the selected value. A minimum and maximum value need to be provided.',
```
georgegillams: Might sound better to say "on either side"

shaundon: Good plan. I copied this from the web nudger page, so I'll change it in both places.

shaundon:
> Should be button instead of buttons in the route.
Maybe best to import the route from ROUTES and use \${ROUTES.button}#theme-props`

We can't do that in the readme unfortunately. I'll fix the broken link though.

shaundon:
@georgegillams I've addressed the feedback and fixed the conflicts - just waiting on CI now.

PR #578, opened 2018-03-08
[BPK-1402] - Fix `closeOnScrimClick` behaviour

matthewdavidson:
I originally though that the custom `onClose` prop type was also broken for the dialog, but it behaves as expected. The intended behaviour is that it warns if you don't provide an `onClose` handler and it's dismissible.

### DIFF
--- a/packages/bpk-component-dialog/stories.js
+++ b/packages/bpk-component-dialog/stories.js
@@ -83,11 +83,11 @@ class DialogContainer extends Component<Props, State> {
           <BpkButton onClick={this.onOpen}>Open dialog</BpkButton>
         </div>
         <BpkDialog
-          closeLabel={this.props.dismissible ? 'Close dialog' : null}
+          closeLabel="Close dialog"
           id="my-dialog"
           className="my-classname"
           isOpen={this.state.isOpen}
-          onClose={this.props.dismissible ? this.onClose : null}
+          onClose={this.onClose}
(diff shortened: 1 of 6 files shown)


packages/bpk-component-dialog/stories.js
```diff
@@ -83,11 +83,11 @@ class DialogContainer extends Component<Props, State> {
           <BpkButton onClick={this.onOpen}>Open dialog</BpkButton>
         </div>
         <BpkDialog
-          closeLabel={this.props.dismissible ? 'Close dialog' : null}
+          closeLabel="Close dialog"
           id="my-dialog"
           className="my-classname"
           isOpen={this.state.isOpen}
-          onClose={this.props.dismissible ? this.onClose : null}
```
matthewdavidson: This logic actually hid the bug from showing up in our stories.

packages/bpk-docs/src/pages/DialogsPage/DialogExamples.js
```diff
@@ -140,6 +140,7 @@ export class NonDismissibleDialogContainer extends Component<
           id="non-dissmissible-dialog"
           className="my-classname"
           isOpen={this.state.isOpen}
+          onClose={this.onClose}
```
matthewdavidson: This omission hid the bug from showing up in our stories.

PR #580, opened 2018-03-09
[NO-JIRA] Flow type Badge component

georgegillams:
I forgot to flow the badge component before I worked on it, so here it is

### DIFF
--- a/packages/bpk-component-badge/src/BpkBadge.js
+++ b/packages/bpk-component-badge/src/BpkBadge.js
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+/* @flow */
+
 import PropTypes from 'prop-types';
 import React from 'react';
 import { cssModules } from 'bpk-react-utils';
@@ -42,7 +44,14 @@ const badgeTypeClassNames = {
   [BADGE_TYPES.outline]: getClassName('bpk-badge--outline'),
 };
 
-const BpkBadge = props => {
+export type Props = {
+  type: $Keys<typeof BADGE_TYPES>,
+  docked: ?string,
+  centered: boolean,
+  className: ?string,
+};
+
+const BpkBadge = (props: Props) => {
   const { type, docked, centered, className, ...rest } = props;
   const classNames = [getClassName('bpk-badge'), badgeTypeClassNames[type]];
 
--- a/packages/bpk-component-badge/BadgeLayout.js
+++ b/packages/bpk-component-badge/BadgeLayout.js
@@ -16,8 +16,10 @@
  * limitations under the License.
  */
 
+/* @flow */
+
 import PropTypes from 'prop-types';
-import React, { Children } from 'react';
+import React, { type Element, Children } from 'react';
 import { cssModules } from 'bpk-react-utils';
 
 import { BADGE_TYPES } from './index';
@@ -31,7 +33,13 @@ const LIGHT_BADGES = [
   BADGE_TYPES.outline,
 ];
 
-const BadgeLayout = ({ docked, children }) => {
+export type Props = {
+  docked: ?string,
+  children: Element<*>,
+};
+
+const BadgeLayout = (props: Props) => {
+  const { docked, children } = props;
   const classNames = [getClassName('bpk-badge-layout__container')];
 
   if (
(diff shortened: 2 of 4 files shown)


packages/bpk-component-badge/src/BpkBadge.js
```diff
@@ -31,6 +33,7 @@ export const BADGE_TYPES = {
   outline: 'outline',
 };
 
+// TODO Does this need to change?
```
shaundon: The rule is to Flow type external interfaces, so you don't need to do this. You can if you want, however! It's up to you.

packages/bpk-component-badge/src/BpkBadge.js
```diff
@@ -42,7 +45,15 @@ const badgeTypeClassNames = {
   [BADGE_TYPES.outline]: getClassName('bpk-badge--outline'),
 };
 
-const BpkBadge = props => {
+type Props = {
+  type: ?string,
+  docked: ?boolean,
+  centered: ?boolean,
```
shaundon: Change the `?boolean` ones to `boolean`, because:

```
?boolean == true | false | null
boolean == true | false
```

So it's more accurate this way.

packages/bpk-component-badge/src/BpkBadge.js
```diff
@@ -42,7 +45,15 @@ const badgeTypeClassNames = {
...
+  className: ?string,
+  ...rest,
```
shaundon: I don't think you need to add `...rest` here but I could be wrong. @k0nserv?

packages/bpk-component-badge/src/BpkBadge.js
```diff
@@ -42,7 +45,15 @@ const badgeTypeClassNames = {
   [BADGE_TYPES.outline]: getClassName('bpk-badge--outline'),
 };
 
-const BpkBadge = props => {
+type Props = {
```
shaundon: We've started a convention of exporting the `type` too, so just add the `export` keyword in front of this line.

Here's an example of how that's surfaced in `index.js`: https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-nudger/index.js

packages/bpk-component-badge/src/BpkBadge.js
```diff
@@ -42,7 +45,15 @@ const badgeTypeClassNames = {
...
+  type: ?string,
```
shaundon: You already have `BADGE_TYPES` - so here you can do `$Keys<typeof BADGE_TYPES>`. Then, consumers will get a Flow error when they try to use an unrecognised one 🙌

packages/bpk-component-badge/src/BpkBadge.js
matthewdavidson: This should yield a runtime error.

georgegillams: My bad. Was copy-pasta

packages/bpk-component-badge/src/BpkBadge.js
```diff
@@ -42,7 +44,14 @@ const badgeTypeClassNames = {
   [BADGE_TYPES.outline]: getClassName('bpk-badge--outline'),
 };
 
-const BpkBadge = props => {
+export type Props = {
+  type: ?$Keys<typeof BADGE_TYPES>,
```
shaundon: One last thing - sorry, I should have mentioned this earlier - this also shouldn't be optional, because there must always be a `type`.

It's a bit confusing and it took me a bit to get my head around, but think of it like this: if it can be `null`, it's ok to be optional. As `type` has a non-null `defaultProps` declaration, it can never be null.

shaundon:
I've added one last comment, otherwise this is 💯 (once squashed).

packages/bpk-component-badge/src/BpkBadge.js
georgegillams: Cool. Have also now done `BadgeLayout.js`

As external consumers don't use the `BadgeLayout`, should I remove the old `prop-types`

shaundon: It's up to you really. Personally I would leave it in, but you're right that it's not technically required now!

PR #583, opened 2018-03-12
Pass "onClose" prop to the wrapped component in withScrim HOC

sogame:
In this [previous PR](https://github.com/Skyscanner/backpack/pull/578) the `onClose` prop is being extracted and not send to `WrappedComponent` anymore, therefore modals can't be closed using the close button.


sogame:
I've just pushed the snapshots, sorry about it.

PR #585, opened 2018-03-12
[BPK-1394] Optimised withDivider logic

shaundon:
* Now uses a single `View` if iOS.


shaundon:
@k0nserv definitely - it does change the look slightly.

PR #587, opened 2018-03-14
[BPK-1141-touchable-native-feedback-docs

matteo-hertel:
Add touchable native feedback docs, as part of this PR I've added 3 minor changes:

1. Add new story to the BPKTouchableNativeFeedback wrapping it in a view to clip ripple (like one would do when using it), this was mostly for screenshot purpose to be in in line with TouchableOverlay
2. rename the story of BPKNudger to be consistent with the rest of the storybook
3. update the TouchableOverlay docs to point to BPKTouchableNativeFeedback instead of the RN one when mentioning TouchableOverlay is iOS only

Screenshot of the docs 

screen shot 2018-03-15 at 08 32 46-fullpage


packages/bpk-docs/src/pages/NativeTouchableNativeFeedbackPage/NativeTouchableNativeFeedbackPage.js
```diff
@@ -0,0 +1,84 @@
...
+    ],
+  },
+];
+
+const NativeTouchableNativeFeedbackPage = () => (
+  <DocsPageBuilder
+    title="Touchable Native Feedback"
+    blurb={[
+      <Paragraph>
+        The Backpack Touchable <em>Native Feedback</em> component is to be used
```
shaundon: The `<em>` should include `Touchable` too. (I know this was probably copied from the Touchable Overlay component but that one is wrong too 😝 )

packages/bpk-docs/src/pages/NativeTouchableNativeFeedbackPage/NativeTouchableNativeFeedbackPage.js
```diff
@@ -0,0 +1,84 @@
...
+        component. The React Native{' '}
+        <BpkLink href={RNTouchableNativeFeedback} blank>
+          Touchable Native Feedback{' '}
+        </BpkLink>{' '}
+        allows the use of the
+        <em> SelectableBackgroundBorderless</em> feature for the ripple effect
+        supported only on Android API level greater than or equal to 21 if that
+        feature is used on a build target with API level less than 21 the app
+        will crash. BpkTouchableNativeFeedback checks the API level and apply
+        the correct option to avoid any crash.
```
shaundon: This sentence is quite long. How about shortening it and breaking it up:


> React Native's Touchable Native Feedback allows the use of 'SelectableBackgroundBorderless' for the ripple effect. However, this is only supported only on Android API level 21 and higher. BpkTouchableNativeFeedback checks the API level first, to prevent your app from crashing on unsupported devices.

matteo-hertel:
@shaundon I've actioned your feedback (in TouchableOverlay too) and update the screeshot in the PR

PR #589, opened 2018-03-14
Fixed hover behaviour on Safari for bpk-component-ticket

mush159:
Before:
safarianimationbug

After: 
safarianimationbugfixed

### DIFF
--- a/packages/bpk-component-ticket/src/bpk-ticket.scss
+++ b/packages/bpk-component-ticket/src/bpk-ticket.scss
@@ -47,6 +47,7 @@
   &__paper {
     position: relative;
     background-color: $bpk-card-background-color;
+    transform-style: preserve-3d; // Fix on hover flicker bug on Safari
 
     @include bpk-box-shadow-sm;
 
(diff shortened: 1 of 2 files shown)


packages/bpk-component-ticket/src/bpk-ticket.scss
```diff
@@ -47,6 +47,7 @@
   &__paper {
     position: relative;
     background-color: $bpk-card-background-color;
+    transform-style: preserve-3d; // FSS-3894
```
shaundon: We try to keep JIRA links out of the Backpack source code, as it's open source so an external viewer won't be able to reference it. Could you replace it with a short comment explaining it's to prevent flickering in Safari? Same for line 67 too.

mush159: Done 😀

shaundon:
This fix has unfortunately surfaced another issue - in Safari, sometimes a line appears next to the ticket. From what we can tell, this doesn't happen all the time, and it doesn't happen on retina screens, only lower-resolution ones. Previously, the flickering behaviour hid this.

screen shot 2018-03-19 at 09 28 34

Do you think that this is worse than the flickering, and therefore should block this from being merged, or is it going to be acceptable for now? (We can make a card to try and fix this in the future though).

shaundon:
We think that this is fixed on High Sierra, so closing this. Will revisit in the future if it turns out to be a problem.

PR #551, opened 2018-02-28
Native picker component contribution

MichaelTkachuk:

PR #599, opened 2018-03-16
Native picker component

shaundon:
A copy of https://github.com/Skyscanner/backpack/pull/551.

There are still some things that need to be done, there are cards for this, and they can be done in follow up stories.

## Details
**All** statefulness has been removed, now, like with modals, users handle whether the picker is open or closed, the value, etc.

There are three public facing components:

* `BpkPicker`: The wrapper component.
* `BpkPickerItem`: On iOS, this is an aliased `Picker.Item`. On Android it is a `View` that will be used in the dialog.
* `BpkPickerTrigger`: This can be optionally used by consumers as a component for rendering the trigger element. It adds the underline and the arrow icon. Can take either a string or an element. This will be split into its own package in a subsequent PR.

The components match the underlying RN `Picker` API as closely as possible.

## Example usage

```
<View>
  <BpkPickerTrigger label={this.state.value} onPress={() => this.setState({isOpen: true})} />
  <BpkPicker 
    isOpen={this.state.isOpen}
    onClose={() => this.setState({isOpen: false})}
    selectedValue={this.state.value}
    onValueChange={value => this.setState({ value})}
    doneLabel="Done"
  >
      <BpkPickerItem label="One" value="1" />
      <BpkPickerItem label="Two" value="2" />
      <BpkPickerItem label="Three" value="3" />
  </BpkPicker>
</View>
```

## Screenshots

screenshot_1521545963

simulator screen shot - iphone 6 - 2018-03-20 at 11 39 22

### DIFF
--- a/native/packages/react-native-bpk-component-picker/package.json
+++ b/native/packages/react-native-bpk-component-picker/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "react-native-bpk-component-picker",
+  "version": "0.0.0",
+  "private": true,
+  "main": "index.js",
+  "description": "Backpack React Native picker component.",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^16.0.0-alpha.1",
+    "react-native": ">= 0.47.0"
+  },
+  "dependencies": {
+    "bpk-tokens": "^26.7.5",
+    "prop-types": "^15.5.8",
+    "react-native-bpk-component-button-link": "^1.1.6",
+    "react-native-bpk-component-icon": "^1.5.0",
+    "react-native-bpk-component-text": "^2.1.32",
+    "react-native-bpk-component-touchable-native-feedback": "^1.1.4",
+    "react-native-bpk-component-touchable-overlay": "^1.0.14"
+  }
+}
--- a/native/packages/react-native-bpk-component-picker/src/BpkPicker-test.common.js
+++ b/native/packages/react-native-bpk-component-picker/src/BpkPicker-test.common.js
@@ -1,0 +1,55 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import TestRenderer from 'react-test-renderer';
+import BpkPicker from './BpkPicker';
+import BpkPickerItem from './BpkPickerItem';
+
+const commonTests = () => {
+  describe('BpkPicker', () => {
+    const emptyFn = () => null;
+
+    it('should render correctly', () => {
+      const testRenderer = TestRenderer.create(
+        <BpkPicker onValueChange={emptyFn} onClose={emptyFn} doneLabel="Done">
+          <BpkPickerItem label="foo" value="foo" />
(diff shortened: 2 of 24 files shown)


native/packages/react-native-bpk-component-picker/src/BpkPicker.js
```diff
@@ -0,0 +1,118 @@
...
+        >
+          {children}
+        </BpkPickerMenu>
+      </View>
+    );
+  }
+}
+
+BpkPicker.propTypes = {
+  children: PropTypes.node.isRequired,
```
shaundon: Lots of duplication here between components. This will be addressed in a follow-up PR.

native/packages/react-native-bpk-component-picker/src/BpkPickerItem.android.js
```diff
@@ -0,0 +1,70 @@
...
+BpkPickerItem.propTypes = {
+  color: PropTypes.string,
+  label: PropTypes.string.isRequired,
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  onPress: PropTypes.func,
+  selected: PropTypes.bool,
+};
+
+BpkPickerItem.defaultProps = {
+  color: null,
```
shaundon: `color` is it's own prop to match its iOS counterpart.

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -0,0 +1,90 @@
...
+  colorWhite,
+  lineHeightBase,
+  spacingBase,
+} from 'bpk-tokens/tokens/base.react.native';
+
+const rowsToDisplay = 6;
+
+const maxListHeight =
+  // eslint-disable-next-line no-mixed-operators
+  (2 * spacingBase + lineHeightBase) * rowsToDisplay;
```
shaundon: To be addressed in a follow-up PR.

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -0,0 +1,90 @@
...
+  );
+  return (
+    <Modal
+      transparent
+      visible={visible}
+      animationType="fade"
+      onRequestClose={() => onClose()}
+    >
+      <View style={styles.overlay}>
+        <ScrollView style={styles.list}>{pickerItems}</ScrollView>
```
shaundon: May be better as `FlatList`, will be addressed in a follow up.

native/packages/react-native-bpk-component-picker/src/BpkPickerItem.android.js
matthewdavidson: I understand you have added this to the android picker item to align with the native behaviour in iOS. But lets drop it and remove it as an explicit prop. Folks can still implement this in userland in both Android and iOS by virtue of ...rest

native/packages/react-native-bpk-component-picker/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "Backpack React Native picker component.",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^16.0.0-alpha.1",
+    "react-native": ">= 0.47.0"
```
matteo-hertel: can we updated this to be inline with the main rn, react version?

native/packages/react-native-bpk-component-picker/src/BpkPicker-test.common.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import React from 'react';
+import TestRenderer from 'react-test-renderer';
+import BpkPicker from './BpkPicker';
+import BpkPickerItem from './BpkPickerItem';
+
+const commonTests = () => {
+  describe('BpkPicker', () => {
+    const emptyFn = () => null;
```
matteo-hertel: we should use `jest.fn()` and potentially check is those function have been called properly

native/packages/react-native-bpk-component-picker/src/BpkPickerTrigger-test.common.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import React from 'react';
+import { View } from 'react-native';
+import TestRenderer from 'react-test-renderer';
+import BpkPickerTrigger from './BpkPickerTrigger';
+
+const commonTests = () => {
+  describe('BpkPickerTrigger', () => {
+    const emptyFn = () => null;
```
matteo-hertel: same as above, `jest.fn` should be used

PR #601, opened 2018-03-16
[BPK-1424] expose small prop in BpkNavigationItem

matteo-hertel:
I've also added inline comments; one thing @shaundon mentioned is to pass the small prop to the Nav components rather than Item, I tough was a good idea in the beginnign, but adding it to the item makes the intent clear so I went for it

screenshot_1521211569
simulator screen shot - iphone 6 - 2018-03-16 at 14 46 08

### DIFF
--- a/native/packages/react-native-bpk-component-horizontal-nav/readme.md
+++ b/native/packages/react-native-bpk-component-horizontal-nav/readme.md
@@ -70,6 +70,7 @@ export default class App extends Component {
 | onPress             | func                                  | true     | -             |
 | title               | string                                | true     | -             |
 | accessibilityLabel  | string                                | false    | props.title   |
+| small               | bool                                  | false    | false         |
 | disabled            | bool                                  | false    | false         |
 | theme               | See [Theme Props](#theme-props) below | false    | null          |
 
--- a/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavItem.js
+++ b/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavItem.js
@@ -23,6 +23,7 @@ import {
   colorGray300,
   colorGray700,
   colorBlue700,
+  spacingSm,
   spacingMd,
   spacingBase,
 } from 'bpk-tokens/tokens/base.react.native';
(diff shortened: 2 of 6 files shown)


native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNavItem.js
```diff
@@ -82,7 +83,7 @@ const BpkHorizontalNavItem = props => {
     : BpkTouchableOverlay;
   const formattedTitle = isAndroid ? title.toUpperCase() : title;
   const platformProps = isAndroid ? { borderlessBackground: false } : {};
-
+  const textSize = small ? 'sm' : 'base';
```
matthewdavidson: `textSize` should be `textStyle` to match the prop that it's being assigned to.

native/packages/react-native-bpk-component-horizontal-nav/readme.md
```diff
@@ -70,6 +70,7 @@ export default class App extends Component {
 | onPress             | func                                  | true     | -             |
 | title               | string                                | true     | -             |
 | accessibilityLabel  | string                                | false    | props.title   |
+| small               | bool                                  | false    | false         |
```
k0nserv: Maybe we should be more explicit about what's actually `small`? It's a bit ambigious as is, something like `smallText` would be a bit clearer IMO

shaundon:
I still feel like `small` should be on the wrapper element and then passed down automatically, to prevent consumers from mixing and matching (on purpose or by accident). 

An example of this is how in the storybook, the disabled item at the end of the list doesn't have `small` 😅

shaundon:
Also, are we ok with just using a boolean for this, or do we want to use an enum to avoid a breaking change if we need to add another size in the future?

willhg:
Yep, I agree @matthewdavidson let's adjust the vertical padding between the type and the baseline to be the same as the default state so they are consistent. Assuming default is using a BPK spacing standard.

matthewdavidson:
@shaundon, 

Perhaps we should expose `textStyle` as a prob, but limit them initially to `sm` and `base`. That would appease @k0nserv's comment.

In reply to your ask about it being on the wrapper element - this is a non issue IMO. We are providing building blocks there is still responsibility on consumers to assemble those blocks correctly. Plus typically with a nav like this the items are mapped from an array.

matteo-hertel:
Pr updated based on feedback (screens in the desc as well)

1.  vertical padding has been updated
2. we agreed on keeping the `small` prop name for consistency with other component
3. We agreed on keeping the prop a boolean

PR #608, opened 2018-03-21
[BPK-1426] Add better Android behaviour

shaundon:
* When there are fewer than six options the picker is now longer full of empty space.
* The scrim is now interactive, it closes the picker when pressed.
* Picker items now have the correct ripple.
* Selected picker item is now always on screen on load, by using `initialScrollIndex` property of `FlatList`.
* Modal width and elevation now matches default Android styles.
* Now works in landscape mode on iOS.

**Note:** I don't love the magic numbers in use to calculate row height, but it's needed for `getItemLayout` and calculating the picker size. In the refactor card this should be optimised to be less fragile to styles changing.

### DIFF
--- a/native/packages/react-native-bpk-component-picker/src/BpkPickerItem.android.js
+++ b/native/packages/react-native-bpk-component-picker/src/BpkPickerItem.android.js
@@ -34,15 +34,24 @@ const styles = StyleSheet.create({
 
 const BpkPickerItem = props => {
   const { value, label, onPress, selected } = props;
+
+  const accessibilityTraits = ['button'];
+  if (selected) {
+    accessibilityTraits.push('selected');
+  }
+
   return (
-    <View style={styles.pickerItem}>
-      <BpkTouchableNativeFeedback
-        onPress={() => onPress && onPress(value)}
-        accessibilityLabel={label}
-      >
+    <BpkTouchableNativeFeedback
+      onPress={() => onPress && onPress(value)}
+      accessibilityComponentType="button"
+      accessibilityLabel={label}
+      accessibilityTraits={accessibilityTraits}
+      borderlessBackground={false}
+    >
+      <View style={styles.pickerItem}>
         <BpkText style={selected ? styles.selected : {}}>{label}</BpkText>
-      </BpkTouchableNativeFeedback>
-    </View>
+      </View>
+    </BpkTouchableNativeFeedback>
   );
 };
 
--- a/native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
+++ b/native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
@@ -17,7 +17,13 @@
  */
 
 import React from 'react';
-import { Modal, View, StyleSheet, ScrollView } from 'react-native';
+import {
+  FlatList,
+  Modal,
+  StyleSheet,
+  TouchableWithoutFeedback,
+  View,
+} from 'react-native';
 import PropTypes from 'prop-types';
 import { setOpacity } from 'bpk-tokens';
 import {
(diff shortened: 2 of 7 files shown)


native/packages/react-native-bpk-component-picker/src/BpkPickerItem.android.js
```diff
@@ -34,15 +34,24 @@ const styles = StyleSheet.create({
...
+  }
+
   return (
-    <View style={styles.pickerItem}>
-      <BpkTouchableNativeFeedback
-        onPress={() => onPress && onPress(value)}
-        accessibilityLabel={label}
-      >
+    <BpkTouchableNativeFeedback
+      onPress={() => onPress && onPress(value)}
```
matthewdavidson: This check for `onPress` isn't actually required. `onPress={onPress}` is good enough. It's a required prop anyways right?

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -27,64 +33,109 @@ import {
...
   spacingBase,
 } from 'bpk-tokens/tokens/base.react.native';
+import BpkPickerItem from './BpkPickerItem';
 
-const rowsToDisplay = 6;
-
-const maxListHeight =
-  // eslint-disable-next-line no-mixed-operators
-  (2 * spacingBase + lineHeightBase) * rowsToDisplay;
+const MAX_ROWS_TO_DISPLAY = 6;
```
matteo-hertel: 🙌

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -27,64 +33,109 @@ import {
...
+  const pickerItems = React.Children.map(children, (child, index) => {
+    const { label, value } = child.props;
+
+    let selected = false;
+    if (selectedValue && selectedValue === value) {
+      selected = true;
+
+      // If selected item will be off-screen, set initialScrollIndex.
+      if (index >= MAX_ROWS_TO_DISPLAY) {
+        initialScrollIndex = index - 2;
```
matteo-hertel: is the -2 to have the selected item centered?

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -27,64 +33,109 @@ import {
...
-    justifyContent: 'center',
   },
   list: {
     backgroundColor: colorWhite,
     borderRadius: borderRadiusSm,
-    maxHeight: maxListHeight,
-    margin: spacingBase,
-    flex: 0,
+    alignSelf: 'center',
+    top: '50%',
```
matthewdavidson: Isn't flexbox much better suited to vertically centering things?

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
shaundon: Yeah, I'll add a screenshot to this thread, one moment 🕙

shaundon: Do you think I should add a comment to the code to make it clearer?

shaundon: Without `- 2` (selected item is at the top)
screenshot_1521627144

With `- 2`: (selected item is centred nicely)
screenshot_1521627160

native/packages/react-native-bpk-component-picker/src/BpkPickerItem.android.js
```diff
@@ -34,15 +34,24 @@ const styles = StyleSheet.create({
 
 const BpkPickerItem = props => {
   const { value, label, onPress, selected } = props;
+
+  const accessibilityTraits = ['button'];
+  if (selected) {
+    accessibilityTraits.push('selected');
+  }
+
```
shaundon: a11y was missing before..

native/packages/react-native-bpk-component-picker/src/BpkPickerItem.android.js
```diff
@@ -34,15 +34,24 @@ const styles = StyleSheet.create({
...
+      accessibilityLabel={label}
+      accessibilityTraits={accessibilityTraits}
+      borderlessBackground={false}
+    >
+      <View style={styles.pickerItem}>
         <BpkText style={selected ? styles.selected : {}}>{label}</BpkText>
-      </BpkTouchableNativeFeedback>
-    </View>
+      </View>
+    </BpkTouchableNativeFeedback>
```
shaundon: This makes the ripple work properly.

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -27,64 +33,109 @@ import {
...
+    minWidth: '50%',
```
shaundon: There may be a flexier way to do this.

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -27,64 +33,109 @@ import {
...
+      }
+    }
+    return { index, label, value, selected };
+  });
+
+  const rowsToDisplay =
+    pickerItems.length > MAX_ROWS_TO_DISPLAY
+      ? MAX_ROWS_TO_DISPLAY
+      : pickerItems.length;
+  const heightOfOneItem = spacingBase * 2 + lineHeightBase;
```
shaundon: To be refactored as this is currently fragile, will break if `BpkPickerItem` style changes.

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -27,64 +33,109 @@ import {
...
       visible={visible}
       animationType="fade"
       onRequestClose={onClose}
     >
-      <View style={styles.overlay}>
-        <ScrollView style={styles.list}>{pickerItems}</ScrollView>
-      </View>
+      <TouchableWithoutFeedback onPress={onClose}>
+        <View style={styles.overlay} />
+      </TouchableWithoutFeedback>
```
shaundon: Scrim's now clickable yo

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.ios.js
```diff
@@ -69,7 +69,12 @@ const PickerMenu = props => {
     React.cloneElement(child, { key: child.props.value }),
   );
   return (
-    <Modal transparent visible={visible} animationType="slide">
+    <Modal
+      supportedOrientations={['portrait', 'landscape']}
```
shaundon: Now works in landscape.

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
shaundon: I struggled to do it in the Y-direction - wanna pair on trying to get it working? 🙏 📿

native/packages/react-native-bpk-component-picker/src/BpkPickerItem.android.js
shaundon: It's actually not required, because of the way the API works. Consumers use `BpkPickerItem` like this:

```
<BpkPicker>
    <BpkPickerItem label="foo" value="1" />
</BpkPicker>
```

And then `BpkPickerMenu` adds the `onPress` behaviour when the children are cloned and it's rendered.

Tbh this is probably a good indication that these shouldn't be the same component. I'll split it into two (internal one used for rendering, external one for consumers to pass data) in the refactor.

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
matteo-hertel: Maybe a tiny comment can help out why the magic 2 but I gueesed it right first time so should be ok if you don't add it 😄

shaundon: I added it anyway to be safe.

PR #611, opened 2018-03-21
[BPK-1432] Add Flow typing to BpkPicker

shaundon:


native/packages/react-native-bpk-component-picker/src/common-types.js
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import { type Node } from 'react';
+import PropTypes from 'prop-types';
+
+export type PickerValue = string | number;
```
k0nserv: Why do we support both string and number?

native/packages/react-native-bpk-component-picker/src/common-types.js
```diff
@@ -0,0 +1,50 @@
...
+
+export type PickerMenu = {
```
k0nserv: This should be `PickerMenuProps` right?

native/packages/react-native-bpk-component-picker/src/common-types.js
shaundon: We're matching what the underlying Picker component uses: https://github.com/facebook/react-native/blob/master/Libraries/Components/Picker/Picker.js#L53

(It says `any` but then right above it, says string or number, so I went with the stricter one).

native/packages/react-native-bpk-component-picker/src/common-types.js
shaundon: Ah yeah, will fix.

shaundon: Done

PR #614, opened 2018-03-22
[No JIRA] Don't botch image styles in nav bar

k0nserv:
Because we were replacing the style network images didn't work as they
always require their own style.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Fixed:**
+- react-native-bpk-component-navigation-bar:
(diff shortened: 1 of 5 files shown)


changelog.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Fixed:**
+- react-native-bpk-component-navigation-bar:
+  - When providing an image for the title we no longer override the style instead we append to it.
```
shaundon: Holy run on sentence, Batman 😜 

How about:

```When providing an image for the title we no longer override the style. Instead, we append to it.```

PR #615, opened 2018-03-22
[BPK-1371] Fix modal issue on iPad

k0nserv:
Mobile safari running on an iPad exhibits many problems with the fixed
scrim used in the modal when the virtual keyboard is present. This
resolves many of these issues by adding `position: fixed` to the body
when the modal and other scrim using components are presented.

Things to consider:

- Does this have a negative effect anywhere we use the scrim?
- Should we make this configurable by the user or just always apply the fix?

### DIFF
--- a/packages/bpk-scrim-utils/src/__snapshots__/withScrim-test.js.snap
+++ b/packages/bpk-scrim-utils/src/__snapshots__/withScrim-test.js.snap
@@ -11,6 +11,7 @@ exports[`BpkScrim render should render correctly 1`] = `
   />
   <TestComponent
     dialogRef={[Function]}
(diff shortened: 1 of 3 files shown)


matthewdavidson:
When the modal opens, because of `position: fixed;`, the scroll position changes to `0` (top). This is probably an acceptable tradeoff. However when we close the modal it should go back to the original scroll position. Here is a gif of the bug:


ios banner alert misaligned text

georgegillams:
This seems to work. Shame that it doesn't just behave with `position: fixed`.

georgegillams:
Have approved so that you can merge once the changelog is updated :)

PR #616, opened 2018-03-23
[BPK-1405] Fix autosuggest id when used inside BpkFieldset

georgegillams:
For purposes of recreating this bug and testing changes, I added an autosuggest example to the Fieldsets page. I haven't removed it again, as there may be value in leaving it there. 

I thought about making the necessary change within `BpkAutosuggest`, but being a wrapper component it is not easy. I have therefore added a fix in the BpkFieldset which uses the value of `children.props.inputProps.id` if available. If not, it will use `children.props.id` as normal.

### DIFF
--- a/packages/bpk-component-fieldset/src/BpkFieldset.js
+++ b/packages/bpk-component-fieldset/src/BpkFieldset.js
@@ -60,9 +60,14 @@ const BpkFieldset = (props: Props) => {
     return null;
   }
 
+  let childId = children.props.id;
+  if (children.props.inputProps && children.props.inputProps.id) {
+    childId = children.props.inputProps.id;
+  }
+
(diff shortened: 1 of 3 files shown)


packages/bpk-component-fieldset/src/BpkFieldset.js
```diff
@@ -60,9 +60,19 @@ const BpkFieldset = (props: Props) => {
     return null;
   }
 
+  let childId = children.props.id;
+  if (
+    children.props.inputProps &&
```
k0nserv: This is redundant `children.props.inputProps && children.props.inputProps.id` is enough

k0nserv:
@matthewdavidson fixed changelog please re-approve

PR #617, opened 2018-03-23
[BPK-1430] style update

matteo-hertel:
Update to the horizontal nav, once figure out what the problem was it was an easy change

I have not added the extra view for android because it'll look odd

viewunderline


native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -36,14 +36,16 @@ const styles = StyleSheet.create({
   horizontalNav: {
     borderColor: 'transparent',
     borderBottomColor: colorGray100,
+    flexDirection: 'column',
     borderWidth: borderSizeSm,
   },
+  flexGrow: {
```
k0nserv: As I remember the reason for this was a bit intricate, could we incldue a comment explaining it so we don't end up confused about why this is here

matteo-hertel: Fair point, will do

matteo-hertel:
One of the comments was the horizontal nav is a bit further  down the screen, I think it's due to the storybook decorator change we did a while ago, here's a screenshot(s) on the left the screenshot on the docs, on the right the current master

slice 1

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -36,14 +36,19 @@ const styles = StyleSheet.create({
...
+  // to allow ScrollView item to be properly positioned in the spaceAround config
+  // the contentContainerStyle needs flexGrow to have the children define the end
+  // width and allow overflow from the view
+  flexGrow: {
+    flexGrow: 1,
+  },
   horizontalNavInner: {
     flexDirection: 'row',
   },
   spaceAround: {
```
matthewdavidson: Can we rename the styles from:
```
horizontalNav: {
},
flexGrow: {
},
horizontalNavInner: {
},
spaceAround: {
}
```

to:

```
nav: {
},
navSpaceAround: {
},
inner: {
},
innerSpaceAround: {
}
```

By prefixing the names it's easier to see what maps where.

matteo-hertel:
@matthewdavidson I've updated based on the feedback, please have a look

PR #619, opened 2018-03-26
[BPK-1437] Remove state from animate height component

matthewdavidson:
The animate height component has a lot of unnecessary renders - this is mostly due to keeping things in state that don't belong in state.

This PR attempts to remove as much from state as possible.

Also removed an extraneous view.

### DIFF
--- a/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
+++ b/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
@@ -16,9 +16,9 @@
  * limitations under the License.
  */
 
-import { View, Platform, Animated } from 'react-native';
 import React from 'react';
 import PropTypes from 'prop-types';
+import { View, Platform, Animated, ViewPropTypes } from 'react-native';
 import { animationDurationSm } from 'bpk-tokens/tokens/base.react.native';
 
 const COLLAPSED_HEIGHT = 0.01;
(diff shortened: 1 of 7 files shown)


native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -28,114 +28,118 @@ class BpkAnimateHeight extends React.Component {
...
     this.animationDuration =
       Platform.OS === 'ios' ? props.animationDuration : 0;
-
-    this.componentWillReceiveProps = this.componentWillReceiveProps.bind(this);
-    this.setExpandedHeight = this.setExpandedHeight.bind(this);
-    this.onHeightAnimationComplete = this.onHeightAnimationComplete.bind(this);
-    this.performAnimation = this.performAnimation.bind(this);
   }
 
   componentWillReceiveProps(nextProps) {
```
shaundon: @k0nserv told me that `componentWillReceiveProp` is marked for deprecation in React 16.3. Can we change it to `componentDidUpdate(prevProps)` instead?

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -28,114 +28,118 @@ class BpkAnimateHeight extends React.Component {
...
     if (typeof this.props.onAnimationComplete === 'function') {
       this.props.onAnimationComplete();
     }
+
     // eslint-disable-next-line no-underscore-dangle
     if (this.state.height._value - 0.01 < 0.00001) {
-      this.setState({ inTree: false });
+      this.inTree = false;
+      this.forceUpdate();
     }
```
shaundon: I'm wondering if we can avoid using `forceUpdate` as React [discourages it](https://reactjs.org/docs/react-component.html#forceupdate). Is `this.setState(this.state)` too hacky?

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
matthewdavidson: Perhaps in the future.

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
matthewdavidson: The React docs indeed discourage usage of `forceUpdate`, but putting things in state that don't belong in state is also discouraged. In this particular example I am using `forceUpdate` to trigger a `render` in order to remove the view from the react tree. Previously this was handled by `setState({ inTree: false });` But `inTree` doesn't really belong in state. 

i.e. previously we'd call `setState({ inTree: true });` in `componentWillReceiveProps` - this actually results in 2 renders instead of one.

matthewdavidson: oh and i guess to answer your specific question - `this.setState(this.state)` is no better than `this.forceUpdate()` lol

matthewdavidson:
@matteo-hertel i had to rebase to fix merge conflicts - can you approve again?

bitmoji

PR #620, opened 2018-03-27
update preinstall, node

matteo-hertel:
quick change to the `package.json`

you might have to run `nvm use` once this gets merged

the preinstall check has problems with windows npx should help that

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -76,10 +76,10 @@ When you are ready to jump into code, use the following workflow to add new comp
 
 ## Getting started
 
-> Backpack development requires Node LTS and npm `^5.5.1`. [Nvm](https://github.com/creationix/nvm) users can run
+> Backpack development requires Node LTS and npm `^5.6.0`. [Nvm](https://github.com/creationix/nvm) users can run
 `nvm use` to switch to `lts/carbon`. [Nave](https://github.com/isaacs/nave) users can use `nave auto`. You can also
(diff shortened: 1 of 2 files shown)


k0nserv:
But now `npx` need to be installed globally? Can't the ensure-node-env check fail because `npx` is missing in the global env you happen to be running?

matteo-hertel:
The latests LTS ships with `npx` installed by default, so if npx failes because is not installed, could be an indicator the node version is not correct, the proof is in the pudding tough, run 
`docker run --rm node:carbon npx -v` should output a version number

k0nserv:
Yeah but the problem is what used to be a nice descriptive error message will now just be `unknown binary npx` on Node < 8 which degrades UX. Further Backpack doesn't strictly speaking support Windows anyway

matteo-hertel:
The correct node version is listed in the getting started guide anyway, if we can remove a brickwall for windows users by having a less user friendly message, I think it's worth it, it also makes the `preinstall` command much neater

k0nserv:
Yeah but in that case can't we use the curl version that works on Windows instead?

matteo-hertel:
I have a version of that curl that works in some windows, beacuse there are tons of different terminal emulators, not all of them implements `curl` the same way, with `npx` will work in a terminal emulator, cmd, power shell etc

matteo-hertel:
Discussed offline, `curl` won over `npx`

PR #621, opened 2018-03-27
[No JIRA] Centralize theme util

k0nserv:
Move some theming utility functions to `react-native-bpk-theming` as
they fit better in there. No changelog entry because we don't expect
consumers to use these.

### DIFF
--- a/native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar.ios.js
+++ b/native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar.ios.js
@@ -21,19 +21,18 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { type Element, StyleSheet, View, ViewPropTypes } from 'react-native';
 import BpkText from 'react-native-bpk-component-text';
-import { withTheme } from 'react-native-bpk-theming';
+import {
+  withTheme,
+  getThemeAttributes,
+  makeThemePropType,
+} from 'react-native-bpk-theming';
 import {
   colorGray50,
   colorGray100,
   colorGray700,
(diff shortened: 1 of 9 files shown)


native/packages/react-native-bpk-theming/src/util.js
```diff
@@ -0,0 +1,71 @@
...
+    return new Error(
+      `Invalid prop \`${propName}\` supplied to \`${componentName}\`. When supplying \`theme\` all the required theming attributes(\`${requiredAttributes.join(
+        ', ',
+      )}\`) must be supplied.`,
+    ); // eslint-disable-line max-len
+  }
+  return false;
+};
+
+export const getThemeStyle = (
```
matthewdavidson: I think this should be called `getThemeAttributes`

k0nserv:
As a follow up we can use this stuff in

+ Horizontal Nav
+ Button
+ Button Link
+ Spinner
+ Switch

native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar.ios.js
```diff
@@ -134,7 +133,10 @@ class BpkNavigationBar extends Component<Props, {}> {
 
   constructor(props) {
     super(props);
-    this.theme = getThemeStyle(IOS_THEME_ATTRIBUTES, this.props.theme || {});
+    this.theme = getThemeAttributes(
+      IOS_THEME_ATTRIBUTES,
+      this.props.theme || {},
```
matthewdavidson: Not a biggy, but it'd be nice if you could just pass `this.props.theme`.

k0nserv: Suppose yeah

PR #622, opened 2018-03-27
[BPK-1421] Add docs for native pagination dots

shaundon:
0 0 0 0_8080_components_native_pagination-dots


shaundon:
@k0nserv updated now.

PR #623, opened 2018-03-27
[BPK-1437] Remove redundant snapshot tests from banner alert

matthewdavidson:
As titled - there is no need to test every prop with each alert type permutation. It's "good enough" to test them with one alert type.

Snapshot tests become less valuable when one slight change ends up failing hundreds of tests. The more cognitive load for the contributor / reviewer means more chance of issues slipping through.


matthewdavidson:
I must have forgotten to commit them ...

bitmoji

PR #625, opened 2018-03-28
[BPK-1450] Icons in navigation bar title

k0nserv:


native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar.android.js
```diff
@@ -181,6 +184,40 @@ class BpkNavigationBar extends Component<Props, {}> {
...
+    // title in the body of the if to `Element`.
+    // While this if is mutually exclusive to the above it
+    // cannot be an else if as Flow seems unable to handle this.
+    if (typeof title === 'object' && title.type && isValidElement(title)) {
+      titleView = (
+        <View style={styles.titleViewOuter}>
+          {React.cloneElement(title, {
+            style: [
+              title.props.style ? title.props.style : null,
+              { maxHeight: 32 },
```
matteo-hertel: what;s the plan for the magic number? we can create a token or even al alias for it

native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar.ios.js
```diff
@@ -167,6 +172,32 @@ class BpkNavigationBar extends Component<Props, {}> {
...
+
+    // This if ensures Flow correctly refines the type of
+    // title in the body of the if to `Element`.
+    // While this if is mutually exclusive to the above it
+    // cannot be an else if as Flow seems unable to handle this.
+    if (typeof title === 'object' && title.type && isValidElement(title)) {
+      titleView = React.cloneElement(title, {
+        style: [
+          title.props.style ? title.props.style : null,
+          { maxHeight: 28 },
```
matteo-hertel: same comment as above for the magic number

native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar.android.js
k0nserv: All of these magic numbers are intentional all though maybe these two shouldn't be. I'd argue it's out of scope for this PR though

k0nserv:
I'll squash at the end before merging

native/packages/react-native-bpk-component-navigation-bar/readme.md
```diff
@@ -79,13 +79,23 @@ import BpkNavigationBar from 'react-native-bpk-component-navigation-bar'
...
-| Property       | PropType               | Required | Default Value |
-| -------------- | ---------------------- | -------- | ------------- |
-| title          | oneOf(string, element) | true     | -             |
-| theme          | object                 | false    | null          |
-| leadingButton  | button element         | false    | null          |
-| trailingButton | button element         | false    | null          |
-| subtitleView   | element                | false    | null          |
+| Property       | PropType                             | Required | Default Value |
+| -------------- | ------------------------------------ | -------- | ------------- |
+| title          | oneOf(string, titleWithIcon element) | true     | -             |
```
shaundon: I think that instead of `titleWithIcon element`, we should say `titleWithIcon shape`. `element` is a bit ambiguous with the element type IMO.

native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar-test.ios.js
```diff
@@ -123,5 +123,23 @@ describe('ios', () => {
...
+                title="Back"
+                showTitle
+                onPress={jest.fn()}
+              />
+            }
+          />,
+        )
+        .toJSON();
+      expect(tree).toMatchSnapshot();
+    });
```
shaundon: Is it worth having a test for `iconPosition: 'trailing'` too?

matthewdavidson:
Are there any screenshots to update? Perhaps pop one in this PR?

native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar-test.ios.js
```diff
@@ -123,5 +123,45 @@ describe('ios', () => {
...
+              <BpkNavigationBarBackButtonIOS
+                title="Back"
+                showTitle
+                onPress={jest.fn()}
+              />
+            }
+          />,
+        )
+        .toJSON();
+      expect(tree).toMatchSnapshot();
```
matthewdavidson: Any reason why these tests are not in common? The additional trailing icon test is now missing from android, for instance.

native/packages/react-native-bpk-component-navigation-bar/src/TitleView-test.common.js
```diff
@@ -0,0 +1,66 @@
...
+  describe('TitleView', () => {
+    it('should render correctly with only a title', () => {
+      const tree = renderer
+        .create(<TitleView title="Backpack" tintColor="rgb(255, 255, 255)" />)
+        .toJSON();
+
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly with leading icon', () => {
```
matthewdavidson: @shaundon / @matteo-hertel  looks like trailing / leading is already being tested here?

native/packages/react-native-bpk-component-navigation-bar/src/TitleView.js
```diff
@@ -0,0 +1,91 @@
...
+  leadingIcon: {
+    marginEnd: spacingSm,
+  },
+  trailingIcon: {
+    marginStart: spacingSm,
+  },
+  textWithIcon: {
+    // This line height ensure the icon lines up
+    // nicely with the baseline of the text.
+    lineHeight: spacingLg,
```
matthewdavidson: I reckon this should be a "magic" number as opposed to a token. 

The use of a token makes it seems like the line height value has meaning. It actually doesn't - the value just so happened to work with lock icons. A different value is required for other icons AFIK. 

Does that make sense?

native/packages/react-native-bpk-component-navigation-bar/src/TitleView-test.common.js
matteo-hertel: damn, that's true, I'll remove the one I added 😞

PR #628, opened 2018-03-29
[BPK-1454] Fix fullscreen modal content scrolling

georgegillams:
Before:
before

After:
fix

iOS Test:
ios_test

Windoze Tests:


packages/bpk-component-modal/src/bpk-modal-dialog.scss
```diff
@@ -66,17 +69,9 @@
     opacity: $bpk-modal-opacity;
   }
 
-  &--iphone-fix {
```
matthewdavidson: FYI there are still instances of `--iphone-fix` in the `withScrim` HOC (see `bpk-scrim-utils`). 

It'd be worth revisiting these...

unreleased.md
```diff
@@ -7,3 +7,5 @@
 **Fixed:**
 - react-native-bpk-component-banner-alert:
   - Fixed devDependencies that should be dependencies.
+- bpk-component-modal:
```
matthewdavidson: Changes in the modal also affect dialogs so we should test them and figure out if they are affected / improved.

Also drawers might suffer a similar fate so i reckon we should check them too (outside the scope of this PR).

packages/bpk-docs/src/pages/ModalsPage/ModalsPage.js
```diff
@@ -121,7 +121,218 @@ const components = [
         buttonText="Open modal"
       >
         <Paragraph>
-          You can put anything you want in here, including forms:
```
matthewdavidson: Hmm I'm not sure about having this as an example in the docs - perhaps a story is enough?

packages/bpk-component-modal/src/bpk-modal-dialog.scss
```diff
@@ -99,5 +94,7 @@
 
   &__content {
     padding: $bpk-modal-content-padding;
+    flex-grow: 1;
+    overflow-y: scroll;
```
matthewdavidson: This now causes a scrollbar gutter to appear on desktop:

screen shot 2018-03-29 at 15 13 16

matthewdavidson: Dialogs are also affected:

image

unreleased.md
georgegillams: Think dialogs are fine (ie behave the same as before 😄 )

georgegillams: And drawers now behave in the same way as the fixed modals, so that's nice

matthewdavidson:
The default modal in mobile safari should appear like this:
image uploaded from ios 1

It currently appears like this:
image uploaded from ios 2

packages/bpk-component-modal/src/bpk-modal-dialog.scss
georgegillams: I think we can remove that too. It doesn't seem to change anything

packages/bpk-component-modal/src/bpk-modal-dialog.scss
matthewdavidson: Drawers are also affected:

image

georgegillams: Weird. I don't get that!

georgegillams: Ok I do on some browsers...

georgegillams: Done - was a 1 line fix

PR #630, opened 2018-03-30
[BPK-1344] New docs nav

shaundon:
Sadly I didn't manage to finish this in time. I'll add as much info as I can so that hopefully somebody else can pick it up:

* I made `NeoSideNav` which is conditionally called instead of `SideNav`. There's some duplication between that and `SideNav`, e.g. the link sorter. This duplication can be removed when the old SideNav is replaced with the new one.

* The new side nav stores a list of the 'main' pages (using backpack, tokens, components and github). The idea is that it opens not expanded, showing just the current page (e.g. components), then when clicked, it'll expand the list to show the other main pages. I've added the toggling functionality but it doesn't seem to be having an effect on making the other pages show, I think I did React wrong or something. When expanded, the other list (e.g. the list of all components) should be greyed out.

* I have to say, I'm not convinced about this interaction model - it's increasing the number of clicks needed to move around, and hiding data away, and I'm not sure what the benefit of this is. Especially as this is a documentation site, which is one place it's ok to be information-dense IMO.

* Currently, the picture and icon for each main page are location in `NeoHomePage`. These should be relocated somewhere higher up the tree, as they'll be used in this sidebar.


matthewdavidson:
RTL is broken:

image

packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
```diff
@@ -26,7 +26,8 @@
   }
 
   &__main {
-    flex: 1;
+    display: flex;
+    flex: 1 0 auto;
```
matthewdavidson: This can just be `flex: 1;`

packages/bpk-docs/src/layouts/DocsLayout/DocsLayout.js
```diff
@@ -198,11 +199,16 @@ const links = [
...
 ];
 
-const DocsLayout = ({ children, ...rest }) => (
-  <SideNavLayout links={links} {...rest}>
-    {children}
-  </SideNavLayout>
-);
+const DocsLayout = ({ children, ...rest }) =>
+  process.env.BPK_NEO ? (
+    <NeoSideNav activeSection="components" links={links} {...rest}>
```
matthewdavidson: I feel like `activeSection` should be derived from the current `location` (from react router). That's how the old side nav works it out...

packages/bpk-docs/src/layouts/NeoSideNav/SideNav.scss
```diff
@@ -0,0 +1,60 @@
...
+    margin-left: 2px;
+  }
+
+  &__sidebar {
+    align-items: flex-start;
+    display: flex;
+    background-color: $bpk-color-gray-900;
+    color: $bpk-color-white;
+    flex-direction: column;
+    min-width: $bpk-one-pixel-rem * 480;
```
matthewdavidson: `20 * $bpk-spacing-base;`

packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
k0nserv: That's not true see the [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/flex). Namely

> The item is given the specified proportion of the free space in the container. This is equivalent to setting "flex: <positive-number> 1 0".

k0nserv:
RTL fixed, I flipped to logo too which just looks dumb. On the other hand if you don't do that it's not great either.

*Flipped*
image


*Not Flipped*
image

packages/bpk-docs/src/layouts/NeoSideNav/NavList.scss
```diff
@@ -87,6 +92,14 @@
       transform: translateY(-50%);
 
       @include bpk-icon(arrow-right, $bpk-color-red-500, small);
+
+      @include bpk-rtl {
+        // HACK: Move sligthly further to the right to account for icon bounding box so
```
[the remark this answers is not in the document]
k0nserv: Do we have access to that somehow?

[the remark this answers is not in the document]
matthewdavidson: We'll have to wait till next week haha

k0nserv:
Resolved in 8f2b5f4. The mix of flexbox and our grid is starting to be painful, but luckily we'll get rid of this when we finalize neo

PR #633, opened 2018-04-02
[BPK-1437] Refactor android banner alerts

matthewdavidson:
- [x] Add flow typing
  - Including a breaking change to lowercase all alert type keys i.e. `ALERT_TYPES.success` instead of `ALERT_TYPES.SUCCESS`
- [x] Add correct screenshot sizes
- [x] Use button link component for android banner alert
- [x] Removed a couple of extraneous views
- [x] Android button links are now borderless by default
- [ ] ~Remove absolute positioning for icon alignment~ I don;t think we'll be able to do this after all. Fundamentally the icons just do not align up to android text size small :/


native/packages/react-native-bpk-component-banner-alert/src/AlertTypes.js
```diff
@@ -1,26 +0,0 @@
-/*
```
matthewdavidson: This has moved to `common-types.js`.

native/packages/react-native-bpk-component-banner-alert/src/AnimateAndFade-test.common.js
```diff
@@ -45,7 +48,7 @@ const commonTests = () => {
     it('should render correctly with animateOnEnter', () => {
       const tree = renderer
         .create(
-          <AnimateAndFade shown animateOnEnter>
+          <AnimateAndFade show animateOnEnter>
```
matthewdavidson: I fixed these typo's

native/packages/react-native-bpk-component-banner-alert/src/AnimateAndFade.js
```diff
@@ -16,15 +16,60 @@
...
-import React, { Component } from 'react';
+import React, { Component, type Node } from 'react';
 import { View, Animated, ViewPropTypes } from 'react-native';
 import { animationDurationSm } from 'bpk-tokens/tokens/base.react.native';
 
 const COLLAPSED_HEIGHT = 0.01;
 
-class AnimateAndFade extends Component {
-  constructor(props) {
+type MeasureCallback = (
```
matthewdavidson: The signature of RN's measure callback -> https://facebook.github.io/react-native/docs/direct-manipulation.html#measurecallback

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert-test.ios.js
```diff
@@ -16,9 +16,10 @@
  * limitations under the License.
  */
 
+/* @flow */
+
 import commonTests from './BpkBannerAlert-test.common';
 
-jest.useFakeTimers();
```
matthewdavidson: I have no idea what this was for, but the tests still pass without it.

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -16,139 +16,111 @@
...
+} from './common-types';
 
-const styles = StyleSheet.create({
-  mainContainer: {
-    paddingVertical: spacingBase,
-    paddingHorizontal: spacingLg,
+import AnimateAndFade from './AnimateAndFade';
+
+const iconAlignmentPadding = {
+  // Current android "sm" text calculated line height is 19.4.
```
matthewdavidson: As mentioned in this PR description - the android icon alignment with text size small is just plain off. That's why i have kept some hard coded magic numbers.

native/packages/react-native-bpk-component-button-link/readme.md
```diff
@@ -59,16 +59,17 @@ export default class App extends Component {
...
+| onPress                             | func                                                                      | true     | -             |
+| title                               | string                                                                    | true     | -             |
+| disabled                            | bool                                                                      | false    | false         |
+| accessibilityLabel                  | string                                                                    | false    | props.title   |
+| icon                                | oneOf(string, element) Strings must be a [BpkIcon](/components/web/icons) | false    | null          |
+| iconAlignment                       | oneOf('leading', 'trailing')                                              | false    | trailing      |
+| large (iOS only)                    | bool                                                                      | false    | false         |
+| borderlessBackground (Android only) | bool                                                                      | false    | true          |
```
matthewdavidson: Added `borderlessBackground ` to this prop table

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
k0nserv: What kind of alignment dose this achieve? Center? Why doens't `alignItems: 'center'` work in that case?

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -168,108 +139,63 @@ const BpkBannerAlert = props => {
...
-      </View>
-      {dismissable && (
-        <DismissableButton
-          dismissButtonLabel={dismissButtonLabel}
-          onDismiss={onDismiss}
-        />
-      )}
-      {expandable && <ExpandableIcon expanded={expanded} />}
-    </View>
+  const ExpandToggleView = expandToggleViewProps => (
```
k0nserv: Can we move this outside? Recreating the function for each render seems redundant and the value of capturing a few values in the closure seems negligable at best

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -159,7 +131,6 @@ const BpkBannerAlert = props => {
     dismissButtonLabel,
     toggleExpandedButtonLabel,
     show,
-    style: userStyle,
```
k0nserv: Why is this getting removed? This is a breaking change

native/packages/react-native-bpk-component-banner-alert/src/common-types.js
```diff
@@ -0,0 +1,87 @@
...
+  onDismiss: ?(event: SyntheticEvent<>) => mixed,
+  onToggleExpanded: ?(event: SyntheticEvent<>) => mixed,
+  show: boolean,
+  toggleExpandedButtonLabel: ?string,
+  bannerStyle: ?any,
+};
+
+export const commonPropTypes = {
+  message: PropTypes.string.isRequired,
+  type: PropTypes.oneOf([
```
k0nserv: This PropType does not match the Flow type. Flow specifies the keys of the `ALERT_TYPES` object whereas this uses the values. Now since they are the same this is not an immediate problem, but it's strange

native/packages/react-native-bpk-component-banner-alert/src/customPropTypes.js
```diff
@@ -18,6 +18,8 @@ import { isString } from 'util';
 * limitations under the License.
 */
 
+/* @flow */
+
 export const dismissablePropType = (props, propName, componentName) => {
```
k0nserv: Shouldn't this been in common-types too? Given that all the other prop types are there

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
matthewdavidson: Styles will still be applied to the outer view as part of `{...rest}`

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert-test.ios.js
matteo-hertel: it was due to using `Date.now` but you're now spying on it so it should be safe to remove

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -16,139 +16,111 @@
...
-  spacingXxl,
+  borderRadiusSm,
 } from 'bpk-tokens/tokens/base.react.native';
-
-import BpkAnimateHeight from 'react-native-bpk-component-animate-height';
-import BpkIcon, { icons } from 'react-native-bpk-component-icon';
 import BpkText from 'react-native-bpk-component-text';
+import { StyleSheet, View } from 'react-native';
+import BpkIcon, { icons } from 'react-native-bpk-component-icon';
+import BpkButtonLink from 'react-native-bpk-component-button-link';
```
matteo-hertel: 🙌

native/packages/react-native-bpk-component-banner-alert/src/__snapshots__/AnimateAndFade-test.android.js.snap
```diff
@@ -79,11 +78,10 @@ exports[`Android AnimateAndFade should render correctly with animateOnEnter 1`]
 exports[`Android AnimateAndFade should render correctly with animateOnLeave 1`] = `
 <View
   collapsable={undefined}
-  shown={true}
   style={
     Object {
-      "height": 0.01,
-      "opacity": 0,
+      "height": null,
```
matteo-hertel: this does not look correct, should it not be an integer?

native/packages/react-native-bpk-component-banner-alert/src/common-types.js
matteo-hertel: you could also use `Object.keys` like in the button component 
https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-button-link/src/common-types.js#L55

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
matthewdavidson: Fixed!

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
matthewdavidson: `alignItems: 'center' work in that case?` does work (sort of - the alignment isn't perfect). But a problem arises when the banner alert message spans multiple lines. Here is a graphic which shows the alignment and what happens with a wrapping message:

image

native/packages/react-native-bpk-component-banner-alert/src/customPropTypes.js
matthewdavidson: There are no custom prop types inside `common-types.js`. I mean the file separation still makes sense...

native/packages/react-native-bpk-component-banner-alert/src/__snapshots__/AnimateAndFade-test.android.js.snap
matthewdavidson: Actually it shouldn't! When the banner alert is expanded, a height is not rendered. This is so that if the content changes inside the height is allowed to expand / collapse accordingly. If we had a hard coded height the contents would be clipped or too short for the height.

native/packages/react-native-bpk-component-banner-alert/src/common-types.js
matthewdavidson: Done!

native/packages/react-native-bpk-component-banner-alert/src/customPropTypes.js
k0nserv: Yeah but all the prop types for the banner alert are in there, it's confusing at the very least

PR #634, opened 2018-04-02
[BPK-1383] Reduce duplication of theming logic

k0nserv:


native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.android.js
```diff
@@ -52,10 +52,11 @@ const BpkButtonLink = (props: Props) => {
...
 
-  const themeStyle =
-    theme && themeAttributesSupplied(theme)
-      ? { color: theme.buttonLinkTextColor }
-      : null;
+  const themeAttributes = getThemeAttributes(REQUIRED_THEME_ATTRIBUTES, theme);
+  let themeStyle = null;
+  if (themeAttributes) {
+    themeStyle = { color: themeAttributes.buttonLinkTextColor };
+  }
```
matthewdavidson: Why not:
```js
const themeStyle = themeAttributes ? { color: themeAttributes.buttonLinkTextColor } : null;
```
?

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.ios.js
```diff
@@ -54,10 +54,11 @@ const BpkButtonLink = (props: Props) => {
...
 
-  const themeStyle =
-    theme && themeAttributesSupplied(theme)
-      ? { color: theme.buttonLinkTextColor }
-      : null;
+  const themeAttributes = getThemeAttributes(REQUIRED_THEME_ATTRIBUTES, theme);
+  let themeStyle = null;
+  if (themeAttributes) {
+    themeStyle = { color: themeAttributes.buttonLinkTextColor };
+  }
```
matthewdavidson: Same as android.

native/packages/react-native-bpk-component-button-link/src/utils-test.js
```diff
@@ -1,56 +0,0 @@
-/*
```
matthewdavidson: 🙌

native/packages/react-native-bpk-component-button-link/src/utils.js
```diff
@@ -1,46 +0,0 @@
-/*
```
matthewdavidson: 🙌

native/packages/react-native-bpk-component-spinner/src/BpkSpinner.js
```diff
@@ -87,8 +86,8 @@ const BpkSpinner = props => {
 
 const propTypes = {
   small: PropTypes.bool,
-  theme: themePropType,
-  type: PropTypes.oneOf(SPINNER_TYPES),
+  theme: makeThemePropType(REQUIRED_THEME_ATTRIBUTES),
+  type: PropTypes.oneOf(Object.keys(SPINNER_TYPES)),
```
matthewdavidson: Minor: `Object.keys(SPINNER_TYPES)` is done three times - could be one.

native/packages/react-native-bpk-component-switch/src/__snapshots__/BpkSwitch-test.android.js.snap
```diff
@@ -64,11 +61,6 @@ exports[`Android BpkSwitch should support theming 1`] = `
   onStartShouldSetResponder={[Function]}
   onTintColor="rgb(230, 228, 235)"
   style={undefined}
-  theme={
```
matthewdavidson: It looks like the theme functionality on the switch has been regressed...

k0nserv: Nope it just used to be spread on the Switch element as part of `rest` before, but now it isn't.

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.android.js
k0nserv: Yeah you're right

native/packages/react-native-bpk-component-switch/src/__snapshots__/BpkSwitch-test.android.js.snap
matthewdavidson: Ok then in that case we must be missing a test for theming? For instance I can't see any application of a "theme" in these snapshots: https://github.com/Skyscanner/backpack/blob/76b1c7b0a028f312c46b8d15ad57c97cae6a094d/native/packages/react-native-bpk-component-switch/src/__snapshots__/BpkSwitch-test.android.js.snap

k0nserv: The `theme` prop never ends up in the snapshot for any of our themeable components instead other props/styles are derived and passed down. In this case this is captured by the different tint colros

PR #635, opened 2018-04-02
[BPK-1404] fix build warning

matteo-hertel:
This fix most of the build warning we have while running tests, on the native side there are no more warnings, on the web side the only warnings are realted to React Router, wich needs to be tackled separately 

I'll add some inline comment to explain some of the decisions

### DIFF
--- a/native/packages/react-native-bpk-component-card/src/BpkCard-test.android.js
+++ b/native/packages/react-native-bpk-component-card/src/BpkCard-test.android.js
@@ -21,11 +21,19 @@
 import commonTests from './BpkCard-test.common';
 
 jest.mock('react-native', () => {
+  const MARSHMALLOW = 23;
   const reactNative = jest.requireActual('react-native');
   jest
     .spyOn(reactNative.Platform, 'select')
     .mockImplementation(obj => obj.android || obj.default);
   reactNative.Platform.OS = 'android';
+
+  // We rely on BpkTouchableNativeFeedback to handle the ripple effect on older devices
+  // so in this test suite we're assuming API > 23 where the ripple is supported
+  Object.defineProperty(reactNative.Platform, 'Version', {
+    value: MARSHMALLOW,
+  });
+
   return reactNative;
 });
 
--- a/native/packages/react-native-bpk-component-nudger/src/BpkNudger-test.common.js
+++ b/native/packages/react-native-bpk-component-nudger/src/BpkNudger-test.common.js
@@ -117,7 +117,8 @@ const commonTests = () => {
     });
 
     it('should disable theming if the required attribute is omitted', () => {
-      const onPressFn = jest.fn();
+      const consoleErrorFn = jest.fn();
+      jest.spyOn(console, 'error').mockImplementation(consoleErrorFn);
       const theme = {};
       const tree = renderer
         .create(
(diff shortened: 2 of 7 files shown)


native/packages/react-native-bpk-component-nudger/src/BpkNudger-test.common.js
```diff
@@ -128,12 +129,13 @@ const commonTests = () => {
...
               decreaseButtonLabel="Decrease"
               increaseButtonLabel="Increase"
-              onChange={onPressFn}
+              onChange={() => null}
             />
           </BpkThemeProvider>,
         )
         .toJSON();
       expect(tree).toMatchSnapshot();
+      expect(consoleErrorFn.mock.calls.length).toBe(1);
```
k0nserv: Why are we expecting this?

native/packages/react-native-bpk-component-card/src/BpkCard-test.android.js
```diff
@@ -21,11 +21,19 @@
...
   const reactNative = jest.requireActual('react-native');
   jest
     .spyOn(reactNative.Platform, 'select')
     .mockImplementation(obj => obj.android || obj.default);
   reactNative.Platform.OS = 'android';
+
+  // We rely on BpkTouchableNativeFeedback to handle the ripple effect on older devices
+  // so in this test suite we're assuming API > 23 where the ripple is supported
+  Object.defineProperty(reactNative.Platform, 'Version', {
+    value: MARSHMALLOW,
```
matteo-hertel: This is to prevent the warning
```
Requested foreground ripple, but it is not available on this version of Android. Consider calling TouchableNativeFeedback.canUseNativeForeground() and using a different Touchable if the result is false.
```
we rely on `BpkToucheableNativeFeedback` to provide the correct ripple effect, so it's safe to test API >= 23

native/packages/react-native-bpk-component-text-input/src/BpkTextInput.js
```diff
@@ -66,7 +66,7 @@ class BpkTextInput extends Component<Props, State> {
     onFocus: PropTypes.func,
     placeholder: PropTypes.string,
     style: ViewPropTypes.style,
-    valid: PropTypes.oneOf(true, false, null),
+    valid: PropTypes.oneOf([true, false, null]),
```
matteo-hertel: `oneOf` expect an array, this fixed all the warning for TextInput and PhoneInput

packages/bpk-component-theme-toggle/src/updateOnThemeChange-test.js
```diff
@@ -17,21 +17,34 @@
...
+import PropTypes from 'prop-types';
 import renderer from 'react-test-renderer';
 import { mount } from 'enzyme';
 import updateOnThemeChange from './updateOnThemeChange';
 import { THEME_CHANGE_EVENT, getHtmlElement } from './utils';
 
-const EnhancedComponent = updateOnThemeChange('div');
+const Dummy = ({ children }) => <div>{children}</div>;
+
+const EnhancedComponent = updateOnThemeChange(Dummy);
```
matteo-hertel: The enhanced component has the PropType `children` required, so I've added a dummy component to comply with the requirement

native/packages/react-native-bpk-component-nudger/src/BpkNudger-test.common.js
matteo-hertel: yes we're expecting an error from the theme provider if the theming object does not contain the appropriate theming properties

PR #636, opened 2018-04-03
[BPK-1311] - Use `react-autosuggest`'s `renderInputComponent` prop correctly

matthewdavidson:
**Background:**

We use a 3rd party component for our autosuggest implementation -> [react-autosuggest](https://github.com/moroshko/react-autosuggest).

Previously, we integrated `BpkInput` into `BpkAutosuggest` by making use of the [`renderInputComponent`](https://github.com/moroshko/react-autosuggest#renderinputcomponent-optional) prop.

As it turns out we were using this incorrectly. Our implementation looked like this:

```js
Autosuggest.defaultProps.renderInputComponent = BpkInput;
```

`react-autosuggest`  expects a function that returns a component, not the component itself.


packages/bpk-component-autosuggest/src/BpkAutosuggest-test.js
```diff
@@ -26,7 +26,12 @@ const onSuggestionsFetchRequested = () => null;
 const onSuggestionsClearRequested = () => null;
 const getSuggestionValue = suggestion => suggestion;
 const renderSuggestion = suggestion => <span>{suggestion}</span>;
-const inputProps = { value: 'Edinburgh', onChange: () => null };
+const inputProps = {
+  id: 'origin',
+  name: 'Origin',
```
matthewdavidson: I had to add `id` and `origin` to the input props because they are required in `BpkInput`. Previously these prop type warnings were not being raised _because_ we were not implementing `renderInputComponent` correctly.

I think the term for what was happening is _squashing_ 🤔

packages/bpk-component-autosuggest/src/BpkAutosuggest-test.js
```diff
@@ -64,23 +69,24 @@ describe('BpkAutosuggest', () => {
...
     const tree = ReactTestUtils.renderIntoDocument(
       <BpkAutosuggest
         suggestions={suggestions}
         onSuggestionsFetchRequested={onSuggestionsFetchRequested}
         onSuggestionsClearRequested={onSuggestionsClearRequested}
         getSuggestionValue={getSuggestionValue}
         renderSuggestion={renderSuggestion}
-        inputProps={inputProps}
-        ref={storeAutosuggestReference}
+        inputProps={{ inputRef: storeAutosuggestReference, ...inputProps }}
```
matteo-hertel: is it worth update the readme/docs with the usage of the inputProps?

matthewdavidson: The docs are here -> https://backpack.github.io/components/web/autosuggest

What would you update in particular?

matteo-hertel: nevermind, it's all covered!

PR #637, opened 2018-04-03
[BPK-1404] fix measue warning

matteo-hertel:
This fixes the warning 
```
  console.warn node_modules/react-native/jest/setup.js:99
    Calling .measure() in the test renderer environment is not supported. Instead, mock out your components that use findNodeHandle with replacements that don't rely on the native environment.
```

jest will load each file in the `__mocks__` folder, in this case the RN Animated module will be mocked.

The Native codebase has no more warning during tests now

michael fassbender perfection gif-source

### DIFF
--- a/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight-test.common.js
+++ b/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight-test.common.js
@@ -23,6 +23,9 @@ import BpkText from 'react-native-bpk-component-text';
 import BpkAnimateHeight from './BpkAnimateHeight';
 
 const commonTests = () => {
+  // Fake timer is needed to prevent Animation warning during the tests
+  jest.useFakeTimers();
+
   describe('BpkAnimateHeight', () => {
     const animateHeightContent = (
       <BpkText>
--- a/native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert-test.common.js
+++ b/native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert-test.common.js
@@ -26,9 +26,10 @@ import { ALERT_TYPES } from './common-types';
 import BpkBannerAlert from './BpkBannerAlert';
 
 const commonTests = () => {
-  describe('BpkBannerAlert', () => {
-    jest.spyOn(Date, 'now').mockImplementation(() => 1503187200000);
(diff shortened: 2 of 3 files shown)


matthewdavidson:
Does this not have to be added to `react-native-bpk-component-animate-height` too?

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert-test.common.js
```diff
@@ -27,6 +27,7 @@ import BpkBannerAlert from './BpkBannerAlert';
 
 const commonTests = () => {
   describe('BpkBannerAlert', () => {
+    jest.useFakeTimers();
     jest.spyOn(Date, 'now').mockImplementation(() => 1503187200000);
```
matthewdavidson: Can `jest.spyOn(Date, 'now').mockImplementation(() => 1503187200000);` be removed then?

matthewdavidson: Also there is an instance of `jest.spyOn(Date, 'now')` in the `AnimateAndFade` tests ... is that still relevant?

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight-test.common.js
```diff
@@ -23,6 +23,7 @@ import BpkText from 'react-native-bpk-component-text';
 import BpkAnimateHeight from './BpkAnimateHeight';
 
 const commonTests = () => {
+  jest.useFakeTimers();
```
matthewdavidson: It pains me to say this, but I think each usage of `jest.useFakeTimers();` warrants a comment...

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert-test.common.js
matteo-hertel: I think it can be removed, I'll see if no snaphots are relying on that

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight-test.common.js
matteo-hertel: this makes me happy, really happy 😄 
will gladly add a comment there

matthewdavidson:
bitmoji

PR #639, opened 2018-04-03
[BPK-1423] - Generate icon font from static codepoints mapping

matthewdavidson:
[`gulp-iconfont`](https://github.com/nfroidure/gulp-iconfont) depends on [`gulp-svgicons2svgfont`](https://github.com/nfroidure/gulp-svgicons2svgfont) which has a [`metadataProvider `](https://github.com/nfroidure/gulp-svgicons2svgfont#optionsmetadataprovider) option.

I use this to roll our own logic for maintaining a static mapping of codepoint <--> icon.

Any time a new icon is added, it will be given a incremented value. See this in action with [the flask icon PR](https://github.com/Skyscanner/backpack/pull/638).


packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
```diff
@@ -0,0 +1,76 @@
...
+    if (codepoints[name]) {
+      return callback(null, {
+        path: file,
+        name,
+        unicode: [String.fromCodePoint(codepoints[name])],
+      });
+    }
+
+    const codepointValues = Object.values(codepoints).map(value =>
+      parseInt(value, 10),
```
matthewdavidson: This is to help flow... any idea how to do this better?

packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
```diff
@@ -0,0 +1,76 @@
...
+    );
+
+    if (codepointValues.length) {
+      const sortedValues = codepointValues.sort();
+      codepoints[name] = sortedValues[sortedValues.length - 1] + 1;
+    } else {
+      codepoints[name] = 59905;
```
matthewdavidson: Same starting codepoint that the [underlying dependency used](https://github.com/nfroidure/gulp-svgicons2svgfont#optionsstartunicode)

packages/bpk-svgs/gulpfile.babel.js
```diff
@@ -317,27 +318,22 @@ gulp.task('icons-font', ['icons-common'], () => {
 
   const saveMapping = generateFont.on('glyphs', glyphs => {
     const baseDir = 'dist/font';
-    // Og all the glyphs generate a key value pair with name and code
     const mapping = glyphs.reduce((acc, glyph) => {
-      // use punycode to get the text representation of the unicode
```
matteo-hertel: I don't agree on removing this comment, looking at the code below it's not clear what's happening, at least the comment explains the output

packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
matteo-hertel: what does `59905` means?

packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
k0nserv: Does it need to be better? I think this is good

packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
```diff
@@ -0,0 +1,76 @@
...
+import path from 'path';
+
+export default (codepointsPath: string) => {
+  const absolutePath = path.join(process.cwd(), codepointsPath);
+
+  const codepoints = fs.existsSync(absolutePath)
+    ? JSON.parse(fs.readFileSync(absolutePath, 'utf8'))
+    : {};
+
+  return (file: string, callback: (err: ?Error, obj: ?Object) => mixed) => {
```
k0nserv: Can this function get called many times and take the later branch? i.e when the codepoint isn't in the file already. If so can't the fact that the reading of the file is memoized cause a bug because we write over it?

matthewdavidson: `codepoints` is read once and kept in memory so it's always up-to-date.

packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
matteo-hertel: we have lodash so you could use the `_.unary` util around `parseInt` and make it with arity 1, the default radix is 10, but it's a massive overkill

packages/bpk-svgs/gulpfile.babel.js
matthewdavidson: I've added it back in

packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
matthewdavidson: I've added a code comment

packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
matthewdavidson: Ok lets leave as is.

matthewdavidson:
@matteo-hertel, @k0nserv 

Can i get some..

bitmoji

PR #642, opened 2018-04-05
SPIKE: [BPK-1431] Changelog generation

k0nserv:
This is the outcome of the spike on generating changelogs. With this I've introduced a `changelog.yml` for the `bpk-component-spinner` package and re-created the changes for that package in it. There are a few new npm scripts here: `npm run generate-changelogs` generates the per packages changelogs from each `changelog.yml`, `npm run upcoming-changes` collects all changes that have not yet been released and prints them to the console. `upcoming-changes` is useful just before doing a release to aid a human in making decisions about the new versions to release. Something I haven't done in this spike is write the script that converts the upcoming section in the packages' changelog.yml to a release or generates the master changelog at the root. For the release I imagine a `postrelease` job that will generate both per package and master changelogs as well as move upcoming changes to the releases section.

`bpk-component-spinner` changelog 

https://github.com/Skyscanner/backpack/blob/9eabf220302dcd52503eec768e08c8df0689ee33/packages/bpk-component-spinner/changelog.yml

## Examples

**Upcoming changes output:**
image

**Changelog in the docs:**
screenshot-2018-4-5 spinners backpack


scripts/release-process/util.js
```diff
@@ -0,0 +1,53 @@
...
+const yaml = require('js-yaml');
+const { promisify } = require('util');
+const path = require('path');
+
+const CHANGELOG_GLOBS = ['packages/*/changelog.yml'];
+const GLOBBY_OPTIONS = {
+  gitignore: true,
+};
+
+const readFileAsync = promisify(fs.readFile);
```
matteo-hertel: 🎖

package.json
```diff
@@ -39,6 +39,8 @@
     "docs:dist": "npm run clean:dist && NODE_ENV=production webpack --progress --display=minimal --bail && npm run sassdoc",
     "start": "npm run build && npm run storybook",
     "native": "./subshell ./native npm run storybook",
+    "generate-changelogs": "node scripts/release-process/generate-changelogs.js",
+    "upcoming-changes": "node scripts/release-process/upcoming-changes.js",
```
matteo-hertel: Not part of the card probably, but we might want to standardise how we handle scripts, we have a mixture of `node {script}.js` and executable files

packages/bpk-component-spinner/changelog.yml
```diff
@@ -0,0 +1,55 @@
+upcoming:
+  changes:
+    minor: We are doing new things here
```
matteo-hertel: I'm assuming this will go before merging?

.gitignore
```diff
@@ -11,3 +11,4 @@ lerna-debug.log
 sassdoc-data.json
 backpack.iml
 yarn.lock
+/packages/*/changelog.md
```
matthewdavidson: Is this just temporary? My understanding is that we'll still have `changelod.md`'s in the repo to link to etc?

scripts/release-process/generate-changelogs.js
```diff
@@ -0,0 +1,58 @@
...
+
+const determineOutputPath = changelogPath => {
+  const basename = path.basename(changelogPath, '.yml');
+  const markdownBasename = `${basename}.md`;
+
+  return path.join(path.dirname(changelogPath), markdownBasename);
+};
+
+const generateHumanReadableChangelog = parsedChangelog =>
+  template(parsedChangelog, 'bpk-component-spinner');
```
matteo-hertel: `bpk-component-spinner` will be a parameter later on correct?

matthewdavidson: This hard coded title `'bpk-component-spinner'` should be derived from the `changelog.yml` right?

.gitignore
k0nserv: Yes this is just temporary, let's get rid of this later

scripts/release-process/generate-changelogs.js
k0nserv: Good catch it should be derived from `package.json`. Let me fix that

packages/bpk-component-spinner/changelog.yml
k0nserv: Eh yeah I'll comment it out

k0nserv:
Comments addressed in 955e385

matteo-hertel:
Flawless! squash and it's good to go!

k0nserv:
I forgot to squash

Giphy

PR #646, opened 2018-04-06
[BPK-1289] Add flow to bpk-component-input

k0nserv:


matthewdavidson:
I have an idea to avoid a breaking change, add this to the `index.js` entry point file:
```js
// remove this on the next major
export const INPUT_TYPES = { TEXT: 'text', blah, blah, blah, ...TYPES };
```

This allows us to use the flow goodness inside the component, whilst maintaining the API for consumers.

We can remove the hackyness on the next warranted major

matthewdavidson:
Can you update all usages to use the new lowercase syntax? `readme.md`'s and usages in `bpk-docs` etc.

unreleased.md
```diff
@@ -3,3 +3,7 @@
 **Added:**
 - react-native-bpk-component-pagination-dots:
   - Introducing the React Native pagination dots component.
+
+**Changed:**
```
matthewdavidson: I don't think this warrants a changelog entry now. Nothing has changed from a consumer POV.

k0nserv: We've deprecated the old keys and intend to remove them. That's definitely something we should communicate IMO

PR #648, opened 2018-04-06
[NO-JIRA] Replace Neo hyperlinks with navlinks

georgegillams:


packages/bpk-docs/src/layouts/NeoSideNav/SideNav.js
```diff
@@ -61,13 +61,13 @@ export default class SideNav extends Component<Props, State> {
     return (
       <section className={getClassName('bpkdocs-side-nav-layout__wrapper')}>
         <section className={getClassName('bpkdocs-side-nav-layout__sidebar')}>
-          <BpkLink href="/">
```
k0nserv: I thought we had a custom router link for Backpack?

georgegillams: A what?

https://media2.giphy.com/media/2MfD8qyUt1t5e/200w.gif?cid=c94812d05ac76183443176736b0f6eae

k0nserv: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-router-link/src/BpkRouterLink.js

georgegillams: Yeah but that applies a load of styling as well as wrapping a NavLink around. Unless we update that component to look different in Neo then it's gonna look wrong

k0nserv: Okay fair point

PR #652, opened 2018-04-10
[NoJira] add android formatter

matteo-hertel:
While updating the tokens in android i found out the wrong value has been used in the `base.raw.android.json` file (see https://github.com/Skyscanner/backpack-android/pull/21)

this fixes it 😄


matteo-hertel:
as discussed offline close this one and move the android codebase

PR #653, opened 2018-04-11
[BPK-1347] Preserve context in portals

matthewdavidson:
Our portals don't preserve react context in their trees.

Previous github issue on react: https://github.com/facebook/react/issues/8262.

Live demo here: https://codesandbox.io/s/2xzrn05rnr

### DIFF
--- a/packages/bpk-react-utils/src/Portal.js
+++ b/packages/bpk-react-utils/src/Portal.js
@@ -16,10 +16,14 @@
  * limitations under the License.
  */
 
+import {
+  unstable_renderSubtreeIntoContainer, // eslint-disable-line camelcase
+  unmountComponentAtNode,
+  findDOMNode,
+} from 'react-dom';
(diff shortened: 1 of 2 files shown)


packages/bpk-react-utils/src/Portal.js
```diff
@@ -243,11 +247,16 @@ class Portal extends Component {
     const missesExpectedTarget = this.props.target && !this.getTargetElement();
 
     if (this.portalElement && !missesExpectedTarget) {
-      render(this.props.children, this.portalElement, () => {
-        if (this.props.isOpen) {
-          this.props.onRender(this.portalElement, this.getTargetElement());
-        }
-      });
+      unstable_renderSubtreeIntoContainer(
```
matthewdavidson: This still works in React 16 - when we upgrade to React 16 and make use of their portals we will be able to delete all this code.

PR #657, opened 2018-04-13
[BPK-1455]: Updated native horizontal nav to match Sketch measurements.

jamesf3rguson:


jamesf3rguson:
Yup, good point and probably the changelog too. Oops!

PR #658, opened 2018-04-13
[NOJIRA] Ensure base stylesheet globals are never scoped by css modules

matthewdavidson:
This could happen when consumers do this in a `backpack-react-scripts` project:

```js
import 'bpk-stylesheets';
```

The `scss` files referenced inside would be processed through css modules and global utility classes (which tbh are frowned upon now) would be scoped.

An example of this in action is here: http://js.skyscnr.com/sttc/oc-registry/components/base-stylesheet/0.0.4/build/static/css/main.1e3e53e5.css

### DIFF
--- a/packages/bpk-stylesheets/index.scss
+++ b/packages/bpk-stylesheets/index.scss
@@ -42,22 +42,22 @@ body {
   font-size: $bpk-font-size-base;
   line-height: $bpk-line-height-base;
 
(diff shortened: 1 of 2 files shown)


k0nserv:
This could end up shadowing other classes with the same names on existing pages thus subtly breaking those pages. How do we roll this out with that in mind?

PR #660, opened 2018-04-16
[BPK-1460] Fix new docs RTL bugs

shaundon:


packages/bpk-mixins/src/mixins/_margins.scss
```diff
@@ -0,0 +1,71 @@
...
+/// @param {string} $margin The desired leading margin.
+/// @param {boolean} $resetUppositeMargin [false] Whether to reset the opposite margin to 0.
+///
+/// @content
+///
+/// @example scss
+///   .selector {
+///     @include bpk-margin-leading($bpk-spacing-sm);
+///   }
+@mixin bpk-margin-leading($margin, $resetOppositeMargin:false) {
```
k0nserv: Come to think of it we should have `resetOppositeMargin` be true by default imo

k0nserv: Could we explore using [keyword arguments](http://thesassway.com/intermediate/leveraging-sass-mixins-for-cleaner-code) too? Especially with the boolean value for `resetOppositeMargin` I think the callsites might be a bit cryptic.

shaundon: I've defaulted it to `true`. Keyword arguments look cool, I didn't know that was a thing.

PR #666, opened 2018-04-18
[BPK-1490] Add disabled styles for bar buttons

k0nserv:
### Android

screenshot_1524050399

### iOS
simulator screen shot - iphone 6 - 2018-04-18 at 12 19 54

## Semver

I conclude that this has to be a breaking change because theming will stop working without intervention from the consumer.

## Todo

- [x] Update the docs screenshot as they will have changed for the modal variant. @jamesf3rguson are you okay with showing the tick icon for the modal example on android?


matthewdavidson:
bitmoji

PR #675, opened 2018-04-20
[BPK-1479] Sort docs tokens by value

georgegillams:


matthewdavidson:
The problem with removing null values means ...

image

matthewdavidson:
Also, sorting by value doesn't with shadow tokens 😕:

image

matteo-hertel:
I've updated the PR with the following:
- refactor a wee bit the the getToken function to make full use of lodash, makes it a bit more readable than before
- If the whole token section is not availabe the token switcher will show "Unavailable"

screenshots below

screen shot 2018-04-24 at 03 07 31-fullpage
screen shot 2018-04-24 at 03 07 45-fullpage

The only section that it's causong problems is shadows, perhaps small bit of custom logic is needed to handle the shadow complex value, but beyond the scope of this PR

packages/bpk-docs/src/helpers/tokens-helper.js
```diff
@@ -69,21 +67,30 @@ export const getTokenValue = (token, platform) => {
...
+  _.chain(keys || Object.keys(tokens))
+    .reduce((acc, key) => {
+      if (!tokens[key]) return acc;
+      acc[key] = tokens[key];
+      return acc;
+    }, {})
+    .sortBy(token => {
+      const numericMatch = `${token.value}`.match(/([0-9]*[.]?[0-9]+)/g);
+      if (!numericMatch) {
+        return -1;
```
matteo-hertel: this puts all the non numeric matches close to each other, will they then be sorted by key

packages/bpk-docs/src/helpers/tokens-helper.js
```diff
@@ -16,14 +16,12 @@
  * limitations under the License.
  */
 
-import union from 'lodash/union';
-import kebabCase from 'lodash/kebabCase';
-
+import _ from 'lodash';
```
tiagoengel: I don't know what is our policy about importing the whole lib rather than individual functions but this means every project using backpack will get the whole lodash lib included too. I see it is necessary for the `chain` logic bellow but this can be written using the functional modules (e.g `lodash/fp/reduce`) that support currying by default. It would become something like this:

`flow(reduce(...), sortBy(...), keyBy(...))(keys || Object.keys(tokens)`

With the added benefit that we will be importing only what's being used.

k0nserv: Good point Tiago, generally we care about this, but this is the docs page which is not consumed by so it's not as important in this case.

tiagoengel: 🤦‍♂️ , forgot it was the doc site. although a smaller bundle size is always better 😀

packages/bpk-docs/src/helpers/tokens-helper.js
matthewdavidson: Web shadows are still in the wrong order here:

image

For non -numeric values sorting by key alphabetically isn't correct because we want `xs` to come before `sm`. Perhaps we should look at an alternative to sorting by value? Adding `order` meta data to token source might be the way to go?

georgegillams:
So following my latest change, the behaviour is as follows:

If all the tokens being displayed have names following the format SOMETHING_LG / SOMETHING_MD etc then sort as follows:
```
xxs => 0
xs => 1
sm => 2
...
```

If they don't all follow that pattern, then sort on numerical value.

This works well for all existing tokens displayed on our docs except for these ones:
screen shot 2018-04-27 at 12 47 45

I wonder whether these should be split out into separated tables anyway, like we do for typesetting.

packages/bpk-docs/src/helpers/tokens-helper.js
```diff
@@ -70,11 +68,54 @@ export const getTokenValue = (token, platform) => {
 };
 
 export const getTokens = (tokens, keys = null) => {
-  const outTokens = {};
-  (keys || Object.keys(tokens)).forEach(key => {
-    outTokens[key] = tokens[key] || null;
-  });
-  return outTokens;
+  const reducedTokens = (keys || Object.keys(tokens)).reduce((acc, key) => {
```
tiagoengel: not: I'd name this `currentTokens` or `validTokens'

packages/bpk-docs/src/helpers/tokens-helper.js
```diff
@@ -70,11 +68,54 @@ export const getTokenValue = (token, platform) => {
...
+    if (!tokens[key]) return acc;
+    acc[key] = tokens[key];
+    return acc;
+  }, {});
+
+  // If all the tokens match our size naming convention then we can
+  // order based on their name. Otherwise we should order on value
+  if (
+    Object.keys(reducedTokens).length > 0 &&
```
tiagoengel: nit: we can replace it with `_.isEmpty({})`

PR #677, opened 2018-04-23
BPK-1520 - Fix android icon alignment

matthewdavidson:
We discovered that our icons on android were not vertically centered in their bounding box for both small and large icons. Turns out there is an android specific prop `includeFontPadding` (see https://facebook.github.io/react-native/docs/text.html#style) that includes extra padding to account for ascenders and descenders. Since these are icons in a 24 / 16 dp bounding box this does not apply so we set it to false.

**Note:** The screenshot for iOS icons looks like it has changed as a result of this but in fact it's just because the screenshot on master was out of date. The first commit illustrates this.

Before | After
------------ | -------------
before | after


k0nserv:
How does this look for the navigation bar back button? That was spot on when I made it. I'm concerned this could have broken that

matthewdavidson:
@k0nserv the screenshots of all affected components have been updated as part of this PR. See https://github.com/Skyscanner/backpack/pull/677/files?short_path=6e2748d#diff-6e2748d79ff505a54d97512254987d92 in you specific example.

k0nserv:
This does indeed break the alignments on Android.

image
*Note that the text baseline doesn't line up with the end of the arrow anymore*

Maybe we need an API to support opting in to the new behaviour?

native/packages/react-native-bpk-component-icon/src/BpkIcon.js
```diff
@@ -32,12 +32,11 @@ const styles = StyleSheet.create({
   icon: {
     fontFamily: 'BpkIcon',
     fontSize: spacingLg,
-    lineHeight: spacingLg,
```
k0nserv: Do we not still need this to ensure we get the exact height we want or is that embedded in the font?

matthewdavidson: Please refer to the images in the desc of this PR. 

The dimensions are now 24 x 24 as expected as a result of removing the line height.

shaundon:
@k0nserv James did say in a previous comment:

> All good from design... we know the alignment is off, but will accept this for now as a re-design of this component is needed.

So maybe the navigation bar misalignment is an acceptable trade-off. I appreciate how hard it was for you to align them in the first place though 😅

k0nserv:
> So maybe the navigation bar misalignment is an acceptable trade-off. I appreciate how hard it was for you to align them in the first place though 😅

No it wasn't actually hard because our icons follow the material spec and the box was the right size. It lined up perfectly from the first go.

PR #678, opened 2018-04-24
BPK-1520 - Align banner alert to sketch

matthewdavidson:
**Note:** this is dependent on https://github.com/Skyscanner/backpack/pull/677

This PR removes the reliance on the line height + padding to achieve overall height and instead uses a min height (48dp) and flex box to achieve vertical centering. There is still padding on the message in case of message overflow.


native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.ios.js
```diff
@@ -77,37 +79,36 @@ const BpkButtonLink = (props: Props) => {
...
-          </BpkText>
-          {typeof icon === 'string' ? (
-            <BpkIcon icon={icon} style={iconStyle} small={!large} />
-          ) : (
-            icon
-          )}
-        </View>
-      </TouchableOpacity>
-    </View>
+    <TouchableOpacity
```
matthewdavidson: I've removed the outer view here - `TouchableOpacity` already wraps `children` in a `<View />`.

native/packages/react-native-bpk-component-button-link/src/styles.js
```diff
@@ -22,28 +22,27 @@ import { StyleSheet } from 'react-native';
...
   colorBlue500,
   colorGray300,
-  borderRadiusSm,
   spacingSm,
-  spacingMd,
+  spacingXl,
 } from 'bpk-tokens/tokens/base.react.native';
 
 const styles = StyleSheet.create({
-  container: {
```
matthewdavidson: This was only ever used on iOS. Moreover it added a border radius to something that doesn't have a visible box therefore it is redundant.

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.ios.js
```diff
@@ -77,37 +79,36 @@ const BpkButtonLink = (props: Props) => {
   }
 
   if (disabled) {
-    textStyle.push(styles.disabled);
+    textStyle.push(styles.textDisabled);
     accessibilityTraits.push('disabled');
   }
 
-  if (style) {
```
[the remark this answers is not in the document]
k0nserv: Ah I missed that

PR #680, opened 2018-04-24
[BPK-1346] Sidebar desegregation and new routes

shaundon:
* Sidebar is no longer split into web/native/utilities, it's just an alphabetical list.
* Routes in the sidebar are now `/components/whatever` instead of `/components/web/whatever`. (Except utilities, wasn't sure if we wanted to do those too so I've left them out initially).

### DIFF
--- a/packages/bpk-docs/src/layouts/NeoSideNavLayout/links-sorter.js
+++ b/packages/bpk-docs/src/layouts/NeoSideNavLayout/links-sorter.js
@@ -22,14 +22,14 @@ export default links =>
     const bHasRoute = !!b.route;
 
     if (aHasRoute === bHasRoute) {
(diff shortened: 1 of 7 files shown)


packages/bpk-docs/src/layouts/NeoSideNavLayout/links-sorter.js
```diff
@@ -22,14 +22,14 @@ export default links =>
...
 
-      if (aId < bId) {
+      if (aName < bName) {
         return -1;
       }
 
-      if (aId > bId) {
+      if (aName > bName) {
         return 1;
       }
```
shaundon: I think we want ordering to be on the name that users see, not the ID, so I changed this.

PR #682, opened 2018-04-24
[BPK-1504] Fix outline buttons on Android 5

k0nserv:
On Android 5(API Level < 23) the outline buttons didn't have an outline.
With this change the outline has moved to the outer view and thus again
works on older versions of Android. An issue for the underlying problem
has also been raised with React Native
https://github.com/facebook/react-native/issues/18999.


k0nserv:
The new Android screenshots were taken on API Level 23 because the background colour on Android 5 is still weird. It does however work on Android 5 now

native/packages/react-native-bpk-component-button/src/BpkButton.ios.js
```diff
@@ -81,13 +81,17 @@ const BpkButton = (props: Props) => {
...
 
   if (disabled) {
     accessibilityTraits.push('disabled');
   }
 
   return (
-    <LinearGradient colors={gradientColors} style={[containerStyle, style]}>
+    <LinearGradient
+      colors={gradientColors}
+      style={[containerStyle, style, containerTheme]}
```
tiagoengel: Shouldn't the user style be the last one?

k0nserv: Good catch, yes they should

matthewdavidson:
@k0nserv can we remove this file and add a gitignore for the folder?

`native/android/captures/com.backpack_2018.04.20_13.50.li`

k0nserv:
> @k0nserv can we remove this file and add a gitignore for the folder?
native/android/captures/com.backpack_2018.04.20_13.50.li

God damnit! Sure I can

matthewdavidson:
The screenshots _currently_ need to be taken on the avd described in the [contributing.md](https://github.com/Skyscanner/backpack/blob/master/contributing.md#react-native).

Notice how they all add a new icon in the status bar?

PR #644, opened 2018-04-06
[BPK-1347] New docs mobile nav

georgegillams:
_Please read inline comments for an overview of this change..._

PR #684, opened 2018-04-25
[BPK-1532] Add navigation bar component

tiagoengel:
- [x] The actual component
- [X] Tests
- [x] Stories
- [x] Theme for links (Should work by default)
- [X] Readme
- [X] unreleased.md 

PS: docs in a different PR

### DIFF
--- a/packages/bpk-component-navigation-bar/package.json
+++ b/packages/bpk-component-navigation-bar/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "bpk-component-navigation-bar",
+  "version": "0.0.0",
+  "description": "Backpack nativation bar component.",
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^15.4.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "bpk-mixins": "^17.8.0",
+    "bpk-react-utils": "^2.5.1",
+    "bpk-component-text": "^1.0.57",
+    "bpk-component-close-button": "^1.0.61",
+    "prop-types": "^15.5.8"
+  },
+  "devDependencies": {
+    "bpk-component-icon": "^3.20.0",
+    "bpk-component-link": "^1.1.15"
+  }
+}
--- a/packages/bpk-component-navigation-bar/readme.md
+++ b/packages/bpk-component-navigation-bar/readme.md
@@ -1,0 +1,75 @@
+# bpk-component-navigation-bar
+
+> Backpack navigation bar component.
+
+## Installation
+
+```sh
+npm install bpk-component-navigation-bar --save-dev
+```
+
+## Usage
+
+### BpkNavigationBar
+
+```js
+import React from 'react';
+import ArrowIcon from 'bpk-component-icon/sm/long-arrow-left';
+import { withRtlSupport } from 'bpk-component-icon';
+import { BpkButtonLink } from 'bpk-component-link';
+
+import BpkNavigationBar, { BpkNavigationBarIconButton } from 'bpk-component-navigation-bar';
+
+const ArrowIconWithRtl = withRtlSupport(ArrowIcon);
+
+export default () => (
+  <BpkNavigationBar
+    title="Backpack"
+    leadingButton={
+      <BpkNavigationBarIconButton
+        onClick={() => {}}
+        icon={ArrowIconWithRtl}
+        label="back"
+      />
+    }
+    trailingButton={
+      <BpkButtonLink onClick={() => {}}>Done</BpkButtonLink>
+    }
+  />
+)
+```
+
+#### Props
+
+| Property              | PropType | Required | Default Value    |
+| --------------------- | -------- | -------- | ---------------- |
+| id                    | string   | true     | -                |
+| title                 | string   | true     | -                |
+| className             | string   | false    | null             |
+| leadingButton         | element  | false    | null             |
+| trailingButton        | element  | false    | null             |
+
+### BpkNavigationBarIconButton
+
+```js
+import React from 'react';
+import ArrowIcon from 'bpk-component-icon/sm/long-arrow-left';
+import { BpkNavigationBarIconButton } from 'bpk-component-navigation-bar';
+
+export default () => (
+  <BpkNavigationBarIconButton
+    onClick={() => {}}
+    icon={ArrowIcon}
+    label="back"
+  />
+)
+```
+
+#### Props
+
+| Property              | PropType | Required | Default Value    |
+| --------------------- | -------- | -------- | ---------------- |
+| icon                  | func     | true     | -                |
+| onClick               | func     | true     | -                |
+| label                 | string     | true     | -              |
+| className             | string   | false    | null             |
--- a/packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
+++ b/packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
@@ -1,0 +1,87 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React, { type Element } from 'react';
+import PropTypes from 'prop-types';
+import { cssModules } from 'bpk-react-utils';
+import BpkText from 'bpk-component-text';
+
+import STYLES from './BpkNavigationBar.scss';
+
+const getClassNames = cssModules(STYLES);
+
+export type Props = {
+  id: string,
+  title: string,
+  className: ?string,
+  leadingButton: ?Element<any>,
+  trailingButton: ?Element<any>,
+};
+
+const cloneWithClass = (elem: Element<any>, newStyle: string) => {
+  const className = getClassNames(elem.props.className, newStyle);
+  return React.cloneElement(elem, { ...elem.props, className });
+};
+
+const BpkNavigationBar = (props: Props) => {
+  const {
+    id,
+    title,
+    className,
+    leadingButton,
+    trailingButton,
+    ...rest
+  } = props;
+
+  return (
+    <nav
+      aria-labelledby={`${id}-bpk-navigation-bar-title`}
+      className={getClassNames('bpk-navigation-bar', className)}
+      {...rest}
+    >
+      {leadingButton &&
+        cloneWithClass(leadingButton, 'bpk-navigation-bar__leading-item')}
+      <BpkText
+        id={`${id}-bpk-navigation-bar-title`}
+        bold
+        className={getClassNames('bpk-navigation-bar__title')}
+      >
+        {title}
+      </BpkText>
+      {trailingButton &&
+        cloneWithClass(trailingButton, 'bpk-navigation-bar__trailing-item')}
+    </nav>
+  );
+};
+
+BpkNavigationBar.propTypes = {
+  title: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  leadingButton: PropTypes.element,
+  trailingButton: PropTypes.element,
+};
+
+BpkNavigationBar.defaultProps = {
+  className: null,
+  leadingButton: null,
+  trailingButton: null,
+};
+
+export default BpkNavigationBar;
--- a/packages/bpk-component-navigation-bar/src/BpkNavigationBarIconButton.js
+++ b/packages/bpk-component-navigation-bar/src/BpkNavigationBarIconButton.js
@@ -1,0 +1,48 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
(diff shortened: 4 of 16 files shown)


packages/bpk-component-navigation-bar/readme.md
```diff
@@ -0,0 +1,55 @@
...
+## Props
+
+| Property              | PropType | Required | Default Value    |
+| --------------------- | -------- | -------- | ---------------- |
+| title                 | string   | true     | -                |
+| className             | string   | false    | null             |
+| leadingButton         | element  | false    | null             |
+| trailingButton        | element  | false    | null             |
+
+## BpkNavigationIconButton Props
```
tiagoengel: not sure if this is the right way to document this secondary component

packages/bpk-react-utils/src/cssModules.js
```diff
@@ -16,5 +16,11 @@
  * limitations under the License.
  */
 
-export default (styles = {}) => className =>
-  styles[className] ? styles[className] : className;
+export default (styles = {}) => (...classNames) =>
```
tiagoengel: can now be used like this:

`
getClassNames('some-class', classNames, disabled && 'disabled')
`

packages/bpk-component-navigation-bar/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^15.4.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "bpk-mixins": "^17.8.0",
+    "bpk-react-utils": "^2.5.1",
+    "bpk-component-text": "^1.0.57",
+    "bpk-component-button": "^2.0.17",
```
matthewdavidson: question this

packages/bpk-component-navigation-bar/readme.md
matthewdavidson: use a third level heading :) `### BpkNavigationBar` and `### BpkNavigationIconButton`

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -0,0 +1,78 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import { cssModules } from 'bpk-react-utils';
+import BpkText from 'bpk-component-text';
+
+import STYLES from './bpk-navigation-bar.scss';
+
+const getClassNames = cssModules(STYLES);
+
+type Props = {
```
matthewdavidson: These should be exported for consumers (remember to re-export in `index.js)

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+/* @flow */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { cssModules } from 'bpk-react-utils';
+import BpkText from 'bpk-component-text';
+
+import STYLES from './bpk-navigation-bar.scss';
```
matthewdavidson: We've started to adopt the same name as the component for co-located stylesheets too i.e. `BpkNavigationBar.scss`

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <nav
+      aria-labelledby="bpk-navigation-bar-title"
+      className={getClassNames('bpk-navigation-bar', className)}
+    >
+      {leadingButton &&
+        cloneWithStyle(leadingButton, 'bpk-navigation-bar__leading-item')}
+      <BpkText
+        id="bpk-navigation-bar-title"
```
matthewdavidson: this ID should be customised by userland i.e `id={``${props.id}-navigation-bar-title``}`

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -0,0 +1,78 @@
...
+const getClassNames = cssModules(STYLES);
+
+type Props = {
+  title: string,
+  className: ?string,
+  leadingButton: ?React$Element<any>,
+  trailingButton: ?React$Element<any>,
+};
+
+const cloneWithStyle = (elem: React$Element<any>, newStyle: string) => {
```
matthewdavidson: Perhaps `cloneWithClassName`?

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -0,0 +1,78 @@
...
+import BpkText from 'bpk-component-text';
+
+import STYLES from './bpk-navigation-bar.scss';
+
+const getClassNames = cssModules(STYLES);
+
+type Props = {
+  title: string,
+  className: ?string,
+  leadingButton: ?React$Element<any>,
```
matthewdavidson: Shouldn't this be `React.Element<any>`? Not sure where the `$` is coming from. Also `type Element` can be imported from the `'react'` package i.e. `import React, { type Element } from 'react';`

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -0,0 +1,78 @@
...
+  trailingButton: ?React$Element<any>,
+};
+
+const cloneWithStyle = (elem: React$Element<any>, newStyle: string) => {
+  const className = getClassNames(elem.props.className, newStyle);
+  return React.cloneElement(elem, { ...elem.props, className });
+};
+
+const BpkNavigationBar = (props: Props) => {
+  const { title, className, leadingButton, trailingButton } = props;
```
matthewdavidson: We should spread `{...rest}` onto the `<nav />`

packages/bpk-component-navigation-bar/src/BpkNavigationIconButton.js
```diff
@@ -0,0 +1,68 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React, { type Node } from 'react';
+import PropTypes from 'prop-types';
+import { cssModules } from 'bpk-react-utils';
+import BpkButton from 'bpk-component-button';
```
matthewdavidson: Im not sure a `BpkButton` should be used here. Technically the configuration you are using it in is unsupported - hence why you have to add a class to it to remove it's padding. 

If consumers did this we'd say that they were violating the internals of the component therefore we shouldn't be doing it either. If we find ourselves doing this then we should ask why and if it's required it should be made an explicit api (i.e. through a react prop.)

Regardless, I think what you really want to use is the `bpk-component-close-button` component - see this comment on the intention with future direction: https://github.com/Skyscanner/backpack/pull/644#discussion_r181335150

packages/bpk-component-navigation-bar/src/bpk-navigation-bar-icon-button.scss
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-navigation-bar-icon-button {
+  padding: unset;
```
matthewdavidson: This is bad :P

packages/bpk-component-navigation-bar/src/bpk-navigation-bar.scss
```diff
@@ -0,0 +1,57 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-navigation-bar {
+  position: relative;
```
matthewdavidson: We should be using flexbox here - it is preferred over absolute positioning - you wouldn't have to handle RTL for instance...

packages/bpk-component-navigation-bar/src/bpk-navigation-bar.scss
```diff
@@ -0,0 +1,57 @@
...
+
+  &__leading-item,
+  &__trailing-item {
+    position: absolute;
+  }
+
+  &__leading-item {
+    left: $bpk-spacing-sm;
+
+    html[dir=rtl] & {
```
matthewdavidson: For future reference - we have a mixin for this ... `@include bpk-rtl {}`

packages/bpk-component-navigation-bar/stories.js
```diff
@@ -0,0 +1,110 @@
...
+import BpkNavigationBar from './index';
+import STYLES from './stories.scss';
+
+const getClassNames = cssModules(STYLES);
+
+const AlignedArrowIcon = withButtonAlignment(withRtlSupport(ArrowIcon));
+const AlignedCloseIcon = withButtonAlignment(withRtlSupport(CloseIcon));
+
+storiesOf('bpk-component-navigation-bar', module)
+  .add('default', () => (
```
matthewdavidson: nit: `Default`

packages/bpk-component-navigation-bar/stories.js
```diff
@@ -0,0 +1,110 @@
...
+        trailingButton={
+          <BpkNavigationIconButton onClick={action('close clicked')}>
+            <AlignedCloseIcon />
+            <span className="visually-hidden">close</span>
+          </BpkNavigationIconButton>
+        }
+      />
+    </div>
+  ))
+  .add('leadingIconOnly', () => (
```
matthewdavidson: Use english in these story names .. i.e. `Leading icon only`

packages/bpk-react-utils/src/cssModules.js
matthewdavidson: bitmoji

packages/bpk-component-navigation-bar/readme.md
```diff
@@ -0,0 +1,56 @@
...
+import BpkNavigationBar, { BpkNavigationIconButton } from 'bpk-component-navigation-bar';
+
+const AlignedArrowIcon = withButtonAlignment(withRtlSupport(ArrowIcon));
+
+export default () => (
+  <BpkNavigationBar
+    title="Backpack"
+    leadingButton={
+      <BpkNavigationIconButton onClick={() => {}}>
+        <AlignedArrowIcon />
```
matthewdavidson: I think alignment should be handled inside the `BpkNavigationIconButton` component, so should any visually hidden text. Perhaps a `label` prop?

packages/bpk-component-navigation-bar/src/BpkNavigationIconButton.js
tiagoengel: The problem with the bpk close button is that we cannot customise the icon, witch is the intention of this component. As for the padding, is being removed because that's how they should look in the nav bar. 

If I were to not use the `BpkButton` component I'd have to implement something similar to the close button that accepts different icons ( which is fine). I just didn't see any reason not to use `BpkButton` and change a bit it's style considering it accepts a `className` prop to do so.

packages/bpk-component-navigation-bar/src/bpk-navigation-bar-icon-button.scss
tiagoengel: using `unset` or removing the padding? :)

packages/bpk-component-navigation-bar/src/bpk-navigation-bar.scss
tiagoengel: I don't think there's a way to implement this with flex because both items can present or not, they can have different sizes and the title need to be always centralised. see:

desktop hd

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
tiagoengel: wouldn't have style then :D.

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
tiagoengel: :shrug: React.Element was not working, I thought that was how I should access the Element type, but I guess import it is the best option

packages/bpk-component-navigation-bar/src/bpk-navigation-bar-icon-button.scss
matthewdavidson: The overriding of style on something that should be considered a private implementation ;)

packages/bpk-component-navigation-bar/src/bpk-navigation-bar.scss
matthewdavidson: Yeah you're right i'd keep as is.

packages/bpk-component-navigation-bar/readme.md
matthewdavidson: Yeah something like that :) or even `icon={Icon}` - that way you wouldn't have to clone the element 😇 .

Consumers already have flexibility by being able to pass anything they want into `leadingButton` i.e.:
```
leadingButton={
  <MyCustomButton />
}
```

georgegillams:
Looking good. Just noticed a couple of things on Android. The extreme title seems to just run off the edge of the screen.

screen shot 2018-04-27 at 16 34 46

Also when I click on the card it causes the app to freeze up. This might just be weirdness on my emulator, but would be worth checking on another emulator or on the Pixel

packages/bpk-component-navigation-bar/stories.scss
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-navigation-bar-story {
+  padding: .09rem .09rem $bpk-spacing-xxl * 2;
```
matthewdavidson: We have `$bpk-one-pixel-rem` for this (See https://backpack.github.io/sassdoc/#spacings-variable-bpk-one-pixel-rem)

packages/bpk-component-navigation-bar/src/BpkNavigationBarIconButton.js
```diff
@@ -0,0 +1,55 @@
...
+import { withRtlSupport } from 'bpk-component-icon';
+
+export type Props = {
+  icon: ComponentType<any>,
+  label: string,
+  onClick: (event: SyntheticEvent<>) => void,
+  className: ?string,
+};
+
+const addRtlSupportIfNeeded = (component: ComponentType<any>) => {
```
matthewdavidson: Can you explain whats happening here?

tiagoengel: Do you mean adding a comment or just explaining here :)? It's basically checking if the icon has been wrapped with rtl support already, to avoid doing so again. Re wrapping wouldn't be a problem though, I just though it'd be better to avoid it if we can.

matthewdavidson: Ah! In that case I think we should remove it altogether - there are some icons that shouldn't be flipped. The rule of thumb is that if the icon conveys direction (i.e. an long-arrow-right) then flip it, if it doesn't (i.e.`chart`) then don't.

We should leave this up to userland :)

PR #686, opened 2018-04-25
[BPK-1501] Implement Neo Redirects

k0nserv:


packages/bpk-docs/src/constants/redirect-routes.js
```diff
@@ -18,6 +18,84 @@
...
+  '/components/native/text': ROUTES.NEO_TEXT,
+  '/components/native/touchable-overlay': ROUTES.NEO_TOUCHABLE_OVERLAY,
+  '/components/native/touchable-native-feedback':
+    ROUTES.NEO_TOUCHABLE_NATIVE_FEEDBACK,
+};
+
+const NEO_REDIRECTS = process.env.BPK_NEO
+  ? {
+      '/components/utilities': ROUTES.COMPONENTS,
+      '/components/web': ROUTES.COMPONENTS,
```
k0nserv: Missing `/components/native`

packages/bpk-docs/src/routes/Routes.js
```diff
@@ -295,59 +323,103 @@ const Routes = (
 
       <Route path={ROUTES.NATIVE_COMPONENTS}>
```
k0nserv: The `oldSite` check could only be done at this level instead of every route below

PR #690, opened 2018-04-27
WIP: [BPK-1533] Introduce navigation stack component

matthewdavidson:
- [x] The actual component
- [ ] Tests
- [x] Storybook examples
- [x] RTL
- [ ] a11y
- [ ] ~Docs~ _future PR_

There is still a problem with the animation if the user adds/pops views too fast, shouldn't change the overall implementation though


k0nserv:
Had a quick look at this locally and it's really nice 👍 Should we have an example where the navigation bar is not part of the navigation stack e.g it doesn't animate. Following on that topic should we have design guidelines for how to use the navigation stack and navigation bar together? cc @jamesf3rguson

PR #693, opened 2018-04-30
[No JIRA] Consume BpkFlatList and BpkSectionList in BpkDialingCodeList

shaundon:


native/packages/react-native-bpk-component-phone-input/src/BpkDialingCodeList.android.js
```diff
@@ -47,20 +40,21 @@ const BpkDialingCodeList = ({
...
           onItemPress(item);
         }}
-        flag={renderFlag(item)}
+        image={<BpkFlag flag={renderFlag(item)} />}
       />
     )}
     keyExtractor={item => item.id}
-    ItemSeparatorComponent={() => <View style={styles.separator} />}
+    ItemSeparatorComponent={BpkFlatListItemSeparator}
+    extraData={selectedId}
```
matthewdavidson: What is `extraData `?

native/packages/react-native-bpk-component-phone-input/src/BpkDialingCodeList.android.js
```diff
@@ -18,25 +18,18 @@
...
 
-import { View, StyleSheet, FlatList } from 'react-native';
 import React from 'react';
-import { spacingMd, colorGray100 } from 'bpk-tokens/tokens/base.react.native';
-import BpkDialingCodeListItem from './BpkDialingCodeListItem';
+import BpkFlatList, {
+  BpkFlatListItem,
+  BpkFlatListItemSeparator,
+} from 'react-native-bpk-component-flat-list';
+import BpkFlag from './BpkFlag';
```
matthewdavidson: Im confused - why were we not using `BpkFlag` before? I cant see it being added here...

shaundon: We were using it inside the list item component, which has been removed in favour of consuming `BpkFlatListItem`. 

It's in use at this higher level because it adds styling to the image passed in by the consumer, which we want to preserve (it sizes the image and adds a border, which is something we want here).

native/packages/react-native-bpk-component-phone-input/src/BpkDialingCodeList.android.js
shaundon: So when I changed the code to use `BpkFlatList`, the stateful storybook example stopped updating on click. 

I looked into it and it seems that it's because of the use of `PureComponent`, and this gives `FlatList` an indication of data that should be used to make re-rendering decisions: https://facebook.github.io/react-native/docs/flatlist.html#extradata (if you do a find on that page for `extraData` there are some more examples and info scattered around too).

shaundon: I've done more digging here and this is what I think is going on, and why this is necessary.

**The problem**
In the storybook, when the selected item changes, a re-render doesn't happen, so it never gets the 'selected' look and feel.

**Why's it happening?**
`FlatList` is a `PureComponent`. In the storybook, we have it inside a stateful wrapper. When state changes, the shallow comparison cannot see that `this.state.selectedId` has changed, so a re-render of the list is not triggered.

This quote from the [`FlatList` docs](https://facebook.github.io/react-native/docs/flatlist.html) explains it better than me:

> By passing extraData={this.state} to FlatList we make sure FlatList itself will re-render when the state.selected changes. Without setting this prop, FlatList would not know it needs to re-render any items because it is also a PureComponent and the prop comparison will not show any changes.

So, ordinarily when `this.state` changes in the wrapper component, the wrapped component (in this case `FlatList`) will be re-rendered. But because it's a `PureComponent`, it doesn't know that `this.state` has changed.

**The solution**
By passing `extraData={selectedId}` into our component, `FlatList` knows to re-render when the `selectedId` prop changes.

PR #696, opened 2018-05-01
[No JIRA] Improve performance for BpkFlatListItem and BpkSectionListItem

shaundon:
* Override `shouldComponentUpdate` to only compare primitive props, as comparing `onPress` was causing long lists to be very slow.
* Add a storybook example for long lists for future performance testing.

### DIFF
--- a/native/packages/react-native-bpk-component-flat-list/stories.js
+++ b/native/packages/react-native-bpk-component-flat-list/stories.js
@@ -51,13 +51,16 @@ const countries = [
 ];
 
 class StatefulBpkFlatList extends React.Component<{
+  extraEntries: number,
   showImages: boolean,
 }> {
   static propTypes = {
+    extraEntries: PropTypes.number,
     showImages: PropTypes.bool,
   };
 
   static defaultProps = {
+    extraEntries: 0,
     showImages: false,
   };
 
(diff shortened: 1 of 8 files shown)


native/packages/react-native-bpk-component-section-list/stories.js
```diff
@@ -78,48 +78,73 @@ const getFlagUriFromCountryCode = countryCode =>
...
+
+  onItemPress = item => {
+    this.setState({ selectedAirport: item.id });
+  };
+
+  getData = () => {
+    const data = airportCities.slice();
+    if (this.props.extraEntries > 0) {
+      const newAirports = [];
+      for (let i = 0; i < this.props.extraEntries; i += 1) {
```
matthewdavidson: You can use `new Array(this.props.extraEntries).fill().map(...)` here...

native/packages/react-native-bpk-component-flat-list/stories.js
```diff
@@ -51,41 +51,57 @@ const countries = [
...
 
   constructor() {
     super();
     this.state = { selectedCountry: 'DJ' };
   }
+
+  getData = () => {
+    const data = countries.slice();
+    if (this.props.extraEntries > 0) {
+      for (let i = 0; i < this.props.extraEntries; i += 1) {
```
matthewdavidson: You can use `new Array(this.props.extraEntries).fill().map(...)` here...

native/packages/react-native-bpk-component-section-list/stories.js
shaundon: Oh yeah, cheers!

PR #699, opened 2018-05-02
[BPK-1552] Make navigation bar themable

tiagoengel:
screen shot 2018-05-02 at 15 02 46
screen shot 2018-05-02 at 15 02 54
screen shot 2018-05-02 at 15 03 04
screen shot 2018-05-02 at 15 03 12

### DIFF
--- a/.storybook/themeableAttributes.js
+++ b/.storybook/themeableAttributes.js
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+/* @flow */
+
 import { themeAttributes as accordionThemeAttributes } from './../packages/bpk-component-accordion';
 import { themeAttributes as barchartThemeAttributes } from './../packages/bpk-component-barchart';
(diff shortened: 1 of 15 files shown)


tiagoengel:
The background is themable too, those screenshots are from the component's stories and I wasn't sure about changing the background there. Do you want me to change it to something else?

jamesf3rguson:
Oh cool. Then all good. We should probably set it up so that when people use the theme switcher on our site the background changes to the primary colour and links, text and icons go white.

tiagoengel:
Actually, that'd be a problem... Right now links have their own theme properties, not related to the navigation bar, so if we want to make the background dark and items white them we'd need to make all links white. I'll update this to have more "general" theme properties that apply to any item.

tiagoengel:
@k0nserv the problem with that is that we need a different style in the navbar than the rest of the page. If we were to style using the default link theme attributes then in the examples above all links would have to be turned white.

PR #706, opened 2018-05-03
[BPK-1552] Fix BpkNavigationBarButtonLink export and update docs

tiagoengel:
I forgot to export the BpkNavigationBarButtonLink in the previous PR 🤦‍♂️ 
I'm also updating the docs page to apply the correct theme.


k0nserv:
Question: How are these imported in `stories.js`? Since the stories worked I guess they weren't imported from `index.js`, if so we should change it there too.

tiagoengel:
Yeah, they were not being imported from `index`. That's a mental note I made to myself "always import from index" :). I copied from another file that was not importing from index and followed the style. I'll update it

PR #708, opened 2018-05-04
New component: Bpk-component-calendar-scroll

martrik:
Hello! I'm Marti from Tyrannosaurus. Some months ago we talked discussed the possibility of creating a new date range picker. We agreed that the first step would be to create a scrollable calendar. Here it is!

We know that there might be some things to change and correct, starting from the component name. Another thing to talk about is wether this component should be focused just for mobile or not. Currently the scroll bar is inside the calendar as this is the correct behavior for mobile, but if used in desktop it should be moved to the side.

Let's start discussing how we can add this component into Backpack :)


matthewdavidson:
Hi @martrik! 

Having had a quick skim over this PR I noticed that the calendar scroll component doesn't depend on `bpk-component-calendar` at all and instead duplicates a lot of the components (see [here](https://github.com/Skyscanner/backpack/tree/67768163f343cc0a360103d402aeaf9f90eda58a/packages/bpk-component-calendar/src)).

I'll close this for now until we get a chance to catch up on the best way forward!

PR #711, opened 2018-05-04
wip: [BPK-1521] Align react and react-native versions

tiagoengel:


native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -72,9 +95,12 @@ class BpkAnimateHeight extends React.Component {
     const toValue = expanded ? height : COLLAPSED_HEIGHT;
     const delay = expanded ? expandDelay : collapseDelay;
 
-    Animated.timing(this.height, { toValue, duration, delay }).start(
-      onAnimationComplete,
-    );
+    this.adjustChildHeight(height, () => {
+      Animated.timing(this.height, { toValue, duration, delay }).start(() => {
+        this.adjustChildHeight.reset();
+        if (onAnimationComplete) onAnimationComplete();
```
k0nserv: Can we have braces here? I'm surprised prettier doesn't do this actually

k0nserv: Can we have braces here? I'm surprised prettier doesn't do this actually

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
```diff
@@ -102,13 +128,16 @@ class BpkAnimateHeight extends React.Component {
           ref={ref => {
             this.innerViewRef = ref;
           }}
-          style={innerStyle}
+          style={[innerStyle, STYLES.measure]}
```
k0nserv: So we render this `View` soley for the side effect of getting a measure? Can we write a comment somewhere explaining this?

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
tiagoengel: sure

native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js
tiagoengel: yeah, after the upgrade the inner view does not render outside its parent height on IOS (unless position is absolute). We need this extra view just to measure the child's height. I've tried with a single child using position absolute but it doesn't behave the same in some cases, especially style wise, if I make the "visible" view absolute it may affect the visual of its childrens.

PR #712, opened 2018-05-04
[BPK-1554] Add themed hero images to neo docs

georgegillams:
This is dependent on `BPK-1454-new-docs-theming`.

My implementation for the hero image works, but is not ideal for the following reasons:
 - The event listening mechanism it uses is not exposed by `bpk-component-theme-toggle`
 - As it relies on the home-page listening to events, switching to the homepage after selecting a theme results in the wrong image being shown.

An alternative would be to change the image using CSS, or to expose our theming through context, so that the HomePage component can see what theme is currently applied


packages/bpk-docs/src/themeableAttributes.js
```diff
@@ -46,6 +46,8 @@ const docsThemeAttributes = [
   'docsSidebarBackground',
   'docsSidebarLink',
   'docsSidebarLinkBorder',
+  'themeName',
+  'logoFillColor',
```
matteo-hertel: This is prop we theme and was throwing an error in the console

georgegillams:
One thing that does need to be changed is the credit and creditHref values for the three images. Will might be able to provide these

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

matteo-hertel:
Nice one @georgegillams I'll add that in

matteo-hertel:
ah! I wrongly assumed we had that info, we can still updated that later on, is not a blocker for this PR
I've updated the render logic to add the a tag only if there is a credit, we'll then update the object to reflect the new info

PR #718, opened 2018-05-08
[No jira] Fix casing on docs

jamesf3rguson:
- Make component anchor links sentence case
- Make app screenshots sub-text sentence case


jamesf3rguson:
This is only for NEO.

PR #720, opened 2018-05-08
[BPK-1546] Web section list components

shaundon:
* Docs to come in a separate PR.
* In future there are going to be more options for this, e.g. title, subtitle, icons. Keeping one eye on that, I think this implementation leaves it loose enough that we can support that in future without making breaking changes.
* I also created three new mixins for adding both a top and bottom border simultaneously. In the end I didn't use them (see below for why) but I figured it's good to have them anyway, now I've already made them.
* One compromise I've made wrt the design is I've only added a bottom border. I originally had a top and bottom one, but then when they stack, you get a double border where the two components touch. I've made the assumption that just a bottom one is ok, because I think there will always be an instance of the header component at the top of the section list (cc @jamesf3rguson to confirm).

screen shot 2018-05-09 at 14 12 54

### DIFF
--- a/packages/bpk-component-section-list/src/BpkSectionList.js
+++ b/packages/bpk-component-section-list/src/BpkSectionList.js
@@ -1,0 +1,38 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import PropTypes from 'prop-types';
+import React, { type Node } from 'react';
+
+type Props = {
+  children: Node,
+};
+
+const BpkSectionList = (props: Props) => {
+  const { children, ...rest } = props;
+
+  return <section {...rest}>{children}</section>;
+};
+
+BpkSectionList.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+export default BpkSectionList;
--- a/packages/bpk-component-section-list/src/BpkSectionListItem.js
+++ b/packages/bpk-component-section-list/src/BpkSectionListItem.js
@@ -1,0 +1,109 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import PropTypes from 'prop-types';
+import React, { type Node } from 'react';
+import { cssModules } from 'bpk-react-utils';
+import BpkLargeChevronRightIcon from 'bpk-component-icon/lg/chevron-right';
+import { withRtlSupport } from 'bpk-component-icon';
+
+import STYLES from './bpk-section-list-item.scss';
+
+const BpkLargeChevronRightIconWithRtlSupport = withRtlSupport(
+  BpkLargeChevronRightIcon,
+);
+
+const getClassName = cssModules(STYLES);
+
+type Props = {
+  blank: boolean,
+  children: Node,
+  className: ?string,
+  href: ?string,
+  onClick: ?(event: SyntheticEvent<>) => void,
+};
+
+const BpkSectionListItem = (props: Props) => {
+  const { blank, children, className, href, onClick, ...rest } = props;
+  const classNames = [
+    getClassName(
+      'bpk-section-list-item',
+      (href || onClick) && 'bpk-section-list-item--interactive',
+      className,
+    ),
+  ];
+
+  if (href) {
+    const target = blank ? '_blank' : null;
+    return (
+      <a
+        href={href}
+        target={target}
+        onClick={onClick}
+        className={classNames.join(' ')}
+        {...rest}
+      >
+        {children}
+        <BpkLargeChevronRightIconWithRtlSupport
+          className={getClassName('bpk-section-list-item__chevron')}
+        />
+      </a>
+    );
+  }
+
+  if (onClick) {
+    return (
+      <button
+        type="button"
+        onClick={onClick}
+        className={classNames.join(' ')}
+        {...rest}
+      >
+        {children}
+        <BpkLargeChevronRightIconWithRtlSupport
+          className={getClassName('bpk-section-list-item__chevron')}
+        />
+      </button>
+    );
+  }
+
+  return (
+    <div className={classNames.join(' ')} {...rest}>
+      {children}
+    </div>
+  );
+};
+
+BpkSectionListItem.propTypes = {
+  children: PropTypes.node.isRequired,
+  blank: PropTypes.bool,
+  className: PropTypes.string,
+  href: PropTypes.string,
+  onClick: PropTypes.func,
+};
+
+BpkSectionListItem.defaultProps = {
+  blank: false,
+  className: null,
+  href: null,
+  onClick: null,
+};
+
+export default BpkSectionListItem;
--- a/packages/bpk-component-section-list/src/bpk-section-list-item.scss
+++ b/packages/bpk-component-section-list/src/bpk-section-list-item.scss
@@ -1,0 +1,52 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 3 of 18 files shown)


packages/bpk-component-section-list/src/BpkSectionListItem.js
```diff
@@ -0,0 +1,101 @@
...
+    </div>
+  );
+};
+
+BpkSectionListItem.propTypes = {
+  children: PropTypes.node.isRequired,
+  blank: PropTypes.bool,
+  className: PropTypes.string,
+  href: PropTypes.string,
+  onClick: PropTypes.func,
```
k0nserv: The prop type of `href` and `onClick` could be stronger to ensure you don't supply both. Same applies for `blank` with `onClick`. Maybe we should have two components?

packages/bpk-component-section-list/src/bpk-section-list-item.scss
```diff
@@ -0,0 +1,52 @@
...
+  color: $bpk-color-gray-700;
+  text-decoration: none;
+
+  @include bpk-border-bottom-sm($bpk-color-gray-100);
+
+  &__chevron {
+    fill: $bpk-color-gray-500;
+
+    @include bpk-rtl {
+      transform: scaleX(-1);
```
k0nserv: We should handle this in JS instead of CSS by using the approriate icons imo

packages/bpk-component-section-list/src/BpkSectionListItem.js
shaundon: > Maybe we should have two components?

We can definitely do that, however I based this off of `BpkLink` which does allow for both `href` and `onclick`: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-link/src/BpkLink.js#L56, so worth bearing that in mind.

packages/bpk-component-section-list/src/bpk-section-list-item.scss
shaundon: Yeah that's fair, I'll do that.

shaundon:
> Should we have a BpkSectionList root item in order to constrain the usage a bit more and make the API better defined. It seems strange to render these straight in a div imo. If nothing else BpkSectionList could enforce the types of children via PropTypes.

Yeah may as well - will do that now.

packages/bpk-component-section-list/src/BpkSectionListItem.js
k0nserv: Fair let's keep it as is then

k0nserv:
Shouldn't `BpkListItem` have `pointer: cursor` in the `button` variant too?

shaundon:
@k0nserv also worth mentioning the future API while this is in flux.

So the plan is (see the jira card) to support a title, subtitle, icon and other things. I was thinking that a good, non-constraining way to do that would be with some more inner components. So some sample usages could be:

```
// With a title.
<BpkSectionListItem>
    <BpkSectionListItem.Title text="Title" />
</BpkSectionListItem>

// With a title and subtitle.
<BpkSectionListItem>
    <BpkSectionListItem.Title text="Title" />
    <BpkSectionListItem.Subtitle text="Subtitle" />
</BpkSectionListItem>
```

..and so on.

We could do it using `<BpkSectionListItem title="Title" subtitle="Subtitle" />`, but some difficulties with this would be:

1. How do we handle the use case of the subtitle being above the title. An `invertTitles` prop?
2. What happens if the user supplies both `title` and `children`?

What do you reckon - we don't need to think about it *too* much right now, but we should ensure we're not going to immediately cause a breaking change if we go down the current route.

shaundon:
> Should we have a BpkSectionList root item in order to constrain the usage a bit more and make the API better defined.

I've now added this.

> Shouldn't BpkListItem have pointer: cursor in the button variant too?

Done.

packages/bpk-component-section-list/src/bpk-section-list-item.scss
shaundon: Actually I'm not sure how to do this in JS, we don't generally access the RTL state in there do we? Do you know of any code samples I can check out?

tiagoengel: Navigation stack is doing this. see: https://github.com/Skyscanner/backpack/blob/dda0a9e869bf3e3936b32efc1d7ca0a5f21b4b4c/packages/bpk-component-navigation-stack/stories.js#L35. There's an internal `isRTL` function here https://github.com/Skyscanner/backpack/blob/dda0a9e869bf3e3936b32efc1d7ca0a5f21b4b4c/packages/bpk-component-navigation-stack/src/utils.js

k0nserv: Hmm we have `withRtlSupport` but that actually just adds a class anyway... Since you are using an icon could you at least change the code to use `withRtlSupport`?

shaundon: Forgot `withRtlSupport` was a thing! Thanks both of you, I've updated the code.

packages/bpk-component-section-list/src/BpkSectionList.js
```diff
@@ -0,0 +1,38 @@
...
+import React, { type Node } from 'react';
+
+type Props = {
+  children: Node,
+};
+
+const BpkSectionList = (props: Props) => {
+  const { children, ...rest } = props;
+
+  return <div {...rest}>{children}</div>;
```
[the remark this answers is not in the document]
k0nserv: Great point Matt 👍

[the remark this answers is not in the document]
shaundon: Clue's in the name I guess 😂

packages/bpk-component-section-list/src/bpk-section-list-item.scss
```diff
@@ -0,0 +1,52 @@
...
+  @include bpk-border-bottom-sm($bpk-color-gray-100);
+
+  &--interactive {
+    cursor: pointer;
+  }
+
+  &__chevron {
+    fill: $bpk-color-gray-500;
+
+    #{$root}:hover & {
```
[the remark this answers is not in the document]
shaundon: bitmoji

packages/bpk-component-section-list/src/BpkSectionListItem.js
```diff
@@ -0,0 +1,109 @@
...
+        <BpkLargeChevronRightIconWithRtlSupport
+          className={getClassName('bpk-section-list-item__chevron')}
+        />
+      </a>
+    );
+  }
+
+  if (onClick) {
+    return (
+      <button
```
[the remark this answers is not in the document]
k0nserv: I thin this is a great point. Personally I think a less flexible API with more semnatic output would be better. Something like 

```javsacript
  <BpkSection>
    <BpkSectionListSection
      header={<BpkSectionListHeader>Cities</BpkSectionListHeader>}
    >
      <BpkSectionListItem>Tokyo</BpkSectionListItem>
      <BpkSectionListItem>Stockholm</BpkSectionListItem>
      <BpkSectionListItem>Edinburgh</BpkSectionListItem>
    </BpkSectionListSection>
  </BpkSection>
```

Maybe?

packages/bpk-component-section-list/src/BpkSectionListItem.js
```diff
@@ -0,0 +1,109 @@
...
+  blank: boolean,
+  children: Node,
+  className: ?string,
+  href: ?string,
+  onClick: ?(event: SyntheticEvent<>) => void,
+};
+
+const BpkSectionListItem = (props: Props) => {
+  const { blank, children, className, href, onClick, ...rest } = props;
+  const classNames = [getClassName('bpk-section-list-item')];
```
tiagoengel: this can be simplified using 
```
getClassName('bpk-section-list-item',
  href || onClick && 'bpk-section-list-item--interactive',
  className
)
```

packages/bpk-mixins/src/mixins/_borders.scss
```diff
@@ -96,6 +109,19 @@
...
+/// Small "1px" top and bottom border.
+///
+/// Applied using box-shadow which preserves vertical rhythm on elements without a fixed height.
+///
+/// @example scss
+///   .selector {
+///     @include bpk-border-top-bottom-sm($bpk-color-gray-100);
+///   }
+
+@mixin bpk-border-top-bottom-sm($color, $inset: inset) {
```
matthewdavidson: We should remove these mixins actually since your no longer using them.

packages/bpk-component-section-list/src/BpkSectionListItem.js
shaundon: Forgot about the new API you made - thanks!

packages/bpk-component-section-list/src/BpkSectionListItem.js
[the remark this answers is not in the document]
shaundon: @k0nserv I like your idea. I think we should compromise our API to preserve semantic output where possible.

To make the API easier to consume, we could make the `header` prop accept both a string or an element, so that most consumers will be able to do this:

```
  <BpkSectionList>
    <BpkSectionListSectionheader="Cities">
      <BpkSectionListItem>Tokyo</BpkSectionListItem>
      <BpkSectionListItem>Stockholm</BpkSectionListItem>
      <BpkSectionListItem>Edinburgh</BpkSectionListItem>
    </BpkSectionListSection>
  </BpkSectionList>
```

If this sounds good I'll go ahead and do it (cc @matthewdavidson)

shaundon:
I've updated the API, it now looks like this:

```
 <BpkSectionList>
    <BpkSectionListSection headerText="Cities">
      <BpkSectionListItem>Tokyo</BpkSectionListItem>
      ...
    </BpkSectionListSection>
  </BpkSectionList>
```

The resulting markup looks like this:

```
<section>
  <section>
    <header>Cities</header>
    <ul>
      <li>Tokyo</li>
      ...
    </ul>
  </section>
</section>
```

I spoke to James who confirmed that headers should only ever contain strings, so I've removed the ability for the user to pass in an element, which simplifies things a lot.

**Note:** Not loving the name `BpkSectionListSection`, but I'm bad at naming things. Any suggestions?

k0nserv:
I think `BpkSectionListSection` is a bit unfortunate as a name, but it's still a good name. `headerText` is required so you can't have a section with a header? Also can you have margins etc between sections?

shaundon:
@k0nserv ones without a header aren't on the spec (see attached) but I'll confirm with James.

examples

shaundon:
> Also can you have margins etc between sections?

It wasn't on the spec, but it has support for `className` via `...rest` so consumers can do it if they want.

PR #721, opened 2018-05-08
[BPK-1557]: Make the first paragraph of a component's intro blurb bigger

jamesf3rguson:


shaundon:
This would technically work, but I'm not sure how BEM it is to use the `:first-child` selector. 

However I can see the rationale in this, as if we only want to make the first paragraph of the blurb larger (i.e. not every part of the blurb), this will take the weight of making that styling happen away from consumers.

The alternative would be to make a new `BlurbParagraph` type component and go through each blurb, changing it to use that. `BlurbParagraph` would implement the same CSS pretty much.

jamesf3rguson:
Yeh I kind of wanted to avoid the new component route to save having to go through every file, but if thats the best way happy to suck it up. :P

matthewdavidson:
Yeah you need to use a different component for this @jamesf3rguson.

georgegillams:
If you used a regex-find-replace to change all the page blurbs to string values instead of nodes, then you could just replace `Blurb` with `BigBlurb` [here](https://github.com/Skyscanner/backpack/blob/c5baa620b358754e3912b811814f2a068dde5c37/packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js#L75) and it could do all the formatting required... Might be easier?

jamesf3rguson:
Righto... leave it with me. Thanks for looking at this.

________________________________________

packages/bpk-docs/src/components/neo/IntroBlurb/IntroBlurb.js
```diff
@@ -0,0 +1,37 @@
...
+
+import BpkText from 'bpk-component-text';
+
+import { cssModules, withDefaultProps } from 'bpk-react-utils';
+
+import STYLES from './IntroBlurb.scss';
+
+const getClassName = cssModules(STYLES);
+
+export const ParagraphNoMargin = withDefaultProps(BpkText, {
```
shaundon: This can be removed as it's not used.

packages/bpk-docs/src/components/neo/IntroBlurb/index.js
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import IntroBlurb, { IntroBlurbNoMargin } from './IntroBlurb';
+
+export { IntroBlurbNoMargin };
```
shaundon: `IntroBlurbNoMargin` isn't being used anywhere, so we can get rid of it.

packages/bpk-docs/src/components/neo/IntroBlurb/IntroBlurb.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import BpkText from 'bpk-component-text';
+
```
shaundon: We don't need breaks between each `import`, we usually bunch them together.

jamesf3rguson:
Thanks @shaundon - hopefully all good now.

PR #722, opened 2018-05-08
[NO-JIRA]: Change anchor links title

jamesf3rguson:
The anchor links title appears on all pages, not just components and therefore does not make sense to name 'Available configurations'. 'In this section' is more generic and can apply to all pages.

### DIFF
--- a/packages/bpk-docs/src/components/neo/PageHead/PageHead.js
+++ b/packages/bpk-docs/src/components/neo/PageHead/PageHead.js
@@ -56,9 +56,9 @@ const PageHead = (props: Props) => {
         <BpkText
           bold
           tagName="h2"
-          className={getClassName('bpkdocs-page-head__available-configs')}
+          className={getClassName('bpkdocs-page-head__in-section')}
         >
-          Available configurations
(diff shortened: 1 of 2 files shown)


packages/bpk-docs/src/components/neo/PageHead/PageHead.js
```diff
@@ -56,9 +56,9 @@ const PageHead = (props: Props) => {
         <BpkText
           bold
           tagName="h2"
-          className={getClassName('bpkdocs-page-head__available-configs')}
+          className={getClassName('bpkdocs-page-head__in-section')}
         >
-          Available configurations
+          In this section:
```
shaundon: I don't think this should end in a colon. It's a lead-in statement to a list, but it isn't a complete sentence on its own, which, as a rule of thumb, dictates when the colon should appear.

https://www.quickanddirtytips.com/education/grammar/formatting-vertical-lists?page=2#introductory-colons

jamesf3rguson:
Good spot! No worries, fixed it and pushed.

PR #724, opened 2018-05-09
[No JIRA] Refine pagehead in neo

shaundon:
* _In this section_ now only appears if `menu.length > 0`.
* `PageHead` only appears if there's a blurb or menu.


k0nserv:
@shaundon The changes look good. Can you clarify which pages we are making these changes for?

shaundon:
@k0nserv yep sure - so it's for some of the token pages, like this one: https://backpack-neo.github.io/tokens/borders

PR #726, opened 2018-05-09
[BPK-1533] Introduce navigation stack component

tiagoengel:


packages/bpk-component-navigation-stack/readme.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Usage
+
+```js
+import React from 'react';
+import { updateOnDirectionChange } from 'bpk-component-rtl-toggle';
+import BpkNavigationStack from 'bpk-component-navigation-stack';
+
+const RtlAwareNavigationStack = updateOnDirectionChange(BpkNavigationStack);
```
matthewdavidson: I'd remove the RTL aware code from the readme. This is mostly there for our storybooks so that we don;t need to reload the page.

packages/bpk-component-navigation-stack/readme.md
```diff
@@ -0,0 +1,53 @@
...
+const StatefulNavigationStack = withNavigationStackState(RtlAwareNavigationStack);
+
+export default () => (
+  <StatefulNavigationStack initialViews={[<div key="1" />]} />
+);
+```
+
+By default the `pushView` and `popView` callbacks are forwarded to the childrens, if you
```
matthewdavidson: `childrens` -> `children`

packages/bpk-component-navigation-stack/src/BpkNavigationStack.js
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  static defaultProps = {
+    className: null,
+  };
+
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      transition: undefined,
```
matthewdavidson: `null` would be better than `undefined` right? `null` explicitly means no value whereas `undefined` is up to interpretation...

packages/bpk-component-navigation-stack/src/BpkNavigationStack.js
```diff
@@ -0,0 +1,125 @@
...
+            })`,
+          }}
+        >
+          {Children.map(views, (view, idx) => (
+            <Transition
+              in
+              key={idx}
+              timeout={parseInt(durationSm, 10)}
+              unmountOnExit
+              onEnter={this.handleViewEnter}
```
matthewdavidson: We usually match these callbacks to the name of the prop being assigned to.. i.e:
```
onEnter={this.onEnter}
// or
onEnter={this.onTransitionEnter}
```

packages/bpk-component-navigation-stack/src/BpkNavigationStack.js
tiagoengel: yeah, I was having problems with flow and put undefined just to make it happy, but forgot to change

packages/bpk-component-navigation-stack/readme.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkNavigationStack from 'bpk-component-navigation-stack';
+
+export default () => (
+  <BpkNavigationStack views={[<div key="1" />]} />
```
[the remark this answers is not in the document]
tiagoengel: It was triggering a waring during tests. As far as I know any array of elements needs keys on React. This can be dynamic and react will use the key to decide if it should remount the item in the DOM or just update it.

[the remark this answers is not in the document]
matthewdavidson: `React.Children.toArray` can be used (see https://reactjs.org/docs/react-api.html#reactchildrentoarray).

I don't think we should force consumers to add keys to the views they are passing.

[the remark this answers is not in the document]
tiagoengel: Not sure why the warning is not being triggered in the storybook examples too.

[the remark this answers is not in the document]
tiagoengel: I think `Children.toArray` is supposed to be used to flatten the child structure into an array not to add keys to an array (it adds but that's a consequence, not the objective). I'll remove the key from the docs though, it doesn't really seem to trigger a warning in the stories, only in the tests.

PR #733, opened 2018-05-10
[No JIRA] Add missing redirects for neo

k0nserv:


packages/bpk-docs/src/constants/redirect-routes.js
```diff
@@ -78,6 +79,7 @@ const NEO_NATIVE_REDIRECTS = {
   '/components/native/picker': ROUTES.NEO_PICKER,
   '/components/native/progress': ROUTES.NEO_PROGRESS,
   '/components/native/section-list': ROUTES.NEO_SECTION_LIST,
+  '/components/native/SELECT': ROUTES.NEO_SELECT,
```
shaundon: Should this be uppercase? Looks wrong compared to the others..

k0nserv: No it shouldn't, but strangely it seems to work

PR #746, opened 2018-05-11
[BPK-1583] Apply max-width to neo screenshots

shaundon:
See attached low-res af gif
kapture 2018-05-11 at 11 04 46


willhg:
Yep! looks great.

PR #747, opened 2018-05-11
[BPK-1584] Add sticky footer to docs layout

georgegillams:
**UPDATE 12/05/2018** (Yes, that's a saturday 😆)

When I first implemented this, the solution did not cater for Safari versions < 11.1 or IE 11.

I fixed IE 11 with a [small hack](https://github.com/Skyscanner/backpack/pull/747/files#diff-4e104caa9fb361ea77529d88c900d991R35) that is applied to IE only.

I fixed the Safari issues with blood, sweat and anger.

Now for screenshots:

Chrome:
chrome

Edge:
edge

Firefox (Quantum):
firefox_quantum

IE:
internet_explorer

IOS:
ios

Safari 10.1:
safari_10 1

Safari 11.1:
safari_11 1

### DIFF
--- a/packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
+++ b/packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
@@ -20,7 +20,21 @@
 
 .bpkdocs-default-layout {
   &__container {
-    background-color: $bpk-color-gray-100;
+    position: absolute;
+    display: flex;
+    width: 100%;
+    min-height: 100%;
+    flex-direction: column;
+  }
+
+  &__content-container {
+    display: flex;
+    flex-direction: column;
+    flex: 1 0 auto;
+
+    @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+      min-height: calc(100vh - 5.25rem); /* Hack for Internet Explorer: */
+    }
   }
 
   &__footer-container {
--- a/packages/bpk-docs/src/layouts/NeoSideNavLayout/SideNavLayout.scss
+++ b/packages/bpk-docs/src/layouts/NeoSideNavLayout/SideNavLayout.scss
@@ -20,6 +20,7 @@
 
 .bpkdocs-side-nav-layout {
   display: flex;
+  flex: 1 0 auto;
 
   &__sidebar-destop-wrapper {
     display: flex;
@@ -36,6 +37,6 @@
 
   &__main {
     min-width: 0;
-    flex: 1;
+    flex: 1 1 auto;
   }
 }
--- a/packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
+++ b/packages/bpk-docs/src/layouts/DefaultLayout/DefaultLayout.js
@@ -75,13 +75,15 @@ class DefaultLayout extends Component {
       <EnhancedThemeProvider
         themeAttributes={themeAttributes}
         id="portal-target"
-        className={
-          process.env.BPK_NEO
-            ? getClassName('bpkdocs-default-layout__container')
-            : null
-        }
       >
-        <div id="application-container">
+        <div
+          id="application-container"
+          className={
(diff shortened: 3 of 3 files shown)


matthewdavidson:
screen shot 2018-05-11 at 12 01 51

packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
```diff
@@ -20,7 +20,21 @@
...
+    min-height: 100%;
+    flex-direction: column;
+  }
+
+  &__content-container {
+    display: flex;
+    flex-direction: column;
+    flex: 1 0 auto;
+
+    @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
```
georgegillams: This is a hack that is required for IE 11 only. It is very hacky, but the worst case scenario is that we change the footer size, and the alignment goes slightly wrong in IE 11.

packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
```diff
@@ -20,7 +20,21 @@
 
 .bpkdocs-default-layout {
   &__container {
-    background-color: $bpk-color-gray-100;
+    position: absolute;
+    display: flex;
+    width: 100%;
```
georgegillams: This is necessary as some browsers (including IE and Edge) do not apply the flex properly in order to fill the browser window, making the webpage narrower than it should be

packages/bpk-docs/src/layouts/NeoSideNavLayout/SideNavLayout.scss
```diff
@@ -20,6 +20,7 @@
 
 .bpkdocs-side-nav-layout {
   display: flex;
+  flex: 1 0 auto;
```
georgegillams: causes the side-nav layout to fill the space available vertically - see Matt's screenshot

tiagoengel:
LGTM, works on safari 11.0 and IE \o/. There is only one thing I've noticed that is a little bit different than before:

Before
screen shot 2018-05-14 at 09 02 10

After
screen shot 2018-05-14 at 09 01 33

Notice the white background. I think is not a big deal and we can fix it in a different ticket, but I'd like someone else to approve this

packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
tiagoengel: can we add this comment in the code? explaining why this is necessary?

tiagoengel:
There's another ticket to fix the background thing https://gojira.skyscanner.net/projects/BPK/issues/BPK-1587?filter=allopenissues&orderby=priority+ASC%2C+updated+DESC

packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
shaundon: Not a fan of this sort of hack, where there's a line that only one browser can see.. If we do go down this road I think we should at least make it a mixin, so it's a bit clearer:

```
@include internet-explorer-only {
  // Code here
}
```

packages/bpk-docs/src/layouts/DefaultLayout/default-layout.scss
```diff
@@ -20,7 +20,21 @@
...
+    flex-direction: column;
+  }
+
+  &__content-container {
+    display: flex;
+    flex-direction: column;
+    flex: 1 0 auto;
+
+    @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+      min-height: calc(100vh - 5.25rem); /* Hack for Internet Explorer: */
```
shaundon: `5.25rem` being the height of the footer? We should set this in a variable somewhere.

shaundon:
Closing this (but keeping the branch around) as we've decided it's not a blocker for release.

PR #748, opened 2018-05-11
[BPK-1580] Fix navigation bar themed colours

georgegillams:


georgegillams:
Woops - didn't realise Tiago was already on this 🤦‍♂️

PR #750, opened 2018-05-11
[BPK-1580] Add missing themes for the neo sidebar

tiagoengel:
screen shot 2018-05-11 at 14 04 37
screen shot 2018-05-11 at 14 04 22
screen shot 2018-05-11 at 14 58 31
screen shot 2018-05-11 at 14 04 01

### DIFF
--- a/packages/bpk-docs/src/layouts/NeoSideNavLayout/SectionsList.scss
+++ b/packages/bpk-docs/src/layouts/NeoSideNavLayout/SectionsList.scss
@@ -41,6 +41,7 @@ $bpk-heading-button-padding: $bpk-spacing-md;
     @include bpk-text;
     @include bpk-text-xl;
     @include bpk-border-bottom-sm($bpk-color-gray-700);
+    @include bpk-border-bottom-sm(var(--bpk-docs-sidebar-link-border, $bpk-color-gray-700));
 
     @include bpk-rtl {
       text-align: right;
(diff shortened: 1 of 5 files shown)


packages/bpk-docs/src/layouts/NeoSideNavLayout/SectionsList.scss
```diff
@@ -40,7 +40,7 @@ $bpk-heading-button-padding: $bpk-spacing-md;
 
     @include bpk-text;
     @include bpk-text-xl;
-    @include bpk-border-bottom-sm($bpk-color-gray-700);
+    @include bpk-border-bottom-sm(var(--bpk-docs-sidebar-link-border, $bpk-color-gray-700));
```
shaundon: `var` doesn't work on IE, so we'll need to do both here:

```
@include bpk-border-bottom-sm($bpk-color-gray-700); // Fallback
@include bpk-border-bottom-sm(var(--bpk-docs-sidebar-link-border, $bpk-color-gray-700));
```

You don't need to do this when using `bpk-themeable-property` because that adds the fallback automatically.

packages/bpk-docs/src/layouts/NeoSideNavLayout/SectionsList.scss
```diff
@@ -93,7 +93,7 @@ $bpk-heading-button-padding: $bpk-spacing-md;
     background: none;
     text-align: left;
 
-    @include bpk-border-bottom-sm($bpk-color-gray-700);
+    @include bpk-border-bottom-sm(var(--bpk-docs-sidebar-link-border, $bpk-color-gray-700));
```
shaundon: Same as above comment.

PR #751, opened 2018-05-11
[BPK-1581] Fix docs title issue

shaundon:
In #704 I removed some things from `NeoDocsPageBuilder` because I thought *all* pages were now 'wrapped' (inside an instance of `DocsPageWrapper`). Turns out this was wrong, because while all _component_ pages are wrapped, tokens and other pages are not.

This PR reverts lots of the previous PR to reinstate the correct appearance. It also fixes the issue where `window.title` wasn't always updated as users moved between pages.

**Before this PR**
screen shot 2018-05-11 at 14 06 13

**After this PR**
screen shot 2018-05-11 at 14 05 58


jamesf3rguson:
The headings are still missing from the page lol

PR #757, opened 2018-05-16
Add a basic map component 

kenneyyang:
add a basic map component and go through a merge request review workflow
we will add maker using another merge request


packages/bpk-component-map/readme.md
```diff
@@ -0,0 +1,117 @@
...
+export default () => (
+  <BpkMap  
+    zoom={15}
+    centerLatitude={55.944357}
+    centerLongitude={-3.1967116}
+    zoomEnabled={false}
+    dragEnabled={false}
+    width="600px"
+    height="600px"
+    language="zh" />
```
micetti: I can't spot a difference between this example and the one above

packages/bpk-component-map/readme.md
```diff
@@ -0,0 +1,116 @@
...
+import BpkMap from 'bpk-component-map';
+
+export default () => (
+  <BpkMap  
+    boundSouth={55.94129273544452}
+    boundWest={-3.2285547854247625}
+    boundNorth={55.952707392208396}
+    boundEast={-3.159632742578083}
+    width="600px"
+    height="600px"
```
micetti: What is the behaviour if the south-west-north-east boundaries don't form the same aspect ratio as width and height? will it be stretched, zoomed...  throw an error?

kenneyyang: it will be zoomed with suitable level.

micetti: thanks.

shaundon:
We think that at this point in the component's development, it makes most sense to make it mostly just a wrapper to the underlying component, and not including most of the custom logic that's been included here. (Although adding an `onMapLoaded` function makes sense, we think).

Take a look at [the code for `BpkAutosuggest`](https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-autosuggest/src/BpkAutosuggest.js). Backpack makes some very small additions and then re-exports the component largely untouched.

It's usually better to let consumers just use the underlying props rather than changing them. This makes the component easier to maintain on our side when `react-google-maps` changes too. It also means that we don't need to worry about maintaining the readme file, we can just direct consumers to the underlying one.

We can add more things to this component in future, for example when custom map markers are created, we can then export a component for this and provide value.

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -0,0 +1,192 @@
...
+      zoom,
+      language,
+      region,
+      zoomEnabled,
+      dragEnabled,
+      onZoom,
+      onDrag,
+      ...rest
+    } = this.props;
+    const InnerMap = withScriptjs(
```
k0nserv: We shouldn't be doing this here, it creates new functions everytime which breaks React's reconciling

shaundon:
Closing in favour of #773.

PR #762, opened 2018-05-18
[BPK-1588] Add image to web nav bar

georgegillams:


packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -59,7 +59,7 @@ class DocsPageWrapper extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = { selected: this.props.nativeSubpage ? 'native' : 'web' };
+    this.state = { selected: 'web' };
```
georgegillams: undo this!

packages/bpk-docs/src/pages/NavigationBarPage/NavigationBarPage.js
```diff
@@ -118,6 +118,65 @@ const components = [
...
+    title: 'With icon in title',
+    blurb: [
+      <Paragraph>The title can include elements other than text.</Paragraph>,
+    ],
+    examples: [
+      <div className={getClassNames('bpk-navigation-bar-example')}>
+        <BpkNavigationBar
+          id="default-bpk-nav"
+          title={
+            <div style={{ display: 'flex', flexDirection: 'row' }}>
```
georgegillams: replace with an image

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -72,7 +72,7 @@ const BpkNavigationBar = (props: Props) => {
 };
 
 BpkNavigationBar.propTypes = {
-  title: PropTypes.string.isRequired,
+  title: PropTypes.node.isRequired,
```
tiagoengel: this should be `element`

packages/bpk-component-navigation-bar/stories.js
```diff
@@ -107,4 +107,35 @@ storiesOf('bpk-component-navigation-bar', module)
...
       />
     </div>
+  ))
+  .add('With icon', () => (
+    <div className={getClassNames('bpk-navigation-bar-story')}>
+      <BpkNavigationBar
+        id="test"
+        title={
+          <div style={{ display: 'flex', alignItems: 'center' }}>
+            <img
```
tiagoengel: Wouldn't it be easier to use `BpkIcon`. If users want to use something else they can but as an example I think we should use an icon.

packages/bpk-component-navigation-bar/stories.js
```diff
@@ -107,4 +107,35 @@ storiesOf('bpk-component-navigation-bar', module)
         }
       />
     </div>
+  ))
+  .add('With icon', () => (
+    <div className={getClassNames('bpk-navigation-bar-story')}>
+      <BpkNavigationBar
+        id="test"
+        title={
+          <div style={{ display: 'flex', alignItems: 'center' }}>
```
tiagoengel: can we get rid of the inline styles and use a class? Also, I think a `span` is semantically more appropriate here.

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -29,7 +29,7 @@ const getClassNames = cssModules(STYLES);
 
 export type Props = {
   id: string,
-  title: string,
+  title: string | Element<any>,
```
tiagoengel: can we add a new test that renders with an element instead of a string?

georgegillams: Done 👍

packages/bpk-component-navigation-bar/readme.md
```diff
@@ -44,7 +44,7 @@ export default () => (
 | Property              | PropType | Required | Default Value    |
 | --------------------- | -------- | -------- | ---------------- |
 | id                    | string   | true     | -                |
-| title                 | string   | true     | -                |
+| title                 | node     | true     | -                |
```
tiagoengel: this needs to be changed to `element` to

unreleased.md
```diff
@@ -3,3 +3,6 @@
 **Added:**
 - react-native-bpk-component-pagination-dots:
   - Introducing the React Native pagination dots component.
+
+- bpk-component-navigation-bar:
+  - The title prop now accepts nodes, so images and other elements can be passed in as well as text.
```
tiagoengel: "accepts elements"

tiagoengel:
@georgegillams I've talked with @jamesf3rguson and we're going to change a little bit the logo to use only one image and no text. That reflects a little bit better the use case they're expecting from it. But otherwise looks good.

PR #763, opened 2018-05-18
[NO-JIRA] Enable deep linking to web vs native pages

georgegillams:
Some things you can try:
http://localhost:8080/components/web/badge redirects to http://localhost:8080/components/badge
http://localhost:8080/components/badge?platform=web#light will open the badge page with web selected, and the # works to take you to the `Light` section

Similarly
http://localhost:8080/components/button/?platform=web#featured
http://localhost:8080/components/button/?platform=native#featured

Unfortunately adding redirects from `/components/web/X` to `/components/X?platform=web` doesn’t work as react router can’t deal with it

### DIFF
--- a/packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
+++ b/packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
@@ -18,6 +18,7 @@
 
 import PropTypes from 'prop-types';
 import React from 'react';
+import { withRouter } from 'react-router';
 import BpkContentContainer from 'bpk-component-content-container';
 import BpkHorizontalNav, {
   BpkHorizontalNavItem,
@@ -46,6 +47,11 @@ const contentShape = PropTypes.oneOfType([PropTypes.string, childrenPropType]);
 class DocsPageWrapper extends React.Component {
   static propTypes = {
     title: PropTypes.string.isRequired,
+    router: PropTypes.shape({
+      isActive: PropTypes.func,
+      replace: PropTypes.func,
+      push: PropTypes.func,
+    }).isRequired,
     blurb: contentShape,
     webSubpage: PropTypes.element,
     nativeSubpage: PropTypes.element,
@@ -61,33 +67,34 @@ class DocsPageWrapper extends React.Component {
     super(props);
 
     this.state = {
-      selected: this.props.nativeSubpage ? 'native' : 'web',
+      selected: this.getSelectedSubpage(),
     };
   }
 
   componentWillMount = () => {
-    if (typeof window !== 'undefined') {
-      const url = new URL(window.location);
-      const platform = url.searchParams.get('platform');
-      if (platform) {
-        this.state = {
-          selected: platform === 'native' ? 'native' : 'web',
-        };
-      }
+    const { router } = this.props;
+    const activePlatform = ['native', 'web'].find(platform =>
+      router.isActive({ query: { platform } }),
+    );
+    if (activePlatform) {
+      this.setState({ selected: activePlatform });
+    } else if (typeof window !== 'undefined') {
+      this.props.router.replace({
+        pathname: window.location.pathname,
+        query: { platform: this.getSelectedSubpage() },
+      });
     }
   };
 
   onClick = selected => {
-    if (typeof window !== 'undefined') {
-      const simpleWindowLocation = `${window.location}`.split('?')[0];
-      window.history.pushState(
-        `platform${selected}`,
-        'Title',
-        `${simpleWindowLocation}?platform=${selected}`,
-      );
-    }
+    this.props.router.push({
+      pathname: window.location.pathname,
+      query: { platform: selected },
+    });
     this.setState({ selected });
   };
+
+  getSelectedSubpage = () => (this.props.nativeSubpage ? 'native' : 'web');
 
   render() {
     const { blurb, nativeSubpage, title, webSubpage } = this.props;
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -185,5 +185,8 @@
       "jest",
       "babel-jest"
     ]
+  },
+  "dependencies": {
+    "query-string": "^6.1.0"
```
shaundon: This should be a dep of `bpk-docs` right?

packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -59,9 +60,31 @@ class DocsPageWrapper extends React.Component {
...
-    this.state = { selected: this.props.nativeSubpage ? 'native' : 'web' };
+
+    this.state = {
+      selected: this.props.nativeSubpage ? 'native' : 'web',
+    };
   }
 
+  componentWillMount = () => {
+    const urlParams = queryString.parse(window.location.search);
+    if (urlParams.platform) {
```
shaundon: I think we should validate this too, just to prevent any silly issues where platform is `foo` and then the page display gets messed up.

package.json
georgegillams: Yes it should - good catch!

packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -59,9 +59,36 @@ class DocsPageWrapper extends React.Component {
...
-    this.state = { selected: this.props.nativeSubpage ? 'native' : 'web' };
+
+    this.state = {
+      selected: this.props.nativeSubpage ? 'native' : 'web',
+    };
   }
 
+  componentWillMount = () => {
+    if (typeof window !== 'undefined') {
+      const url = new URL(window.location);
```
shaundon: No IE support 😢 https://caniuse.com/#search=url

shaundon:
@tiagoengel turns out we do need to check if `window` is defined 😂 (this was on CI)

screen shot 2018-05-22 at 09 01 59

shaundon:
@tiagoengel I liked your method and have implemented it, take another look please 😇

shaundon:
> For me the ideal would be to replace all routes to always have the platform param by default, but that's probably a lot more work.

I think we can do a follow-up PR soon so that it loads the component smartly, using the selected platform in the sidebar.

packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -59,9 +63,50 @@ class DocsPageWrapper extends React.Component {
...
+        pathname: window.location.pathname,
+        query: { platform: selected },
+      });
+    }
+    this.setState({ selected });
+  };
+
+  getSelectedSubpage = () => (this.props.nativeSubpage ? 'native' : 'web');
+
+  getUrlParameter = paramName => {
```
tiagoengel: are we still using this function?

packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -59,9 +63,50 @@ class DocsPageWrapper extends React.Component {
...
+    };
   }
 
+  componentWillMount = () => {
+    const { router } = this.props;
+    const activePlatform = ['native', 'web'].find(platform =>
+      router.isActive({ query: { platform } }),
+    );
+    if (activePlatform) {
+      this.state = { selected: activePlatform };
```
tiagoengel: I think this should be `setState` right?

packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -46,6 +47,9 @@ const contentShape = PropTypes.oneOfType([PropTypes.string, childrenPropType]);
 class DocsPageWrapper extends React.Component {
   static propTypes = {
     title: PropTypes.string.isRequired,
+    // https://github.com/ReactTraining/react-router/issues/5129#issuecomment-302131982
+    // eslint-disable-next-line react/forbid-prop-types
+    router: PropTypes.object.isRequired,
```
tiagoengel: can't we use a shape here? i.e. `PropTypes.shape({ isActive: PropType.func, replace: PropType.func, push: PropTypes.func })`

packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -59,9 +63,50 @@ class DocsPageWrapper extends React.Component {
...
+    if (typeof window !== 'undefined') {
+      this.props.router.push({
+        pathname: window.location.pathname,
+        query: { platform: selected },
+      });
+    }
+    this.setState({ selected });
+  };
+
+  getSelectedSubpage = () => (this.props.nativeSubpage ? 'native' : 'web');
```
tiagoengel: nit: no need for the extra parenthesis here

shaundon: Prettier did that 🤷‍♂️

packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
shaundon: Yep, will do.

packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
shaundon: Whoops no, I'll remove.

PR #771, opened 2018-05-22
[NO-JIRA] Update `update-owners` script

tiagoengel:
We can now provide the user we want to add and it ignores private packages (as those fail and break the script)


scripts/npm/update-owners.sh
```diff
@@ -16,10 +16,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-set -e
```
matteo-hertel: Why removing the `set -e` the fail early it's always a good thing, what's the reason for removing it?

PR #772, opened 2018-05-22
[BPK-1530] Introduce native carousel component

tiagoengel:
TODO:

- [x] The actual component
- [x] a11y
- [x] RTL?
- [x] tests
- [x] readme
- [ ] ~docs (new PR)~

*No more kittens :/*

### DIFF
--- a/native/packages/react-native-bpk-component-carousel/readme.md
+++ b/native/packages/react-native-bpk-component-carousel/readme.md
@@ -1,0 +1,63 @@
+# react-native-bpk-component-carousel
+
+> Backpack React Native carousel component.
+
+## Installation
+
+```sh
+npm install react-native-bpk-component-carousel --save-dev
+```
+
+## Usage
+
+```js
+import React, { Component } from 'react';
+import { View, StyleSheet } from 'react-native';
+import { spacingXl } from 'bpk-tokens/tokens/base.react.native';
+import BpkCarousel, { BpkCarouselItem } from 'react-native-bpk-component-carousel';
+
+const styles = StyleSheet.create({
+  carousel: {
+    width: spacingXl * 10,
+    height: spacingXl * 10
+  },
+  image: {
+    width: '100%',
+    height: '100%'
+  }
+})
+
+export default () => (
+  <BpkCarousel
+    accessibilityLabel={(page, total) => `${page + 1} of ${total}`}
+    style={styles.carousel}
+  >
+    <BpkCarouselItem>
+      <Image
+        style={styles.image}
+        source={require('./path/to/image.jpg')}
+      />
+    </BpkCarouselItem>
+    <BpkCarouselItem>
+      <Image
+        style={styles.image}
+        source={require('./path/to/image.jpg')}
+      />
+    </BpkCarouselItem>
+  </BpkCarousel>
+);
+```
+
+## Props
+
+| Property           | PropType                              | Required | Default Value |
+| ------------------ | ------------------------------------- | -------- | ------------- |
+| accessibilityLabel | oneOfType(string, func)               | true     | -             |
+| children           | node                                  | true     | -             |
+| showIndicator      | bool                                  | false    | true          |
+
+## BpkCarouselItem props
+
+| Property           | PropType                              | Required | Default Value |
+| ------------------ | ------------------------------------- | -------- | ------------- |
+| children           | node                                  | true     | -             |
--- a/native/packages/react-native-bpk-component-carousel/src/BpkCarousel-test.common.js
+++ b/native/packages/react-native-bpk-component-carousel/src/BpkCarousel-test.common.js
@@ -1,0 +1,89 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React from 'react';
+import renderer from 'react-test-renderer';
+import BpkText from 'react-native-bpk-component-text';
+import BpkCarousel from './BpkCarousel';
+import BpkCarouselItem from './BpkCarouselItem';
+
+const accessibilityLabel = "I'm accessible";
+
+// Manually triggers the onLayout callback with the value provided
+const renderWithLayout = (
+  component: any,
+  layout: { width?: number, height?: number } = { width: 300, height: 300 },
+) => {
+  const rendered = renderer.create(component);
+  const mockNativeEvent = {
+    nativeEvent: {
+      layout,
+    },
+  };
+  rendered.toJSON().props.onLayout(mockNativeEvent);
+  return rendered;
+};
+
+const commonTests = () => {
+  describe('BpkCarousel', () => {
+    it('should render correctly', () => {
+      const tree = renderWithLayout(
+        <BpkCarousel
+          accessibilityLabel={accessibilityLabel}
+          style={{ width: 300 }}
+        >
+          <BpkCarouselItem>
+            <BpkText>View 1</BpkText>
+          </BpkCarouselItem>
+        </BpkCarousel>,
+      ).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly with an "accessibilityLabel" function', () => {
+      const tree = renderWithLayout(
+        <BpkCarousel
+          accessibilityLabel={(page, total) => `${page} of ${total}`}
+        >
+          <BpkCarouselItem>
+            <BpkText>View 1</BpkText>
+          </BpkCarouselItem>
+        </BpkCarousel>,
+      ).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly without indicators', () => {
+      const tree = renderWithLayout(
+        <BpkCarousel
+          accessibilityLabel={accessibilityLabel}
+          showIndicator={false}
+        >
+          <BpkCarouselItem>
+            <BpkText>View 1</BpkText>
+          </BpkCarouselItem>
+        </BpkCarousel>,
+      ).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+  });
+};
+
+export default commonTests;
--- a/native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
+++ b/native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
@@ -1,0 +1,139 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React, { Fragment, type Element, type ChildrenArray } from 'react';
+import PropTypes from 'prop-types';
+import {
+  ScrollView,
+  type StyleObj,
+  ViewPropTypes,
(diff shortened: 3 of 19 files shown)


native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,116 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React, { Fragment, type Element, type ChildrenArray } from 'react';
+import PropTypes from 'prop-types';
+import { ScrollView, type StyleObj, ViewPropTypes } from 'react-native';
```
k0nserv: `StyleObj` 🙌

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,116 @@
...
+    const currentIndex = this.state.width
+      ? contentOffset.x / this.state.width
+      : 0;
+    this.setState({ currentIndex });
+  };
+
+  render() {
+    const { style, children, renderPagination } = this.props;
+    const { width, height, currentIndex } = this.state;
+    const shouldRenderChildren = width !== null && height !== null;
```
k0nserv: So this means we do one render pass that has no output just to get layout values at which point we actually render the component?

k0nserv: Can we have a newline here to introduce some "air" in the code? The other methods could also do with some more vertical whitespace imo

native/packages/react-native-bpk-component-carousel/src/BpkCarouselItem.js
```diff
@@ -0,0 +1,43 @@
...
+import PropTypes from 'prop-types';
+import { View, type StyleObj, ViewPropTypes } from 'react-native';
+
+export type Props = {
+  children: Node,
+  style: ?StyleObj,
+};
+
+const BpkCarouselItem = ({ children, style }: Props) => (
+  <View style={style}>{children}</View>
```
k0nserv: Wouldn't it be good if these `View`s had `width: '100%'` and `height: '100%` as that's always what we want?

native/packages/react-native-bpk-component-carousel/src/BpkCarousel-test.common.js
```diff
@@ -0,0 +1,93 @@
...
+const commonTests = () => {
+  describe('BpkCarousel', () => {
+    it('should render correctly', () => {
+      const tree = renderWithLayout(
+        <BpkCarousel
+          accessibilityLabel={accessibilityLabel}
+          style={{ width: 300 }}
+        >
+          <BpkCarouselItem>
+            <Text>View 1</Text>
```
shaundon: We should use `BpkText` here for dogfooding (also applies to other usages of `Text`.)

native/packages/react-native-bpk-component-carousel/readme.md
```diff
@@ -0,0 +1,58 @@
...
+);
+```
+
+## Props
+
+| Property           | PropType                              | Required | Default Value |
+| ------------------ | ------------------------------------- | -------- | ------------- |
+| accessibilityLabel | oneOfType(string, func)               | true     | -             |
+| children           | node                                  | true     | -             |
+| renderPagination   | func                                  | false    | null          |
```
k0nserv: Is this the best API design? It seems like `BpkPaginationDots` should be more first party even if we support custom pagination indiactors too

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,116 @@
...
+  handleScroll = (event: any) => {
+    const { contentOffset } = event.nativeEvent;
+    const currentIndex = this.state.width
+      ? contentOffset.x / this.state.width
+      : 0;
+    this.setState({ currentIndex });
+  };
+
+  render() {
+    const { style, children, renderPagination } = this.props;
```
shaundon: Can we put these in alphabetic order to satisfy my obsessive tendencies?

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,116 @@
...
+    const { width, height, currentIndex } = this.state;
+    const shouldRenderChildren = width !== null && height !== null;
+    return (
+      <Fragment>
```
shaundon: This is a React 16 feature right? Just want to confirm it's ok to use this, I can't remember if we officially support 16.

native/packages/react-native-bpk-component-carousel/src/BpkCarouselItem-test.common.js
```diff
@@ -0,0 +1,41 @@
...
+import renderer from 'react-test-renderer';
+import BpkCarouselItem from './BpkCarouselItem';
+
+const commonTests = () => {
+  describe('BpkCarouselItem', () => {
+    it('should render correctly', () => {
+      const tree = renderer
+        .create(
+          <BpkCarouselItem style={{ width: 300 }}>
+            <Text>View 1</Text>
```
shaundon: Use `BpkText` for dogfooding.

native/packages/react-native-bpk-component-carousel/readme.md
matthewdavidson: Agreed I think the pagination dots should be baked in with the ability to customise removed completely.

native/packages/react-native-bpk-component-carousel/src/BpkCarouselItem.js
matthewdavidson: Yeah I am wondering what the purpose of a `BpkCarouselItem ` is ...

native/packages/react-native-bpk-component-carousel/stories.js
```diff
@@ -0,0 +1,122 @@
...
+import {
+  spacingMd,
+  colorGray300,
+  colorGray500,
+} from 'bpk-tokens/tokens/base.react.native';
+import range from 'lodash/range';
+import CenterDecorator from '../../storybook/CenterDecorator';
+import BpkCarousel, { BpkCarouselItem } from './index';
+
+const STYLES = StyleSheet.create({
```
shaundon: Lowercase variable name needed here.

native/packages/react-native-bpk-component-carousel/stories.js
```diff
@@ -0,0 +1,122 @@
...
+    maxHeight: 300,
+  },
+});
+
+const renderKittenImages = () =>
+  [
+    'https://placekitten.com/g/300/300',
+    'https://placekitten.com/300/300',
+    'https://placekitten.com/g/400/400',
+    'https://placekitten.com/g/450/450',
```
shaundon: As much as I love these, we should probably use something more _us_. `BpkImage` gets things from our CDN, can we do that here too? https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-image/stories.js#L29

native/packages/react-native-bpk-component-carousel/src/BpkCarouselItem.js
shaundon: Tbf I advised having an item component, maybe in hindsight that wasn't the correct choice. I was thinking that it gives us a wrapper in case there's functionality we want to introduce after the fact.

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
matthewdavidson: Yeah we do 👍

native/packages/react-native-bpk-component-carousel/stories.js
matthewdavidson: Huh? We use this pattern in other components ...

native/packages/react-native-bpk-component-carousel/readme.md
shaundon: Lol what a terrible example from me! Ah well..

native/packages/react-native-bpk-component-carousel/stories.js
matthewdavidson: https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-animate-height/src/BpkAnimateHeight.js#L40

native/packages/react-native-bpk-component-carousel/src/BpkCarouselItem.js
tiagoengel: I don't know if `width/height: 100%` work with the scroll view. I can try again but as far as I remember it doesn't work.

tiagoengel: The `BpkCarouselItem` really seems to be without purpose here. I could replace it for a view but as @shaundon said it makes it much easier to evolve it in the future. e.g. 

```
<BpkCarouselItem onVisible={changeTab} />
```

native/packages/react-native-bpk-component-carousel/readme.md
k0nserv: Your example in the stories are doing more than you say here. For example setting the style


```
 <BpkPaginationDots
    style={{
     alignSelf: 'center',
     marginTop: -spacingMd * 4,
   }}
```

This definitely feels like something a consumer shouldn't have to care about. At the very least if we make it configurable the `BpkPaginationDots` should render in the correct place by default

tiagoengel: Yeah what I meant with my example is that with the current implementation we don't need to care/impose this style. In any case I'll chat with @willhg tomorrow to check what are his views on this and what is the default style we want to have

native/packages/react-native-bpk-component-carousel/readme.md
```diff
@@ -0,0 +1,58 @@
...
+
+```js
+import React, { Component } from 'react';
+import { View, StyleSheet } from 'react-native';
+import BpkCarousel, { BpkCarouselItem } from 'react-native-bpk-component-carousel';
+
+export default () => (
+  <BpkCarousel 
+    accessibilityLabel={(page, total) => `${page + 1} of ${total}`} 
+    style={{ width: 300, height: 300 }}
```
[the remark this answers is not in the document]
tiagoengel: what token would I use here? I can't find anything appropriate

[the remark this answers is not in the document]
tiagoengel: I think will just put percentages here instead of fixed values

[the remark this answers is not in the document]
shaundon: > which token would I use here? I can't find anything appropriate

I'd just do something like `$bpk-spacing-xl * 10`

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
tiagoengel: I'll do it but I have to voice my disagreement :D. I honestly don't think this is relevant or desired, and if it is, the linter should be the one to catch/force this, otherwise it will never be a pattern throughout the code base. I also don't think it makes sense in most situations. Let's say I a have the props `width, height, style` and `alertText, errorText, successText`. I would group them together, by their "function" and not alphabetically.

shaundon: That makes sense - if you feel strongly about it I'm happy to back down!

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,139 @@
...
+                ...child.props,
+                style: [child.props.style, { width, height }],
+              }),
+            )}
+        </ScrollView>
+        {shouldRenderChildren &&
+          renderPagination && (
+            <BpkPaginationDots
+              style={styles.pagination}
+              accessibilityLabel={this.getAccessibilityLabel()}
```
tiagoengel: Not really sure this should be accessible (or what to put in here). The pagination is not focusable nor interactive, it doesn't seem something the screen reader should care about. What you guys think? @shaundon @matthewdavidson @k0nserv

matthewdavidson: I tend to agree here - this is a purely visual indicator. The accessibility of a carousel should be handled by the scrollview itself. I can remove these props as part of my card.

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,139 @@
...
+  height: ?number,
+};
+
+class BpkCarousel extends React.Component<Props, State> {
+  static propTypes = {
+    accessibilityLabel: PropTypes.oneOfType([PropTypes.string, PropTypes.func])
+      .isRequired,
+    children: PropTypes.node.isRequired,
+    style: ViewPropTypes.style,
+    renderPagination: PropTypes.bool,
```
shaundon: @matthewdavidson is renaming `BpkPaginationDots` to `BpkCarouselIndicator` (I think, I may have gotten the name wrong), so once that happens, calling this prop `renderPagination` will be a bit confusing. How about `showIndicator` instead?

tiagoengel: yeah, I think that's probably better. will do.

PR #773, opened 2018-05-22
[BPK-1601] Add web map component

shaundon:
I've taken the map component from #757 and pared it back so that it's a light wrapper around `react-google-maps`.

The storybook has been updated to reflect the new format. It means that consumers have to do slightly more work, but they have more flexibility and it matches the underlying API, meaning there's less code to maintain and less documentation to write.

**To do**
* [x] Snapshot tests aren't right. We probably need to use Enzyme to get the map to load. (See this: https://github.com/tomchentw/react-google-maps/issues/410).
* [x] Readme needs to be updated.
* [x] Commits need to be squashed (this can be done later).

**Open questions**
* Is `export * from 'react-google-maps` ok? I don't know of anywhere else we do this. Alternatively we could export everything from the package manually.
* In `BpkMap` we're applying the HOCs. Maybe even this is too much and we shouldn't do that either, and just do `export default GoogleMap`.

### DIFF
--- a/packages/bpk-component-map/index.js
+++ b/packages/bpk-component-map/index.js
@@ -1,0 +1,42 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import BpkMap from './src/BpkMap';
+
+export default BpkMap;
+export {
+  withGoogleMap,
+  withScriptjs,
+  GoogleMap,
+  Circle,
+  Marker,
+  Polyline,
+  Polygon,
+  Rectangle,
+  InfoWindow,
+  OverlayView,
+  GroundOverlay,
+  DirectionsRenderer,
+  FusionTablesLayer,
+  KmlLayer,
+  TrafficLayer,
+  StreetViewPanorama,
+  BicyclingLayer,
+} from 'react-google-maps';
--- a/packages/bpk-component-map/readme.md
+++ b/packages/bpk-component-map/readme.md
@@ -1,0 +1,81 @@
+# bpk-component-map
+
+> Backpack map component.
+
+## Installation
+
+```sh
+npm install bpk-component-map --save-dev
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkMap from 'bpk-component-map';
+
+export default () => (
+  <BpkMap
+    containerElement={<div style={{ height: '400px' }} />}
+    mapElement={<div style={{ height: `100%` }} />}
+    defaultZoom={15}
+    defaultCenter={{
+      lat: 27.9881,
+      lng: 86.925,
+    }}
+    options={{
+      zoomControl: false,
+      dragEnabled: false,
+    }}
+  />
+);
+```
+
+## Accompanying HOCs
+
+### withScriptjs
+
+`withScriptjs` is a HOC that loads the Google Maps Javascript, then loads the map. This is useful for when you don't already have the Google Maps Javascript loaded.
+
+If you intend to include multiple maps on one page, it's better to load the Google Maps Javascript elsewhere and not use this HOC, as it downloads the script every time it's used.
+
+```js
+import React from 'react';
+import BpkMap, { withScriptjs } from 'bpk-component-map';
+
+const MAP_URL = 'https://maps.googleapis.com/maps/api/js?v=3.exp&libraries=geometry,drawing,places';
+
+const BpkMapWithScript = withScriptjs(BpkMap);
+
+export default () => (
+  <BpkMapWithScript
+    googleMapURL={MAP_URL}
+    loadingElement={<div />}
+    containerElement={<div style={{ height: '400px' }} />}
+    mapElement={<div style={{ height: `100%` }} />}
+    defaultZoom={15}
+    defaultCenter={{
+      lat: 27.9881,
+      lng: 86.925,
+    }}
+    options={{
+      zoomControl: false,
+      dragEnabled: false,
+    }}
+  />
+);
+```
+
+## Props
+
+### BpkMap
+
+| Property	      | PropType	| Required                	| Default Value |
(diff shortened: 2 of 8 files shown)


packages/bpk-component-map/index.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+/* @flow */
+
+import BpkMap from './src/BpkMap';
+import BpkMapMarker from './src/BpkMapMarker';
+
+export default BpkMap;
+export { BpkMapMarker };
+export * from 'react-google-maps';
```
matthewdavidson: How come we need to export all of the exports in `react-google-maps`?

packages/bpk-component-map/readme.md
```diff
@@ -0,0 +1,60 @@
...
+    }}
+  >
+    <BpkMapMarker
+      position={{
+        lat: 27.9881,
+        lng: 86.925,
+      }}
+      mapPaneName="overlayMouseTarget"
+    >
+      Everest summit
```
matthewdavidson: Love this

packages/bpk-component-map/src/__snapshots__/BpkMap-test.js.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`BpkMap should render correctly 1`] = `<div />`;
```
matthewdavidson: These snapshots are pretty useless - how can we test this component better?

packages/bpk-component-map/readme.md
shaundon: Thought you'd like that haha. I originally did Buckingham Palace but then I thought I should be less colonial about it

packages/bpk-component-map/index.js
shaundon: Well, I felt that it would be beneficial to give consumers access to everything `react-google-maps` offers. There's quite a lot of exports (it's pretty powerful). 

I'm thinking of a situation where say, a consumer wants to draw a circle on the map (let's say it's to draw a radius around a hotel). If we're not exporting that, they'd have to pull it in from `react-google-maps` instead which doesn't feel quite right to me.

I am totally open to opinions here - I don't think it's something we've done before and we should make sure we get it right. I mentioned it in the 'open questions' part of the PR description:

> Is `export * from 'react-google-maps` ok? I don't know of anywhere else we do this. Alternatively we could export everything from the package manually.

packages/bpk-component-map/src/__snapshots__/BpkMap-test.js.snap
shaundon: This is on the todo list:

>  Snapshot tests aren't right. We probably need to use Enzyme to get the map to load. (See this: tomchentw/react-google-maps#410).

My bad as I should have marked this PR as WIP in this aspect (will do now).

packages/bpk-component-map/src/BpkMap-test.js
```diff
@@ -0,0 +1,47 @@
...
+          done();
+        }}
+      />,
+    );
+  });
+
+  it('should render without throwing', () => {
+    expect(wrapper.exists()).toBeTruthy();
+  });
+});
```
shaundon: I'm struggling to provide value with this test. As we're just delegating all functionality to `react-google-maps`, there isn't much we can test.

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+/* @flow */
+
+import React from 'react';
+import { withScriptjs, withGoogleMap, GoogleMap } from 'react-google-maps';
+
+const BpkMap = withScriptjs(
+  withGoogleMap(({ mapRef, ...rest }) => <GoogleMap ref={mapRef} {...rest} />),
+);
```
shaundon: @matteo-hertel and I talked this over and feel that it provides a nicer API to consumers if `BpkMap` wraps the component in a HOC for them, but I'm highlighting it as maybe something we don't want to do. I'm trying to think of use cases where a user *doesn't* wish to do this.

packages/bpk-component-map/src/BpkMapMarker.js
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import { OverlayView } from 'react-google-maps';
+
+export default OverlayView;
```
shaundon: I'm actually not sure that this should exist at all now. The way you use `OverlayView` is like this:

```
<OverlayView
  position={{ lat: 55.944357, lng: -3.1967116 }} 
  mapPaneName="overlayMouseTarget"
>
  // Content goes here, e.g a BpkIcon or a popover or whatever.
</OverlayView>
```

As things stand, consumers will need to pass in children to this component to have it do anything.

So, maybe we should shelve this until we can provide more value. For example, if we define some marker types, we could provide an API like this:

```
<BpkMapMarker
  position={{ lat: 55.944357, lng: -3.1967116 }} 
  type={MARKER_TYPES.ATTRACTION}
  onClick={() => {}}
/>
```

If we leave this as-is, then change it later, we'll have to make breaking changes, which is why I'm leaning towards leaving it for now..

packages/bpk-component-map/stories.js
```diff
@@ -0,0 +1,147 @@
...
+const RtlSmallHotelsIcon = withRtlSupport(HotelsSmallIcon);
+
+const StoryMap = props => {
+  const { language, ...rest } = props;
+  return (
+    <BpkMap
+      googleMapURL={`https://maps.googleapis.com/maps/api/js?v=3.exp&language=${language}&libraries=geometry,drawing,places`}
+      loadingElement={<div />}
+      containerElement={<div style={{ height: '400px' }} />}
+      mapElement={<div style={{ height: `100%` }} />}
```
shaundon: Should maybe put this in scss

packages/bpk-component-map/index.js
matthewdavidson: Sounds good to me 👍

packages/bpk-component-map/src/BpkMap-test.js
matthewdavidson: So don't test it? :)

In all seriousness we should only be testing our code and from the looks of it `BpkMap.js` doesn't have much of our code in it.

shaundon: No complaints from me!

packages/bpk-component-map/readme.md
```diff
@@ -0,0 +1,81 @@
...
+    }}
+  />
+);
+```
+
+## Props
+
+### BpkMap
+
+| Property	      | PropType	| Required                	| Default Value |
```
matthewdavidson: apart from this misalignment 

bitmoji

shaundon: Weird! It's not misaligned in my editor:

screen shot 2018-05-24 at 15 50 02

PR #774, opened 2018-05-24
BPK-1505 RN card api 21 fix

matteo-hertel:
This fixes the card component on an android API level 21

Screenshot from an API 21 emulator, updated screenshot for the current emulator have been updated as part of the PR
default
focused
with-divider-arranged-vertically
with-divider
without-padding


native/package.json
```diff
@@ -7,7 +7,8 @@
     "start": "react-native start",
     "ios": "react-native run-ios",
     "android:emulator": "$ANDROID_HOME/tools/emulator -avd bpk-avd",
-    "android": "npm run android:emulator & react-native run-android",
+    "android:build": "react-native run-android",
```
matthewdavidson: This is useful if you don't want to run the emulator

matteo-hertel:
Still testing on API 21, will update and squashed once done

PR #775, opened 2018-05-24
[BPK-1569] Native alert component

amolgupta:
Documentation is pending


native/packages/react-native-bpk-component-alert/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "react-native-bpk-component-alert",
+  "version": "0.0.01",
```
shaundon: This should be `0.0.0`. When we do a release, Lerna increments it automatically.

native/packages/react-native-bpk-component-alert/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "main": "index.js",
+  "description": "Backpack React Native switch component.",
```
shaundon: Should say alert, not switch.

native/packages/react-native-bpk-component-alert/readme.md
```diff
@@ -0,0 +1,36 @@
+# react-native-bpk-component-switch
+
+> Backpack React Native switch component.
```
shaundon: These two lines should say alert, not switch.

native/packages/react-native-bpk-component-alert/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^16.0.0-alpha.1",
+    "react-native": ">= 0.47.0"
+  },
+  "dependencies": {
+    "bpk-tokens": "^27.0.0",
+    "prop-types": "^15.5.8",
+    "react-native-bpk-theming": "^1.0.53"
```
shaundon: I don't think `react-native-bpk-theming` is being used in this project, so it doesn't need to be listed here.

native/packages/react-native-bpk-component-alert/src/__snapshots__/BpkSwitch-test.android.js.snap
```diff
@@ -0,0 +1,69 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
shaundon: These snapshot tests can be deleted, as there are no tests to generate them.

native/packages/react-native-bpk-component-alert/stories.js
```diff
@@ -0,0 +1,121 @@
...
+        text: 'No',
+        onPress: () => action('negative button press'),
+        style: 'cancel',
+      },
+      { text: 'Yes', onPress: () => action('positive button press') },
+    ],
+    { cancelable: true },
+  );
+};
+
```
shaundon: We don't have an example of a [destructive alert for iOS](https://facebook.github.io/react-native/docs/alert.html#ios), I think this should also be included in storybook and screenshots.

shaundon: img_1119

It shows it in red - I think we should have an example that shows each of the styles, like in the above screenshot.

shaundon:
The code for this looks great now, the only other thing I'm thinking is that we should change the storybooks (and screenshots) to be less opinionated, which should make it more obvious for consumers what the different arguments to `BpkAlert.alert` do.

For example, instead of:

screen shot 2018-05-25 at 14 04 12

I think it's better for consumers to do this:

40534796-159fd95a-5fff-11e8-8852-4862e31dcf16

That way, consumers can tell from the picture what each style of button does.

amolgupta:
Yes. That makes sense. We can also follow that pattern across all stories.

shaundon:
It looks great now 😄 

The only thing left to do is squash the commits, so that they're combined into one single commit. One of us can show you how to do that if you're unfamiliar.

PR #777, opened 2018-05-25
[BPK-1609] Fix deep linking navigation in IE

georgegillams:
This was borked in IE as we were using the `Array.find()` method which is not supported. I have lifted MDN's polyfil to work round this in browsers that do not have `Array.find()`.

### DIFF
--- a/packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
+++ b/packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
@@ -19,6 +19,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { withRouter } from 'react-router';
+import { find } from 'lodash';
 import BpkContentContainer from 'bpk-component-content-container';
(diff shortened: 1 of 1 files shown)


packages/bpk-docs/src/components/neo/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -96,6 +96,27 @@ class DocsPageWrapper extends React.Component {
 
   getSelectedSubpage = () => (this.props.nativeSubpage ? 'native' : 'web');
 
+  compatibleFind = (arr, val) => {
```
tiagoengel: can't we just use lodash.find?

georgegillams: Oh maybe. Will investigate :)

georgegillams: Haha yeah that works a treat 😄

PR #779, opened 2018-05-25
[BPK-1601] Add docs for map component

shaundon:
I've named the route `NEO_MAP` so as not to confuse people, but this is a Neo only thing (seeing as we now only have Neo!).

Right now there's a very minimal example, we can add more later when we add functionality on top of `react-google-maps`.

0 0 0 0_8080_components_blockquote_platform web

### DIFF
--- a/packages/bpk-docs/src/pages/MapPage/MapPage.js
+++ b/packages/bpk-docs/src/pages/MapPage/MapPage.js
@@ -1,0 +1,79 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


packages/bpk-docs/src/pages/MapPage/MapPage.js
```diff
@@ -0,0 +1,79 @@
...
+
+import mapReadme from 'bpk-component-map/readme.md';
+
+import DocsPageBuilder from './../../components/DocsPageBuilder';
+import DocsPageWrapper from './../../components/neo/DocsPageWrapper';
+import IntroBlurb from './../../components/neo/IntroBlurb';
+
+const BpkMapWithScript = withScriptjs(BpkMap);
+const MAP_URL =
+  'https://maps.googleapis.com/maps/api/js?v=3.exp&libraries=geometry,drawing,places';
```
k0nserv: What kind of usage limits are there for unauthenticated requests to Google Maps?

shaundon:
> Do we have a card to follow up and rip out the old doc site?

We do indeed! Can't wait to do that 😄

packages/bpk-docs/src/pages/MapPage/MapPage.js
shaundon: We should be alright hopefully: https://developers.google.com/maps/documentation/javascript/usage. Looks like 25,000 map loads per 24 hours. If this becomes an issue, we can always get Skyscanner's API key.

PR #780, opened 2018-05-25
[No JIRA] Fix minor issues with map component docs

shaundon:


packages/bpk-docs/src/pages/MapPage/MapPage.js
```diff
@@ -60,7 +60,7 @@ const blurb = [
 
 const MapSubpage = ({ ...rest }) => (
   <DocsPageBuilder
-    title="Blockquote"
+    title="Map"
```
matteo-hertel: 😱

shaundon: Yeah my bad

shaundon: png

It me

PR #790, opened 2018-05-30
[BPK-1611] Add web phone input

tiagoengel:
- [x] a11y
- [x] RTL
- [x] Tests
- [x] Stories
- [x] Readme

Docs and integration with the @matteo-hertel's work in a different PR

### DIFF
--- a/packages/bpk-component-phone-input/readme.md
+++ b/packages/bpk-component-phone-input/readme.md
@@ -1,0 +1,83 @@
+# bpk-component-phone-input
+
+> Backpack phone input component.
+
+## Installation
+
+```sh
+npm install bpk-component-phone-input --save-dev
+```
+
+## Usage
+
+```js
+import React, { Component } from 'react';
+import BpkPhoneInput from 'bpk-component-phone-input';
+
+export default class extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { dialingCode: '44', phoneNumber: '' }
+  }
+
+  onChange = (evt) => {
+    this.setState({ phoneNumber: evt.target.value });
+  }
+
+  onDialingCodeChange = (evt) => {
+    this.setState({ dialingCode: evt.target.value });
+  }
+
+
+  render() {
+    return (
+      <BpkPhoneInput
+        id="phone-input-id"
+        name="Telephone input"
+        placeholder="Telephone number"
+        onChange={this.onChange}
+        onDialingCodeChange={this.onDialingCodeChange}
+        value={this.state.phoneNumber}
+        dialingCode={this.state.dialingCode}
+        dialingCodes={[
+          { code: '44', description: '+44' },
+          { code: '55', description: '+55' },
+        ]}
+        dialingCodeProps={{
+          id: 'dialing-code',
+          name: 'Dialing code',
+          'aria-label': 'Dialing code',
+        }}
+      />
+    )
+  }
+}
+```
+
+## Props
+
+| Property              | PropType                                              | Required   | Default Value    |
+| --------------------- | ----------------------------------------------------- | ---------- | ---------------- |
+| dialingCode           | string                                                | true       | -                |
+| dialingCodeProps      | shape({ id: string, name: string })                   | true       | -                |
+| dialingCodes          | arrayOf(shape({ code: string, description: string })) | true       | -                |
+| id                    | string                                                | true       | -                |
+| name                  | string                                                | true       | -                |
+| onChange              | func                                                  | true       | -                |
+| onDialingCodeChange   | func                                                  | true       | -                |
+| value                 | string                                                | true       | -                |
+| className             | string                                                | false      | null             |
+| disabled              | boolean                                               | false      | false            |
+| large                 | boolean                                               | false      | false            |
+| valid                 | boolean                                               | false      | null             |
+| wrapperProps          | object                                                | false      | {}               |
+
+### dialingCodeProps
+
+Note that `id` and `name` are required but more properties can be provided, e.g. `dialingCodeProps={{ id: 'id', name: 'name', className: 'some-class' }}`. All
+properties will be forwarded to the underlying `BpkSelect` component.
+
+### dialingCodes
+
+The same is true for `dialingCodes`. Each object **must** have a `code` and `description`, but can have more properties and those
+will be forwarded the the `option` element they represent.
--- a/packages/bpk-component-phone-input/src/BpkPhoneInput.js
+++ b/packages/bpk-component-phone-input/src/BpkPhoneInput.js
@@ -1,0 +1,141 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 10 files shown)


packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -0,0 +1,144 @@
...
+      >
+        {dialingCodes.map(({ code, description, ...extraDialingProps }) => (
+          <option key={code} value={code} {...extraDialingProps}>
+            {description}
+          </option>
+        ))}
+      </BpkSelect>
+      <BpkInput
+        {...commonProps}
+        {...rest}
```
georgegillams: `...rest` should generally be placed on the outermost container except for some very specific circumstances

tiagoengel: Yeah, this is one of the those specific circumstances :D. When used with the FieldSet, it passes trough some a11y props that we need to forward to the input (to make it accessible). The options were to either filter all those properties and pass it to the input, which would be a bit confusing, or this. I talked with @matthewdavidson and we decided to go this way and have a property called `containerProps` so users can still pass in extra props to the container.

georgegillams: Ah fair enough. We'll leave it as it is then. The only thing I might suggest is pulling `style` out and applying it to the outer container, as some consumers use style to adjust positioning etc

packages/bpk-component-phone-input/stories.js
```diff
@@ -0,0 +1,112 @@
...
+    );
+  }
+}
+
+storiesOf('bpk-component-phone-input', module)
+  .add('Default', () => <StoryContainer />)
+  .add('Large', () => <StoryContainer large />)
+  .add('With Validation', () => (
+    <StoryContainer
+      validationMessage="Please inform a valid phone number"
```
shaundon: This is quite strange phrasing - how about `Please enter a valid phone number`?

packages/bpk-component-phone-input/readme.md
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+```
+
+## Props
+
+| Property              | PropType                                              | Required   | Default Value    |
+| --------------------- | ----------------------------------------------------- | ---------- | ---------------- |
+| dialingCode           | string                                                | true       | -                |
+| dialingCodeProps      | shape({ id: string, name: string })                   | true       | -                |
```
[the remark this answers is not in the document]
tiagoengel: `id` and `name` are required and the rest is optional, similar to how we use`...rest`. `classname` is in the flow declaration because I'm using it directly and otherwise flow whines about it. Do you have any idea how to type it better? Maybe I can mention bellow that `id` and `name` are required but any other property can be passed and will be forwarded to the component.

packages/bpk-component-phone-input/stories.js
tiagoengel: I'll do

packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -0,0 +1,144 @@
...
+
+type ChildProps = { [string]: any };
+
+export type Props = {
+  dialingCode: string,
+  dialingCodeProps: { id: string, name: string, className?: string },
+  dialingCodes: Array<{ code: string, description: string }>,
+  id: string,
+  name: string,
+  onChange: (string, string) => void,
```
k0nserv: This should return mixed

packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -0,0 +1,144 @@
...
+  value: string,
+  className: ?string,
+  containerProps: ChildProps,
+  disabled: ?boolean,
```
k0nserv: In general we don't have nulalble boolean props instead we explicitly make them `true` or `false` with `defaultProps`. In this case it seems like these should be `boolean` and have a default of `false`. Note this does not apply to `valid` beacuse it needs to be tri-state boolean

packages/bpk-component-phone-input/src/BpkPhoneInput.scss
```diff
@@ -0,0 +1,37 @@
...
+
+.bpk-phone-input {
+  display: flex;
+
+  &__dialing-code {
+    width: auto;
+  }
+
+  &__phone-number {
+    margin-left: $bpk-spacing-md;
```
k0nserv: There's a helper called `bpk-leading-margin` we could use instead here

packages/bpk-component-phone-input/readme.md
```diff
@@ -0,0 +1,67 @@
...
+| dialingCodes          | arrayOf(shape({ code: string, description: string })) | true       | -                |
+| id                    | string                                                | true       | -                |
+| name                  | string                                                | true       | -                |
+| id                    | string                                                | true       | -                |
+| onChange              | func                                                  | true       | -                |
```
k0nserv: Should we have two onChange functions instead of one? Personally think it'd be nicer with one for each case e.g

+ `onNumberChange`
+ `onDialingCodeChange`

tiagoengel: We could, although I'd prefer `onChange` and `onDialingCodeChange` in this case. Not sure what would be the advantage of that though.

packages/bpk-component-phone-input/readme.md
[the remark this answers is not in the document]
shaundon: Yeah maybe mentioning it below is a good idea.

Things like this are making me think that we should re-think how we expose prop types to users..

packages/bpk-component-phone-input/src/BpkPhoneInput.scss
tiagoengel: noice

packages/bpk-component-phone-input/readme.md
```diff
@@ -0,0 +1,82 @@
...
+| dialingCode           | string                                                | true       | -                |
+| dialingCodeProps      | shape({ id: string, name: string })                   | true       | -                |
+| dialingCodes          | arrayOf(shape({ code: string, description: string })) | true       | -                |
+| id                    | string                                                | true       | -                |
+| name                  | string                                                | true       | -                |
+| onChange              | func                                                  | true       | -                |
+| onDialingCodeChange   | func                                                  | true       | -                |
+| value                 | string                                                | true       | -                |
+| className             | string                                                | false      | null             |
+| containerProps        | object                                                | false      | {}               |
```
shaundon: `container` has [connotations](https://medium.com/@learnreact/container-components-c0e67432e005) in React so we usually use `wrapper` instead.

(Sorry, I should have said this before but I only just noticed)

packages/bpk-component-phone-input/stories.js
```diff
@@ -0,0 +1,117 @@
...
+}
+
+storiesOf('bpk-component-phone-input', module)
+  .add('Default', () => <StoryContainer />)
+  .add('Large', () => <StoryContainer large />)
+  .add('With Validation', () => (
+    <StoryContainer
+      validationMessage="Please enter a valid phone number"
+      validNumber="99999999999"
+      description="Inform 99999999999"
```
shaundon: I think this should say `Enter 99999999999` too, to match the validation message.

Actually, could we change `99999999999` to something easier to type in, like `0123456789`?

PR #798, opened 2018-06-01
[BPK-1626] RN rename pagination dots

georgegillams:


georgegillams:
@k0nserv think I've addressed all of those now 👍

georgegillams:
On second thoughts, I may have now actually addressed all of them 😆

georgegillams:
Ok just ignore me

PR #799, opened 2018-06-01
[BPK-1612] select with image

matteo-hertel:
- [x] Update unrelased.md
- [x] Update Readme

Docs will follow as separate PR

This adds an image render prop to the select box, if the image is set the select box will look like this
```
+------------------------+
|+----+ +--------------+ |
||    | |              | |
|+----+ +--------------+ |
+------------------------+
ASCII art By Matt
```

if no image is provided, it'll render the selecbox as it currently is (fully backward compatible)

In the storybook there is an example of usage

Note: The docked config has not been modified as we need design input for that

### DIFF
--- a/packages/bpk-component-select/src/BpkSelect-test.js
+++ b/packages/bpk-component-select/src/BpkSelect-test.js
@@ -20,6 +20,9 @@ import React from 'react';
 import renderer from 'react-test-renderer';
 import BpkSelect from './BpkSelect';
 
+const svgPlaceholder = `data:image/svg+xml;charset=utf-8,%3Csvg xmlns%3D'http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg' viewBox%3D'0 0 200 150'%2F%3E`;
+const selectImage = <img alt="placeholder" src={svgPlaceholder} />;
+
 describe('BpkSelect', () => {
   it('should render correctly', () => {
     const tree = renderer
@@ -33,8 +36,54 @@ describe('BpkSelect', () => {
           <option value="apples">Apples</option>
           <option value="oranges">Oranges</option>
           <option value="pears">Pears</option>
-          <option value="tomatos" disabled>
-            Tomatos
+          <option value="tomatoes" disabled>
+            Tomatoes
+          </option>
+        </BpkSelect>,
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('should render correctly with image', () => {
+    const tree = renderer
+      .create(
+        <BpkSelect
+          id="fruits"
+          name="fruits"
+          value="oranges"
+          image={selectImage}
+          onChange={() => null}
+        >
+          <option value="apples">Apples</option>
+          <option value="oranges">Oranges</option>
+          <option value="pears">Pears</option>
+          <option value="tomatoes" disabled>
+            Tomatoes
+          </option>
+        </BpkSelect>,
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('should render correctly with image and custom classes', () => {
+    const tree = renderer
+      .create(
+        <BpkSelect
+          id="fruits"
+          name="fruits"
+          value="oranges"
+          imageWrapperClassName="imageWrapperClass--test"
+          wrapperClassName="wrapperClass--test"
+          image={selectImage}
+          onChange={() => null}
+        >
+          <option value="apples">Apples</option>
+          <option value="oranges">Oranges</option>
+          <option value="pears">Pears</option>
+          <option value="tomatoes" disabled>
+            Tomatoes
           </option>
         </BpkSelect>,
       )
@@ -55,8 +104,8 @@ describe('BpkSelect', () => {
           <option value="apples">Apples</option>
           <option value="oranges">Oranges</option>
           <option value="pears">Pears</option>
-          <option value="tomatos" disabled>
-            Tomatos
+          <option value="tomatoes" disabled>
+            Tomatoes
           </option>
         </BpkSelect>,
       )
@@ -77,8 +126,8 @@ describe('BpkSelect', () => {
           <option value="apples">Apples</option>
           <option value="oranges">Oranges</option>
           <option value="pears">Pears</option>
-          <option value="tomatos" disabled>
-            Tomatos
+          <option value="tomatoes" disabled>
+            Tomatoes
           </option>
         </BpkSelect>,
       )
@@ -99,8 +148,8 @@ describe('BpkSelect', () => {
           <option value="apples">Apples</option>
           <option value="oranges">Oranges</option>
           <option value="pears">Pears</option>
-          <option value="tomatos" disabled>
-            Tomatos
+          <option value="tomatoes" disabled>
+            Tomatoes
           </option>
         </BpkSelect>,
       )
@@ -121,8 +170,8 @@ describe('BpkSelect', () => {
           <option value="apples">Apples</option>
           <option value="oranges">Oranges</option>
           <option value="pears">Pears</option>
-          <option value="tomatos" disabled>
-            Tomatos
+          <option value="tomatoes" disabled>
+            Tomatoes
           </option>
         </BpkSelect>,
       )
@@ -143,8 +192,8 @@ describe('BpkSelect', () => {
           <option value="apples">Apples</option>
           <option value="oranges">Oranges</option>
           <option value="pears">Pears</option>
-          <option value="tomatos" disabled>
-            Tomatos
+          <option value="tomatoes" disabled>
+            tomatos
           </option>
         </BpkSelect>,
       )
@@ -165,8 +214,8 @@ describe('BpkSelect', () => {
           <option value="apples">Apples</option>
           <option value="oranges">Oranges</option>
           <option value="pears">Pears</option>
-          <option value="tomatos" disabled>
-            Tomatos
+          <option value="tomatoes" disabled>
+            Tomatoes
           </option>
         </BpkSelect>,
       )
@@ -187,8 +236,8 @@ describe('BpkSelect', () => {
           <option value="apples">Apples</option>
           <option value="oranges">Oranges</option>
           <option value="pears">Pears</option>
-          <option value="tomatos" disabled>
-            Tomatos
+          <option value="tomatoes" disabled>
+            Tomatoes
           </option>
         </BpkSelect>,
       )
--- a/packages/bpk-component-select/src/BpkSelect.js
+++ b/packages/bpk-component-select/src/BpkSelect.js
@@ -15,8 +15,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* @flow */
 
-import PropTypes from 'prop-types';
+import PropTypes, { type Node } from 'prop-types';
 import React from 'react';
 import { cssModules } from 'bpk-react-utils';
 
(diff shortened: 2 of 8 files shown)


packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -15,25 +15,48 @@
...
 
 const getClassName = cssModules(STYLES);
 
-const BpkSelect = props => {
+export type Props = {
+  id: string,
+  name: string,
+  value: string,
+  className: ?string,
+  docked: ?boolean,
```
k0nserv: These `?boolean` types are not correct, they should be `boolean` because they are all filled in by React via the `defaultProps` before the function is called.

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -59,37 +86,58 @@ const BpkSelect = props => {
...
   className: null,
-  valid: null,
-  large: false,
   docked: false,
   dockedFirst: false,
-  dockedMiddle: false,
   dockedLast: false,
+  dockedMiddle: false,
+  image: null,
+  imageContainerClassName: '',
```
k0nserv: Classnames should be `null` by default not `''`

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -42,6 +65,10 @@ const BpkSelect = props => {
   const isInvalid = valid === false;
 
   if (large) {
+    wrapperClassNames.push(getClassName('bpk-select-wrapper--large'));
+    imageContainerClassNames.push(
+      getClassName('bpk-select-image-container--large'),
```
k0nserv: Shouldn't this be `bpk-select__image-container--large`

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -15,25 +15,48 @@
...
+  image: ?() => Node,
+  imageContainerClassName: ?string,
+  large: ?boolean,
+  valid: ?boolean,
+  wrapperClassName: ?string,
+};
+
+const BpkSelect = (props: Props) => {
   const classNames = [getClassName('bpk-select')];
+  const imageContainerClassNames = [getClassName('bpk-select-image-container')];
```
k0nserv: Shouldn't this be `bpk-select__image-container`?

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -15,25 +15,48 @@
...
+  const wrapperClassNames = [getClassName('bpk-select-wrapper')];
```
k0nserv: Shouldn't this be `bpk-select__wrapper`?

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -59,37 +86,58 @@ const BpkSelect = props => {
...
     <select
       className={classNames.join(' ')}
       aria-invalid={isInvalid}
       {...rest}
     />
   );
+
+  if (image) {
+    return (
+      <div className={wrapperClassNames.join(' ')}>
```
k0nserv: Instead of using `wrapperClassNames` you can use Tiago's extensions to `getClassNames` and write this as

```
className={getClassName('bpk-select__wrapper', large && 'bpk-select__wrapper--large', wrapperClassName)}
```
Same goes for the image container

k0nserv:
We also need to test this on IE11, but I'll do that after we address the question of the image placement

matteo-hertel:
@k0nserv as per the design the image should be inside the box, but we can't pass arbitrary nodes to the option tag, the agreement was to render the image in a div adjacent to the select box, I'll action all the other comments, I wasn't aware of the `getClassName` bit 😅

k0nserv:
The image is much smaller in IE11

image

packages/bpk-component-select/src/BpkSelect.js
matteo-hertel: it's following the same principle as the large config e.g. https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-select/src/__snapshots__/BpkSelect-test.js.snap#L72

with the `--` rather than `__`

matteo-hertel: Ignore that, I didn't understand your point 😅

matteo-hertel:
IE and Edger are working properly now

bs_win10_ie_11 0 1
bs_win10_ie_11 0
bs_win10_edge_16 0 1
bs_win10_edge_16 0

packages/bpk-component-select/src/BpkSelect-test.js
```diff
@@ -42,6 +45,52 @@ describe('BpkSelect', () => {
...
+          imageContainerClassName="imageContainerClass--test"
+          wrapperClassName="wrapperClass--test"
+          image={imageRenderProp}
+          onChange={() => null}
+        >
+          <option value="apples">Apples</option>
+          <option value="oranges">Oranges</option>
+          <option value="pears">Pears</option>
+          <option value="tomatos" disabled>
+            Tomatos
```
shaundon: I know this is just a test file, but it's `tomatoes` with an `e` 🙊

packages/bpk-component-select/stories.js
```diff
@@ -17,11 +17,96 @@
...
+    value: 'pear',
+    label: 'Pear',
+    image: 'http://via.placeholder.com/250x50/00d775/000000',
+    disabled: false,
+  },
+  {
+    key: 3,
+    value: 'tomato',
+    label: 'Tomato',
+    image: 'http://via.placeholder.com/250x50/ff5432/000000',
```
shaundon: I'm wondering if we should refrain from using another external service here. 

In the `FlatList` component [we use our own CDN to grab images for use in the storybook](https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-flat-list/stories.js#L39), maybe that'd be good here too (we could change the fruits to countries). It's still an 'external' service, but then we'd only be reliant on one service.

What do you think?

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -15,25 +15,46 @@
...
+export type Props = {
+  id: string,
+  name: string,
+  value: string,
+  className: ?string,
+  docked: boolean,
+  dockedFirst: boolean,
+  dockedLast: boolean,
+  dockedMiddle: boolean,
+  image: ?() => Node,
```
shaundon: I'm not convinced that this needs to be a render prop. 

I know that it's so users can change it based on what's selected, but they can do that already by wrapping it in a parent component which stores the selected item in state, which I imagine is how this component is already used by consumers. The fact that we aren't passing any arguments into the render function indicates to me that it isn't necessary.

Having said that, the convention for render props is to name them `renderThing` - so I think `renderImage` would be a better name here **if we do keep it as a render prop**.

packages/bpk-component-select/src/BpkSelect-test.js
matteo-hertel: well then..

packages/bpk-component-select/src/BpkSelect.js
matteo-hertel: 2bkm4v

packages/bpk-component-select/src/BpkSelect-test.js
matteo-hertel: it was a test wide typo, not my fault 😄

packages/bpk-component-select/src/bpk-select.scss
```diff
@@ -41,3 +41,48 @@
...
+
+@mixin bpk-select-changeme {
+  display: flex;
+  padding-right: 0;
+  flex-direction: row;
+  align-items: center;
+  background: none;
+}
+
+.bpk-select-changeme {
```
matteo-hertel: Need a better name for this, ideas?

matteo-hertel:
After a bit of back and forth, I have something that works now, here some screens


screen shot 2018-06-04 at 3 40 57 pm
screen shot 2018-06-04 at 3 41 10 pm
screen shot 2018-06-04 at 3 41 21 pm
screen shot 2018-06-04 at 3 41 38 pm

The default version with the flag looks a bit odd, It's preobably because of the pic

IE testing has been done as well

bs_win10_ie_11 0
bs_win10_ie_11 0 1
bs_win10_ie_11 0 2

@jamesf3rguson what do your designer-eyes see?

packages/bpk-component-select/src/bpk-select.scss
tiagoengel: It sounds like this should be a modifier, eg `bpk-select--with-image`. You wouldn't need to redefine all the other modifiers in this case. e.g. `className={getClassName('bpk-select', 'bpk-select--with-image')}`

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -15,81 +15,124 @@
...
-      className={classNames.join(' ')}
+      className={getClassName(
+        'bpk-select',
+        large && 'bpk-select--large',
+        docked && 'bpk-select--docked',
+        dockedFirst && 'bpk-select--docked-first',
+        dockedMiddle && 'bpk-select--docked-middle',
+        dockedLast && 'bpk-select--docked-last',
+        className,
+      )}
```
shaundon: This is beautiful.

packages/bpk-component-select/src/bpk-select.scss
```diff
@@ -40,4 +40,24 @@
...
+    flex-direction: row;
+    align-items: center;
+    background: none;
+
+    .bpk-select {
+      height: auto;
+      padding-top: 0;
+      padding-bottom: 0;
+      border: transparent;
+    }
```
shaundon: This is clever but I think it's kind of contravening BEM (slightly). How come we changed away from doing it as a mixin?

Also - this is probably not a modifier (my bad), it's probably better as an element like `bpk-select-image-wrapper`.

packages/bpk-component-select/stories.js
```diff
@@ -17,11 +17,84 @@
...
+  { key: 5, id: 'GD', name: 'Grenada', disabled: false },
+  { key: 6, id: 'HT', name: 'Haiti', disabled: false },
+  { key: 7, id: 'IT', name: 'Italy', disabled: false },
+  { key: 8, id: 'US', name: 'USA', disabled: true },
+];
+class SelectWithImage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      selected: 'IT',
```
shaundon: screen shot 2018-06-04 at 17 22 24

packages/bpk-component-select/stories.js
```diff
@@ -17,11 +17,84 @@
...
+    };
+  }
+
+  getItemByValue = () => {
+    const { options } = this.props;
+    return val => {
+      const items = options.filter(o => o.id === val);
+      if (!items.length) throw new Error('Item does not exists');
```
shaundon: Is eslint ok with this? I didn't think it liked inline `if` statements..

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -15,81 +15,124 @@
...
+      <div
+        className={getClassName(
+          'bpk-select',
+          'bpk-select--with-image',
+          large && 'bpk-select--large',
+          docked && 'bpk-select--docked',
+          dockedFirst && 'bpk-select--docked-first',
+          dockedMiddle && 'bpk-select--docked-middle',
+          dockedLast && 'bpk-select--docked-last',
+          wrapperClassName,
```
shaundon: This isn't valid BEM, as the class names are being re-used on this and the select element - so I think we need to introduce some repetition in the Sass to make these unique.

packages/bpk-component-select/stories.js
matteo-hertel: travis is happy so I'm assuming so 😄

packages/bpk-component-select/src/BpkSelect-test.js
```diff
@@ -33,8 +36,54 @@ describe('BpkSelect', () => {
...
+  });
+
+  it('should render correctly with image and custom classes', () => {
+    const tree = renderer
+      .create(
+        <BpkSelect
+          id="fruits"
+          name="fruits"
+          value="oranges"
+          imageContainerClassName="imageContainerClass--test"
```
matthewdavidson: this should be `imageWrapperClassName`

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -15,81 +15,124 @@
...
+  id: string,
+  name: string,
+  value: string,
+  className: ?string,
+  docked: boolean,
+  dockedFirst: boolean,
+  dockedLast: boolean,
+  dockedMiddle: boolean,
+  image: ?Node,
+  imageContainerClassName: ?string,
```
matthewdavidson: Should be `imageWrapperClassName`

matteo-hertel:
All the feedback are dealt with, also @jamesf3rguson was happy with the screenshot provided offline

shaundon:
png

PR #801, opened 2018-06-01
[No-Jira] POC dark docs examples

georgegillams:
Question for design: Should the background be themed? Or always be the same dark color?

screenshot_2018-06-01 badge backpack 1

### DIFF
--- a/packages/bpk-docs/src/components/DocsPageBuilder/NeoDocsPageBuilder.js
+++ b/packages/bpk-docs/src/components/DocsPageBuilder/NeoDocsPageBuilder.js
@@ -82,7 +82,10 @@ const ComponentExample = component => {
   );
 
   const examples = (component.examples || []).length ? (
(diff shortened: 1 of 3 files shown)


packages/bpk-docs/src/pages/BadgePage/BadgePage.js
```diff
@@ -136,42 +135,38 @@ const components = [
...
-          Bananas
-        </BpkBadge>
-        <BpkBadge type={BADGE_TYPES.light} className={badgeClassName}>
-          Strawberries
-        </BpkBadge>
-        <BpkBadge type={BADGE_TYPES.light} className={badgeClassName}>
-          Pears
-        </BpkBadge>
-      </BadgeLayout>,
+    examplesDark: [
```
tiagoengel: I'm not really sure about this. Wouldn't you think a new property, like `dark: true` would be better?

PR #804, opened 2018-06-01
Allow "node" as "children" in BpkBreakpoint

sogame:
About `children` prop in BpkBreakpoint:
* readme.md says it's a `node`: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-breakpoint/readme.md
* The code says it’s a `func`: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-breakpoint/src/BpkBreakpoint.js#L46
* `react-responsive` code says it any of both: https://github.com/contra/react-responsive/blob/master/src/index.js#L12

The inconsistency between readme.md and the code must be fixed.
Also, I guess there’s no reason in `BpkBreakpoint` for force just one of the two.


sogame:
Thanks @k0nserv! Rebased (conflicts in `unreleased.md` only)

PR #814, opened 2018-06-05
[NoJira] update select RTL

matteo-hertel:
While implementing the select component @tiagoengel found a problem with RTL, this fixes it

screen shot 2018-06-05 at 3 39 04 pm

### DIFF
--- a/packages/bpk-component-select/src/bpk-select.scss
+++ b/packages/bpk-component-select/src/bpk-select.scss
@@ -45,6 +45,10 @@
     height: auto;
     padding: 0;
     border: transparent;
+
+    @include bpk-rtl {
+      padding: 0;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/bpk-component-select/src/bpk-select.scss
```diff
@@ -91,6 +95,10 @@ $select-image-large-height: 22 * $bpk-one-pixel-rem;
     height: $select-image-height;
     margin-right: $bpk-spacing-xs;
 
+    @include bpk-rtl {
+      margin-left: $bpk-spacing-xs;
```
matthewdavidson: You should also have to reset the `margin-right` you previously set i.e:
```
@include bpk-rtl {
  margin-right: 0;
  margin-left: $bpk-spacing-xs;
}
```

packages/bpk-component-select/src/bpk-select.scss
```diff
@@ -59,6 +59,10 @@ $select-image-large-height: 22 * $bpk-one-pixel-rem;
   flex-direction: row;
   align-items: center;
   background: none;
+
+  @include bpk-rtl {
+    padding-left: 0;
```
matthewdavidson: You should also have to reset the margin-right you previously set i.e:

@include bpk-rtl {
  padding-right: ${some value};
  padding-left: 0;
}

matteo-hertel:
@matthewdavidson good shout, I've done it

PR #815, opened 2018-06-05
[BPK-1611] Add image to the web phone input

tiagoengel:


packages/bpk-component-select/readme.md
```diff
@@ -38,6 +38,7 @@ export default () => (
 | onChange                  | func       | true     | -             |
 | valid                     | bool       | false    | null          |
 | large                     | bool       | false    | false         |
+| disable                   | bool       | false    | false         |
```
shaundon: typo - should say `disabled`.

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -30,6 +30,7 @@ export type Props = {
   name: string,
   value: string,
   className: ?string,
+  disabled: ?boolean,
```
shaundon: I think this should just be `boolean`, as it isn't possible for it to be `null` due to having a defaultProp of `false`. Same as with `docked` etc.

tiagoengel:
We should've done it in the select card. I'm not really adding the disabled property, the select already accepts and styles the 'disabled' version (https://github.com/Skyscanner/backpack/blob/master/packages/bpk-mixins/src/mixins/_forms.scss#L325). 

The problem is that the image was not being correctly styled (as it has now a wrapper outside the original select).

shaundon:
@tiagoengel gotcha, thanks for clarifying 👍

tiagoengel:
Oh right, I meant to do that but forgot 🤦‍♂️

packages/bpk-component-phone-input/readme.md
```diff
@@ -13,6 +13,26 @@ npm install bpk-component-phone-input --save-dev
...
+import BpkImage from 'bpk-component-image';
+
+const DIALING_CODE_TO_ID_MAP = {
+  '44': 'uk',
+  '55': 'br',
+};
+
+const getFlag = dialingCode => {
+  const countryCode = DIALING_CODE_TO_ID_MAP[dialingCode];
+  const url = `https://images.skyscnr.com/images/country/flag/header/${countryCode}.png`;
```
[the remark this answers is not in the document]
tiagoengel: done

PR #819, opened 2018-06-06
[BPK-1611] Focus displays correctly on selects with image

matthewdavidson:
screen shot 2018-06-06 at 15 34 20
screen shot 2018-06-06 at 15 35 04

### DIFF
--- a/packages/bpk-component-phone-input/stories.js
+++ b/packages/bpk-component-phone-input/stories.js
@@ -32,15 +32,7 @@ const DIALING_CODE_TO_ID_MAP = {
 const getFlag = dialingCode => {
   const countryCode = DIALING_CODE_TO_ID_MAP[dialingCode];
   const url = `https://images.skyscnr.com/images/country/flag/header/${countryCode}.png`;
-  return (
-    <BpkImage
-      altText="Flag"
-      height={38}
-      width={50}
-      style={{ width: '100%' }}
-      src={url}
-    />
-  );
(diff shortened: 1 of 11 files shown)


packages/bpk-component-phone-input/stories.js
```diff
@@ -32,15 +32,7 @@ const DIALING_CODE_TO_ID_MAP = {
...
-  return (
-    <BpkImage
-      altText="Flag"
-      height={38}
-      width={50}
-      style={{ width: '100%' }}
-      src={url}
-    />
-  );
+  return <BpkImage altText="Flag" height={38} width={50} src={url} />;
```
tiagoengel: we should probably do the same change in the readme too now that the element doesn't need the extra style.

packages/bpk-component-select/src/bpk-select.scss
```diff
@@ -41,71 +41,65 @@
...
-  &--borderless {
-    height: auto;
-    padding: 0;
-    border: transparent;
+  &--with-image {
+    padding-left: $bpk-spacing-xl;
 
-    @include bpk-rtl {
-      padding: 0;
+    @media screen\0 {
```
matthewdavidson: These are to combat the IE fixes located in the `bpk-select` mixin -> https://github.com/Skyscanner/backpack/blob/d01d46016361a1d7f5cc28bdd27811341428a367/packages/bpk-mixins/src/mixins/_forms.scss#L321

PR #822, opened 2018-06-07
[NoJira] add e2e test for docs site

matteo-hertel:
This adds cypress for e2e testing, at the minute used just for sanity check while refactoring the docs site

Later on we can follow up with proper testing and travis integration


matteo-hertel:
After a discussion offline I'll re-create this with all the bells and whistles later on

PR #823, opened 2018-06-07
[No JIRA] Improve text input docs

shaundon:


native/packages/react-native-bpk-component-text-input/readme.md
```diff
@@ -100,3 +100,5 @@ export default class App extends Component {
 | validationMessage           | string                                                      | false    | null          |
 | style                       | style                                                       | false    | null          |
 | accessoryView               | node                                                        | false    | null          |
+
+**Note:** `BpkTextInput` wraps React Native's [Text Input component](https://facebook.github.io/react-native/docs/textinput.html) and exposes its props.
```
matthewdavidson: 🤔 Do we think this is really necessary? If we do this then we should probably do the same for:
- Alert
- Button
- ButtonLink
- Spinner
- Switch
- Text
etc

My thinking is that this is _implied_. The same for web i.e. buttons, links, inputs etc are all styling wrappers around "native" underlying components

shaundon: I don't know if the implication is clear enough - we had somebody yesterday not realise this, and I think it's easy to think it's implied from our experienced point of view, but consumers may not know this.

On some components we used to include this in the blurb on the docs site page. I'm certain we did this for buttons, and I think also switch and spinner, but when the docs site was rewritten these blurbs were removed. So there is a precedent for it too.

matthewdavidson: Good spot - however I think we should find out why they were removed first before we add them in again.

shaundon: @jamesf3rguson you wrote the blurbs - could you clarify?

jamesf3rguson: yeh I removed as I  didn't think it was necessary and we kept repeating it. I vote not to have it

shaundon: Ok fair enough, I'll go with the majority opinion here and leave it out 😄 

I do however reserve the right to use the 'I told you so' Bitmoji if this issue crops up again 😜

matthewdavidson:
bitmoji

PR #827, opened 2018-06-07
Add "overflow: hidden" to "bpk-border-radius" mixins

sogame:
We're using `bpk-border-radius` mixins sometimes to show rounded corners in images, and this requires adding `overflow: hidden` also.

It would be good if this is done automatically by the mixins. Do you see any possible side effect with this change?

### DIFF
--- a/packages/bpk-mixins/src/mixins/_radii.scss
+++ b/packages/bpk-mixins/src/mixins/_radii.scss
@@ -31,6 +31,7 @@
 
 @mixin bpk-border-radius-xs {
   border-radius: $bpk-border-radius-xs;
+  overflow: hidden;
 }
 
 /// Small border radius.
@@ -42,6 +43,7 @@
 
 @mixin bpk-border-radius-sm {
   border-radius: $bpk-border-radius-sm;
+  overflow: hidden;
 }
 
 /// Pill shaped border radius.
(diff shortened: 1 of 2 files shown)


tiagoengel:
I'd be agains this change @sogame because I think it has a high change of breaking existing code. I'm sure at least a few people are using this mixin inside a div that has overflowing items (and needs those to be visible). It also seems to me that people will not expect a border radius mixin to also add `overflow: hidden` by default, as this is not directly related to border radius style.

sogame:
Thanks @tiagoengel makes total sense, I'll close this PR :-)

One question though: do you think it would make sense to create `bpk-border-radius-image` mixins that do both things?
One one hand, it would be handy, especially for people that could struggle to understand why the border radius is not being applied to the corners of images. But on the other, maybe it adds too much complexity, with so many border-radius mixins...

What do you think about this? Other pros/cons?

matthewdavidson:
The difference between...
```scss
@include bpk-border-radius-sm;
@include bpk-border-radius--with-image;
```
...and...
```scss
overflow: hidden;

@include bpk-border-radius-sm;
```

...is negligible. We aren't really expanding the Sass mixin library too much just now. I think it's fine for this to stay in userland.

sogame:
I agree, I was thinking of just `@include bpk-border-radius--with-image;`, and this one doing both things, but it's true that the pattern used in Backpack is the one you both mentioned, so fine for me.

But I think it's a good idea what Tiago mentioned, about documenting this, so it can be easier to understand for people who can struggle to find the fix :-)

PR #828, opened 2018-06-07
Lint scss files on commit

sogame:
I just noticed that lint-staged is used to lint only js files. I think it would be very useful (at least for contributors) to lint also scss files.


matteo-hertel:
Thanks for the contribution you deserve a medal 🎖

PR #830, opened 2018-06-08
[NO-JIRA] Android - Fix unresolved dependency on common.jar

georgegillams:
On a clean machine, following the instructions in contributing.md leads to a number of build errors in the Android project. Making the changes in this PR makes the instructions work flawlessly. I think that we should merge these changes in so as to align the gradle project with the contributing.md file.

### DIFF
--- a/native/android/build.gradle
+++ b/native/android/build.gradle
@@ -16,6 +16,7 @@ buildscript {
 allprojects {
     repositories {
         mavenLocal()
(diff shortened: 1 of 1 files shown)


k0nserv:
I think we should revert this. With these changes we are using the Android P build tools, sdk and target SDK. Android P is still in beta and we shouldn't be using it yet

PR #831, opened 2018-06-08
[BPK-1568] clean up part 2

matteo-hertel:
This is part 2 (almost there I promise)

The only bit left to remove are the webpack config and package.json, I'll follow up with another PR.

I 've also fixed a bunch of broken links and console warning

it'll be quite hard to review, sorry about that, have a cookie


packages/bpk-docs/src/pages/ThemingPage/ThemingPage.js
```diff
@@ -85,18 +80,16 @@ const webBlurb = [
...
     </BpkListItem>
     <BpkListItem>
-      <BpkRouterLink to={`${TYPOGRAPHY}#blockquotes`}>
-        Blockquotes
-      </BpkRouterLink>
+      <BpkRouterLink to={BLOCKQUOTE}>Blockquotes</BpkRouterLink>
     </BpkListItem>
     <BpkListItem>
-      <BpkRouterLink to={BUTTONS}>Button</BpkRouterLink>
+      <BpkRouterLink to={BUTTON}>Button</BpkRouterLink>
```
tiagoengel: Any reason we're not linking to the web tab (platform=web) here too? Same question for all bellow that have a native version.

matteo-hertel: Good shout, at the moment the link on both native and web goes to native so I'll update that to improve it

matteo-hertel:
@tiagoengel thanks for the feedback I've updated all the link to go to the correct section now!

PR #834, opened 2018-06-11
Bpk 1568 post neo cleanup part3

matteo-hertel:
This gets rid of all the Neo mentions, moves the neo components and makes them the default 

unfortunately is a humongous PR 😅 

I'll squash and whatnot if travis is happy and somebody casts an eye on it


packages/bpk-docs/src/components/DocsPageBuilder/DocsPageBuilder.js
```diff
@@ -134,55 +153,75 @@ const CustomSection = section => [
...
-          [...props.components, ...props.customSections].map(
-            ExampleNavListItem,
-          ),
-        )}
-      </BpkList>
-    )}
-    {props.tokenMap ? <TokenSwitcher tokens={props.tokenMap} /> : null}
-    {props.usageTable
-      ? flatten([
+const NeoDocsPageBuilder = props => {
```
k0nserv: This shouldn't be called `NeoDocsPageBuilder` anymore should it?

packages/bpk-docs/src/pages/HomePageOld/HomePage.js
```diff
@@ -0,0 +1,104 @@
+/*
```
k0nserv: Why do we need this?

packages/bpk-docs/src/components/DocsPageBuilder/DocsPageBuilder.js
matteo-hertel: Ah I missed that, thanks!

packages/bpk-docs/src/pages/HomePageOld/HomePage.js
matteo-hertel: we don't, thanks for spotting it

PR #841, opened 2018-06-13
[BPK-1642] Use RecyclerListView in BpkCarousel

shaundon:


native/package.json
```diff
@@ -30,7 +30,7 @@
     "verbose": true,
     "testRegex": ".*-test(\\.ios)?(\\.android)?\\.js$",
     "transformIgnorePatterns": [
-      "node_modules/(?!react-native|bpk|@skyscanner)"
+      "node_modules/(?!react-native|bpk|@skyscanner|lodash-es)"
```
shaundon: https://github.com/Flipkart/recyclerlistview/issues/171

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -107,26 +126,21 @@ class BpkCarousel extends React.Component<Props, State> {
...
-        >
-          {shouldRenderChildren &&
-            React.Children.map(children, child =>
-              React.cloneElement(child, {
-                ...child.props,
-                style: [child.props.style, { width, height }],
-              }),
-            )}
-        </ScrollView>
+      <View style={style} onLayout={this.onLayout}>
```
shaundon: I switched `Fragment` to `View` as `onLayout` isn't being called on the `RecyclerListView`, which we rely on. This was recently fixed, so when the next update is released we may be able to change this back: https://github.com/Flipkart/recyclerlistview/pull/193

native/packages/react-native-bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -68,7 +69,15 @@ class BpkCarousel extends React.Component<Props, State> {
 
   constructor(props: Props) {
     super(props);
-    this.state = { width: null, height: null, currentIndex: 0 };
+    const dataProvider = new DataProvider((r1, r2) => r1 !== r2);
```
tiagoengel: shouldn't we compare props here? i.e. `shallow(r1.props, r2.props)`

shaundon: Tbh I just copied this from the sample code in the repo: https://github.com/Flipkart/recyclerlistview/tree/master/docs/guides/samplecode

The current implementation in the app does it the same way too (I can't link to it here because it's a private repo)

tiagoengel: The thing is in both cases they are comparing simple objects, in the app's case it was image urls and in the example just a simple array of integers, so `r1 !== r2` is fine. We are comparing react elements here so I'm not sure if this is the correct way of deciding if they've changed and/or are different.

shaundon: Yeah good point. I'm not sure if there's a better way to compare them though, as the component could be anything.. any ideas?

shaundon: FWIW this was the implementation we used when doing performance testing and it seemed pretty performant too.

tiagoengel: Yeah I'm not sure about this either, I guess if the performance is good then it's ok

PR #842, opened 2018-06-13
[BPK-1644] Update ios text style screenshots on master

matthewdavidson:
This is a prerequisite for https://github.com/Skyscanner/backpack/pull/843.

I discovered that the iOS text style screenshots were different to the source on a fresh checkout on master. This is due to the old screenshots being taken on a simulator running iOS 10.3 (as opposed to what we are running now, 11.2).

Difference (10.3 on left, 11.2 on right):
difference

This PR updates the screenshots to provide a new baseline.


matthewdavidson:
I'm going to close this in light of the issue we have discovered with React Native and iOS 11.2 `font-weight` rendering. Please see https://github.com/Skyscanner/backpack/pull/843 for more details.

PR #844, opened 2018-06-13
[BPK-1630] Enhance map component

shaundon:
Previously, `BpkMap` was a thin layer around `GoogleMap`. Now, I've whitelisted props on it to make the API closer to the map we use in React Native. For example, `onZoomChanged()` from `GoogleMap` is now `onZoom(newZoomLevel)`.

I've also removed all the exports from `react-google-map` that we were re-exporting previously. Because one of these exports, `OverlayView`, was already in use by consumers, I've created `BpkOverlayView` to serve this use case. The code for it is based on the code that Blue Dolphin did in their app.

I've deliberately called the component `BpkOverlayView` and **not** `BpkMapMarker` to leave scope for adding a simpler marker component in future.

The readme, storybook and docs have been updated to suit this new way of doing things.

**Other things**
* Added a default loading spinner.
* Added defaults for `mapElement` and `containerElement`.
* Added the ability to achieve functionality that previously required usage of `refs` using props (e.g. `fitBounds()`.
* Added new types for `LatLong` and `Bounds`. `LatLong` is abstracted into `common-types` and exported.
* Added a `showControls` prop which is passed down to `disableDefaultUI`.
* Hard coded some options to false, based on how consumers currently use the component (we can amend this later if necessary).

### DIFF
--- a/packages/bpk-component-map/package.json
+++ b/packages/bpk-component-map/package.json
@@ -13,11 +13,14 @@
     "react": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
+    "bpk-component-spinner": "^2.2.34",
+    "bpk-mixins": "^17.10.4",
+    "bpk-react-utils": "^2.6.1",
+    "prop-types": "^15.5.8",
     "react-google-maps": "^9.4.5"
   },
   "devDependencies": {
-    "bpk-component-icon": "^3.21.4",
-    "bpk-component-tooltip": "^3.1.18",
-    "prop-types": "^15.5.8"
+    "prop-types": "^15.5.8",
+    "bpk-component-text": "^1.0.64"
   }
 }
--- a/packages/bpk-component-map/readme.md
+++ b/packages/bpk-component-map/readme.md
@@ -12,56 +12,52 @@ npm install bpk-component-map --save-dev
 
 ```js
 import React from 'react';
-import BpkMap from 'bpk-component-map';
+import BpkText from 'bpk-component-text';
+import BpkMap, { BpkOverlayView } from 'bpk-component-map';
 
 export default () => (
   <BpkMap
-    containerElement={<div style={{ height: '400px' }} />}
-    mapElement={<div style={{ height: `100%` }} />}
-    defaultZoom={15}
-    defaultCenter={{
-      lat: 27.9881,
-      lng: 86.925,
+    zoom={15}
+    showControls={false}
+    panEnabled={false}
+    center={{
+      latitude: 27.9881,
+      longitude: 86.925,
     }}
-    options={{
-      zoomControl: false,
-      dragEnabled: false,
-    }}
-  />
+  >
+    <BpkOverlayView position={{ latitude: 27.9881, longitude: 86.925 }}>
+      <BpkText>Shibuya Crossing</BpkText>
+    </BpkOverlayView>
+  <BpkMap>
 );
 ```
 
 ## Accompanying HOCs
 
-### withScriptjs
+### withGoogleMapsScript
 
-`withScriptjs` is a HOC that loads the Google Maps Javascript, then loads the map. This is useful for when you don't already have the Google Maps Javascript loaded.
+`withGoogleMapsScript` is a HOC that loads the Google Maps Javascript, then loads the map. This is useful for when you don't already have the Google Maps Javascript loaded.
 
 If you intend to include multiple maps on one page, it's better to load the Google Maps Javascript elsewhere and not use this HOC, as it downloads the script every time it's used.
 
 ```js
 import React from 'react';
-import BpkMap, { withScriptjs } from 'bpk-component-map';
+import BpkMap, { withGoogleMapsScript } from 'bpk-component-map';
 
 const MAP_URL = 'https://maps.googleapis.com/maps/api/js?v=3.exp&libraries=geometry,drawing,places';
 
-const BpkMapWithScript = withScriptjs(BpkMap);
+const BpkMapWithScript = withGoogleMapsScript(BpkMap);
 
 export default () => (
   <BpkMapWithScript
     googleMapURL={MAP_URL}
-    loadingElement={<div />}
-    containerElement={<div style={{ height: '400px' }} />}
-    mapElement={<div style={{ height: `100%` }} />}
-    defaultZoom={15}
-    defaultCenter={{
-      lat: 27.9881,
-      lng: 86.925,
+    zoom={15}
+    center={{
+      latitide: 27.9881,
+      longitude: 86.925,
     }}
-    options={{
-      zoomControl: false,
-      dragEnabled: false,
-    }}
+    showControls={false}
+    panEnabled={false}
   />
 );
 ```
@@ -70,12 +66,33 @@ export default () => (
 
 ### BpkMap
 
-| Property	      | PropType	| Required                	| Default Value |
-| --------------- | --------- | ------------------------- | ------------- |
-| mapRef          | func      | false                     | null          |
+| Property	       | PropType                                                          | Required                 | Default Value                    |
+| ---------------- | ----------------------------------------------------------------- | ------------------------ | -------------------------------- |
+| bounds           | shape({north: number, east: number, south: number, west: number}) | false                    | null                             |
+| center           | shape({latitude: number, longitude: number})                      | false                    | null                             |
+| containerElement | node                                                              | false                    | <div style={{height: '100%'}} /> |
+| mapElement       | node                                                              | false                    | <div style={{height: '100%'}} /> |
+| mapRef           | func                                                              | false                    | null                             |
+| onRegionChange   | func                                                              | false                    | null                             |
+| onZoom           | func                                                              | false                    | null                             |
+| panEnabled       | bool                                                              | false                    | true                             |
+| showControls     | bool                                                              | false                    | true                             |
+| zoom             | number                                                            | false                    | 15                               |
 
-When using `withScriptjs`, some additional props are required. Refer to [`withScriptjs` from `react-google-maps`](https://tomchentw.github.io/react-google-maps/#withscriptjs).
+Note: One of `bounds` and `center` must be provided.
 
-Refer to [`GoogleMap` from `react-google-maps`](https://tomchentw.github.io/react-google-maps/#withgooglemap) for all other props.
+#### withGoogleMapsScript
 
-> Note: `bpk-component-map` also exports everything that `react-google-maps` does, such as `InfoBox` and `OverlayView`.
+When using `withGoogleMapsScript`, some additional props are available:
+
+| Property	       | PropType                                     | Required                  | Default Value  |
+| ---------------- | -------------------------------------------- | ------------------------- | -------------- |
+| googleMapURL     | string                                       | true                      | -              |
+| loadingElement   | node                                         | false                     | BpkSpinner     |
+
+### BpkOverlayView
+
+| Property	       | PropType                                     | Required                 | Default Value      |
+| ---------------- | -------------------------------------------- | ------------------------ | ------------------ |
+| children         | node                                         | true                     | -                  |
+| position         | shape({latitude: number, longitude: number}) | true                     | -                  |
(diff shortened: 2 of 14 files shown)


packages/bpk-component-map/readme.md
```diff
@@ -70,12 +71,39 @@ export default () => (
 
 ### BpkMap
 
-| Property	      | PropType	| Required                	| Default Value |
-| --------------- | --------- | ------------------------- | ------------- |
-| mapRef          | func      | false                     | null          |
+| Property	       | PropType                                     | Required                 | Default Value |
+| ---------------- | -------------------------------------------- | ------------------------ | ------------- |
+| containerElement | node                                         | true                     | -             |
+| mapElement       | node                                         | true                     | -             |
```
matthewdavidson: Can we provide default implementations for `containerElement` and `mapElement` ?

packages/bpk-component-map/readme.md
```diff
@@ -70,12 +71,39 @@ export default () => (
...
 
-When using `withScriptjs`, some additional props are required. Refer to [`withScriptjs` from `react-google-maps`](https://tomchentw.github.io/react-google-maps/#withscriptjs).
+### BpkOverlayView
 
-Refer to [`GoogleMap` from `react-google-maps`](https://tomchentw.github.io/react-google-maps/#withgooglemap) for all other props.
+| Property	       | PropType                                     | Required                 | Default Value      |
+| ---------------- | -------------------------------------------- | ------------------------ | ------------------ |
+| children         | node                                         | true                     | -                  |
+| position         | shape({latitude: number, longitude: number}) | true                     | -                  |
+| mapPaneName      | oneOf(MAP_PANE_NAMES)                        | false                    | overlayMouseTarget |
```
matthewdavidson: Lets remove this from the docs...

packages/bpk-component-map/readme.md
```diff
@@ -70,12 +71,39 @@ export default () => (
...
 ### BpkMap
 
-| Property	      | PropType	| Required                	| Default Value |
-| --------------- | --------- | ------------------------- | ------------- |
-| mapRef          | func      | false                     | null          |
+| Property	       | PropType                                     | Required                 | Default Value |
+| ---------------- | -------------------------------------------- | ------------------------ | ------------- |
+| containerElement | node                                         | true                     | -             |
+| mapElement       | node                                         | true                     | -             |
+| region           | shape({latitude: number, longitude: number}) | true                     | -             |
```
matthewdavidson: Switch this to something like `center`...

packages/bpk-component-map/readme.md
```diff
@@ -70,12 +71,39 @@ export default () => (
...
 
-Refer to [`GoogleMap` from `react-google-maps`](https://tomchentw.github.io/react-google-maps/#withgooglemap) for all other props.
+| Property	       | PropType                                     | Required                 | Default Value      |
+| ---------------- | -------------------------------------------- | ------------------------ | ------------------ |
+| children         | node                                         | true                     | -                  |
+| position         | shape({latitude: number, longitude: number}) | true                     | -                  |
+| mapPaneName      | oneOf(MAP_PANE_NAMES)                        | false                    | overlayMouseTarget |
+
+
+## Methods
```
matthewdavidson: Remove these (values can be retrieved using callbacks) and add alternative prop for `fitBounds`

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -18,11 +18,141 @@
...
+  scrollEnabled: boolean,
+  zoom: number,
+  zoomEnabled: boolean,
+  bounds: ?{
+    south: number,
+    west: number,
+    north: number,
+    east: number,
+  },
+  center: ?{
```
k0nserv: This object is going to be so common that it's probably a good idea to create a distinc type for it instead of repeating it in many places

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -18,11 +18,141 @@
...
+    north: number,
+    east: number,
+  }) => void,
+};
+
+type Props = {
+  scrollEnabled: boolean,
+  zoom: number,
+  zoomEnabled: boolean,
+  bounds: ?{
```
k0nserv: This could also have a distinc type

packages/bpk-component-map/src/BpkOverlayView.js
```diff
@@ -0,0 +1,53 @@
...
+    latitude: number,
+    longitude: number,
+  },
+};
+
+const BpkOverlayView = (props: Props) => {
+  const { children, position, ...rest } = props;
+  return (
+    <OverlayView
+      mapPaneName="overlayMouseTarget"
```
k0nserv: What's this used for and is it safe to hardcode it like this?

packages/bpk-component-map/src/BpkOverlayView.js
```diff
@@ -0,0 +1,53 @@
...
+      position={{ lat: position.latitude, lng: position.longitude }}
```
k0nserv: This seems like a common enough operation that we might want to have a function for it?

packages/bpk-component-map/src/BpkMap.js
shaundon: Good shout - I've done that now.

packages/bpk-component-map/src/BpkMap.js
shaundon: Good shout - I've done that now.

packages/bpk-component-map/src/BpkOverlayView.js
shaundon: @matthewdavidson and I discussed it and think it's ok. In the original PR this was a prop, but we decided to leave it hard-coded, and let consumers override it with `...rest` if they really need to. 

This is the documentation for it: https://developers.google.com/maps/documentation/javascript/reference/3.exp/overlay-view#MapPanes.floatPane

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -18,11 +18,133 @@
...
+    onZoom,
+    center,
+    scrollEnabled,
+    zoom,
+    zoomEnabled,
+  } = props;
+  let ref: MapRef = null;
+
+  if (!bounds && !center) {
+    throw new Error('BpkGoogleMap: Provide either `bounds` or `center` props.');
```
sogame: This should be `BpkMap`, isn't it?

packages/bpk-component-map/package.json
```diff
@@ -13,11 +13,13 @@
...
+    "bpk-react-utils": "^2.6.1",
+    "prop-types": "^15.5.8",
     "react-google-maps": "^9.4.5"
   },
   "devDependencies": {
-    "bpk-component-icon": "^3.21.1",
-    "bpk-component-tooltip": "^3.1.15",
-    "prop-types": "^15.5.8"
+    "bpk-component-spinner": "^2.2.31",
+    "bpk-component-text": "^1.0.61"
```
sogame: One question: should `react-google-maps` peer-dependencies (`@types/googlemaps`, `@types/markerclustererplus`, `@types/react`) be added as `devDependencies` to isolate the clients from those internal details?

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -18,11 +18,133 @@
...
+          ? {
+              lat: center.latitude,
+              lng: center.longitude,
+            }
+          : null
+      }
+      defaultZoom={zoom}
+      options={{
+        gestureHandling: scrollEnabled ? 'auto' : 'none',
+        zoomControl: zoomEnabled,
```
sogame: We would need to be able to enable/disable more controls: `draggable`, `mapTypeControl`, `streetViewControl`, `fullscreenControl`, `rotateControl`.

This is the wrapper we use around the current `BpkMap`, and the options we set: https://gitlab.skyscannertools.net/hotels-web/hotels-website/blob/master/client/src/components/MapWrapper/MapWrapper.jsx#L132-140

Also, what do you think about adding a `dynamic` prop that when set to false disables all controls and interactions? This is useful when using the map as a static map (in our case, to show an "image" of the hotel location).

You can see our current map usage, both static and dynamic here: http://hotels-website.slingshot.eu-west-1.sandbox.aws.skyscanner.local/47016457

packages/bpk-component-map/src/BpkMap.js
shaundon: Whoops! Yes, it should 😅

packages/bpk-component-map/src/BpkMap.js
shaundon: Thanks for this. I'll update it to support this. The links are very helpful too.

> Also, what do you think about adding a dynamic prop that when set to false disables all controls and interactions? This is useful when using the map as a static map (in our case, to show an "image" of the hotel location).

I like the intention of this, but I think we should hold off on it for now. The reason I say this is that we may want to achieve the same behaviour in a different way, for example a component that displays a static map image (so it would look the same, but actually just show an image), which may help us use fewer API calls. We're talking to Gadget squad about doing something like this in the future.

sogame: Great! It would be very good to be able to use just an image when showing an static map :-)

Keep us updated on that component, as you can see we'll use it also :-P
For now we'll keep disabling all interactions and controls when "simulating" the static map.

shaundon: `draggable` is [deprecated](https://developers.google.com/maps/documentation/javascript/reference/3.exp/map#MapOptions.draggable), they recommend using `gestureHandling: 'none'`, which we're doing in this file on line 98.

shaundon: I think we can use [`disableDefaultUI`](https://developers.google.com/maps/documentation/javascript/reference/3.exp/map#MapOptions.disableDefaultUI) as a way to toggle all of those controls at once (except draggable). What do you think about us using that instead - can you foresee any issues on your end? For example, a situation where you'd want to disable some but not others.

packages/bpk-component-map/package.json
shaundon: I'm not sure - the thing is those peer dependencies are all TypeScript definition packages, which aren't needed here. 

It would be good to not have the warning that they aren't there, but I'm not sure that we should be installing packages we aren't using just for the sake of that.

matthewdavidson: If it means that we swallow the errors as opposed to multiple consumers having to deal with them then it might be the lesser of two evils.

We can even follow up with PR to address.

shaundon: > We can even follow up with PR to address.

@matthewdavidson you mean follow up with a PR back to `react-google-maps`?

matthewdavidson: Yeah, if it bugs us so much :) We cant be the only ones that don't use typescript

shaundon: Yeah sounds good. I'm gonna leave it out of this PR though because scope creep, if you don't mind. I'll follow up with another after this.

packages/bpk-component-map/src/BpkMap.js
sogame: Good one about `draggable` :-D

About `disableDefaultUI`, as we discussed (and just for the record here), we need to be able to disable just some controls, like full screen, satellite view, streetview... so we'll still require fine grained access to the different controls.

packages/bpk-component-map/readme.md
```diff
@@ -12,56 +12,53 @@ npm install bpk-component-map --save-dev
...
 
-`withScriptjs` is a HOC that loads the Google Maps Javascript, then loads the map. This is useful for when you don't already have the Google Maps Javascript loaded.
+`withGoogleMapsScript` is a HOC that loads the Google Maps Javascript, then loads the map. This is useful for when you don't already have the Google Maps Javascript loaded.
 
 If you intend to include multiple maps on one page, it's better to load the Google Maps Javascript elsewhere and not use this HOC, as it downloads the script every time it's used.
 
 ```js
 import React from 'react';
-import BpkMap, { withScriptjs } from 'bpk-component-map';
+import { BpkSpinner, SPINNER_TYPES } from 'bpk-component-spinner';
```
k0nserv: Why is this imported? It's not used AFAICT

shaundon: Good spot! I'll remove it.

PR #845, opened 2018-06-14
small tweaks in documentation

arnaugm:
- datepicker
- pagination
- select
- alignment

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -202,6 +202,7 @@ The “page” modules themselves contain introductory blurbs and examples for t
 You can run the docs app locally using:
 
 ```sh
+npm run build
 npm run docs
 ```
 
--- a/packages/bpk-component-datepicker/readme.md
+++ b/packages/bpk-component-datepicker/readme.md
@@ -15,6 +15,7 @@ import React, { Component } from 'react';
 import BpkDatepicker from 'bpk-component-datepicker';
 import format from 'date-fns/format';
 
+const formatDate = date => format(date, 'DD/MM/YYYY');
 const formatDateFull = date => format(date, 'dddd, Do MMMM YYYY');
 const formatMonth = date => format(date, 'MMMM YYYY');
 const daysOfWeek = [
--- a/packages/bpk-component-select/readme.md
+++ b/packages/bpk-component-select/readme.md
@@ -19,7 +19,7 @@ export default () => (
     id="fruits"
     name="fruits"
     value="oranges"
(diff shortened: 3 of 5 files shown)


contributing.md
```diff
@@ -202,7 +202,8 @@ The “page” modules themselves contain introductory blurbs and examples for t
 You can run the docs app locally using:
 
 ```sh
-npm run docs
+npm run build
```
arnaugm: is not obvious that you need to create the build first, you may have done it before if you follow the whole guide but if you jump directly to documentation you may have missed the step

packages/bpk-component-datepicker/readme.md
```diff
@@ -15,6 +15,7 @@ import React, { Component } from 'react';
 import BpkDatepicker from 'bpk-component-datepicker';
 import format from 'date-fns/format';
 
+const formatDate = date => format(date, 'DD/MM/YYYY');
```
arnaugm: that was used in the example but the declaration was missing, probably because is copied from calendar where this is not needed

packages/bpk-component-select/readme.md
```diff
@@ -19,7 +19,7 @@ export default () => (
     id="fruits"
     name="fruits"
     value="oranges"
-    onChange={() => console.log('select changed')}
+    onChange={(e) => console.log(`select changed to ${e.target.value}`)}
```
arnaugm: makes evident what you receive in the callback, I though it was directly the value

packages/bpk-docs/src/pages/AlignmentPage/AlignmentPage.js
```diff
@@ -163,6 +165,41 @@ const AlignedSpan = withAlignment(
       </BpkText>,
     ],
   },
+  {
+    id: 'smallWithIcon',
+    title: 'Text with textStyle "sm" alignment with an icon',
+    blurb: [
+      <Paragraph>
+        Similarly, to use a small text, line height and the BpkText wrapper need
+        to use the small size:
```
arnaugm: not all the cases are covered in examples so important to inform that the elements to play with are the size in the BpkText and the line height

contributing.md
```diff
@@ -202,7 +202,8 @@ The “page” modules themselves contain introductory blurbs and examples for t
...
+npm run docs:neo
```
k0nserv: This script has been removed now it should just be `npm run docs`

arnaugm: changed

PR #846, opened 2018-06-14
[BPK-1640] Improve carousel indicator performance

tiagoengel:
Part of this fix was done in the transitiongroup lib: see https://github.com/Skyscanner/react-native-transitiongroup/commit/da2ef6e710c85f1466cd933bb9cbd774b86a6b67


shaundon:
This looks good, but the tests are failing. It looks like snapshots just need to be updated.

PR #849, opened 2018-06-18
[BPK-1629] Add eslint-plugin-backpack to Backpack

shaundon:


k0nserv:
I feel like we should fix it

PR #850, opened 2018-06-18
Bpk 1517 update android api version

amolgupta:


contributing.md
```diff
@@ -286,23 +286,25 @@ Accept the SDK licences:
 $ANDROID_SDK_ROOT/tools/bin/sdkmanager --licenses
 ```
 
-Download an Android system image. Note that you may get a warning about a `.cfg` file not being present.
-You're safe to ignore this.
+Download Android system images for the minimum and targeted API levels. Note that you may get a warning about a `.cfg` file not being present.You're safe to ignore this.
```
matthewdavidson: We need a space after first sentence. i.e. `present. You're safe to ignore this.`

amolgupta:
- completed testing of components on level 21 emulator
- changed gradle `compile` to `implementation`
- Build tools and and minSDKVersion aligned with main app
- Changed window background for consistency

native/android/app/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,19 @@
...
+     Unless required by applicable law or agreed to in writing, software
+     distributed under the License is distributed on an "AS IS" BASIS,
+     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     See the License for the specific language governing permissions and
+     limitations under the License.
+-->
+
+<resources>
+
+    <color name="accent">#00b2d6</color>
```
matthewdavidson: Can't we use https://github.com/Skyscanner/backpack-android here?

native/android/app/build.gradle
```diff
@@ -162,9 +162,9 @@ android {
 }
 
 dependencies {
-    compile fileTree(dir: "libs", include: ["*.jar"])
-    compile "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
-    compile "com.facebook.react:react-native:+" // From node_modules
+    implementation fileTree(dir: "libs", include: ["*.jar"])
```
matthewdavidson: Whats the deal here? can you link some docs? :)

amolgupta: done.

native/android/app/src/main/res/values/colors.xml
amolgupta: Sounds good. I will add the dependency and use references

native/android/app/build.gradle
matthewdavidson: Where are the docs? :)

native/android/app/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <color name="accent">@color/bpkBlue500</color>
+
+    <!--default background on API level 26 onwards -->
+    <color name="background">#FAFAFA</color>
```
matthewdavidson: This should use a backpack token too right? Also whats the requirement for this file? It just redefines backpack tokens under a new alias, cant we just use them directly in `native/android/app/src/main/res/values/styles.xml` ?

native/android/app/build.gradle
amolgupta: I have linked to them in the comment. Will mention on the PR merge notes also.

native/android/app/src/main/res/values/colors.xml
amolgupta: tokens do not have a value for `#FAFAFA` which is the default window background on new Android versions. Also `styles.xml` contains styles and `colors.xml` contains only color values. Just an Android convention.

matthewdavidson: > tokens do not have a value for #FAFAFA which is the default window background on new Android versions.

We want to use a backpack token for this though :) I suggest `colorWhite`.

> Also styles.xml contains styles and colors.xml contains only color values. Just an Android convention.

If we are just consuming backpack colours though, we don't need `colors.xml` do we? We just consume the colours from backpack directly in `styles.xml`? Or is it common to redeclare resources from dependencies?

native/android/app/build.gradle
matthewdavidson: Gotcha, comment being [this one](https://github.com/Skyscanner/backpack/pull/850#issuecomment-398094488)

The pertinent bit:
> The `compile` configuration still exists but should not be used as it will not offer the guarantees that the api and `implementation` configurations provide.

native/android/app/src/main/res/values/colors.xml
shaundon: > We want to use a backpack token for this though :) I suggest colorWhite.

Bear in mind that if we *do* use something other than #fafafa, we'll need to retake every Android screenshot.

Not saying we should, just something to be aware of as it'll add scope.

matthewdavidson: I'm happy to do it in another PR to chunk up the blast radius.

In almost all cases we should be using a backpack token. 🎒🙌

PR #855, opened 2018-06-20
[BPK-1679] Add dark version of scrim

k0nserv:
Supersedes #854


packages/bpk-scrim-utils/src/withScrim.js
```diff
@@ -38,6 +39,29 @@ const getClassName = cssModules(STYLES);
 
 const withScrim = WrappedComponent => {
   class WithScrim extends Component {
+    static propTypes = {
+      dark: PropTypes.bool,
+      getApplicationElement: PropTypes.func.isRequired,
```
shaundon: Should put the required one at the top to match convention.

k0nserv: Yeah, good catch

PR #856, opened 2018-06-20
[BPK-1623] Fix icon mapping generation

tiagoengel:
To test this you can:

1. add a test icon (duplicate existing file)
2. run `npm run build`
3. run git add (do NOT commit)
4. run `npm run build` again
5. check there are no new changes to be committed.

### DIFF
--- a/packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
+++ b/packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
@@ -61,18 +61,16 @@ export default (codepointsPath: string) => {
         return obj;
       }, {});
 
-    const newFileContents = `${JSON.stringify(sortedCodepoints, null, 2)}\n`;
-
-    return fs.writeFile(absolutePath, newFileContents, err => {
-      const data = err
-        ? null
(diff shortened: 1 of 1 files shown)


packages/bpk-svgs/tasks/getIconFontMetadataProvider.js
```diff
@@ -61,18 +61,16 @@ export default (codepointsPath: string) => {
...
-        ? null
-        : {
-            path: file,
-            name,
-            unicode: [String.fromCodePoint(codepoints[name])],
-          };
-
-      callback(err, data);
-    });
+    try {
```
tiagoengel: A little bit of context:

- This logic handles new files (that is, files that are not in the `codepoints.json` file)
- Line 36 handles files that are already present in that file
- The first time `run build` is executed for a new file, it will be handled by this code
- The next times it will be handled in line 36
- Because of how gulp streams work calling the callback asynchronously means it will be "emitted" in a random order (in this case always last as all other files are handled synchronously). 

Handling both situations synchronously ensures we have consistent results between executions.

PR #857, opened 2018-06-20
Greenkeeper/gulp iconfont 10.0.1

matthewdavidson:
In favour of https://github.com/Skyscanner/backpack/pull/839

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10420,30 +10420,41 @@
         }
       }
     },
-    "gulp-cond": {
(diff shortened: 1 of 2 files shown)


matthewdavidson:
I checked this out locally and noticed that the iconfont build output was different (there were missing files / unstaged changes).

I just rebased this on top of master with the new pristine check (https://github.com/Skyscanner/backpack/pull/858) which should cause the build to fail .

PR #862, opened 2018-06-21
[BPK-1598] Contribution guide improvements

shaundon:
Diff is going to be quite hard to read, so I recommend just looking at the file directly: https://github.com/Skyscanner/backpack/blob/BPK-1598-contribution-guide/contributing.md

It's been separated into more of a flow. Previously it was very web heavy, with a React Native bit in its own section. With this PR, RN is integrated into the setup instructions throughout.

I'm liberally using `details` elements. May be controversial.

I've also tried to make this a lot more focused on the code side of things. I've removed a lot of the philosophical side (e.g. should `x` be in Backpack) because I feel that's covered well by our internal FAQ. 

The 'adding a new component' part has been largely left alone, as I think it's quite good already, and I'd also like to implement @k0nserv's in a future PR and make a sample/template package for users to copy and reference. I think having that will make a lot of the process around creating a new component self-documenting, so we may be able to remove more of this section then.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -1,34 +1,26 @@
 # Contributing to Backpack
 
-So you want to help us enable Skyscanner to create beautiful, coherent products at scale? That's awesome! ♥
+You want to help us enable Skyscanner to create beautiful, coherent products at scale? That's awesome! :heart:
 
-This document describes how to go about it.
+## Table of contents
 
-<!-- TOC depthFrom:2 depthTo:3 withLinks:1 updateOnSave:1 orderedList:0 -->
+* [Prerequisites](#prerequisites)
+* [Getting started](#getting-started)
+* [Adding a new component](#adding-a-new-component)
+* [How to](#how-to)
 
-- [License](#license)
-- [Before you jump in](#before-you-jump-in)
-- [Getting started](#getting-started)
-- [Development tasks](#development-tasks)
-- [Decisions](#decisions)
-- [Adding a new component](#adding-a-new-component)
-- [Design](#design)
-- [Tokens](#tokens)
-- [Sass mixins](#sass-mixins)
-- [React component](#react-component)
-- [Documentation](#documentation)
-- [React Native](#react-native)
-- [Publishing packages (Backpack squad members only)](#publishing-packages-backpack-squad-members-only)
+## Prerequisites
 
-<!-- /TOC -->
-
-## License
+### Licence
 
 By contributing your code, you agree to license your contribution under the terms of the [APLv2](./LICENSE).
 
-All files are released with the Apache 2.0 license.
+All files are released with the Apache 2.0 licence.
 
-If you are adding a new file it should have a header comment like this:
+If you are adding a new file it should have a header comment containing licence information:
+
+<details>
+<summary>Show/hide licence header</summary>
 
 ```
 Backpack - Skyscanner's Design System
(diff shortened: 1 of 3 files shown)


contributing.md
```diff
@@ -48,69 +40,136 @@ See the License for the specific language governing permissions and
...
+### Decisions
 
-> Is this really going to be used by multiple products?
+Conventions and squad decisions are kept in the [decisions folder](/decisions). We recommend familiarising yourself with these.
 
-If you use it multiple times yourself, it might as well be a component, but maybe Backpack is the wrong place for it and it should instead stay in your own codebase. If in doubt, don’t hesitate to get in touch with us!
+### Installing Node
 
-> I’m sure this should be in Backpack!
+Backpack is developed using Node, using LTS for the Node version and `^5.6.0` for npm.
```
matthewdavidson: Node LTS is 8.11.3 now! Also can we link to https://github.com/Skyscanner/ensure-node-env?

contributing.md
```diff
@@ -48,69 +40,136 @@ See the License for the specific language governing permissions and
...
+
+##### APK signing
+
+For members of Backpack we have a keystore tied to our Google Maps API key in LastPass. Retrieve this key and place it in `native/android/backpack.keystore`. For contributors who are not members of Backpack nothing needs to be done, but Google Maps will not work. If you need Google Maps to work you'll need to supply your own Google Maps Api Key and possible keystore.
+
+</details>
+
+### Build the code
+
+Build the code with `npm run build`.
```
matthewdavidson: I think this is misleading - we are not building the code, we are build run time dependencies for our javascript code. These dependencies are: tokens, svgs, fonts etc.

contributing.md
```diff
@@ -48,69 +40,136 @@ See the License for the specific language governing permissions and
...
+
+#### Web
+
+Run `npm start` to start the storybook server, then go to [http://localhost:9001](http://localhost:9001) in a web browser to view it.
+
+#### React Native
+
+1. Run `npm run native` to start the storybook server.
+2. Open another terminal tab/window.
+3. Run `npm run ios` to install and run the Backpack app on an iPhone simulator.
```
matthewdavidson: > to install and run
There is no installation process here ... just "run" ?

Same with Android below

contributing.md
shaundon: Will do!

contributing.md
shaundon: Yeah maybe `build the code` is an oversimplification.

contributing.md
shaundon: It installs the Backpack app onto the virtual device. But I'll change it - the description doesn't need to be this nuanced, 'run' is fine here.

shaundon:
@matthewdavidson I've addressed your comments.

contributing.md
```diff
@@ -48,69 +40,143 @@ See the License for the specific language governing permissions and
...
+<details>
+<summary>iOS</summary>
 
-To get started run:
+Install XCode from the [App Store](https://itunes.apple.com/gb/app/xcode/id497799835?mt=12). Once installed, open it and accept the licence agreement. You're free to close it after that.
 
-```sh
-npm install
-```
+We use [Cocoapods](https://cocoapods.org) to install some iOS-specific dependencies. Cocoapods uses Ruby, so you'll need to install that too. [rbenv](https://github.com/rbenv/rbenv) and [rvm](https://rvm.io/) are both good ways to get Ruby. The version of Ruby you'll need is specified in `native/ios/.ruby-version`.
```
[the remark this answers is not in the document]
shaundon: Ah yeah - I meant to ask you about bundler, does it need to be manually installed or does it come with Ruby?

shaundon:
https://github.com/Skyscanner/backpack/pull/863 should fix the failing CI.

contributing.md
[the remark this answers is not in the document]
shaundon: Never mind, I figured it out, and I've added a section about bundler.

georgegillams:
Love this new structure
https://media1.giphy.com/media/3o7qE8cQUUCsrRQ1ry/200w.gif?cid=c94812d05b2cb7796a4e694d6bb6d231

PR #869, opened 2018-06-22
[BPK-1596] Fix calendar docs page on narrow devices

georgegillams:
[Edit by Matteo]

I've updated the css slightly to center the calendar on mobile view, it also gave me the change to finally understand how flex works:

Screenshot:
screen shot 2018-06-25 at 10 37 27 am

### DIFF
--- a/packages/bpk-docs/src/pages/CalendarPage/CalendarPage.js
+++ b/packages/bpk-docs/src/pages/CalendarPage/CalendarPage.js
@@ -24,6 +24,8 @@ import BpkCalendar, {
   BpkCalendarDate,
   BpkCalendarGridHeader,
 } from 'bpk-component-calendar';
+import BpkMobileScrollContainer from 'bpk-component-mobile-scroll-container';
+import { cssModules } from 'bpk-react-utils';
 
 import calendarReadme from 'bpk-component-calendar/readme.md';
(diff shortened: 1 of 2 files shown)


native/ios/Podfile.lock
```diff
@@ -87,7 +87,7 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
-  BVLinearGradient: 4b3f3bdf1ac1a52a5ef5e65c9585d793a7c3f825
+  BVLinearGradient: 6d8909e3fc6b089defa4b89d967441fa6827a2ee
```
shaundon: @k0nserv I'm not very well-versed with Cocoapods - is this something we'd expect to see change?

georgegillams: FYI this is done automatically when I run `pod install`

georgegillams: Like, I feel like it doesn't belong in this PR anyway, as the PR has nothing to do with native.

shaundon: Yeah, it's probably not harmful, but I think it should be reverted anyway for the sake of keeping it out this PR.

PR #873, opened 2018-06-25
[No JIRA] Add boilerplate package

shaundon:
The idea here is that we can point users to this package from `contributing.md` and then it's a bit more self documenting. We can also use this ourselves.

We'll have to keep this in line with our conventions etc, but we don't change things that often, and we already have to keep the written-down guide in `contributing.md` up to date. Keeping some code up to date should be easier than keeping documentation up to date, too.

My plan is to do two more PRs after this:

1. The same as this but for React Native.
2. Refactor `contributing.md` a little more to point to these packages.

### DIFF
--- a/packages/bpk-component-boilerplate/index.js
+++ b/packages/bpk-component-boilerplate/index.js
@@ -1,0 +1,25 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* @flow */
+
+import BpkBoilerplate, {
+  type Props as BpkBoilerplateProps,
+} from './src/BpkBoilerplate';
+
+export type { BpkBoilerplateProps };
+export default BpkBoilerplate;
--- a/packages/bpk-component-boilerplate/package.json
+++ b/packages/bpk-component-boilerplate/package.json
@@ -1,0 +1,20 @@
+{
+  "name": "bpk-component-boilerplate",
+  "version": "0.0.0",
+  "description": "Backpack boilerplate component. For reference only, not intended to be used or published.",
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^15.0.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "bpk-mixins": "^17.10.4",
+    "bpk-react-utils": "^2.6.1",
+    "prop-types": "^15.5.8"
+  }
+}
--- a/packages/bpk-component-boilerplate/readme.md
+++ b/packages/bpk-component-boilerplate/readme.md
@@ -1,0 +1,28 @@
+# bpk-component-boilerplate
+
+> Backpack example component.
+
+## Installation
+
+```sh
+npm install bpk-component-boilerplate --save-dev
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkBoilerplate from 'bpk-component-code';
+
+export default () => (
+  <div>
+    <BpkBoilerPlate />
+  </div>
+);
+```
+
+## Props
+
+| Property  | PropType | Required | Default Value |
+| --------- | -------- | -------- | ------------- |
+| className | string   | false    | null          |
--- a/packages/bpk-component-boilerplate/src/BpkBoilerplate-test.js
+++ b/packages/bpk-component-boilerplate/src/BpkBoilerplate-test.js
@@ -1,0 +1,41 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* @flow */
+
+import React from 'react';
+import renderer from 'react-test-renderer';
+import BpkBoilerplate from './BpkBoilerplate';
+
+describe('BpkBoilerplate', () => {
+  it('should render correctly', () => {
+    const tree = renderer.create(<BpkBoilerplate />).toJSON();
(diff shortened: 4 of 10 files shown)


packages/bpk-component-boilerplate/src/BpkBoilerplate.js
```diff
@@ -0,0 +1,64 @@
...
+const getClassName = cssModules(STYLES);
+
+type Props = {
+  title: string,
+  alternate: boolean,
+  className: ?string,
+};
+const BpkBoilerplate = (props: Props) => {
+  const { alternate, className, title, ...rest } = props;
+  const classNames = [getClassName('bpk-boilerplate')];
```
tiagoengel: Let's use the new `className` syntax here. i.e. `getClassName('bpk-boilerplate', alternate && 'bpk-boilerplate--alternate', className)`

packages/bpk-component-boilerplate/src/BpkBoilerplate.js
```diff
@@ -0,0 +1,64 @@
...
+import React from 'react';
+import { cssModules } from 'bpk-react-utils';
+import BpkText from 'bpk-component-text';
+
+import STYLES from './BpkBoilerplate.scss';
+
+const getClassName = cssModules(STYLES);
+
+type Props = {
+  title: string,
```
tiagoengel: Do we really need to have `title` and `alternate` here? It feels like we will be removing this most of the times.

packages/bpk-component-boilerplate/src/BpkBoilerplate.scss
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-boilerplate {
+  padding: $bpk-spacing-md;
```
tiagoengel: I think this can be removed too, maybe just add `display: flex` so it won't complain about empty declarations. The rest of the styles seem like something we will end up removing for most new components

packages/bpk-component-boilerplate/readme.md
```diff
@@ -0,0 +1,31 @@
...
+    <BpkBoilerPlate title="Hello world" alternate />
+  </div>
+);
+```
+
+## Props
+
+| Property  | PropType | Required | Default Value |
+| --------- | -------- | -------- | ------------- |
+| title     | Node     | true     | -             |
```
tiagoengel: I can see the values of having those props to show how to document them, but I think it'd be better to just remove them (except className)  and leave a TODO here. e.g. `TODO: document propTypes - required props first in alphabetical order then optional`

packages/bpk-component-boilerplate/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "type": "git",
+    "url": "git@github.com:Skyscanner/backpack.git"
+  },
+  "author": "Backpack Design System <backpack@skyscanner.net>",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": "^15.0.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "bpk-component-text": "^1.0.64",
```
tiagoengel: I can see myself copying this boilerplate and forgetting to remove this 😄 . I think it's better we avoid adding other bpk components in the boilerplate.

packages/bpk-component-boilerplate/index.js
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* @flow */
+
+import BpkBoilerplate from './src/BpkBoilerplate';
+
+export default BpkBoilerplate;
```
tiagoengel: We need to export the props definition (flow) here too.

shaundon: Are you sure? I thought we only exported Flow types that are likely to be used by consumers, e.g. if there was a `ButtonType` type.

packages/bpk-component-boilerplate/src/BpkBoilerplate.scss
shaundon: Bear in mind that this package serves two purposes:

1) To be useful for us.
2) To show contributors our conventions and how we structure code.

In this case I figured this would be a good way to show contributors how to use our mixins and follow BEM structure.

packages/bpk-component-boilerplate/src/BpkBoilerplate.js
shaundon: Bear in mind that this package serves two purposes:

1) To be useful for us.
2) To show contributors our conventions and how we structure code.

In this case I figured this would be a good way to show contributors how we use Flow types and use props. Maybe that's too much hand-holding though, what do you think?

packages/bpk-component-boilerplate/index.js
tiagoengel: I think so, props are public so types should be visible. If consumers want to extend our components and we don't export the types they would have to redefine all props.

packages/bpk-component-boilerplate/src/BpkBoilerplate.scss
shaundon: Having had more time to think about this, I think you're right and I should strip this back. If prospective contributors want to see examples of how we do things, they can easily find it from looking at other packages.

packages/bpk-component-boilerplate/src/BpkBoilerplate.js
tiagoengel: I agree with that but I think className is enough to show it. In any case I don't have a strong opinion about this so if you think we should keep it that's fine.

packages/bpk-component-boilerplate/src/BpkBoilerplate-test.js
```diff
@@ -0,0 +1,43 @@
...
+import renderer from 'react-test-renderer';
+import BpkBoilerplate from './BpkBoilerplate';
+
+describe('BpkBoilerplate', () => {
+  it('should render correctly', () => {
+    const tree = renderer.create(<BpkBoilerplate />).toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('should support custom class names', () => {
```
tiagoengel: We need to remove this and update the snapshots. Otherwise LGTM

shaundon: We want the boilerplate example to support custom class names though, don't we?

tiagoengel: Sorry I meant the `title` prop

shaundon: Ahh yeah! Now fixed 👍

packages/bpk-component-boilerplate/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "bpk-component-boilerplate",
+  "version": "0.0.0",
+  "description": "Backpack boilerplate component. For reference only, not intended to be used or published.",
```
matthewdavidson: > For reference only, not intended to be used or published.

I bet that this copy gets accidentally published in a new component 😏

packages/bpk-component-boilerplate/readme.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkBoilerplate from 'bpk-component-code';
+
+export default () => (
+  <div>
```
matthewdavidson: Why does this needs a wrapping `<div />`

shaundon: Cos I'm so extra

(I'll remove it)

packages/bpk-component-boilerplate/package.json
shaundon: This can be removed once my super new CLI is up and running.

PR #877, opened 2018-06-26
[BPK-1685] Fix calendar transition for custom font size

tiagoengel:
Tested on Safari, FF, Chrome and IE11.

### DIFF
--- a/packages/bpk-component-calendar/src/utils.js
+++ b/packages/bpk-component-calendar/src/utils.js
@@ -21,21 +21,35 @@ import {
   calendarDaySpacing,
 } from 'bpk-tokens/tokens/base.es6';
 
-const remToPx = value => {
-  let parsed = null;
+const CSS_UNIT_REGEX = /(^[+-]?(?:\d*\.)?\d+)(.+)/i;
 
-  if (/rem$/.test(value)) {
-    parsed = parseFloat(value.replace(/rem/, '')) * 16;
+const splitToken = value => {
+  const match = value.match(CSS_UNIT_REGEX);
+  if (!match) {
+    throw new Error(
+      `Invalid token value. Expecting a valid css unit, got ${value}`,
+    );
   }
-
-  return parsed || null;
+  const [_, val, unit] = match; // eslint-disable-line no-unused-vars
+  return [parseFloat(val), unit];
 };
 
-export const getCalendarGridWidth = () =>
-  7 * (remToPx(calendarDaySize) + remToPx(calendarDaySpacing));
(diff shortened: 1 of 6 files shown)


packages/bpk-component-calendar/src/utils.js
```diff
@@ -21,21 +21,32 @@ import {
...
+    throw new Error('Invalid token');
+  }
+  const [_, val, unit] = match; // eslint-disable-line no-unused-vars
+  return [parseFloat(val), unit];
+};
 
-  if (/rem$/.test(value)) {
-    parsed = parseFloat(value.replace(/rem/, '')) * 16;
+export const getCalendarGridWidth = (multiplier = 1) => {
+  const [sizeValue, sizeUnit] = splitToken(calendarDaySize);
```
tiagoengel: I'm wondering if we should move this outside the function as this is a static value and will never change. It would also fail as soon as the JS file loads if the tokens are not correct.

packages/bpk-component-calendar/src/utils.js
```diff
@@ -21,21 +21,33 @@ import {
...
+  const [spacingValue, spacingUnit] = splitToken(calendarDaySpacing);
+
+  if (sizeUnit !== spacingUnit) {
+    throw new Error(
+      `'calendarDaySize' and 'calendarDaySpacing' must use the same unit. Got ${sizeUnit} and ${spacingUnit}`,
+    );
   }
 
-  return parsed || null;
+  const width = multiplier * (7 * (sizeValue + spacingValue));
```
tiagoengel: Using `calc` would be ideal but it's not working on IE11 (even though it says it does). I reckon the problem is `translateX` plus `calc`.

packages/bpk-component-calendar/src/utils.js
```diff
@@ -21,21 +21,33 @@ import {
   calendarDaySpacing,
 } from 'bpk-tokens/tokens/base.es6';
 
-const remToPx = value => {
-  let parsed = null;
+const splitToken = value => {
+  const match = value.match(/(^[+-]?(?:\d*\.)?\d+)(.+)/i);
```
k0nserv: This regex should be pulled out of this function to avoid it being recreated at all times

tiagoengel: good point, I'll do

tiagoengel: done

tiagoengel:
I tested on all supported browsers except mobile Safari @shaundon. Do you mind doing that?

PR #879, opened 2018-06-26
[No JIRA] Disable font scaling for native icons

k0nserv:
When font scaling is enabled for `BpkIcon` it breaks icons by increasing their size when accessibility text sizes are used. Arguably this is a good thing for most text, but icons probably shouldn't scale


k0nserv:
Closing this for now

PR #880, opened 2018-06-27
[BPK-1688] Add native chip components

shaundon:
**To do**

* [] Readme
* [x] Tests
* [] Implementation
* [] Stories
* [] Changelog
* [x] Flow


native/packages/react-native-bpk-component-chip/src/BpkChip.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  if (style) {
+    wrapperStyle.push(style);
+  }
+
+  const touchableProps: { [string]: any } = {
+    accessibilityLabel,
+  };
+  let Touchable;
```
k0nserv: Can we use `Platform.select` and a `const` value here instead of this method?

native/packages/react-native-bpk-component-chip/src/BpkChipDismissButton.js
```diff
@@ -0,0 +1,52 @@
...
+
+import React from 'react';
+import { Platform } from 'react-native';
+import BpkIcon from 'react-native-bpk-component-icon';
+import BpkTouchableNativeFeedback from 'react-native-bpk-component-touchable-native-feedback';
+import BpkTouchableOverlay from 'react-native-bpk-component-touchable-overlay';
+
+const BpkChipDismissButton = ({ ...rest }: { [string]: any }) => {
+  const Touchable =
+    Platform.OS === 'android'
```
k0nserv: Can we use `Platform.select` here?

native/packages/react-native-bpk-component-chip/src/BpkChipDismissable.js
```diff
@@ -0,0 +1,101 @@
...
+
+type Props = {
+  dismissButtonLabel: string,
+  label: string,
+  large: boolean,
+  onDismiss: ?() => mixed,
+  style: ?StyleObj,
+};
+
+const BpkChipDismissable = (props: Props) => {
```
k0nserv: Should this be `BpkDismissableChip`? That sounds a lot better to me. Do we have a naming convention that dictates this?

native/packages/react-native-bpk-component-chip/stories.js
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+/* @flow */
+
+import React, { Fragment } from 'react';
+import { StyleSheet, View } from 'react-native';
+import PropTypes from 'prop-types';
+import { storiesOf } from '@storybook/react-native';
+import { action } from '@storybook/addon-actions';
+import BpkText from 'react-native-bpk-component-text';
```
k0nserv: It'd be nice if these we split a bit between third party packages, other Backpack packages(and storybook), and code import

native/packages/react-native-bpk-component-chip/src/BpkChipDismissable.js
shaundon: It certainly does sound better that way. I'll see if we have a convention.

native/packages/react-native-bpk-component-chip/src/BpkChip.js
k0nserv: Yeah there probably is a tiny one, but thinking about that is premature optimization. You can write it with a single `Platform.select` as

```javascript
const { Touchable, touchableProps } = Platform.select({
  android: {
    Touchable: BpkTouchableNativeFeedback,
    touchableProps: {
      accessibilityLabel,
      accessibilityComponentType: "button",
      color: "red"
    }
  },
  ios: {
    Touchable: BpkTouchableOverlay,
    touchableProps: {
      accessibilityLabel,
      accessibilityTraits: selected ? ["button", "selected"] : ["button"],
      borderRadius: "sm"
    }
  }
});

```

native/packages/react-native-bpk-component-button/readme.md
```diff
@@ -23,7 +23,7 @@ pod 'BVLinearGradient', :path => '../node_modules/react-native-bpk-component-but
 ## Usage
 
 ```js
-import { View } from 'react-native';
+import { StyleSheet, View } from 'react-native';
```
shaundon: Unrelated to this PR, but I noticed it.

native/packages/react-native-bpk-component-chip/readme.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+## Props
+
+### BpkChip
+
+| Property              | PropType                                                                  | Required | Default Value |
+| --------------------- | ------------------------------------------------------------------------- | -------- | ------------- |
+| accessibilityLabel    | string                                                                    | true     | -             |
+| onPress               | func                                                                      | true     | -             |
```
matthewdavidson: Why is `onPress` required for a normal chip?

native/packages/react-native-bpk-component-chip/src/BpkChip.js
```diff
@@ -0,0 +1,144 @@
...
+    alignItems: 'center',
+    flexDirection: 'row',
+    height: spacingSm * 5,
+    justifyContent: 'space-between',
+    paddingHorizontal: spacingSm,
+  },
+  innerLarge: {
+    height: spacingLg + spacingSm,
+  },
+  wrapperSelected: {
```
matthewdavidson: Pop this style up beside its `wrapper` counterpart :)

native/packages/react-native-bpk-component-chip/src/BpkChip.js
```diff
@@ -0,0 +1,144 @@
...
+  return (
+    <View style={wrapperStyle}>
+      <Touchable {...touchableProps} {...rest}>
+        <View style={innerStyle}>
+          <BpkText textStyle="xs" style={textStyle}>
+            {label}
+          </BpkText>
+        </View>
+      </Touchable>
+    </View>
```
matthewdavidson: 2 `<View />`'s (3 for ios) and a `<Text />` just for a chip? Seems excessive.

native/packages/react-native-bpk-component-chip/src/BpkChipDismissButton.js
```diff
@@ -0,0 +1,70 @@
...
+      touchableProps: {
+        accessibilityTraits: ['button'],
+        borderRadius: 'sm',
+      },
+    },
+  });
+
+  return (
+    <Touchable {...touchableProps} {...rest}>
+      <View style={style}>
```
matthewdavidson: This inner view is only required for android right?

native/packages/react-native-bpk-component-chip/src/BpkDismissibleChip.js
```diff
@@ -0,0 +1,104 @@
...
+    backgroundColor: colorGray100,
+    borderRadius: borderRadiusSm,
+  },
+  wrapperLarge: {
+    height: spacingLg + spacingSm,
+  },
+  text: {
+    color: colorGray700,
+  },
+  dismissBtn: {
```
matthewdavidson: Whats wrong with `dismissButton` 😇 ?

native/packages/react-native-bpk-component-chip/readme.md
shaundon: If they don't supply `onPress`, they've just got a `BpkBadge` then, so I thought BpkChip should be opinionated about this.

native/packages/react-native-bpk-component-chip/src/BpkChip.js
shaundon: Yeah let me see if I can get it down.

native/packages/react-native-bpk-component-chip/readme.md
matthewdavidson: Gotcha :)

native/packages/react-native-bpk-component-chip/src/BpkChipDismissButton.js
shaundon: Yeah correct (that's why I was having issues with the ripple not firing the other day too, I hadn't added an inner view!).

I was avoiding splitting the components into `.android` and `.ios` but if it'll save a view it's probably the way to go.

native/packages/react-native-bpk-component-chip/src/BpkDismissibleChip.js
shaundon: Nowt, I'll change it 👍

native/packages/react-native-bpk-component-chip/src/BpkChipDismissButton.js
matthewdavidson: Yeah TBH the android touchable component always causes developer confusion, that decision to not include a wrapper view inside will plague us 👎

shaundon: Maybe a quick note in the readme will alleviate some of it.

native/packages/react-native-bpk-component-chip/src/BpkChip.js
```diff
@@ -0,0 +1,144 @@
...
+  },
+  inner: {
+    alignItems: 'center',
+    flexDirection: 'row',
+    height: spacingSm * 5,
+    justifyContent: 'space-between',
+    paddingHorizontal: spacingSm,
+  },
+  innerLarge: {
+    height: spacingLg + spacingSm,
```
k0nserv: What are our thoughts with regards to Dynamic Type here?

native/packages/react-native-bpk-component-chip/src/BpkChipDismissButton.js
```diff
@@ -0,0 +1,70 @@
...
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React from 'react';
+import { Platform, type StyleObj, View, ViewPropTypes } from 'react-native';
+import PropTypes from 'prop-types';
+
+import BpkIcon from 'react-native-bpk-component-icon';
```
k0nserv: Nice grouping of imports 👍

native/packages/react-native-bpk-component-chip/src/BpkChipDismissButton.js
```diff
@@ -0,0 +1,70 @@
...
+        accessibilityTraits: ['button'],
+        borderRadius: 'sm',
+      },
+    },
+  });
+
+  return (
+    <Touchable {...touchableProps} {...rest}>
+      <View style={style}>
+        <BpkIcon icon="close" small />
```
k0nserv: What's the tap area of this component? Is it sufficiently large?

native/packages/react-native-bpk-component-chip/stories.js
```diff
@@ -0,0 +1,331 @@
...
+      flights: false,
+      hotels: true,
+      carHire: false,
+      trains: false,
+    };
+  }
+
+  toggleChip = chip => {
+    action(`Toggling ${chip}`);
+    this.setState({
```
k0nserv: This should use the callback version since it relies on previous state

native/packages/react-native-bpk-component-chip/src/BpkChipDismissButton.js
shaundon: It's not large enough, we're gonna change it :)

native/packages/react-native-bpk-component-chip/src/BpkChip.js
```diff
@@ -0,0 +1,55 @@
...
+type Props = {
+  ...$Exact<CommonProps>,
+  large: boolean,
+  selected: boolean,
+};
+
+const BpkChip = (props: Props) => {
+  const { ...rest } = props;
+
+  return <BpkChipWrapper {...rest} dismissible={false} />;
```
k0nserv: This could just be a function call instead of a render which reduces the virtual DOM tree by one level

```javascript
  return BpkChipWrapper({ dismissible: false, ...rest });
```

k0nserv:
With large accessibility sizes on iOS. Notice how the close icon scales too

image

shaundon:
Is that good? I feel like it looks ok 😅

k0nserv:
Yeah I'm not sure how I feel about it

shaundon:
Maybe I'm being ignorant but I'm seeing it as a good thing. If a user wants to have larger text, it stands to reason that the icon should increase in size accordingly too. What would the downsides of this approach be?

matthewdavidson:
> Maybe I'm being ignorant but I'm seeing it as a good thing. If a user wants to have larger text, it stands to reason that the icon should increase in size accordingly too. What would the downsides of this approach be?

Agreed it should scale.

native/packages/react-native-bpk-component-chip/src/BpkChipWrapper.js
```diff
@@ -0,0 +1,135 @@
...
+  },
+  text: {
+    color: colorGray700,
+  },
+  textSelected: {
+    color: colorWhite,
+  },
+  icon: {
+    color: colorGray500,
+    marginLeft: spacingSm,
```
k0nserv: This should be `marginStart` for RTL support

native/packages/react-native-bpk-component-chip/stories.js
```diff
@@ -0,0 +1,272 @@
...
+
+import CenterDecorator from '../../storybook/CenterDecorator';
+import BpkChip, { BpkDismissibleChip } from './index';
+
+const styles = StyleSheet.create({
+  bottomMargin: {
+    marginBottom: spacingMd,
+  },
+  chip: {
+    marginRight: spacingSm,
```
k0nserv: `marginEnd` for RTL

native/packages/react-native-bpk-component-chip/src/BpkChip.js
matthewdavidson: Note that prop types are ignored in this case... not sure if it's worth it.

k0nserv: Only the `BpkChipWrapper` prop types are ignored right? They are internal anyway?

k0nserv: Strictly speaking we don't need `propTypes` for `BpkChipWrapper` since we have Flow and it's internal?

matthewdavidson: Regardless if it's internal or not, the same props are declared for this component anyways so ...

I still think this stands though:
> Also I think it's kind of surprising for devs who haven't came across squashing before.

shaundon: I do like the idea of a smaller DOM tree but I wouldn't want to prematurely optimise at the expense of principle of least surprise (hey I'm a poet)

k0nserv: Okay let's not do that then :)

shaundon:
I've addressed all the comments 🙌

native/packages/react-native-bpk-component-chip/index.js
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import BpkChip from './src/BpkChip';
+import BpkDismissibleChip from './src/BpkDismissibleChip';
+
+export default BpkChip;
+export { BpkDismissibleChip };
```
matteo-hertel: Do we need to export the type for the props?

shaundon: I'm not sure where that convention came from 🤔 is that something we always do? My understanding was that we exported things like `buttonTypes`, as I can see when consumers might need those, but I'm not sure when they'd need these.

shaundon: It's easy for me to do so I'm happy to do it, I just wanna understand what our convention is.

matteo-hertel: that's why I've asked, not 100% sure 😅

PR #881, opened 2018-06-28
[No JIRA] Fix BpkImage onload when server-side rendering

dominikkrejcik:
When BpkImage is used with server-side rendering, there was a race
condition where if the underlying img element loaded before the React
bundle executed and attached the onload listener, the image would appear
to be loading forever.

Fix it by checking whether the image is already loaded when the
component mounts and triggering the callback if it is.


packages/bpk-component-image/src/BpkImage.js
```diff
@@ -42,39 +42,62 @@ type BpkImageProps = {
...
+    if (hidden) {
+      imgClassNames.push(getClassName('bpk-image__img--hidden'));
+    }
+
+    return (
+      <img
+        className={imgClassNames.join(' ')}
+        alt={altText}
+        onLoad={onImageLoad}
+        ref={el => this.setImgRef(el)}
```
dominikkrejcik: One thing I wanted to check, do do we need to do `this.setImageRef = this.setImageRef.bind(this)` for this to work? I had that in my original code I tested in our project but Flow was complaining here so I took it out.

k0nserv: No, but you could change it to `ref={this.setImgRef}` in which case it would need to be bound. We don't explicitly bind instead preferring 

```javascript
setImageRef = (el) => {
  this.img = el;
}
```

Which achieves the same thing

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -42,39 +42,62 @@ type BpkImageProps = {
...
+        {...rest}
+      />
+    );
+  }
+}
 
 Image.propTypes = {
```
k0nserv: Can we move this inside the class too?

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -42,39 +42,62 @@ type BpkImageProps = {
...
+  img: ?HTMLImageElement;
 
-  const imgClassNames = [getClassName('bpk-image__img')];
+  static defaultProps = {
+    hidden: false,
+  };
 
-  if (hidden) {
-    imgClassNames.push(getClassName('bpk-image__img--hidden'));
+  constructor() {
```
k0nserv: Please explicitly take and pass the props here

packages/bpk-component-image/src/BpkImage.js
sogame: Just out of curiosity, why you prefer this approach?

In the team we took at look at pros/cons between arrow function and explicitly bind and didn't get much 😅 
Performance wise seems to be more or less the same, arrow function is less code but on the other side not all functions in the class can be created with arrow functions, leading to an inconsistent file... so we're still undecided 😂

unreleased.md
```diff
@@ -1 +1,4 @@
 # Unreleased
+
+- bpk-component-image
```
matthewdavidson: NIT: this should be:
```
**Fixed:**
- bpk-component-image:
+  - Fixed `onload` race condition when using server-side rendering.
```

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -42,39 +42,62 @@ type BpkImageProps = {
...
-const Image = (props: ImageProps): Node => {
-  const { hidden, altText, onImageLoad, ...rest } = props;
+class Image extends Component<ImageProps> {
+  img: ?HTMLImageElement;
 
-  const imgClassNames = [getClassName('bpk-image__img')];
+  static propTypes = {
+    altText: PropTypes.string.isRequired,
+    hidden: PropTypes.bool,
+    onImageLoad: PropTypes.func.isRequired,
```
matthewdavidson: `onImageLoad` is required in props but optional in flow types .. which is it?

dominikkrejcik: I think it was like this before, I only moved props and flow types around. Made it required now.

packages/bpk-component-image/src/BpkImage.js
arnaugm: Arrow functions don't go into the prototype, which means that you create them every time the component is instantiated. That could have a performance penalty, but in general is negligible.
They can also be problematic for testing but this depends on your testing approach. In general use or not arrow functions here is a matter of taste. If you use the object method approach and you need to use it as a callback (and only in this case) you need to bind the this.

k0nserv: We think it's nicer than having a bunch of `binds` in the constructor. I'm not sure about the performance between creating arrow functions in render, using instance properties with arrow functions or calling `bind` in the constructor. In any case it's probably a matter of premature optimization to worry too much about it

PR #882, opened 2018-06-29
[BPK-1680]-initial-detox-integration

matteo-hertel:
This adds detox to both Android and iOS, to provide some visible value I've copied over the switch component testing from the work we did as part of the designer offsite https://github.com/Skyscanner/backpack/tree/story-control

The main change is on the Android side, we used to have a mixture of two namespaces `com.backpack` and `com.native` now is consistent `com.backapck` and some folder have been renamed, in case something goes wrong during the build open android studio and run a Gradle sync 

One thing that's worth a bit of discussion is the folder structure:
Ideally, I wanted the e2e and detox config in the native folder and run the commands from the root with subshell, however, in the native package.json we don't have `jest` as direct dependency and detox requires it, there are two solutions: 

1. we move the config to the native folder and we add jest as a dependency,
2. we leave as it is

For testing this, please follow the instructions in the detox repo
https://github.com/wix/detox/blob/master/docs/Introduction.GettingStarted.md#step-1-install-dependencies
you should be all sorted with the current setup, the only bit missing could be `applesimutils` so run
```
brew tap wix/brew
brew install applesimutils
```
you then need to have the native storybook running (`npm run native`) with the 'all type' switch story selected [http://localhost:7007/?selectedKind=react-native-bpk-component-switch&selectedStory=All%20types&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel](http://localhost:7007/?selectedKind=react-native-bpk-component-switch&selectedStory=All%20types&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)
and you're good to go

then you can test Android
```
npm run detox:build:android
npm run detox:test:android
```

replace `android` with `ios` to test iOS

> Note for iOS, the simulator used will be an iPhone8, based on your OS you might get iOS 11.1 or 11.4, this will be consistent between `npm run ios` and `npm run detox:test:ios`


native/android/build.gradle
```diff
@@ -22,6 +22,11 @@ allprojects {
             // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
             url "$rootDir/../node_modules/react-native/android"
         }
+        buildscript {
```
amolgupta: this might not be required.

matteo-hertel: is part of the docs https://github.com/wix/detox/blob/master/docs/Introduction.Android.md#setup
so I'll leave it there for now  when it'll be all working I might check if I can remove it, thanks for pointing it out tough 👍

e2e/native/BpkSwitch.spec.js
```diff
@@ -0,0 +1,49 @@
+/*
```
matthewdavidson: I think the detox specs should live with the component

k0nserv: Yeah that seems reasonable imo

package.json
```diff
@@ -189,5 +195,24 @@
...
     ]
+  },
+  "detox": {
+    "test-runner": "jest",
+    "runner-config": "e2e/native/config.json",
+    "specs": "e2e",
+    "configurations": {
+      "ios.sim.debug": {
+        "binaryPath": "native/ios/build/Build/Products/Debug-iphonesimulator/native.app",
+        "build": "xcodebuild -project native/ios/native.xcodeproj -scheme native -configuration Debug -sdk iphonesimulator -derivedDataPath native/ios/build",
```
k0nserv: This works? Strange if so because it doesn't include the pods. Should be referencing the workspace not the project

matteo-hertel: we can reference the workspace with the --workspace flag, but it works, given we have to install gems and pod to have `npm run ios` working I think we can safely leave it like this

matteo-hertel:
I've update the PR desc with the missing info, I'll update the contributor.md now

matteo-hertel:
I've update the PR moving the e2e test to the component folder and named it the same way we name the others, I've also updated the contrib.md [here](https://github.com/Skyscanner/backpack/blob/aee87e596ec9367e328af89b6803f35242ffd557/contributing.md) the rendered version

package.json
k0nserv: I'd be more comfortable if we used the workspace reference I feel like it working is a fluke

matteo-hertel: It did work on Matt's machine freshly updated to high sierra and no detox installed at all, it's super easy to update tough so I'll update it 😄

contributing.md
```diff
@@ -112,6 +112,19 @@ $ANDROID_SDK_ROOT/tools/bin/avdmanager create avd --name "bpk-avd-min" --package
...
+
+Backpack uses [Detox](https://github.com/wix/detox) for e2e testing of React Native components, it requires `applesimutils` to be installed, this can be added by running
+
+```
+brew tap wix/brew
+brew install applesimutils
+```
+
+for any issue with Detox please refer the the [offical documentation](https://github.com/wix/detox/tree/master/docs).
+
```
shaundon: I would put this in the above 'React Native' section as a sub-section.

So you'd have:

```
### React Native

> Skip this section if you don't intend to develop React Native components.

* iOS 
* Android

#### Detox

...etc

```

package.json
```diff
@@ -34,6 +34,11 @@
     "build": "npm run build:tokens && npm run build:svgs && lerna run build && (cd ./native && npm run link)",
     "storybook:dist": "build-storybook -c .storybook -o dist/storybook",
     "storybook": "start-storybook -p 9001",
+    "detox": "detox",
+    "detox:build:android": "detox build --configuration android.emu.debug",
+    "detox:build:ios": "detox build --configuration ios.sim.debug",
+    "detox:test:android": "detox test --configuration android.emu.debug --reuse",
```
matthewdavidson: We need to remove the `--reuse` flag remember

contributing.md
matteo-hertel: Done 👍

package.json
matteo-hertel: Done 👍

PR #883, opened 2018-06-29
[BPK-1688] Docs for native chip components

shaundon:


packages/bpk-docs/src/pages/NativeChipsPage/NativeChipsPage.js
```diff
@@ -0,0 +1,87 @@
...
+        src: `/${iosScreenshotDefault}`,
+        altText: 'iOS Default Chip Component',
+        subText: '(iPhone 8 simulator)',
+      },
+      {
+        title: 'Android',
+        width: 1080,
+        height: 1920,
+        src: `/${androidScreenshotDefault}`,
+        altText: 'Android Default Spinner Component',
```
k0nserv: This alt text is not correct

willhg:
Yep all looks good :)

packages/bpk-docs/src/pages/NativeChipsPage/NativeChipsPage.js
shaundon: Good catch!

PR #889, opened 2018-07-02
[NoJira] testing parallel builds

matteo-hertel:
testin'


.travis.yml
```diff
@@ -5,24 +5,48 @@ env:
...
+
+jobs:
+  include:
+    - stage: install dependencies
+      script:
+        - npm install -g greenkeeper-lockfile@1
+        - greenkeeper-lockfile-update
+        - npm install
+        - npm run build
+        - greenkeeper-lockfile-upload
```
matthewdavidson: Why isn't `npm install` etc run in the install phase? i.e. where you currently have `echo 'Backpacking, hang on!'`

.travis.yml
```diff
@@ -5,24 +5,48 @@ env:
...
+      script:
+        - npm run bootstrap
+        - npm run jest
+
+    - stage: shared cache
+      script:
+        - npm run bootstrap
+        - npm run test:native
+
+    - state: shared cache
```
matthewdavidson: why is this `state` as opposed to `stage`?

.travis.yml
```diff
@@ -5,24 +5,48 @@ env:
...
+      script:
+        - npm install -g greenkeeper-lockfile@1
+        - greenkeeper-lockfile-update
+        - npm install
+        - npm run build
+        - greenkeeper-lockfile-upload
+
+    - stage: shared cache
+      script:
+        - npm run bootstrap
```
matthewdavidson: `npm run bootstrap` is happening 4 times here ... that cant be right?

.travis.yml
```diff
@@ -31,7 +55,6 @@ branches:
 
 deploy:
   provider: pages
-  skip_cleanup: true
```
matthewdavidson: Why remove this?

.travis.yml
shaundon: This was mentioned as an issue in the PR description:

> The only problem I have so far is I have to run npm run bootstrap in every step to have lerna happy and I can't test a merge onto master so that might break

.travis.yml
matteo-hertel: oh right, I ran the travis linter and  the `skip_cleanup` is added twice in the travis file

https://github.com/Skyscanner/backpack/blob/master/.travis.yml#L34
https://github.com/Skyscanner/backpack/blob/master/.travis.yml#L39

I'm removing one, I guess it's kind of hard to spot that in the diff github shows, good shout 👍

.travis.yml
matteo-hertel: From the test I've executed, if there isn't a global install it'll default to `npm install` and the install phase will be executed for every single job, so just echoing in the install phase will not run the default `npm install` and will be executed every single job

.travis.yml
matteo-hertel: 😱 typo! I wonder how's that passing 🤔

.travis.yml
matteo-hertel: just to double check I'm work, this is a build without the bootstrap step
https://travis-ci.org/Skyscanner/backpack/jobs/400023027
lerna is not happy 😄

matteo-hertel: bootstrap runs as post install hook so it does run once, however it creates symlinks and I don't think those are preserved between steps

PR #891, opened 2018-07-03
[No JIRA] Add CLI for creating components

shaundon:
Now we have the boilerplate component, here's a basic CLI for creating new components.

It does the following:
* Makes a copy of `bpk-component-boilerplate` or `react-native-bpk-component-boilerplate` with the new component name.
* Renames the files within appropriately.
* Replaces instances of `boilerplate` within files.
* Adds it to the storybook config.
* Lints the new folder and the storybook config with `--fix` to fix any formatting errors and run Prettier.
* Tells the user how to run tests and storybook.

You run it like this:

screen shot 2018-07-03 at 20 27 15

The cool thing about this is that not only is it useful for us, it also means we can remove some of `contributing.md` because it's self evident by using this CLI (for example, the part where we tell them how to add to the storybook config).

My Bash skillz aren't that great so if you see anywhere this can be improved, please feel free.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -73,8 +73,8 @@ configure(() => {
   require('./../packages/bpk-component-drawer/stories');
   require('./../packages/bpk-component-fieldset/stories');
   require('./../packages/bpk-component-form-validation/stories');
-  require('./../packages/bpk-component-grid/stories');
   require('./../packages/bpk-component-grid-toggle/stories');
+  require('./../packages/bpk-component-grid/stories');
   require('./../packages/bpk-component-heading/stories');
   require('./../packages/bpk-component-horizontal-nav/stories');
   require('./../packages/bpk-component-icon/stories');
--- a/contributing.md
+++ b/contributing.md
@@ -203,7 +203,6 @@ If you want to add a new component, we will need the following:
 - Tests
 - Documentation (Including main `readme.md`)
 
-
 ### Design
 
 Sketch is the preferred format for non-technical folks. We’d appreciate if you could provide an exact match of your component in Sketch format together with folders for each state e.g. disabled, expanded etc.
@@ -224,21 +223,9 @@ If you add a new file of mixins, for example for a new *atom*, make sure you add
 
 ### React component
 
-Each React component has its own subdirectory `packages/bpk-component-<component-name>`, where the npm module is stored. For it to be properly published, create a `package.json` starting at version `0.0.0` and take inspiration from existing packages in terms of structure. We use [CSS Modules](https://github.com/css-modules/css-modules) along with [BEM](http://getbem.com/) to prevent collisions and accidental overwrites in CSS.
+Use `npm run create-component` to create a new skeleton React component. Once this is created, use existing components for code style inspiration.
 
-#### Storybook
-
-Create a `stories.js` file and use it to list all the different possible states of your component. Check out Storybook's documentation for more details, and also check out the `stories.js` files of existing components.
-
-> Make sure you import your new component's stories into Storybook by adding it to `.storybook/config.js`.
-
-#### Testing
-
-React components need to be properly unit tested. We use [jest](https://facebook.github.io/jest/)'s snapshot testing, but depending on the complexity of your component, you may want to add additional tests.
-
-#### Readme
-
-Create a `readme.md` file that shows how to install and use your component. It should also include a table of the component's props, PropTypes and their default values. Take inspiration from existing components.
+We use [CSS Modules](https://github.com/css-modules/css-modules) along with [BEM](http://getbem.com/) to prevent collisions and accidental overwrites in CSS.
 
 ### Documentation
 
--- a/native/storybook/storybook.js
+++ b/native/storybook/storybook.js
@@ -53,8 +53,8 @@ const hideWarnings = () => {
 configure(() => {
   require('../packages/react-native-bpk-component-alert/stories');
   require('../packages/react-native-bpk-component-animate-height/stories');
-  require('../packages/react-native-bpk-component-banner-alert/stories');
   require('../packages/react-native-bpk-component-badge/stories');
+  require('../packages/react-native-bpk-component-banner-alert/stories');
   require('../packages/react-native-bpk-component-button-link/stories');
   require('../packages/react-native-bpk-component-button/stories');
   require('../packages/react-native-bpk-component-card/stories');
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -20,6 +20,7 @@
     "clean": "npm run clean:dist && npm run clean:lerna && npm run clean:node_modules && npm run clean:node_modules_native",
     "lint:scss": "stylelint 'packages/**/*.scss' --syntax scss",
     "lint:js": "eslint . .storybook --ext .js,.jsx",
+    "lint:simple:fix": "eslint --fix",
```
georgegillams: Surely there's a way to run this from CLI without this witchery!

.storybook/config.js
```diff
@@ -73,8 +73,8 @@ configure(() => {
   require('./../packages/bpk-component-drawer/stories');
   require('./../packages/bpk-component-fieldset/stories');
   require('./../packages/bpk-component-form-validation/stories');
-  require('./../packages/bpk-component-grid/stories');
   require('./../packages/bpk-component-grid-toggle/stories');
+  require('./../packages/bpk-component-grid/stories');
```
shaundon: This file was reordered slightly by the sort function implemented in `create-component.js`. We think this is probably fine.

native/storybook/storybook.js
```diff
@@ -76,12 +76,12 @@ configure(() => {
...
   require('../packages/react-native-bpk-component-switch/stories');
-  require('../packages/react-native-bpk-component-text/stories');
   require('../packages/react-native-bpk-component-text-input/stories');
-  require('../packages/react-native-bpk-component-touchable-overlay/stories');
+  require('../packages/react-native-bpk-component-text/stories');
   require('../packages/react-native-bpk-component-touchable-native-feedback/stories');
-  require('../packages/react-native-bpk-theming/stories');
+  require('../packages/react-native-bpk-component-touchable-overlay/stories');
   require('../packages/react-native-bpk-styles/stories');
+  require('../packages/react-native-bpk-theming/stories');
```
shaundon: This file was reordered slightly by the sort function implemented in create-component.js. We think this is probably fine.

contributing.md
```diff
@@ -224,21 +223,9 @@ If you add a new file of mixins, for example for a new *atom*, make sure you add
...
-Create a `stories.js` file and use it to list all the different possible states of your component. Check out Storybook's documentation for more details, and also check out the `stories.js` files of existing components.
-
-> Make sure you import your new component's stories into Storybook by adding it to `.storybook/config.js`.
-
-#### Testing
-
-React components need to be properly unit tested. We use [jest](https://facebook.github.io/jest/)'s snapshot testing, but depending on the complexity of your component, you may want to add additional tests.
-
-#### Readme
+Use `npm run new-component` to create a new skeleton React component. Once this is created, use existing components for code style inspiration.
```
georgegillams: Shall we rename to `create-component`. It seems more logical to name it using a verb rather than a pro-noun!

scripts/npm/create-component.js
```diff
@@ -0,0 +1,156 @@
...
+    name: {
+      description: "Enter the component name, e.g. 'banner-alert'",
+      pattern: /^[a-z-]+$/,
+      message: 'Use snake case, e.g. "banner-alert".',
+      required: true,
+    },
+  },
+};
+
+const snakeCaseToPascalCase = input => {
```
matteo-hertel: we have lodash installed for this you can use something like
```
_.mixin({ 'pascalCase': _.flow(_.camelCase, _.upperFirst) });

console.log(_.pascalCase('this-is-snake-case'))
```

scripts/npm/create-component.js
```diff
@@ -0,0 +1,156 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+/* eslint-disable no-console */
+
+const { execSync } = require('child_process');
```
matteo-hertel: `sync` 👎

scripts/npm/create-component.js
```diff
@@ -0,0 +1,156 @@
...
+    storybookConfigFile = `native/storybook/storybook.js`;
+    storybookImport = `require('../packages/react-native-bpk-component-${name}/stories');`;
+  }
+
+  const pascalCaseName = snakeCaseToPascalCase(name);
+
+  const replaceInFiles = (oldWord, newWord) =>
+    `LC_CTYPE=C; LANG=C; find ${newComponentPath} -type f -exec sed -i "" "s/${oldWord}/${newWord}/g" {} +`;
+
+  const renameFiles = (oldWord, newWord) => {
```
matteo-hertel: I would argue that a Stream can do a much better job here that a bunch of commands, the problem with commands is they can behave differently on different machines, a stream will make the code neater and consistent mostly everywhere.
I might be too opinionated here, see what the other thinks

scripts/npm/create-component.js
shaundon: Ooh that looks cool

scripts/npm/create-component.js
shaundon: I'm not well-versed in Streams, have you got a link or anything I can look at?

scripts/npm/create-component.js
shaundon: I get you, but what's the benefit of doing this asynchronously?

matteo-hertel: in this particular case there probably no benefit, but it's not best practice, the power of node is in the event loop and async I/O

It's just me ranting you can safely ignore me 😄

scripts/npm/create-component.js
matteo-hertel: We use them in the svg transforming https://github.com/Skyscanner/backpack/blob/b88e0116f2bcfde3e380800cbad69fbb15ba7733/packages/bpk-svgs/tasks/svg2datauri.js#L72

I'm more than happy to pair with you if we want to go down the Streams route

scripts/npm/create-component.js
```diff
@@ -0,0 +1,206 @@
...
+  return 0;
+};
+
+const createComponent = async (err, { platform, name }) => {
+  if (err) {
+    console.error(err);
+    return;
+  }
+
+  let boilerplateComponentPath = `packages/bpk-component-boilerplate`;
```
k0nserv: `let` 😢 

Would it be nicer to write it like this? I appreciate not everyone is as alergic to `let` as I am
```javascript
const { boilerplateComponentPath, newComponentPath, storybookConfigFile, storybookImport } = platform === 'native' ? {
  boilerplateComponentPath:`native/packages/react-native-bpk-component-boilerplate`,
  newComponentPath: `native/packages/react-native-bpk-component-${name}`,
  storybookConfigFile: `native/storybook/storybook.js`,
  storybookImport:`require('../packages/react-native-bpk-component-${name}/stories');`,
} : {
  boilerplateComponentPath:`packages/bpk-component-boilerplate`,
  newComponentPath: `packages/bpk-component-${name}`,
  storybookConfigFile: `./storybook/config.js`,
  storybookImport:`require('./../${newComponentPath}/stories');`,
}
```

```

matteo-hertel: TBH I don't like that at all, there are a few use cases for let and this one is probably a good one, way more readable than the ternary operator

PR #892, opened 2018-07-03
[NO-JIRA] Add auto import and fix for new eslint rules

tiagoengel:


packages/bpk-component-barchart/src/__snapshots__/BpkChartAxis-test.js.snap
```diff
@@ -4,7 +4,7 @@ exports[`BpkChartAxis X should render band scale 1`] = `
 <g
   className="bpk-chart__axis"
   textAnchor="middle"
-  transform="translate(0, 120)"
+  transform="translate(0, NaN)"
```
tiagoengel: what! Looking into this

tiagoengel:
Addressed all feedback @shaundon and also dialed back the changes in the stories files too. Those have a high change of breaking things so I guess is better to ignore for now. We can start using the rule and and go back to fix those in the future.

.eslintrc
```diff
@@ -29,6 +29,7 @@
   },
   "rules": {
     "backpack/use-tokens": 2,
+    "backpack/auto-import-tokens": 2,
```
matthewdavidson: This config should live in `eslint-config-skyscanner` instead. Then we can upgrade this codebase to use the new version of that package in this PR.

tiagoengel: How should we deal with the native/web configuration though? We probably still need the config in the native folder.

matthewdavidson: sure, but the defaults for:
```
"backpack/use-tokens": 2,
"backpack/auto-import-tokens": 2,
```

Already live in `eslint-config-skyscanner` so we should use em (thats what the shared config is for). Overrides for native can stay there as they are intentional

tiagoengel: Right, I vote for doing it on a follow up PR. Don't want to add more things to this one.

native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar.ios.js
```diff
@@ -52,20 +52,21 @@ const statusBarPadding = isIphoneX ? 44 : 20;
...
   },
   barOuterWithSubtitle: {
-    paddingBottom: 16,
+    paddingBottom: 16, // eslint-disable-line backpack/use-tokens
   },
   barInner: {
     paddingTop: statusBarPadding, // Status bar
-    height: 64, // 44 for the bar + 20 for the status bar
+    // 44 for the bar + 20 for the status bar
+    height: 44, // eslint-disable-line backpack/use-tokens
```
shaundon: Looks like a typo - it's changed from `64` to `44`?

tiagoengel: fixing it 😞

tiagoengel: done

PR #895, opened 2018-07-03
[No JIRA] Add boilerplate component for React Native

shaundon:


native/packages/react-native-bpk-component-boilerplate/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "react-native-bpk-component-boilerplate",
+  "version": "0.0.0",
```
georgegillams: should this have `private: true` set?

shaundon: I think we're ok without it. The web one doesn't have it, and it's ignored in `lerna.json` so it can't actually get published.

georgegillams: Tru dat!

PR #896, opened 2018-07-04
[BPK-1489] Add RN Image component

georgegillams:
TODO:
 - [x] Readme
 - [x] Tests
 - [x] Improve performance
 - [x] Unreleased.md
 - [ ] Default loading state HOC
 - [ ] Docs
 - [ ] Release
 - [ ] Comms
 - [ ] Update Carousel stories and docs to use `BpkImage`

### DIFF
--- a/native/packages/react-native-bpk-component-image/src/BpkImage-test.common.js
+++ b/native/packages/react-native-bpk-component-image/src/BpkImage-test.common.js
@@ -1,0 +1,86 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import { StyleSheet } from 'react-native';
+import renderer from 'react-test-renderer';
+
+import BpkImage from './BpkImage';
+
+const requiredProps = {
+  alt: 'test',
+  source: {
+    uri:
+      'https://images.unsplash.com/photo-1518225853900-cc01b1092211?ixlib=rb-0.3.5&ixid=eyJhcHBfaWQiOjEyMDd9&s=d75dd482ea9e2ed4d53017c326497143&dpr=1&auto=format&fit=crop&w=1000&q=80&cs=tinysrgb',
+  },
+};
+
+const commonTests = () => {
+  describe('BpkImage', () => {
+    it('should render correctly', () => {
+      const tree = renderer.create(<BpkImage {...requiredProps} />).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly with custom style', () => {
+      const styles = StyleSheet.create({
+        custom: {
+          flex: 1,
+        },
+      });
+
+      const tree = renderer
+        .create(<BpkImage style={styles.custom} {...requiredProps} />)
+        .toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly with arbitrary props', () => {
+      const tree = renderer
+        .create(<BpkImage testID="123" {...requiredProps} />)
+        .toJSON();
+
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly when not loaded', () => {
+      const tree = renderer
+        .create(<BpkImage loaded={false} {...requiredProps} />)
+        .toJSON();
+
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly without border radius', () => {
+      const tree = renderer
+        .create(<BpkImage rounded={false} {...requiredProps} />)
+        .toJSON();
+
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('should render correctly when unloaded and out of view', () => {
+      const tree = renderer
+        .create(<BpkImage inView={false} loaded={false} {...requiredProps} />)
+        .toJSON();
+
+      expect(tree).toMatchSnapshot();
+    });
+  });
+};
+export default commonTests;
--- a/native/packages/react-native-bpk-component-image/src/BpkImage.js
+++ b/native/packages/react-native-bpk-component-image/src/BpkImage.js
@@ -1,0 +1,165 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import {
+  Animated,
+  StyleSheet,
+  View,
+  ViewPropTypes,
+  type ImageSourcePropType,
+  type StyleObj,
+} from 'react-native';
+import BpkSpinner from 'react-native-bpk-component-spinner';
+
+import {
+  borderRadiusSm,
+  animationDurationBase,
+  colorGray300,
+} from 'bpk-tokens/tokens/base.react.native';
+
+const styles = StyleSheet.create({
+  outer: {
+    width: '100%',
+    height: '100%',
+  },
+  outerWithBorderRadius: {
+    borderRadius: borderRadiusSm,
+    overflow: 'hidden',
+  },
+  image: {
+    width: '100%',
+    height: '100%',
+  },
+  loadingIndicatorView: {
+    flex: 1,
+    width: '100%',
+    height: '100%',
+    backgroundColor: colorGray300,
+    justifyContent: 'center',
+    alignItems: 'center',
+    position: 'absolute',
+  },
+});
+
+export type Props = {
+  alt: string,
+  source: ImageSourcePropType,
+  inView: boolean,
+  loaded: boolean,
+  onLoad: ?() => mixed,
+  rounded: boolean,
+  style: ?StyleObj,
+};
+
+class BpkImage extends Component<Props> {
+  showLoadingIndicator: boolean;
+
+  imageOpacity: number;
+
+  loadingIndicatorOpacity: number;
+
+  static propTypes = {
+    alt: PropTypes.string.isRequired,
+    source: PropTypes.oneOfType([PropTypes.string, PropTypes.object])
+      .isRequired,
+    inView: PropTypes.bool,
+    loaded: PropTypes.bool,
+    onLoad: PropTypes.func,
+    rounded: PropTypes.bool,
+    style: ViewPropTypes.style,
+  };
+
+  static defaultProps = {
+    inView: true,
+    loaded: true,
+    onLoad: null,
+    rounded: true,
+    style: null,
+  };
+
+  constructor(props: Props) {
+    super(props);
+
+    this.showLoadingIndicator = !props.loaded;
+    this.imageOpacity = new Animated.Value(props.loaded ? 1 : 0);
+    this.loadingIndicatorOpacity = new Animated.Value(props.loaded ? 0 : 1);
+  }
+
+  componentDidUpdate = (prevProps: Props) => {
+    if (!prevProps.loaded && this.props.loaded) {
+      Animated.sequence([
+        Animated.timing(this.imageOpacity, {
+          toValue: 1,
+          duration: animationDurationBase,
+        }),
+        Animated.timing(this.loadingIndicatorOpacity, {
+          toValue: 0,
+          duration: animationDurationBase,
+        }),
+      ]).start(() => {
+        this.showLoadingIndicator = false;
+      });
+    }
+  };
+
+  onImageLoad = () => {
+    if (this.props.onLoad) {
+      this.props.onLoad();
+    }
+  };
+
+  render() {
+    const { style: userStyle, inView, rounded, loaded, ...rest } = this.props;
+
+    const outerStyle = [styles.outer];
+    if (rounded) {
+      outerStyle.push(styles.outerWithBorderRadius);
+    }
+    if (userStyle) {
+      outerStyle.push(userStyle);
+    }
+
+    return this.showLoadingIndicator ? (
+      <View style={outerStyle}>
+        <Animated.View
+          style={[
+            styles.loadingIndicatorView,
+            { opacity: this.loadingIndicatorOpacity },
+          ]}
+        >
+          {inView && <BpkSpinner small type="dark" />}
+        </Animated.View>
+        {inView && (
+          <Animated.Image
+            onLoad={this.onImageLoad}
+            style={[styles.image, { opacity: this.imageOpacity }]}
+            {...rest}
+          />
+        )}
+      </View>
+    ) : (
+      <Animated.Image onLoad={this.onImageLoad} style={outerStyle} {...rest} />
+    );
+  }
+}
+
+export default BpkImage;
--- a/native/packages/react-native-bpk-component-image/stories.js
+++ b/native/packages/react-native-bpk-component-image/stories.js
@@ -1,0 +1,124 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React, { Component } from 'react';
+import { View, StyleSheet, type ImageSourcePropType } from 'react-native';
+import BpkButton from 'react-native-bpk-component-button';
+import { spacingBase, spacingLg } from 'bpk-tokens/tokens/base.react.native';
+import { storiesOf } from '@storybook/react-native';
+
+import CenterDecorator from '../../storybook/CenterDecorator';
+import BpkImage from './index';
+
+const styles = StyleSheet.create({
+  button: {
+    marginBottom: spacingBase,
+  },
+  image: {
+    width: '100%',
+    height: spacingLg * 10,
+  },
+});
+
+type Props = {
+  source: ImageSourcePropType,
+};
+type State = { loaded: boolean };
+
+class BpkImageLoadingStory extends Component<Props, State> {
+  static propTypes = {};
+
+  static defaultProps = {};
+
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      loaded: false,
+    };
+  }
+
+  onLoadToggle = () => {
+    this.setState(prevState => ({ loaded: !prevState.loaded }));
+  };
+
+  render() {
+    const { source, ...rest } = this.props;
+
+    return (
+      <View>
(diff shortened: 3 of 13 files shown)


native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,170 @@
...
+  colorGray300,
+  animationDurationBase,
+} from 'bpk-tokens/tokens/base.react.native';
+
+const styles = StyleSheet.create({
+  outer: {
+    width: '100%',
+    height: '100%',
+  },
+  withBorderRadius: {
```
shaundon: The convention we follow here is a BEM-like structure, so this would be better as `outerWithBorderRadius`.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,170 @@
...
+    borderRadius: borderRadiusSm,
+    overflow: 'hidden',
+  },
+  image: { width: '100%', height: '100%', position: 'absolute' },
```
shaundon: We should get this on multiple lines to match the other blocks.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,170 @@
...
+
+export type Props = {
+  style: StyleObj,
+  inView: ?boolean,
+  borderRadius: ?boolean,
+  loaded: ?boolean,
+  alt: string,
+  source: string,
+  onLoad: ?func,
+};
```
shaundon: A few small Flow issues:

* Convention with is required first (alphabetised), then optional (alphabetised). Same as we do with prop types.
* We avoid using `func` because it's too vague. Flow lets you define exactly what arguments a function takes, plus the return type. As `onLoad` takes no arguments and can return anything (it's up to the consumer), the type would be `?() => mixed`.
* Flow types don't map 1:1 with prop types. Rule of thumb is that it's only optional if the `defaultProps` declaration is `null`. In this case, the `boolean` types all have non-null defaults so they're not optional. As `style` **does** default to `null`, it should be optional.

Put it all together and you get:

```
type Props = {
  alt: string,
  borderRadius: boolean,
  inView: boolean,
  loaded: boolean,
  source: string,
  onLoad: ?() => mixed,
  style: ?StyleObj,
};
```

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,170 @@
...
+  spinnerViewOpacity: number,
+};
+
+class BpkImage extends Component<Props, State> {
+  static propTypes = {
+    style: ViewPropTypes.style,
+    inView: PropTypes.bool,
+    loaded: PropTypes.bool,
+    onLoad: PropTypes.func,
+    borderRadius: PropTypes.bool,
```
shaundon: * Should be alphabetised.
* `alt` is missing.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,170 @@
...
+  };
+
+  static defaultProps = {
+    style: null,
+    inView: true,
+    loaded: true,
+    onLoad: null,
+    borderRadius: true,
```
shaundon: Should be alphabetised.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,170 @@
...
+  inView: ?boolean,
+  borderRadius: ?boolean,
+  loaded: ?boolean,
+  alt: string,
+  source: string,
+  onLoad: ?func,
+};
+
+type State = {
+  renderSpinnerView: boolean,
```
shaundon: To me this name implies that it's going to be a function - think of `render()` for example. Maybe `showSpinnerView` instead?

shaundon: Actually, let's not use the term `spinner` because the fact that we're using a spinner is an internal implementation detail. If we changed to use something else that wasn't a spinner, we'd have to change this variable too. How about `showLoadingIndicator`?

Note: we should remove `spinner` elsewhere too like in the stylesheet naming.

native/packages/react-native-bpk-component-image/stories.js
```diff
@@ -0,0 +1,151 @@
...
+    <BpkImage
+      alt="test"
+      source={{
+        uri:
+          'https://content.skyscnr.com/96508dbac15a2895b0147dc7e7f9ad30/canadian-rockies-canada.jpg',
+      }}
+      style={styles.image}
+    />
+  ))
+  .add('default png', () => (
```
shaundon: We have a convention to name the ones we're going to use for docs site screenshots like `docs:thing`. So I reckon `docs:default` and `docs:no-border-radius` would work here.

native/packages/react-native-bpk-component-image/stories.js
```diff
@@ -0,0 +1,151 @@
...
+  .add('loading png', () => (
+    <BpkImageLoadingStory
+      source={{
+        uri:
+          'https://upload.wikimedia.org/wikipedia/commons/thumb/3/36/Two_red_dice_01.svg/2000px-Two_red_dice_01.svg.png',
+      }}
+    />
+  ))
+
+  .add('control', () => (
```
shaundon: We can probably replace this one now. A performance testing story would be good, I think. Bang 100 images in a ScrollView and check scroll performance vs doing the same with regular images.

native/packages/react-native-bpk-component-image/src/BpkImage-test.common.js
```diff
@@ -0,0 +1,124 @@
...
+import React from 'react';
+import { StyleSheet } from 'react-native';
+import renderer from 'react-test-renderer';
+
+import BpkImage from './BpkImage';
+
+const commonTests = () => {
+  describe('BpkImage', () => {
+    it('should render correctly', () => {
+      const tree = renderer.create(<BpkImage />).toJSON();
```
shaundon: Could you add `alt` and `source` to this test too? We try to make sure all tests include the required props, as this matches our philosophy of not supporting/considering components with missing required props.

native/packages/react-native-bpk-component-image/src/BpkImage-test.common.js
```diff
@@ -0,0 +1,124 @@
...
+    it('should render correctly when unloaded and out of view', () => {
+      const tree = renderer
+        .create(
+          <BpkImage
+            inView={false}
+            loaded={false}
+            alt="test"
+            source={{
+              uri:
+                'https://content.skyscnr.com/200946ddb82b7c026e6e186a7037b1f8/machu-picchu.jpg',
```
shaundon: Because you're using the same `alt` and `source` props for every test, you could make them a bit DRYer and reduce visual noise by putting these into a variable that you can reuse in each test:

```
const requiredProps = {
  alt: 'test',
  source: {
    uri: 'https....",
  },
};

...

<BpkImage {...requiredProps} inView={false} />
```

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,176 @@
...
+    height: '100%',
+    position: 'absolute'
+  },
+  loadingIndicatorView: {
+    width: '100%',
+    height: '100%',
+    backgroundColor: colorGray300,
+    justifyContent: 'center',
+    alignItems: 'center',
+    position: 'absolute',
```
shaundon: I'm not certain but I reckon the number of style props can be reduced here - is `position: absolute` needed? I'll have a play locally and see if I can optimise it, if you don't mind?

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,177 @@
...
+});
+
+export type Props = {
+  onLoad: ?() => mixed,
+  style: ?StyleObj,
+  alt: string,
+  borderRadius: boolean,
+  inView: boolean,
+  loaded: boolean,
+  source: { uri: string } | (string => mixed),
```
shaundon: Types still need to be re-ordered so that optional comes at the end.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,177 @@
...
+type State = {
+  imageOpacity: number,
+  loadingIndicatorOpacity: number,
+  showLoadingIndicator: boolean,
+};
+
+class BpkImage extends Component<Props, State> {
+  static propTypes = {
+    alt: PropTypes.string.isRequired,
+    source: PropTypes.oneOfType([PropTypes.string, PropTypes.object])
```
shaundon: Just thinking, we could maybe trade `object` for `shape`, which will let us enforce that it has a `uri` property (and anything else that `source` accepts.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,177 @@
...
+      .isRequired,
+    borderRadius: PropTypes.bool,
```
shaundon: Sorry, I should have raised this yesterday but it only just occurred to me.

I'm thinking it would be better to name this prop `rounded` (or something to that effect), because the prop should describe the outcome (rounded corners) as opposed to the specific implementation that'll achieve it. 

Let's say that RN changes `borderRadius` to something by another name - now the prop type is incorrectly named. But if we use `rounded`, it'll always be accurate.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,177 @@
...
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      showLoadingIndicator: !props.loaded,
+      imageOpacity: new Animated.Value(props.loaded ? 1 : 0),
+      loadingIndicatorOpacity: new Animated.Value(props.loaded ? 0 : 1),
+    };
+  }
+
+  componentWillReceiveProps = newProps => {
```
shaundon: Convention is to use `nextProps` as the arg name, plus because we have a Flow type for props, you can wrap it in that:

```... (nextProps: Props) => {```

File: native/packages/react-native-bpk-component-image/stories.js [hunk 16]
```diff
@@ -0,0 +1,147 @@
...
+  button: {
+    marginBottom: spacingBase,
+  },
+  image: {
+    width: '100%',
+    height: 10 * spacingLg,
+  },
+});
+
+class BpkImageLoadingStory extends Component<Props, State> {
```
shaundon: `Props` and `State` aren't defined..

native/packages/react-native-bpk-component-image/stories.js
```diff
@@ -0,0 +1,147 @@
...
+import CenterDecorator from '../../storybook/CenterDecorator';
+import BpkImage from './index';
+
+const styles = StyleSheet.create({
+  button: {
+    marginBottom: spacingBase,
+  },
+  image: {
+    width: '100%',
+    height: 10 * spacingLg,
```
shaundon: Nitpick, but I think this is slightly more readable as `spacingLg * 10`.

native/packages/react-native-bpk-component-image/src/BpkImage.js
shaundon: Just had a play with this locally and this achieves the same thing in fewer lines of code:

```
  loadingIndicatorView: {
    flex: 1,
    backgroundColor: colorGray300,
    justifyContent: 'center',
    alignItems: 'center',
  },
```

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,178 @@
...
+    const outerStyle = [styles.outer];
+    if (borderRadius) {
+      outerStyle.push(styles.outerWithBorderRadius);
+    }
+    if (userStyle) {
+      outerStyle.push(userStyle);
+    }
+
+    return (
+      <View style={outerStyle} {...rest}>
```
shaundon: As you suggested, I just checked and we need to put `...rest` on the image component, because then consumers can make use of props from the underlying image component like `blurRadius`.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,178 @@
...
+          >
+            {inView && <BpkSpinner small type="dark" />}
+          </Animated.View>
+        )}
+        {inView && (
+          <Animated.Image
+            onLoad={this.onImageLoad}
+            style={[styles.image, { opacity: this.state.imageOpacity }]}
+            source={source}
+            alt={alt}
```
shaundon: Because we're putting `...rest` on here now, you can remove source={source} and alt={alt} because `...rest` will encapsulate those.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,178 @@
...
+  showLoadingIndicator: boolean,
+};
+
+class BpkImage extends Component<Props, State> {
+  static propTypes = {
+    alt: PropTypes.string.isRequired,
+    source: PropTypes.oneOfType([PropTypes.string, PropTypes.object])
+      .isRequired,
+    borderRadius: PropTypes.bool,
+    inView: PropTypes.bool,
```
shaundon: I don't understand the point of this prop, I get that it stops it from rendering, but I think that's something that can be left to userland to determine. Feels a bit unneeded to me..

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,178 @@
...
+    style: null,
+  };
+
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      showLoadingIndicator: !props.loaded,
+      imageOpacity: new Animated.Value(props.loaded ? 1 : 0),
+      loadingIndicatorOpacity: new Animated.Value(props.loaded ? 0 : 1),
+    };
```
shaundon: This component can be done without state by bringing these variables up to be class level properties, below `defaultProps` and before `constructor`:

```
  showLoadingIndicator: boolean = !this.props.loaded;
  imageOpacity: number = new Animated.Value(this.props.loaded ? 1 : 0);
  loadingIndicatorOpacity: number = new Animated.Value(this.props.loaded ? 0 : 1);
```

Then they can be referenced later with `this.showLoadingIndicator` etc as opposed to `this.state.showLoadingIndicator`.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,178 @@
...
+      setTimeout(() => {
+        Animated.timing(this.state.loadingIndicatorOpacity, {
+          toValue: 0,
+          duration: animationDurationBase,
+        }).start();
+      }, animationDurationBase);
+      setTimeout(() => {
+        this.setState({ showLoadingIndicator: false });
+      }, 2 * animationDurationBase);
+    }
```
shaundon: We can avoid all this funky `setTimeout` logic by using the AWESOME POWER of the Animated API:

```
      Animated.parallel([
        Animated.timing(this.imageOpacity, {
          toValue: 1,
          duration: animationDurationBase,
        }),
        Animated.timing(this.loadingIndicatorOpacity, {
          toValue: 0,
          duration: animationDurationBase,
        })
      ]).start(() => {
        this.showLoadingIndicator = false;
      });
```

native/packages/react-native-bpk-component-image/src/BpkImage.js
shaundon: GH has collapsed my above comment but it still stands (just so we don't forget)

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,178 @@
...
+    height: '100%',
+  },
+  outerWithBorderRadius: {
+    borderRadius: borderRadiusSm,
+    overflow: 'hidden',
+  },
+  image: {
+    width: '100%',
+    height: '100%',
+    position: 'absolute',
```
k0nserv: Does this really need to absolutely positioned?

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,178 @@
...
+    justifyContent: 'center',
+    alignItems: 'center',
+    position: 'absolute',
+  },
+});
+
+export type Props = {
+  onLoad: ?() => mixed,
+  style: ?StyleObj,
+  alt: string,
```
k0nserv: Can we put the required props before the optional ones like we do with propTypes?

native/packages/react-native-bpk-component-image/src/BpkImage.js
k0nserv: We can't fish this out of React Native itself I assume?

native/packages/react-native-bpk-component-image/src/BpkImage.js
k0nserv: I don't think we should be using `componentWillReceiveProps` it's advised against in the [React documentation](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops).

native/packages/react-native-bpk-component-image/src/BpkImage.js
georgegillams: My thinking was that when we come to create a lazyLoading HOC (or however we achieve it) we will need some way to tell the component whether it is `inView`. Not entirely sure how this mecahnism will work. Need to investigate this some more.

native/packages/react-native-bpk-component-image/src/BpkImage.js
georgegillams: oh cool. I didn't know about those features 🎉

native/packages/react-native-bpk-component-image/src/BpkImage.js
georgegillams: I'm not sure how else we can make the `Image` render on top of the loading `View`

native/packages/react-native-bpk-component-image/stories.js
```diff
@@ -0,0 +1,147 @@
...
+  ))
+  .add('docs:loading', () => (
+    <BpkImageLoadingStory
+      source={{
+        uri:
+          'https://content.skyscnr.com/8a8ac17b591b61e6fe5d8f63414561cd/amsterdam-the-netherlands.jpg',
+      }}
+    />
+  ))
+  .add('performance-test-Image', () => (
```
tiagoengel: nit: use perf:{desc} the same way we use docs:{desc}. 

I also don't see the point of testing the base Image component here. Is it to compare it with our component? If so, I think I can be removed, I imagine the base component will be faster and that's ok, we only need to check if ours is fast enough.

georgegillams: That's exactly why it's there. I don't think it's a bad thing for it to stay. At the moment `BpkImage` (once fully rendered) uses no more views than `Image`. If someone changes the component in the future they should check the same again. As there is still a lot of work to do on this component (lazy loading etc) I think it's worth leaving in

tiagoengel: I'd agree if this was actually measuring and comparing both, and if keeping the same performance as the base image component was a prerequisite. But it's bit hard to manually test both and decide if they're performing the same way. 

We really only need to check if our implementation is keeping 60 fps and not dropping frames. Also, we are actually testing the performance of the `ScrollView` here, it would be different we use a `RecyclerView`, for example.

The implementation seems simple enough to require a performance test at the moment, this seems a bit of premature optimisation IMO.

native/packages/react-native-bpk-component-image/src/BpkImage.js
shaundon: We'll investigate this today and see if we can remove this. Will report back!

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,163 @@
...
+});
+
+export type Props = {
+  alt: string,
+  source: ImageSourcePropType,
+  inView: boolean,
+  loaded: boolean,
+  onLoad: ?() => mixed,
+  rounded: boolean,
+  style: ?StyleObj,
```
shaundon: This still needs to be slightly reordered - `rounded` isn't optional so should be above `onLoad`.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -0,0 +1,163 @@
...
+    this.imageOpacity = new Animated.Value(props.loaded ? 1 : 0);
+    this.loadingIndicatorOpacity = new Animated.Value(props.loaded ? 0 : 1);
+  }
+
+  componentWillReceiveProps = (nextProps: Props) => {
+    if (!this.props.loaded && nextProps.loaded) {
+      Animated.sequence([
+        Animated.timing(this.imageOpacity, {
+          toValue: 1,
+          duration: animationDurationBase,
```
shaundon: Just want to double check that `animationDurationBase` is correct here and not `animationDurationSm`. Worth checking with Will.

native/packages/react-native-bpk-component-image/stories.js
georgegillams: Yeah actually I agree!

native/packages/react-native-bpk-component-image/src/BpkImage.js
georgegillams: Have sent a message to Will about this so that he can check when he's in

PR #898, opened 2018-07-04
[BPK-1682] videos on docs site

matteo-hertel:
The original card was to add the Alert component video on the docsite but given this PR will be quite big I don't want to add too much noise and have the focus around the important files.

some if not most of the code has been ~stolen~ borrowed from the PR we made at the designer offsite (https://github.com/Skyscanner/backpack/compare/story-control?expand=1).

One important thing to notice is the different platforms recorders, the iOS one is pretty straightforward and because I need the same API to work seemlessly on testing iOS and Android the android one has some code stolen from my repo https://github.com/matteo-hertel/android-commander/blob/master/src/platforms/android.js

*Note* the `adb` cli is needed for the android recording to work, to check if you have it installed run
`adb --version` if it fails, install the cli with `brew cask install android-platform-tools` (thanks @shaundon for spotting this)

To test this:

- run `npm install` or `npm run clean && npm i` if you're brave or you need a ☕️
  - FYI this will add `puppeteer` which is roughly 70MB 
- run the native storybook `npm run native`
- run the detox as before `npm run detox:test:android` (or `npm run detox:test:ios` if you're into it)
- everything should succeed and you should not see any kind of error

once the above is done, to further prove your awesomeness and thoroughly test the whole thing you can run from the root
- `rm  native/packages/react-native-bpk-component-switch/videos/android/*` or `rm  native/packages/react-native-bpk-component-switch/videos/ios/*`
- re-run the test above
- the video files should be back in place

The docs site can be tested by running `npm run docs` and once that is done visit (http://0.0.0.0:8080/components/switch?platform=native)[http://0.0.0.0:8080/components/switch?platform=native]

one maim thing to notice is the usage of `sleep(time)` across the board, all the times are made up (trying to be sensible) and they can potentially be changed if needed


native/e2e/recorders/android-recorder.js
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const { rawCommand, command, sleep } = require('./../utils');
+
+function video(filename) {
```
shaundon: I think we have a convention to use the `const video = (filename) => {` format. I prefer using arrow functions myself for the scoping and consistency, but I know you prefer the `function` style. Anybody else have an opinion here?

native/e2e/recorders/android-recorder.js
```diff
@@ -0,0 +1,77 @@
...
+      'screenrecord',
+      ' --bit-rate',
+      '6000000',
+      `/sdcard/${filename}.mp4`,
+    );
+  }
+  function stop() {
+    if (!handler) {
+      throw new Error(
+        'Cannot call stop video if the process has not being started',
```
shaundon: Typo: "has not **been** started"

native/e2e/recorders/android-recorder.js
```diff
@@ -0,0 +1,77 @@
...
+  async startRecording() {
+    this.handle.start();
+  }
+
+  async finishRecording() {
+    this.handle.stop();
+    await sleep(1000);
+    await this.handle
+      .get(this.filePath.replace(this.fileName, ''))
+      .then(() => this.handle.remove());
```
shaundon: As you're calling this with no arguments, I think you can use this instead:

`.then(this.handle.remove);`

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
```diff
@@ -0,0 +1,96 @@
...
+      subText: PropTypes.string.isRequired,
+    }),
+  ),
+};
+
+ComponentVideos.defaultProps = {
+  videos: [],
+};
+
+export default ComponentVideos;
```
shaundon: Why you no Flow, bro?

native/e2e/record.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const { sleep, cleanFileName } = require('./utils');
+
+const AndroidRecorder = require('./recorders//android-recorder');
```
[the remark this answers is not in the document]
matteo-hertel: is not good shout 👍

native/e2e/recorders/android-recorder.js
matteo-hertel: That's a good point, I've copy-pasted that from my library (see PR description) if the arrow function is a deal breaker I'll change it

native/e2e/recorders/android-recorder.js
matteo-hertel: you can use point-free style but technically the return value of the previous end block will be passed as an argument to the remove function that does not accepts any, can lead to confusion when reading the code in the future

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
matteo-hertel: ah! forgot about Flow 😅

native/e2e/recorders/android-recorder.js
shaundon: Ah fair enough!

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
matteo-hertel: What I meant is, the other components like `ComponentScreenshot.js` `DocsPageBuilder.js` are not flowtyped so for consistency I'll not add Flow

native/e2e/recorders/android-recorder.js
matthewdavidson: Yeah arrow functions over function declarations please 🙏

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
```diff
@@ -0,0 +1,96 @@
...
+      >
+        <BpkPanel>
+          <video
+            loop
+            controls
+            className={getClassName('bpkdocs-component-videos__item-video')}
+            autoPlay
+            {...rest}
+          >
+            <track kind="captions" />
```
shaundon: Why are we including subtitles? It makes this appear in Safari (see screenshot) which I imagine we don't want.

screen shot 2018-07-04 at 15 00 30

native/e2e/utils.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const { spawn } = require('child_process');
+
+function command(commandString, ...args) {
```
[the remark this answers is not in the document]
matteo-hertel: 🤦‍♂️

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
matteo-hertel: eslint fails if no track is provided, so we either blacklist the rule or add an empty caption that we might fill in later for accessibility purposes, preferences?

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
```diff
@@ -0,0 +1,96 @@
...
+        className={getClassName(
+          'bpkdocs-component-videos__item-video-container',
+        )}
+      >
+        <BpkPanel>
+          <video
+            loop
+            controls
+            className={getClassName('bpkdocs-component-videos__item-video')}
+            autoPlay
```
shaundon: Can we also add `playsInline` and `muted` too?

`playsInline` will mean that on small screen devices the video plays inline (without this, on iPhones the video *only* play in full screen).

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
matthewdavidson: Can we not just add a caption? :) Describe what is being shown in the video?

shaundon: If the video has the `muted` prop, the rule shouldn't fail: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/media-has-caption.md

However, I just tried this and it does still fail for me. This may be because we have an older version as this was a recent fix: https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/357

So I say opt out of the rule for now and when we upgrade eslint (which I believe is very soon) we can revisit this.

shaundon: > Can we not just add a caption? :) Describe what is being shown in the video?

We could do that too, it can just use the alt text we were using previously for the image.

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
matteo-hertel: You're awesome! yes will do!

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
shaundon: F(low)air enough

native/e2e/record.js
```diff
@@ -0,0 +1,47 @@
...
+  kind,
+  story,
+  basePath,
+  testFuction,
+) => {
+  const videoName = cleanFileName(story, kind);
+  const recorder = new Recorder(`${basePath}/${videoName}`, videoName);
+  const testEmitter = new EventEmitter();
+
+  testEmitter.on('test:started', () => recorder.startRecording());
```
k0nserv: Aren't event names like these usually in present tense? e.g `test:start` and `test:end`?

native/e2e/record.js
```diff
@@ -0,0 +1,47 @@
...
+  testEmitter.on('test:ended', async () => {});
+
+  await storybookController.start();
+  await storybookController.activateStory(kind, story);
+
+  await testFuction(testEmitter);
+
+  await recorder.finishRecording();
+  await storybookController.done();
+};
```
matteo-hertel: Quick note, the two functions above cannot be excuted as callback to an event, the origin implementation was

```
estEmitter.on('test:ended', async () => {
    await recorder.finishRecording();
    await storybookController.done();
});
```

but the process exits before that is completed and the result is videos are not saving and the test command hangs forever, so I've moved it and (a)waiting for it to complete

I'll remove the test ended in the next commit, for now it's just a reference

native/e2e/storybook-controller.js
```diff
@@ -0,0 +1,50 @@
...
+    this.storybookBaseUrl = storybookBaseUrl;
+  }
+
+  async start() {
+    this.browser = await puppeteer.launch();
+    this.page = await this.browser.newPage();
+  }
+
+  async done() {
+    console.log('done');
```
k0nserv: Can we remove this?

native/packages/react-native-bpk-component-switch/src/BpkSwitch-test.e2e.js
```diff
@@ -20,30 +20,50 @@
...
+  });
+
   beforeEach(async () => {
     await device.reloadReactNative();
   });
 
   it('docs:switch', async () => {
-    try {
+    const switchTest = async controllerEmitter => {
+      controllerEmitter.emit('test:started');
```
k0nserv: It would be nice if we could get rid of this repition in every testcase also I feel like we should await some common element being avaiable on the screen before we emit `test:started`

native/e2e/record.js
matteo-hertel: there are all sort of discussion about this, the side I've picked comes from DDD, an event is a reaction to something that happened, so the callback runs when the action has already been executed (in this case the test started), we might use events in the future so it'lll be good to make a decision on this, toughts? @shaundon @tiagoengel @amolgupta @georgegillams @matthewdavidson

k0nserv: With all Streams you have `on('data')` and `on('end')`

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
matteo-hertel: further to this, the `track` element require a `src` property to be a url, (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track) so I'll opt out from the rule as the video is `muted` anyway

native/packages/react-native-bpk-component-switch/src/BpkSwitch-test.e2e.js
matteo-hertel: it's a good shout, but it's definitely not part of this PR, need some proper investigation, where to add that common element etc

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.scss
```diff
@@ -0,0 +1,65 @@
...
+    @include bpk-breakpoint-above-mobile {
+      width: calc(50% - #{$bpk-spacing-base});
+      max-width: $bpk-one-pixel-rem * 540; // 1080 (width of Android screenshot) / 2 (pixel density)
+
+      @include bpk-margin-leading($bpk-spacing-base);
+
+      @include bpk-rtl {
+        margin-right: 0;
+
+        @include bpk-margin-trailing($bpk-spacing-base);
```
k0nserv: This is not what you want I think the marging should always be on the leading side so you can remove this `bpk-rtl` block and just keep `@include bpk-margin-leading($bpk-spacing-base)`

k0nserv: You also might want to pass `true` for the second argument to ensure the opposite margin is zeroed

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.scss
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      &:last-child {
+        margin-right: 0;
```
k0nserv: Nitpic: This whole select can just be `@include bpk-margin-trailing(0)`

matteo-hertel: with 0 breaks the margin between the two videos, `@include bpk-margin-trailing($bpk-spacing-base)` does the job

native/e2e/init.js
```diff
@@ -1,5 +1,4 @@
 /*
- *
```
shaundon: 💯 👌 👌

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
```diff
@@ -0,0 +1,98 @@
...
+        <BpkPanel>
+          <video
+            playsInline
+            muted
+            loop
+            controls
+            className={getClassName('bpkdocs-component-videos__item-video')}
+            {...rest}
+          >
+            <track kind="captions" />
```
shaundon: I thought we were removing this because it needs a URL?

packages/bpk-docs/src/components/DocsPageBuilder/DocsPageBuilder.js
```diff
@@ -283,6 +288,16 @@ DocsPageBuilder.propTypes = {
           subText: PropTypes.string.isRequired,
         }),
       ),
+      videos: PropTypes.arrayOf(
+        PropTypes.shape({
+          title: PropTypes.string.isRequired,
+          src: PropTypes.string.isRequired,
+          width: PropTypes.number.isRequired,
+          height: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
```
shaundon: I'm surprised this works, the spec says you can't use strings/percentages for height, it must be a number of pixels.

packages/bpk-docs/src/pages/NativeSwitchPage/NativeSwitchPage.js
```diff
@@ -30,22 +30,20 @@ const components = [
...
     title: 'Default',
-    screenshots: [
+    videos: [
       {
         title: 'iOS',
         width: 750,
-        height: 1334,
-        src: `/${iosScreenshotDefault}`,
-        altText: 'iOS Default Switch Component',
+        height: '100%',
```
shaundon: I just commented out the height part and it still works for me on Chrome. I reckon it's ignoring `100%` because that's not a valid value for height (see my above comment). So maybe we can remove this.

matteo-hertel: I'll remove it see what happens, cheers!

packages/bpk-docs/src/components/DocsPageBuilder/ComponentVideos.js
matteo-hertel: 😞

packages/bpk-docs/src/pages/NativeSwitchPage/NativeSwitchPage.js
matteo-hertel: perfection

PR #899, opened 2018-07-04
[NOJIRA] Upgrade eslint-config-airbnb

matthewdavidson:
> Note: I'm doing this in the Backpack codebase first due to the large test area it provides. I am able to do this because of the use of `peerDependencies`. The only consequence is a warning about an invalid peer dependency on `eslint-config-airbnb` from `eslint-config-skyscanner` which will be resolved once we fix the Skyscanner config in a follow up PR.

💣💣💣💣💣💣

The biggest changes:
- `no-useless-path-segments`:
This changes `./../path/to/component` to `../path/to/component`.
- Enforcing use of callbacks in `setState` calls.
This changes `this.setState({ expanded: !this.state.expanded });` to `this.setState(prevState =>({ expanded: !prevState.expanded }));`.
- `forbid-foreign-prop-types`:
Warns on use of `MyComponent.propTypes` or `MyComponent.defaultProps` in other components.


.eslintrc
```diff
@@ -40,6 +41,12 @@
         "allowRequiredDefaults": true
       }
     ],
+    "react/require-default-props": [
+      2,
+      {
+        "forbidDefaultForRequired": false
+      }
+    ],
```
matthewdavidson: This addition is in a similar vain to the exception above, i.e:
> We want to keep prop-types alongside flow types so we relax this rule to allow us to specify defaults for required flow props

.eslintrc
```diff
@@ -29,6 +29,7 @@
   },
   "rules": {
     "backpack/use-tokens": 2,
+    "react/destructuring-assignment": 0,
```
matthewdavidson: This one could be controversial (and in particular involves making hundreds of changes to our codebase). I'm opting to tackle it in a follow up PR / decision.

packages/bpk-docs/src/components/Header/HamburgerButton.js
```diff
@@ -44,7 +44,6 @@ const HamburgerButton = props => {
       onClick={onClick}
       className={classNames.join(' ')}
       aria-expanded={expanded}
-      aria-controls="menu-list"
```
matthewdavidson: This `id` wasn't even specified anywhere so 🤷‍♂️

packages/bpk-component-fieldset/index.js
```diff
@@ -19,11 +19,11 @@
...
-  BpkFieldsetDefaultPropTypes,
+  propTypes,
+  defaultProps,
 } from './src/BpkFieldset';
 
 export type BpkFieldsetProps = Props;
 
 export default BpkFieldset;
-export { BpkFieldsetPropTypes, BpkFieldsetDefaultPropTypes };
+export { propTypes, defaultProps };
```
tiagoengel: isn't this a breaking change?

packages/bpk-theming/src/BpkThemeProvider-test.js
```diff
@@ -95,6 +95,7 @@ describe('BpkThemeProvider', () => {
 
   it('should warn about missing theme attributes', () => {
     expect(
+      // eslint-disable-next-line react/forbid-foreign-prop-types
```
tiagoengel: why not doing the same thing here, i.e. exporting the propTypes?

package.json
```diff
@@ -95,16 +95,16 @@
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-15": "^1.0.5",
     "enzyme-to-json": "^3.3.0",
-    "eslint": "^5.0.1",
-    "eslint-config-airbnb": "^16.1.0",
+    "eslint": "^4.19.1",
```
shaundon: Why has eslint's version dropped from 5.x to 4.x?

packages/bpk-component-fieldset/index.js
shaundon: I don't really understand why the names here have changed, could you clarify? I'm assuming it's just a new convention.

packages/bpk-component-input/src/customPropTypes.js
```diff
@@ -1,60 +0,0 @@
...
-          ? null
-          : createError(`\`onClear\` must be a function`);
-      default:
-        return null;
-    }
-  }
-  return null;
-};
-
-export default clearablePropType;
```
shaundon: Is there an advantage to putting all this in one file? I feel we're getting awfully close to `common-types.js` becoming a utils dumping ground..

packages/bpk-component-loading-button/src/BpkLoadingButton.js
```diff
@@ -34,7 +34,8 @@ const getPropsIcon = props => {
 
   if (loading) {
     return iconLoading;
-  } else if (disabled) {
+  }
+  if (disabled) {
```
shaundon: Isn't this technically less performant because now this check will fire every time?

packages/bpk-component-popover/src/BpkPopoverPortal-test.js
```diff
@@ -71,7 +73,8 @@ describe('BpkPopoverPortal', () => {
       .create(
         <BpkPopoverPortal
           id="my-popover"
-          target={() => document.createElement('button')}
+          // $FlowFixMe
+          target={() => document.createElement('button', { type: 'button' })}
```
shaundon: I think there should be a quick note on this comment to explain why we're opting out of Flow here? Same as the convention we have to justify eslint opt outs.

packages/bpk-docs/src/components/DocsPageBuilder/TokenSwitcher.js
```diff
@@ -119,12 +119,14 @@ TokenSwitcher.defaultProps = {
...
 export const connect = (tokenSwitcher, table) => {
-  /* eslint-disable react/no-multi-comp */
+  /* eslint-disable react/no-multi-comp, react/no-this-in-sfc */
   class SwitchAwareTable extends Component {
     constructor(props) {
       super(props);
+
       this.state = { platform: null };
       this.onSwitch = platform => this.setState({ platform });
+
```
shaundon: I reckon we could refactor this to not use `this` (and then won't need the opt out) fairly easily - did you try that already though?

packages/bpk-docs/src/pages/HomePage/HomePage.js
```diff
@@ -186,7 +186,7 @@ const HomePage = (props, context) => {
 };
 
 HomePage.contextTypes = {
-  theme: PropTypes.object,
+  theme: PropTypes.object, // eslint-disable-line react/forbid-prop-types
```
shaundon: How come we didn't have to opt out here before? Seems like it should have already been causing an error.

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert-test.common.js
```diff
@@ -154,48 +154,6 @@ const commonTests = () => {
       expect(tree).toMatchSnapshot();
     });
 
-    it('should error if dismissable and expandable', () => {
```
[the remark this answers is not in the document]
matthewdavidson: Because it's already covered in the tests for the custom prop types.

packages/bpk-component-button/src/BpkButton.js
```diff
@@ -128,6 +128,7 @@ const BpkButton = (props: Props) => {
 
   const buttonType = submit ? 'submit' : 'button';
 
+  /* eslint-disable react/button-has-type */
```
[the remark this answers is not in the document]
matthewdavidson: `<button type="button" />` ;)

packages/bpk-component-fieldset/index.js
matthewdavidson: I changed them because the convention I seen elsewhere was to use `propTypes` and `defaultProps` instead of component specific names. But actually for packages that export multiple components perhaps it needs revisiting?

Either way because they are not documented then I consider them non breaking 😇

packages/bpk-theming/src/BpkThemeProvider-test.js
matthewdavidson: I have no idea maybe i got lazy 😂 

Will fix.

package.json
matthewdavidson: Because `eslint-config-airbnb` has a peer on 4.19.1 and hasn't upgraded to 5 yet - we prematurely upgraded to 5 in a PR the other day.

packages/bpk-component-input/src/customPropTypes.js
matthewdavidson: They are all related to types tho. More often than not we end up with wierd cyclic dependencies so its easier to have em all in the one file.

packages/bpk-component-loading-button/src/BpkLoadingButton.js
matthewdavidson: Nah because line 36 has a return statement. here is the eslint docs https://eslint.org/docs/rules/no-else-return

packages/bpk-component-popover/src/BpkPopoverPortal-test.js
matthewdavidson: Agreed - ill add it here and in the code for brevity:

"Flow doesn't yet support a secondary argument for document.createElement".

> Same as the convention we have to justify eslint opt outs.

Where is this convention? :P

packages/bpk-docs/src/components/DocsPageBuilder/TokenSwitcher.js
matthewdavidson: This PR is big enough :)

packages/bpk-component-banner-alert/src/withBannerAlertState.js
```diff
@@ -118,12 +118,15 @@ const withBannerAlertState = (WrappedComponent: ComponentType<any>) => {
...
     onExpandToggle = () => {
-      const expanded = !this.state.expanded;
-      this.setState({ expanded });
+      this.setState(prevState => {
+        const expanded = !prevState.expanded;
 
-      if (this.props.onExpandToggle) {
-        this.props.onExpandToggle(expanded);
-      }
+        if (this.props.onExpandToggle) {
```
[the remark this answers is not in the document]
matthewdavidson: but `expanded` needs to be derived from `prevState` ...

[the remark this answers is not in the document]
tiagoengel: ahh right, I didn't see it was an argument to the function. Maybe calling all of them in the setState callback would be better then, but tbh I don't have a strong opinion about this so it's fine by me.

packages/bpk-component-fieldset/index.js
shaundon: > Either way because they are not documented then I consider them non breaking 😇

That's fair

tiagoengel:
LGTM now, needs rebase and squashing

package.json
shaundon: Ahh gotcha

packages/bpk-component-loading-button/src/BpkLoadingButton.js
shaundon: Ah so it does!

packages/bpk-component-button/src/BpkButton.js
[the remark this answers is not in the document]
k0nserv: Where is this coming from? I feel like we should raise an issue or PR to stop having to disable this

native/packages/react-native-bpk-component-banner-alert/src/BpkBannerAlert.android.js
```diff
@@ -185,8 +185,8 @@ const BpkBannerAlert = (props: Props) => {
   );
 };
 
-BpkBannerAlert.propTypes = commonPropTypes;
+BpkBannerAlert.propTypes = { ...propTypes };
```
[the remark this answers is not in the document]
matthewdavidson: For immutability, if consumers modify that object that is now exported i.e. `delete propTypes.disabled` then that affects the same reference used by the component.

PR #900, opened 2018-07-04
Update this.theme on componentDidUpdate

ejulianova:


ejulianova:
@k0nserv added a note in unreleased.md and amended the commit :-)

PR #901, opened 2018-07-05
react-native-bpk-component-navigation-bar check if title is null

ejulianova:
Otherwise if `null` is passed as title it would throw an error `cannot read property icon of null`.


ejulianova:
That's expected behaviour as discussed with Hugo.

PR #903, opened 2018-07-06
[BPK-1695] Link to the sketch folder and not individual files

tiagoengel:
screen shot 2018-07-06 at 11 13 49


tiagoengel:
I've created a file there alert people the folder is public @matthewdavidson

PR #912, opened 2018-07-11
[BPK-1727] Add font weight documentation

shaundon:
This adds a 'font weight' section to the _Typesetting_ page. It's already been approved by Will in person.

In order to make the copy change when the platform does, I piggybacked off of the `tokenSwitcher` logic to add a `platformSwitchingContent` prop to `DocsPageBuilder.components`. It functions exactly like `tokenMap`, except it can take any arbitrary content. It's a bit messy feeling but, y'know..it's `DocsPageBuilder.js`...

screen shot 2018-07-11 at 14 42 33

**Note:** We could do a refactor to make `tokenMap` use this new `PlatformSwitchingContent` component, but it would be a relatively large refactor, so I think it best if we handle that another time.

### DIFF
--- a/packages/bpk-docs/src/components/DocsPageBuilder/PlatformSwitchingContent.js
+++ b/packages/bpk-docs/src/components/DocsPageBuilder/PlatformSwitchingContent.js
@@ -1,0 +1,37 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import { type Node } from 'react';
+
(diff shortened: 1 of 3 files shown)


packages/bpk-docs/src/components/DocsPageBuilder/PlatformSwitchingContent.js
```diff
@@ -0,0 +1,50 @@
...
+  platform: string,
+  content: {
+    android: Node,
+    ios: Node,
+    web: Node,
+  },
+};
+
+const PlatformSwitchingContent = (props: Props) => {
+  const { platform, content } = props;
```
tiagoengel: can we replace this with `return content[platform] || content.web`?

shaundon: If we were playing code golf, you'd have just scored a hole-in-one. Nice one 😄

shaundon: Done.

packages/bpk-docs/src/pages/TypesettingPage/TypesettingPage.js
```diff
@@ -45,6 +47,48 @@ const components = [
...
+      ios: (
+        <section>
+          <Paragraph>
+            When designing for iOS, use <strong>SF Pro Text</strong>.
+          </Paragraph>
+          <BpkList>
+            <BpkListItem>Regular</BpkListItem>
+            <BpkListItem>Semi-bold (for emphasised text)</BpkListItem>
+            <BpkListItem>
+              Heavy (optional, for <BpkCode>font-size-xxl</BpkCode> only)
```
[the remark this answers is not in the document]
shaundon: Have updated the copy after speaking to Will.

PR #915, opened 2018-07-12
[No JIRA] Remove large PR check from Danger

shaundon:
screen shot 2018-07-12 at 08 06 20

I think we all feel that the above message is just visual noise at this point. We've tried a few different thresholds for it appearing and can't seem to get it right. The reality is that a lot of things we do require a lot of files to be changed! 

For example, a new component tends to involve 20-30 changes, adding docs can be ~15 now we have the native/web wrappers plus screenshots, and updating eslint involves at least 9,000,000 changes.

So, I think we should remove this check. Benefits of this:

* We do a good job of tracking PR size ourselves, so we should keep Danger for things that humans are less good at spotting (e.g. 'have you updated `unreleased.md`).
* Reducing noise from Danger means that the times it does have something to say are more powerful.

bitmoji

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -131,18 +131,6 @@ if (unlicensedFiles.length > 0) {
   );
 }
 
-// Encourage smaller PRs.
-const bigPRThreshold = 8;
-if (fileChanges.length > bigPRThreshold) {
-  warn(
-    `This PR contains ${fileChanges.length} files (${
-      createdFiles.length
-    } new, ${
-      modifiedFiles.length
-    } modified). Consider splitting it into multiple PRs.`,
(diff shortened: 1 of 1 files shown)


georgegillams:
I agree with everything you've said above. Have we looked at a threshold for number of lines changed instead of number of files?

shaundon:
> Have we looked at a threshold for number of lines changed instead of number of files?

We haven't, and it's a good idea, but I'm a bit concerned it would suffer for all the same reasons. E.g. snapshot updated, Prettier changes, they all involve changing a lot of lines..

georgegillams:
tbh I agree that we are quite good at policing this ourselves like you said. Let's just get rid 👍

PR #918, opened 2018-07-13
[BPK-1671] 🐶🥙

k0nserv:


shaundon:
I'm running it locally and it seems good, the only thing is that my `package-lock.json` seems to change when `npm install`. Could you check and see if yours does?

k0nserv:
Right you are. Fixed now

PR #923, opened 2018-07-13
[BPK-1729] Optimise videos

shaundon:
I've added a script to convert videos to a format that plays in every browser. It also reduces the file size massively - the one in this PR went from 1.8mb to 106kb.

I've tested it in the following:

* Chrome (Windows, macOS, Android)
* Safari (macOS, iOS)
* Firefox (Windows, macOS)
* Edge 16
* IE11

To do
- [x] Integrate the script into the recording process. The Android ones are already good so I think we can just do this for iOS.
- [x] Add `brew install ffmpeg` to the prerequisites in `contributing.md`.


native/e2e/recorders/ios-recorder.js
```diff
@@ -27,6 +27,10 @@ const recordSimulator = fileName =>
     fileName,
   ]);
 
+const optimiseRecordedVideo = filename => {
+  exec(`scripts/video/convert-to-mp4.sh ${filename}`);
```
georgegillams: should this be execSync just to ensure that the video conversion is complete by the time the process exits, or are we confident that the script runs perfectly separately?

georgegillams: Also should we change the name of this folder to `detox` instead of `video` so that it matches the commands we run

Or maybe we're planning to change the npm commands to `generate-videos` or something?

shaundon: > should this be execSync just to ensure that the video conversion is complete by the time the process exits, or are we confident that the script runs perfectly separately?

Yeah could do - maybe sync is better here. @matteo-hertel what do you think?

shaundon: > Also should we change the name of this folder to detox instead of video so that it matches the commands we run

Good shout.

matteo-hertel: Either `exec` or `execSync` works as long as some sort of error handling is provided, for instance while testing my ffmpeg failed and no error was surfaced in the console

so something like
```
exec(script, (error, stdout, stderr) => {
  if (error) {
    console.error(`😱 : ${error}`);
    return;
  }`);
});
```

or if you want to go above an beyond use spawn

```
const { spawn } = require('child_process');
const awesomeScript = spawn('script', ['flags']);

awesomeScript.stderr.on('data', (data) => {
  console.log(`stderr: ${data}`);
});
```

scripts/video/convert-to-mp4.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+
+# Video resolution depends on platform.
+RESOLUTION="750x1334"
+if [[ $1 = *"videos/android/"* ]]; then
+  RESOLUTION="1080x1920"
+fi
+
+# Create the new file.
+# Flags explained:
```
matteo-hertel: True hero 🎖

native/e2e/recorders/ios-recorder.js
matteo-hertel: Actually, ignore that, I've added some utils to help with that 
https://github.com/Skyscanner/backpack/blob/master/native/e2e/utils.js#L29

you can use `command` for a promise base `spawn` or `rawCommand` as a simple wrapper of `spawn`

k0nserv: Following on this. If optimisation fails should we clean up the state by removing the video so we don't accidentally end  up using the unoptimised video?

shaundon:
> Following on this. If optimisation fails should we clean up the state by removing the video so we don't accidentally end up using the unoptimised video?

Yeah, good idea!

shaundon:
> Out of curiousty what are the videos like? Is the quality still pretty good?

To my eyes it looks the same!

k0nserv:
> To my eyes it looks the same!

Could you upload two comparisons?

shaundon:
@k0nserv 

This is the unoptimised one from the current live site: https://backpack.github.io/all-types-react-native-bpk-component-switch_2347215ed5db33cc7bd99b1deae65173.mp4

And see the file in this PR for the converted version.

native/e2e/recorders/ios-recorder.js
```diff
@@ -15,17 +15,25 @@
...
     'io',
     'booted',
     'recordVideo',
     '--type=fmp4',
     fileName,
-  ]);
+  );
+
+const optimiseRecordedVideo = filename => {
+  command(`scripts/video/convert-to-mp4.sh`, filename).catch(error => {
```
k0nserv: Shoudl we exit with a non-zero status code if this happens too?

shaundon: I dunno - when the tests fail currently we don't do this and the errors still get surfaced, so I'm not sure what difference it would make.

matteo-hertel: I don't think we should as exiting here might cause a problem with the detox process hanging

k0nserv: So then the only way to know the process has failed is to look at the output?

shaundon: Maybe I'm missing the point here, but if we're running this stuff manually, we're going to be looking at the output anyway right?

k0nserv: No you're right someone will be looking at it

shaundon: Ok cool - is this ok to merge then?

matteo-hertel: :shipit:

PR #925, opened 2018-07-16
Change ratio comparison from equal to greater than.

ShakMR:
This covers when the ratio is over 1 (almost 1)


packages/bpk-component-infinite-scroll/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bpk-component-infinite-scroll",
-  "version": "1.0.5",
+  "version": "1.0.6",
```
matteo-hertel: The update on the package.json is not needed as lerna will do that at release time

ShakMR: changed

matteo-hertel:
@ShakMR super thanks! if you can elaborate a bit more, in a sentence or so what the change is about, it'll be super awesome, just for future reference people can know why and what has changed

PR #931, opened 2018-07-17
[BPK-1707] Fix RTL in native horizontal nav

shaundon:
I had to add a wrapper view to the selected indicator so that it aligns correctly when using RTL.

It took me far too long to figure this out 🤦‍♂️

### DIFF
--- a/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
+++ b/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
@@ -16,7 +16,13 @@
  * limitations under the License.
  */
 
-import { ScrollView, StyleSheet, View, ViewPropTypes } from 'react-native';
+import {
+  I18nManager,
+  ScrollView,
+  StyleSheet,
+  View,
+  ViewPropTypes,
+} from 'react-native';
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
@@ -50,6 +56,9 @@ const styles = StyleSheet.create({
   },
   innerSpaceAround: {
     justifyContent: 'space-around',
+  },
+  indicatorWrapper: {
+    flexDirection: I18nManager.isRTL ? 'row-reverse' : 'row',
   },
 });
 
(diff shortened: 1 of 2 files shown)


k0nserv:
Couldn't you avoid the extra view with math? i.e. change how we calculate the xOffset in the case when when the direction is RTL? `indicatorOffsetX: componentWidth -x` should do it but requires knowing the full width of the component

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -121,10 +130,12 @@ class BpkHorizontalNav extends React.Component {
         return null;
       }
       return (
-        <AnimatedIndicator
-          xOffset={this.state.indicatorOffsetX}
-          width={this.state.indicatorWidth}
-        />
+        <View style={styles.indicatorWrapper}>
```
georgegillams: What if we applied the style to here instead?
https://github.com/Skyscanner/backpack/blob/a2ec249ab31ad13abed7f16a70837b8d6140af4c/native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js#L150

Then would we not need the additional view?

shaundon:
> Couldn't we avoid the extra view with math? i.e. change how we calculate the xOffset in the case when when the direction is RTL? indicatorOffsetX: componentWidth -x should do it but requires knowing the full width of the component

This is what I tried initially, but I ran into issues:
* As you mentioned, we'd need to know the width of the whole component. I added an `onLayout` handler to do this, but it doesn't always fire before the `onChildLayout` ones. So, we'd need to add extra logic to make sure the right things have fired.
* We could maybe put that logic in state, but then we'd have an extra render.
* I then thought we could sum the widths of the child components to get the total width, but this won't work with the `spaceAround` prop because there are gaps between them that won't be accounted for.

So given this, an extra view seemed like the best trade-off.

shaundon:
> Y no tests bruh?

I looked into this, it's because the indicator doesn't render if `state.indicatorOffsetX` isn't set. That doesn't get set if `onLayout` doesn't fire, which I guess doesn't happen in Jest.

native/packages/react-native-bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: I'll try that 👍

PR #932, opened 2018-07-17
[BPK-1699] Update react-slider version

k0nserv:
The bug in BPK-1699 is resolved by upgrading to the latest version of `react-slider`. `react-slider` now has a peer dependency on React of `^16.2`, but it still works fine with React 15 in my tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5630,6 +5630,7 @@
         "require-from-string": "2.0.1",
         "rfc6902": "2.2.2",
         "supports-hyperlinks": "1.0.1",
+        "vm2": "github:patriksimek/vm2#468bc1e54e75e766b842830ea775669992a979e0",
         "voca": "1.4.0"
       },
       "dependencies": {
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -5630,6 +5630,7 @@
         "require-from-string": "2.0.1",
         "rfc6902": "2.2.2",
         "supports-hyperlinks": "1.0.1",
+        "vm2": "github:patriksimek/vm2#468bc1e54e75e766b842830ea775669992a979e0",
```
shaundon: This keeps appearing for me when I `npm install`. Hopefully this will fix it once and for all!

packages/bpk-component-slider/src/BpkSlider-test.js
```diff
@@ -20,6 +20,8 @@ import React from 'react';
 import renderer from 'react-test-renderer';
 import BpkSlider from './BpkSlider';
 
+jest.mock('react-slider', () => 'react-slider');
+
```
shaundon: Is this just a good practice thing or is there a specific reason we're now mocking this out?

k0nserv: It relies on a fuller browser env than is provided by `jsdom` so it breaks if not mocked

shaundon: Ok cool 👍 Thanks for clarifying

PR #941, opened 2018-07-19
[BPK-1733] Ticket images

shaundon:
I haven't made a story for this as it's pretty trivial, but here's the one I tested it with in development to prove this works:

screen shot 2018-07-19 at 13 10 18


shaundon:
I've only applied the fix to the main/left side of the ticket. Is there any reason we can think of **not** to do it to both sides?

PR #943, opened 2018-07-19
[BPK-1724] Fix image loading behaviour in mobile safari

tiagoengel:
I have a video comparing before and after if you want check the difference 🙂

### DIFF
--- a/packages/bpk-component-image/src/withLazyLoading.js
+++ b/packages/bpk-component-image/src/withLazyLoading.js
@@ -20,6 +20,7 @@
 import React, { type Node, type ComponentType } from 'react';
 import PropTypes from 'prop-types';
 import { wrapDisplayName } from 'bpk-react-utils';
+import throttle from 'lodash/throttle';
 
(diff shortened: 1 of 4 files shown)


packages/bpk-component-image/src/withLazyLoading.js
```diff
@@ -104,11 +103,11 @@ export default function withLazyLoading(
       documentRef.removeEventListener('fullscreenchange', this.checkInView);
     };
 
-    checkInView = (): void => {
+    checkInView = throttle(() => {
```
tiagoengel: This is not part of fix per se, just a general good practice and because it was causing some performance issues in my tests.

georgegillams: 👍

georgegillams:
Docs page images seem to be broken now 😢 

screen shot 2018-07-20 at 13 34 12

tiagoengel:
Should be ok now @georgegillams

PR #946, opened 2018-07-20
[No JIRA] Use consistent casing for standard files

k0nserv:


shaundon:
Also, I'd recommend doing a release as soon as this is merged as it'll cause so many components to change.

PR #947, opened 2018-07-20
[BPK-1692] Debugging

georgegillams:
The issue seems to be that, in Safari, the Portal element is handling the click events that have already happened after opening the Popover.

In other words the chronological order of events is:
 - mouseDown on date field
 - mouseUp on date filed -> results in Popover opening
 - portal is rendered, adding event listeners for mouse events
 - portal handles mouseDown and mouseUp events from above -> results in Popover closing 

There seem to be three sensible ways forward in fixing this:
1. Add a delay between the date field click and rendering the Calendar. This will prevent the Portal handing the same mouse/touch events
2. Add a delay between rendering a portal and adding the mouse/touch event listeners. Similar idea to (1), but hopefully solves the issue for other popup components too.
3. Decide that this doesn't happen that often, and that it only affects mobile safari, and therefore don't fix

### DIFF
--- a/packages/bpk-component-datepicker/src/BpkDatepicker.js
+++ b/packages/bpk-component-datepicker/src/BpkDatepicker.js
@@ -41,12 +41,17 @@ class BpkDatepicker extends Component {
   }
 
   onOpen = () => {
-    this.setState({
-      isOpen: true,
-    });
+    setTimeout(() => {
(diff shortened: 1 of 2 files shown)


georgegillams:
Not reproducible on actual iOS device (Tested iOS 12.0)

PR #953, opened 2018-07-23
[BPK-1734] Add `withNotches` prop to `BpkTicket`

shaundon:
To do
* [x] Logic
* [x] Add stories
* [x] Rebase
* [x] Add example to docs
* [x] Add tests
* [x] Update snapshots
* [x] Refactor the code because it's gross
* [x] Update changelog
* [x] Readme
* [x] Fix it in IE and Edge

Also includes:
* Add Flow
* Remove IE9 fallback code
* Refactor className logic

**Implementation notes**
This wasn't as simple as it first appeared. Just removing the notch elements and extending the punchline isn't optimal, because there's a small gap in the box shadow where the punchline is. 

This is because in the original implementation, the box shadow isn't applied to the outermost element, it's applied to each side of the inner parts (_main_ and _stub_).

Also, the punchline was covering up the box shadow from the middle before, but the punchline is only 2px wide now.

So, this PR tweaks the logic so that the box shadow is applied on the outermost element, unless the `withNotches` prop is applied, and then we use the original behaviour.

**Screenshot from the docs site**
screen shot 2018-07-23 at 13 57 55

**IE and Edge**
Previously, due to me using `box-shadow: unset`, I broke the existing tickets (thanks to @k0nserv for spotting that!). I've changed it to `box-shadow: none` and it works now. Here's some screenshots:

IE
screen shot 2018-07-24 at 08 56 03

Edge
screen shot 2018-07-24 at 08 53 58

### DIFF
--- a/packages/bpk-component-ticket/src/BpkTicket.js
+++ b/packages/bpk-component-ticket/src/BpkTicket.js
@@ -32,6 +32,7 @@ type Props = {
   stubProps: { [string]: any },
   padded: boolean,
   vertical: boolean,
+  withNotches: boolean,
   className: ?string,
   stubClassName: ?string,
   href: ?string,
@@ -44,6 +45,7 @@ const BpkTicket = (props: Props) => {
     padded,
     stub,
     vertical,
+    withNotches,
     className,
     stubClassName,
     stubProps,
@@ -54,6 +56,7 @@ const BpkTicket = (props: Props) => {
     'bpk-ticket',
     className,
     vertical && 'bpk-ticket--vertical',
+    withNotches && 'bpk-ticket--with-notches',
   );
 
   const mainClassNames = getClassName(
@@ -62,6 +65,7 @@ const BpkTicket = (props: Props) => {
     padded && 'bpk-ticket__main--padded',
     vertical && 'bpk-ticket__main--vertical',
     !vertical && 'bpk-ticket__main--horizontal',
+    withNotches && 'bpk-ticket__paper--with-notches',
   );
 
   const mainInnerClassNames = getClassName(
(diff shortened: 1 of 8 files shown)


packages/bpk-component-ticket/src/BpkTicket.js
```diff
@@ -16,132 +16,144 @@
...
-  const mainInnerClassNames = [getClassName('bpk-ticket__main-inner')];
-  const stubClassNames = ['bpk-ticket__paper', 'bpk-ticket__stub'].map(
-    getClassName,
+
+  const mainClassNames = getClassName(
+    'bpk-ticket__paper',
+    'bpk-ticket__main',
+    padded && 'bpk-ticket__main--padded',
+    vertical && 'bpk-ticket__main--vertical',
+    !vertical && 'bpk-ticket__main--horizontal',
```
sogame: What about `vertical ? 'bpk-ticket__main--vertical' : 'bpk-ticket__main--horizontal',`? It can be weird as it's the only place this would be used, but it also feels weird to have `vertical && ...` and in the next line `!vertical && ...` 😅

packages/bpk-component-ticket/src/BpkTicket.js
```diff
@@ -16,132 +16,144 @@
...
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { type Node } from 'react';
 import { cssModules } from 'bpk-react-utils';
 
 import STYLES from './bpk-ticket.scss';
 
 const getClassName = cssModules(STYLES);
 
-const isIE9 = () => {
```
georgegillams: 💥

shaundon: I can't tell if this is a positive or negative comment 😂

PR #958, opened 2018-07-25
[BPK-1765] Add `react-native-backpack` package

k0nserv:


native/packages/react-native-backpack/android/settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = 'Backpack2'
```
k0nserv: We should have a TODO here to change this

amolgupta: done

PR #959, opened 2018-07-25
Added scrollable calendar date component

boris-barath:
Implemented a scrollable calendar date component building on top of BpkCalendarDate

the component uses all the same props, but when isOutside is passed, it only returns an empty span tag in order to only display dates within a specific grid's month.

### DIFF
--- a/packages/bpk-component-calendar/src/BpkCalendarDate.js
+++ b/packages/bpk-component-calendar/src/BpkCalendarDate.js
@@ -139,7 +139,7 @@ class BpkCalendarDate extends PureComponent {
   }
 }
 
-BpkCalendarDate.propTypes = {
+export const BpkCalendarDatePropTypes = {
   // Required
   date: PropTypes.instanceOf(Date).isRequired,
   // Optional
@@ -155,7 +155,7 @@ BpkCalendarDate.propTypes = {
   preventKeyboardFocus: PropTypes.bool,
 };
 
-export const BpkCalendarDatePropTypes = BpkCalendarDate.propTypes;
+BpkCalendarDate.propTypes = BpkCalendarDatePropTypes;
 
 BpkCalendarDate.defaultProps = {
   isBlocked: false,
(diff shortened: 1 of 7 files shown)


packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarDate.js
```diff
@@ -0,0 +1,59 @@
...
+import PropTypes from 'prop-types';
+import React, { PureComponent } from 'react';
+import { BpkCalendarDate, CustomPropTypes } from 'bpk-component-calendar';
+
+class BpkScrollableCalendarDate extends PureComponent {
+  render() {
+    const { date, isOutside, ...rest } = this.props;
+
+    // Returning empty span when isOutside is true ensures only focused month displays in grid
+    return !isOutside ? <BpkCalendarDate date={date} {...rest} /> : <span />;
```
k0nserv: Wouldn't it be better to just return `null` instead?

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarDate.js
```diff
@@ -0,0 +1,59 @@
...
+  // Optional
+  isBlocked: PropTypes.bool,
+  isFocused: PropTypes.bool,
+  isOutside: PropTypes.bool,
+  isSelected: PropTypes.bool,
+  isToday: PropTypes.bool,
+  modifiers: CustomPropTypes.DateModifiers,
+  onClick: PropTypes.func,
+  onDateKeyDown: PropTypes.func,
+  preventKeyboardFocus: PropTypes.bool,
```
k0nserv: Shouldn't we spread all `CustomPropTypes` here?

boris-barath: Could you explain what you mean? I am using the same the same props as BpkCalendarDate

packages/bpk-component-calendar/src/BpkCalendarDate.js
```diff
@@ -155,6 +155,8 @@ BpkCalendarDate.propTypes = {
   preventKeyboardFocus: PropTypes.bool,
 };
 
+export const BpkCalendarDatePropTypes = BpkCalendarDate.propTypes;
```
k0nserv: This will not work with an optimizer that removes these in production. Instead of doing it this way declare `PropTypes` as constant, export that and assign it to `BpkCalednarDate.propTypes`

PR #960, opened 2018-07-25
[NO-JIRA] Fix modal scrim animation

georgegillams:
The animation for BpkScrim behind modals is not visible as the animation duration is so short. Increasing it makes the scrim fade instead of just appearing, making it less visually jarring (especially for mobile modals)

I've attached some videos below with the scrim turned red for clarity. Thoughts?

Before:
https://www.dropbox.com/s/utwkq6w5hn2fdj2/before.mov?dl=0

With no scrim:
﻿https://www.dropbox.com/s/vpopfkmb6d8a1eh/after.mov?dl=0

With a fading scrim:
﻿https://www.dropbox.com/s/n5iana9bh8xxs6m/after_2.mov?dl=0

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,5 +1,7 @@
 # Unreleased
 
 **Fixed:**
-- bpk-component-calendar:
-  - Fixed issue with DST transition during midnight in Safari.
+- bpk-component-modal:
+- bpk-component-dialog:
+- bpk-component-datepicker:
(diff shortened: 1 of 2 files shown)


shaundon:
I think this should have a changelog entry too, just so that consumers are aware. I know it isn't going to affect them code-wise, but they may notice that the animation feels different, and then they'll be able to see our changelog and know they didn't break anything.

PR #964, opened 2018-07-26
New Infinite Scroll DataSource

ShakMR:
This adds a DataSource class for infinite scroll, so we can trigger updates on the infinite list when data changes.


UNRELEASED.md
```diff
@@ -1,3 +1,5 @@
 # Unreleased
 
-_Nothing yet..._
\ No newline at end of file
+**Added:**
```
tiagoengel: This needs to be "Breaking". Also can you mention the `fetchFunction` prop was replaced in favour of the new data source approach?

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,57 @@
...
+  offDataChange: (callback: () => void) => void;
+
+  triggerListeners: (...args: Array<any>) => void;
+
+  constructor() {
+    this.listeners = [];
+  }
+
+  onDataChange(callback: () => void) {
+    // need to check if not present already
```
tiagoengel: can we remove this?

packages/bpk-component-infinite-scroll/src/withInfiniteScroll-test.js
```diff
@@ -42,68 +57,69 @@ describe('withInfiniteScroll', () => {
...
-  it('renders items after the first render', async () => {
+  it('renders items after the first render', done => {
     const tree = mount(
-      <InfiniteList onItemsFetch={onItemsFetch(elementsArray)} />,
+      <InfiniteList dataSource={new MyDS(onItemsFetch(elementsArray))} />,
     );
 
-    await intersect();
-    tree.update();
+    setTimeout(async () => {
```
ShakMR: this timeout is needed to allow the `setState` in the `componentWillMount` to reflect in the component.

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,57 @@
...
+
+  fetchItems: OnItemsFetchFunction;
+
+  onDataChange: (callback: () => void) => void;
+
+  offDataChange: (callback: () => void) => void;
+
+  triggerListeners: (...args: Array<any>) => void;
+
+  constructor() {
```
tiagoengel: I think we can make things easier and accept a "fetchFunction" as a constructor param here.

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,57 @@
...
+/* @flow */
+
+import { type OnItemsFetchFunction } from './onItemsFetch';
+
+class DataSource {
+  listeners: Array<() => void>;
+
+  fetchItems: OnItemsFetchFunction;
+
+  onDataChange: (callback: () => void) => void;
```
tiagoengel: Can we return a boolean here? true if added and false if not (when it's already present)

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,57 @@
...
+
+  offDataChange: (callback: () => void) => void;
```
tiagoengel: Same here. True if removed or false if not

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import { type OnItemsFetchFunction } from './onItemsFetch';
```
tiagoengel: Let's move the type definition to this file as this is now the "source" of this

packages/bpk-component-infinite-scroll/src/onItemsFetch.js
```diff
@@ -18,6 +18,11 @@
 
 /* @flow */
 
+export type OnItemsFetchFunction = (
```
tiagoengel: I think we should remove this function and replace it whit an `ArrayDataSource`, something like this. 

```
class ArrayDataSource extends DataSource {
   constructor(elementsArray) {
      this.elementsArray = elementsArray;
   }
   
   fetchItems(index, nElements) {
     if (n <= 0) {
       return resolve([]);	     
     }
     return resolve(elementsArray.slice(0, index + nElements));
  }
}
```

Can be defined in the same file you're defining the `DataSource` class

packages/bpk-component-infinite-scroll/src/withInfiniteScroll-test.js
```diff
@@ -42,68 +57,69 @@ describe('withInfiniteScroll', () => {
...
-      }
-      observe() {} // eslint-disable-line
-      unobserve() {} // eslint-disable-line
-    };
-  });
 
   it('renders an empty list for the first render', () => {
     const tree = renderer.create(
-      <InfiniteList onItemsFetch={onItemsFetch(elementsArray)} />,
+      <InfiniteList dataSource={new MyDS(onItemsFetch(elementsArray))} />,
```
tiagoengel: We can use the `ArrayDataSource` here

packages/bpk-component-infinite-scroll/src/withInfiniteScroll-test.js
```diff
@@ -42,68 +57,69 @@ describe('withInfiniteScroll', () => {
...
     const tree = mount(
-      <InfiniteList onItemsFetch={onItemsFetch(elementsArray)} />,
+      <InfiniteList dataSource={new MyDS(onItemsFetch(elementsArray))} />,
     );
 
-    await intersect();
-    tree.update();
-    await intersect();
-    tree.update();
+    setTimeout(async () => {
```
tiagoengel: Why is this necessary now? Also we can replace it with `setImmediate`

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,108 +114,111 @@ const withInfiniteScroll = (
...
-        this.observer.observe(this.sentinel);
-      }
+      this.props.dataSource.onDataChange(this.updateData);
     }
 
-    componentDidUpdate() {
-      if (this.sentinel && this.state.index > 0) {
-        this.observer.observe(this.sentinel);
-      }
+    componentWillMount() {
```
tiagoengel: We should use `componentDidMount`. `componentWillMount` has been removed from the new versions and it can be potentially called more than once.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,108 +114,111 @@ const withInfiniteScroll = (
...
+        this.setState(newState);
+      });
     }
 
-    componentWillUnMount() {
-      if (this.sentinel) {
-        this.observer.unobserve(this.sentinel);
-      }
-    }
+    updateData = () => {
```
tiagoengel: That was not how I was thinking about this tbh. I think `updateData` should receive all the new data from the `DataSource` and the callback here would only set the state

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,108 +114,111 @@ const withInfiniteScroll = (
...
-      this.observer = new IntersectionObserver(this.handleIntersection, {
-        threshold: 1,
-      });
-    }
-
-    componentDidMount() {
-      if (this.sentinel) {
-        this.observer.observe(this.sentinel);
-      }
+      this.props.dataSource.onDataChange(this.updateData);
```
tiagoengel: this needs to be done again in `componentWillReceiveProps` if the DataSource changes, and we need to call `offDataChange` in the old data source 

We also need to call `offDataChange` on `componentWillUnmount`

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,108 +114,111 @@ const withInfiniteScroll = (
...
+              ref={onRef}
               className={
                 renderLoadingComponent ? null : getClassNames('bpk-sentinel')
               }
             >
               {renderLoadingComponent && renderLoadingComponent()}
             </div>
           );
+          if (index > 0) {
+            LoadingComponent = withIntersectionObserver(LoadingComponent);
```
tiagoengel: This is not a very good idea performance wise. We are recreating this HOC at every render. Is there a problem with the previous implementation?

packages/bpk-component-infinite-scroll/stories.js
```diff
@@ -93,11 +105,43 @@ storiesOf('bpk-component-infinite-scroll', module)
...
+  .add('Force update data', () => {
+    class StoryDataSource extends DataSource {
+      constructor(elements) {
+        super();
+        this.fetchItems = onItemsFetch(elements);
+      }
+
+      updateElements(newElements) {
+        this.fetchItems = onItemsFetch(newElements);
+        this.triggerListeners();
```
tiagoengel: My idea here is that triggerListeners would be called with the `newElements` and so the components that have registered the lister will receive that new data and be able to rerender.

packages/bpk-component-infinite-scroll/stories.js
```diff
@@ -93,11 +105,43 @@ storiesOf('bpk-component-infinite-scroll', module)
...
+        dataSource={new MyDS(delayed)}
         elementsPerScroll={10}
         renderLoadingComponent={() => (
           <BpkSpinner type={SPINNER_TYPES.primary} />
         )}
       />
     );
+  })
+  .add('Force update data', () => {
+    class StoryDataSource extends DataSource {
```
tiagoengel: Can we move class definition outside this function? I think if we implement the `ArrayDataSource` we can add the `updateData` method there too.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll-test.js
ShakMR: trying to do the same. I didn't know about the `setImmediate`. Changing it.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
ShakMR: but in the dataset we don't have knowledge about current index and number of elements per scroll.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
ShakMR: isn't `componentWillReceiveProps` going to be deprecated?

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
ShakMR: not a problem just wanted to abstract it

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
tiagoengel: True. I'm thinking if it wouldn't be better to have this info in the datasource then. I.e. the component will only call `dataSource.fetchMore()` and then `elementsPerScroll` and `index` will be part of the dataSource state.

For example: `new ArrayDataSource(data, { pageSize: 10 })`.

What do you think?

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
tiagoengel: it will - we should probably use `componentDidUpdate` then, and also reset the whole scroll state if that happens as changing the datasource possibly means completely different data.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
ShakMR: what about 
```
      this.props.dataSource.onDataChange((updatedElements: Array<any>) => {
        const { index } = this.state;
        const { elementsPerScroll } = this.props;
        this.setState({
          elementsToRender: updatedElements.slice(
            index - elementsPerScroll,
            elementsPerScroll),
        });
      });
```
I prefer to keep the in the Infinite Scroll component instead of moving it to DataSource (to maintain it small/simple). However moving this to DataSource may allow us to reset the index as we want.

UNRELEASED.md
ShakMR: Changed

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: Done

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: Done

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: Done.

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: Done

packages/bpk-component-infinite-scroll/stories.js
ShakMR: done

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,108 +112,108 @@ const withInfiniteScroll = (
         showSeeMore: false,
       };
 
-      this.onItemsFetch = props.onItemsFetch;
-      this.observer = new IntersectionObserver(this.handleIntersection, {
-        threshold: 1,
+      this.props.dataSource.onDataChange((updatedElements: Array<any>) => {
+        const { index } = this.state;
```
tiagoengel: To be honest I think if we don't move this to the dataSource is better to keep the version you had before (calling fetch again) as this will give more flexibility in case someone is implementing a dataSource that only gets notified of new data but doesn't get the actual data and needs to fetch it again

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,87 @@
...
+  onDataChange: (callback: (updatedElements: Array<any>) => void) => boolean;
+
+  offDataChange: (callback: (updatedElements: Array<any>) => void) => boolean;
+
+  triggerListeners: (...args: Array<any>) => void;
+
+  constructor() {
+    this.listeners = [];
+  }
+
```
tiagoengel: can we have the fetchItems function defined here? The implementation can be just `throw new Error('Not implemented')`

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,51 +110,84 @@ const withInfiniteScroll = (
...
     }
 
-    componentDidUpdate() {
+    componentDidUpdate(_, prevState) {
       if (this.sentinel && this.state.index > 0) {
         this.observer.observe(this.sentinel);
       }
+      if (prevState.index !== 0 && this.state.index === 0) {
+        this.loadMore().then(newState => {
+          this.setState(newState);
```
tiagoengel: Since we are calling `setState` here anyway I think we should move all the `componentWillReceiveProps` here, we can change load more to accept the `index` and `elementsPerScroll` as argument.
```
if (prevState.dataSource !== this.state.dataSource) {
   prevState.dataSource.offDataChange(this.updateData);
   this.props.dataSource.onDataChange(this.updateData);
   this.loadMore(this.state.index, this.state.elementsPerScroll).then(newState => {
          this.setState({ ...newState, isListFinished: false });
   });
}

Comment (ShakMR) [reply packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js]: yeah, that was another option. Changing it 👍

File: packages/bpk-component-infinite-scroll/src/DataSource.js [hunk 20]
```diff
@@ -0,0 +1,87 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+class DataSource {
+  listeners: Array<() => void>;
```
k0nserv: The type of listeners should be `() => mixed` for maximum flexibility.

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,87 @@
...
+
+  onDataChange: (callback: (updatedElements: Array<any>) => void) => boolean;
+
+  offDataChange: (callback: (updatedElements: Array<any>) => void) => boolean;
```
k0nserv: Why are these two members listed twice? Both here and below with their implementations.

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,87 @@
...
+
+  onDataChange(callback: () => void): boolean {
+    if (this.listeners.indexOf(callback) === -1) {
+      this.listeners.push(callback);
+      return true;
+    }
+    return false;
+  }
+
+  offDataChange(callback: () => void): boolean {
```
k0nserv: Can we have a better name for this method? Just from reading this it's not clear at all when these callbacks get called

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,87 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+class DataSource {
```
k0nserv: Can we make this type generic over the type of the elements in the `DataSource`? i.e.

```javascript
class DataSource<T=any> {
  fetchItems(index: number, nElements: number): Promise<Array<T>> {
    throw new Error('Not implemented');
  }
}
```

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+    return false;
+  }
+
+  triggerListeners(...args: Array<any>) {
+    this.listeners.forEach(cb => cb(...args));
+  }
+}
+
+export class ArrayDataSource extends DataSource {
```
k0nserv: This should be `export class ArrayDataSource<T=any> extends DataSource<T>` with the generic treatment

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,51 +110,84 @@ const withInfiniteScroll = (
...
 
     componentDidMount() {
-      if (this.sentinel) {
-        this.observer.observe(this.sentinel);
+      this.loadMore().then(newState => {
+        this.setState(newState);
+      });
+    }
+
+    componentWillReceiveProps(newProps) {
```
k0nserv: We should avoid using `componentWillReceiveProps` if we can. Looks like we could use `getDerivedStateFromProps` instead here? If we do use `getDerivedStateFromProps` we should make sure that the peer depedency on `React` is correct

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: `addDataChangeListener` and `removeDataChangeListener`?

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: what's the difference between using `T` and `any`?

k0nserv: If we support a generic argument that defaults to `any` it will behave the same as you've written by default. However if the consumer explicitly specialises the type with e.g. `DataSource<number` Flow can help catch more errors in the code because it knows the type. For example on `ArrayDataSource<number>` the specialised type of the `Promise` will prevent incorrect usage of the return values

packages/bpk-component-infinite-scroll/src/DataSource.js
k0nserv: `onDataChange` and `removeDataChangeListener` maybe?

k0nserv: or just `removeListener`

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
ShakMR: change it. Working on the docs now

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: changed

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: removed this two

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: changed to `removeListener`

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: 👍 changed

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: done

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
ShakMR: moved to componentDidUpdate

packages/bpk-component-infinite-scroll/README.md
```diff
@@ -38,13 +38,11 @@ const elementsArray = [
 ]
 
 const InfiniteList = withInfiniteScroll(SomeList);
+const dataSource = new ArrayDataSource(elementsArray);
```
k0nserv: Now we can specialise here `ArrayDataSource<string>`

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,83 @@
...
+    return false;
+  }
+
+  triggerListeners(...args: Array<any>): void {
+    this.listeners.forEach(cb => cb(...args));
+  }
+}
+
+export class ArrayDataSource<T = any> extends DataSource<T> {
+  elements: Array<any>;
```
k0nserv: This should be `Array<T>` now

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,83 @@
...
+
+  constructor(elementsArray: Array<any>) {
```
k0nserv: `Array<T>`

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -0,0 +1,83 @@
...
+      const totalElements = elements.length;
+      const n = totalElements - index;
+      if (n <= 0) {
+        return resolve([]);
+      }
+      return resolve(elements.slice(0, index + nElements));
+    });
+  }
+
+  updateData(newElementsArray: Array<any>) {
```
k0nserv: This should take `Array<T>`

packages/bpk-component-infinite-scroll/README.md
ShakMR: I though we weren't using flow in the `README.md`. Added

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: done

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: done

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: done

jordigomezskyscanner:
@tiagoengel @k0nserv Please, could you take another look at this, please?

packages/bpk-component-infinite-scroll/README.md
k0nserv: Didn't realise this was in the README.md yeah we should probably remove it?

packages/bpk-component-infinite-scroll/README.md
```diff
@@ -73,36 +90,45 @@ const SomeList = ({elements}) => (
...
-| onItemsFetch           | func     | true     | -             |
-| elementsPerScroll      | number   | false    | 5             |
-| onScroll               | func     | false    | null          |
-| onScrollFinished       | func     | false    | null          |
-| renderLoadingComponent | func     | false    | null          |
-| renderSeeMoreComponent | func     | false    | null          |
-| seeMoreAfter           | number   | false    | null          |
+| Property               | PropType                | Required | Default Value |
+| ---------------------- | ----------------------- | -------- | ------------- |
+| dataSource             | instanceOf `DataSource` | true     | -             |
```
tiagoengel: we should use `instanceOf(DataSource)` here

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,51 +108,73 @@ const withInfiniteScroll = (
...
+        .fetchItems(index - elementsPerScroll, elementsPerScroll)
+        .then(updatedElements => {
+          this.setState({
+            elementsToRender: updatedElements,
+          });
+        });
+    };
+
+    loadMore(indexParam) {
+      const index = indexParam || this.state.index;
```
tiagoengel: This is not working as expected as `0` will be treated as false here and state will be used. If you test in the stories you'll see that changing stories is not reseting the datasource state. A possible solution is `indexParam != null ? indexPara : this.state.index`.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,51 +108,73 @@ const withInfiniteScroll = (
...
-    componentDidUpdate() {
+    componentDidUpdate(prevProps) {
       if (this.sentinel && this.state.index > 0) {
         this.observer.observe(this.sentinel);
       }
+      if (this.props.dataSource !== prevProps.dataSource) {
+        prevProps.dataSource.removeListener(this.updateData);
+        this.props.dataSource.onDataChange(this.updateData);
+        this.loadMore(0).then(newState => {
+          this.setState(newState);
```
tiagoengel: We should probably do `setState({ isListFinished: false, ...newState })` here in case the previous dataSource had reached the end of the list.

ShakMR: changed

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
ShakMR: true, changed

PR #965, opened 2018-07-30
Added scrollable calendar date component

boris-barath:
Implemented a scrollable calendar date component building on top of BpkCalendarDate

the component uses all the same props, but when isOutside is passed, it only returns an empty span tag in order to only display dates within a specific grid's month.


unreleased.md
```diff
@@ -0,0 +1,3 @@
+# Unreleased
+
+_Nothing yet..._
```
k0nserv: This file shouldn’t exists. The file is called `UNRELEASED.md` and that’s the one you want to change

packages/bpk-component-scrollable-calendar/stories.js
```diff
@@ -20,8 +20,16 @@
...
+import { BpkScrollableCalendarDate } from './index';
 
-storiesOf('bpk-component-scrollable-calendar', module).add('Default', () => (
-  <BpkScrollableCalendar />
-));
+storiesOf('bpk-component-scrollable-calendar', module).add(
+  'BpkScrollableCalendarDate',
+  () => (
+    <div>
+      <h3> Middle date is an empty span - outside of current month </h3>
```
k0nserv: This references the span which we removed

packages/bpk-component-calendar/src/BpkCalendarDate.js
```diff
@@ -155,6 +155,8 @@ BpkCalendarDate.propTypes = {
   preventKeyboardFocus: PropTypes.bool,
 };
 
+BpkCalendarDate.propTypes = propTypes;
```
emmajam: Out of interest, whats the advantage of having a separate `propTypes` object?  Why not follow the same pattern for defaultProps or vice versa?

boris-barath: Since this component uses all the same props as `BpkCalendarDate` I was advised to export those to avoid duplication as much as possible

k0nserv: @emmajam There are babel extensions that strip out `propTypes` and if you try to do `...Component.propTypes` that can cause issues with that. Having a dedicated object helps we've found.

UNRELEASED.md
```diff
@@ -1,2 +1,9 @@
 # Unreleased
 
+**New:**
+ - bpk-component-scrollable-calendar:
+   - implemented BpkScrollableCalendarDate, built on top of BpkCalendarDate
```
matthewdavidson: Whilst the scrollable calendar remains "private", there is no need to add changelog / unreleased entries :). So I'd remove:
```
**New:**
 - bpk-component-scrollable-calendar:
   - implemented BpkScrollableCalendarDate, built on top of BpkCalendarDate
```

packages/bpk-component-calendar/src/BpkCalendarDate.js
matthewdavidson: Great question, see this rule https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-foreign-prop-types.md 👍

packages/bpk-component-scrollable-calendar/stories.js
```diff
@@ -20,8 +20,14 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 
-import BpkScrollableCalendar from '.';
+import { BpkScrollableCalendarDate } from './index';
 
-storiesOf('bpk-component-scrollable-calendar', module).add('Default', () => (
-  <BpkScrollableCalendar />
```
matthewdavidson: Why remove the default story? This should remain unless I am missing something...

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarDate.js
```diff
@@ -0,0 +1,44 @@
...
+import PropTypes from 'prop-types';
+import React, { PureComponent } from 'react';
+import {
+  BpkCalendarDate,
+  BpkCalendarDatePropTypes,
+} from 'bpk-component-calendar';
+
+class BpkScrollableCalendarDate extends PureComponent {
+  render() {
+    const { date, isOutside, ...rest } = this.props;
```
matthewdavidson: No need to pull our `date` explicitly, it'll be passed through `{...rest}`, right?

packages/bpk-component-scrollable-calendar/index.js
```diff
@@ -17,9 +17,9 @@
  */
 /* @flow */
 
-import BpkScrollableCalendar, {
-  type Props as BpkScrollableCalendarProps,
-} from './src/BpkScrollableCalendar';
+import BpkScrollableCalendar from './src/BpkScrollableCalendar';
+import BpkScrollableCalendarDate from './src/BpkScrollableCalendarDate';
 
-export type { BpkScrollableCalendarProps };
```
matthewdavidson: Why remove the exporting of flow types for the main component?

packages/bpk-component-calendar/index.js
```diff
@@ -40,4 +42,5 @@ export {
   composeCalendar,
   withCalendarState,
   themeAttributes,
+  BpkCalendarDatePropTypes,
```
matthewdavidson: Usually we give component's prop type objects a name of `propTypes` (as you have in `BpkCalendarDate.js`). But we haven't yet come across exporting multiple prop type objects in larger components - let me get input from others in the squad before settling on that name (`BpkCalendarDatePropTypes `).

UNRELEASED.md
```diff
@@ -1,2 +1,9 @@
...
+
+**Changed:** 
+ - bpk-component-calendar:
+   - exposed propTypes as BpkCalendarPropTypes for reuse in BpkScrollableCalendarDate
```
matthewdavidson: This should say: "Exposed `BpkCalendarDatePropTypes` for reuse in other components."

^ pending a name change of course :P

packages/bpk-component-scrollable-calendar/stories.js
```diff
@@ -20,8 +20,14 @@
...
-));
+storiesOf('bpk-component-scrollable-calendar', module).add(
+  'BpkScrollableCalendarDate',
+  () => (
+    <div>
+      <h3> An example BpkScrollableCalendarDate component</h3>
```
matthewdavidson: No need to give this a heading, the story name should be a enough

packages/bpk-component-calendar/src/BpkCalendarDate.js
emmajam: Thanks guys ❤️  sorry for delayed response!

PR #967, opened 2018-07-31
Native iOS Panel

k0nserv:


native/packages/react-native-backpack/libraries/Panel/ios/react/classes/BPKPanelWrapper.m
```diff
@@ -0,0 +1,83 @@
...
+- (void)setup {
+    BPKPanel *panel = [[BPKPanel alloc] initWithFrame:self.bounds];
+    self.inner = panel;
+
+    [self addSubview:panel];
+    panel.translatesAutoresizingMaskIntoConstraints = NO;
+    [panel.layoutMarginsGuide.leadingAnchor constraintEqualToAnchor:self.leadingAnchor].active = YES;
+    [panel.layoutMarginsGuide.trailingAnchor constraintEqualToAnchor:self.trailingAnchor].active = YES;
+    [panel.layoutMarginsGuide.topAnchor constraintEqualToAnchor:self.topAnchor].active = YES;
+    [panel.layoutMarginsGuide.bottomAnchor constraintEqualToAnchor:self.bottomAnchor].active = YES;
```
shaundon: Is this the same as using `Ctrl + drag` to create constraints through Interface Builder?

native/packages/react-native-backpack/libraries/Panel/index.js
```diff
@@ -16,5 +16,30 @@
...
  */
 /* @flow */
+import React, { Component, type Node } from 'react';
+import PropTypes from 'prop-types';
+import { requireNativeComponent } from 'react-native';
 
-// Not much to see here
+type Props = {
+  children: Node,
+  padded: boolean,
```
shaundon: Is `padded` non-functional right now? I tried applying it in the Storybook and through Interface Builder and it doesn't seem to have an effect.

native/packages/react-native-backpack/libraries/Panel/ios/react/classes/BPKPanelWrapper.m
k0nserv: Yup more or less

native/packages/react-native-backpack/libraries/Panel/index.js
k0nserv: Yeah pretty much, the React Native part of this implementation doesn't really work.

k0nserv:
Very good point about the storybook. Will have to think of something

shaundon:
Other than the Storybook thing we discussed, I think this looks good, based on my still rudimentary iOS knowledge (although a week ago I'd have struggled to review this at all!). Only other things are:

* I think we should document somewhere (readme, comments in code, somewhere else) that this is currently non-functional. I know that **we** know, but if we're merging non-working code into `master` I think that's a special case to be noted for anybody else viewing this repo.

* `eslint` is broken, you just need to add some opt-outs.

k0nserv:
There's a big warning in the `README.md` already which states this is all WIP. Do you think that is enough or would you want to add something more? Will fix `eslint`

native/ios/native/ToggleNavigationBarMenuModule.h
```diff
@@ -0,0 +1,15 @@
+//
+//  ToggleNavigationBarMenuModule.h
+//  Backpack Native
+//
+//  Created by Hugo Tunius on 01/08/2018.
+//  Copyright © 2018 Facebook. All rights reserved.
+//
```
shaundon: 🙈 Is there something you aren't telling us? 😉

PR #973, opened 2018-08-01
[BPk-1776] RN bindings for native iOS badge

k0nserv:


native/packages/react-native-backpack/libraries/Badge/index.js
```diff
@@ -0,0 +1,66 @@
...
+import PropTypes from 'prop-types';
+import { requireNativeComponent } from 'react-native';
+
+type BadgeType =
+  | 'badgeTypeSuccess'
+  | 'badgeTypeWarning'
+  | 'badgeTypeDestructive'
+  | 'badgeTypeLight'
+  | 'badgeTypeInverse'
+  | 'badgeTypeOutline';
```
shaundon: Currently consumers can use `success`, `warning` etc, so this would be breaking, right? I'm thinking it may make more sense to keep the user-facing API the same and then convert `success => badgeTypeSuccess` internally.

Or alternatively, keep using `success` instead of `badgeTypeSuccess`. Is there a reason this changed? I'm assuming it's something to do with the Obj-C enum.

native/packages/react-native-backpack/libraries/Badge/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+BpkBadge.propTypes = {
+  message: PropTypes.string.isRequired,
+  type: PropTypes.oneOf([
+    'badgeTypeSuccess',
+    'badgeTypeWarning',
+    'badgeTypeDestructive',
+    'badgeTypeLight',
+    'badgeTypeInverse',
+    'badgeTypeOutline',
```
shaundon: There's a lot of repetition here that we could get rid of too.

native/packages/react-native-backpack/libraries/Badge/index.js
```diff
@@ -0,0 +1,66 @@
...
+  ]),
+};
+
+BpkBadge.defaultProps = {
+  type: BADGE_TYPES.warning,
+};
+
+BpkBadge.types = BADGE_TYPES;
```
shaundon: I don't really understand what this line is for - we're already exporting `BADGE_TYPES` from this file?

native/packages/react-native-backpack/libraries/Badge/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+const NativeBadge = requireNativeComponent('BPKBadgeWrapper', BpkBadge);
+
+export default BpkBadge;
```
shaundon: We should probably make this fit our convention of just using `index.js` to do exports and then do the heavy lifting in a `Badge.js` file.

native/packages/react-native-backpack/libraries/Badge/ios/react/classes/BPKBadgeWrapper.m
```diff
@@ -0,0 +1,104 @@
...
+
+#import "BPKBadgeWrapper.h"
+
+#import <Backpack2/Badge.h>
+
+#import <React/RCTUIManagerUtils.h>
+#import <React/RCTUIManager.h>
+#import <React/RCTShadowView.h>
+#import <React/RCTComponent.h>
+//#import <React/UIView+React.h>
```
shaundon: Should we remove this?

native/packages/react-native-backpack/libraries/Badge/ios/react/classes/BPKBadgeWrapper.m
```diff
@@ -0,0 +1,104 @@
...
+
+// Badge Type
+- (void)setType:(BPKBadgeType)type;
+
+- (void)setup;
+@end
+
+@implementation BPKBadgeWrapper
+
+- (instancetype)initWithBridge:(RCTBridge *)bridge {
```
shaundon: Is this the RN equivalent of `initWithFrame` and `initWithCoder`?

native/packages/react-native-backpack/libraries/Badge/ios/react/classes/BPKBadgeWrapper.m
```diff
@@ -0,0 +1,104 @@
...
+
+#pragma mark - Setters/Getters
+
+- (void)setMessage:(nullable NSString *)message {
+    self.inner.message = message;
+}
+
+- (nullable NSString *)message {
+    return self.inner.message;
+}
```
shaundon: Why does `message` have both getters and setters, but `type` only has a setter?

native/packages/react-native-backpack/stories.js
```diff
@@ -31,3 +31,7 @@ storiesOf('react-native-backpack/BPKPanel', module)
       <BpkText>Label</BpkText>
     </BpkPanel>
   ));
+
+storiesOf('react-native-backpack/BPKBadge', module)
+  .addDecorator(CenterDecorator)
+  .add('docs:default', () => <BpkBadge message="Success" />);
```
shaundon: How come we're running two separate sets of stories for this package?

native/packages/react-native-backpack/libraries/Badge/ios/react/classes/BPKBadgeWrapper.h
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <React/RCTView.h>
+
+NS_ASSUME_NONNULL_BEGIN
+@class RCTBridge;
+@interface BPKBadgeWrapper : RCTView
+- (instancetype)initWithBridge:(RCTBridge *)bridge NS_DESIGNATED_INITIALIZER;
```
shaundon: What is `NS_DESIGNATED_INITIALIZER`?

shaundon:
CI is failing because of Flow btw

native/packages/react-native-backpack/libraries/Badge/index.js
k0nserv: Will fix

native/packages/react-native-backpack/libraries/Badge/index.js
k0nserv: Nah I think this is fine for now, would be uneccessary complexity really

native/packages/react-native-backpack/libraries/Badge/ios/react/classes/BPKBadgeWrapper.m
k0nserv: No, we just need a reference to the bridge for this view

native/packages/react-native-backpack/libraries/Badge/ios/react/classes/BPKBadgeWrapper.m
k0nserv: I don't think the getter is ever called

native/packages/react-native-backpack/libraries/Badge/ios/react/classes/BPKBadgeWrapper.h
k0nserv: See https://stackoverflow.com/questions/26185239/ios-designated-initializers-using-ns-designated-initializer

PR #975, opened 2018-08-03
Added Scrollable calendar grid item

boris-barath:
Added the month component for scrollable calendar component. The idea is to display the month's name above every grid element in `BpkScrollableCalendar` for clarity.

The month name is a `BpkText` component. The missing days in comparison with `BpkCalendarGrid` are `null` instead

As with `BpkScrollableCalendarDate`, the props are the same as `BpkCalendarGrid`'s 

BpkScrollableCalendarGrid:
screen shot 2018-08-06 at 11 36 27

BpkCalendarGrid: 
screen shot 2018-08-06 at 11 36 19


packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGrid.js
```diff
@@ -0,0 +1,68 @@
...
+
+import STYLES from './bpk-scrollable-calendar-grid.scss';
+
+const getClassName = cssModules(STYLES);
+
+/*
+  BpkScrollableCalendarGrid - the grid representing a whole month
+*/
+
+class BpkScrollableCalendarGrid extends BpkCalendarGrid {
```
shaundon: This can just be a functional component as it doesn't use state.

Making it functional will negate this, but for future reference, React is more about composition than inheritance, so `extends React.Component` would be more appropriate here. See this for more context: https://reactjs.org/docs/composition-vs-inheritance.html

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGrid.js
```diff
@@ -0,0 +1,68 @@
...
+  BpkCalendarGridPropTypes,
+} from 'bpk-component-calendar';
+
+import STYLES from './bpk-scrollable-calendar-grid.scss';
+
+const getClassName = cssModules(STYLES);
+
+/*
+  BpkScrollableCalendarGrid - the grid representing a whole month
+*/
```
shaundon: We probably don't need this comment here - it's self documenting.

packages/bpk-component-scrollable-calendar/src/_variables.scss
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+$shadow-bottom: 0 1px 0 0 $bpk-color-gray-100; /* stylelint-disable-line unit-blacklist */
+$shadow-left: -1px 0 0 0 $bpk-color-gray-100; /* stylelint-disable-line unit-blacklist */
+$shadow-right: 1px 0 0 0 $bpk-color-gray-100; /* stylelint-disable-line unit-blacklist */
```
shaundon: Adding this many opt-outs is often a good indication that there's a better way to do this. We have `$bpk-one-pixel-rem` which is probably more suitable than `1px` here.

packages/bpk-component-scrollable-calendar/src/_variables.scss
```diff
@@ -0,0 +1,25 @@
+/*
```
shaundon: Are these variables being used anywhere? I can't see anywhere..

PR #977, opened 2018-08-07
Added Scrollable calendar grid item

boris-barath:
Added the month component for scrollable calendar component. The idea is to display the month's name above every grid element in `BpkScrollableCalendar` for clarity.

The month name is a `BpkText` component. The missing days in comparison with `BpkCalendarGrid` are `null` instead

As with `BpkScrollableCalendarDate`, the props are the same as `BpkCalendarGrid`'s 

BpkScrollableCalendarGrid:
screen shot 2018-08-06 at 11 36 27

BpkCalendarGrid: 
screen shot 2018-08-06 at 11 36 19


packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGrid.js
```diff
@@ -0,0 +1,62 @@
...
+} from 'bpk-component-calendar';
+
+import STYLES from './bpk-scrollable-calendar-grid.scss';
+
+const getClassName = cssModules(STYLES);
+
+const BpkScrollableCalendarGrid = props => {
+  const { month, className, ...rest } = props;
+
+  const classNames = [getClassName('bpk-scrollable-calendar-grid')];
```
shaundon: `classNames` isn't used anywhere - did you mean to apply it to the outermost div maybe?

Also - we have a nicer way to do this common class name logic now, this will produce the same effect:

```
const classNames = getClassName('bpk-scrollable-calendar-grid, className)
```

packages/bpk-component-scrollable-calendar/src/bpk-scrollable-calendar-grid.scss
```diff
@@ -0,0 +1,38 @@
...
+@import '~bpk-mixins/index';
+
+.bpk-scrollable-calendar-grid {
+  width: 100%;
+  border-collapse: separate;
+  border-spacing: 0;
+
+  &--animating {
+    z-index: 1;
+    transition: transform .2s ease-in-out;
```
shaundon: There's a token for this - `$bpk-duration-sm`.

packages/bpk-component-scrollable-calendar/src/bpk-scrollable-calendar-grid.scss
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  &__title {
+    margin: $bpk-spacing-md 0 $bpk-spacing-sm $bpk-spacing-sm;
+  }
+
+  &__month {
+    @include bpk-visually-hidden;
+  }
+}
```
shaundon: From what I can tell, `__animating` and `__month` aren't used, are they supposed to be in here?

Also, in the accompanying JS file a `__caption` class is used, but that isn't defined here.

packages/bpk-component-scrollable-calendar/stories.js
```diff
@@ -18,14 +18,30 @@
...
 
-import BpkScrollableCalendar, { BpkScrollableCalendarDate } from './index';
+import { weekDays, formatMonth, formatDateFull } from './test-utils';
+import BpkScrollableCalendar, {
+  BpkScrollableCalendarDate,
+  BpkScrollableCalendarGrid,
+} from './index';
 
 storiesOf('bpk-component-scrollable-calendar', module)
   .add('Default', () => <BpkScrollableCalendar />)
```
georgegillams: Could we remove this story as it is just the default example one included in the boilerplate?

georgegillams: Other than that, LGTM  🎉 

I'll leave @shaundon to review as he knows more about this than I do 👍

shaundon: I think it's ok to leave for now, under the assumption that it'll be fleshed out before this is released. If it's still there by the time we get close to 1.0.0 we can address it.

matthewdavidson: I actually asked Boris to keep this in.

Why would we remove it? We should always have a default story.

shaundon: I think we've just crossed wires - the story itself is fine, it's just that the component underneath that it displays just says "This is an example component".

When I said "_it's ok to leave for now, under the assumption that it'll be fleshed out before this is released_", I meant the component underneath.

So in summary I'm good with it!

matthewdavidson: Ah gotcha - yeah I mean this whole component is a work in progress so before it gets released we can tie up stuff like this.

PR #980, opened 2018-08-08
[BPK-1783] Fix test snapshot differences between timezones

georgegillams:
I think an `unreleased.md` entry is overkill as this should not affect consumers.

To test this out, you can run this command which will set the system timezone between test runs:
`npm run jest Calendar && sudo systemsetup -settimezone America/Campo_Grande && sleep 3 && npm run jest Calendar && sudo systemsetup -settimezone Australia/Adelaide && sleep 3 && npm run jest Calendar && sudo systemsetup -settimezone Europe/London`

And to run all tests:
`npm run test && sudo systemsetup -settimezone America/Campo_Grande && sleep 3 && npm run test && sudo systemsetup -settimezone Australia/Adelaide && sleep 3 && npm run test && sudo systemsetup -settimezone Europe/London`

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "jest:native:watch": "(cd ./native && jest --watch)",
     "jest:update": "jest --updateSnapshot",
     "jest:watch": "jest --watch",
-    "jest": "jest --coverage",
+    "jest": "TZ=Europe/London jest --coverage",
     "flow": "flow",
     "flow-typed": "flow-typed",
     "build:tokens": "lerna run tokens --scope bpk-tokens",


packages/bpk-component-calendar/src/BpkCalendarGrid-test.js
```diff
@@ -32,7 +32,7 @@ describe('BpkCalendarGrid', () => {
     const tree = renderer
       .create(
         <BpkCalendarGrid
-          month={new Date('2016-10')}
+          month={new Date(Date.UTC(2016, 9, 30, 3, 0, 0))}
```
matthewdavidson: We use a similar method elsewhere (see [composeCalendar-test.js](https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-calendar/src/composeCalendar-test.js#L44))wich doesn;t specify hours, mins, seconds or milliseconds ... How come you specify them here?

georgegillams:
@amolgupta has found this and suggested I take a look. Will do that now and update on here either way
https://stackoverflow.com/questions/40935886/components-using-date-objects-produce-different-snapshots-in-different-timezones/46980825#46980825

packages/bpk-component-calendar/src/BpkCalendarGrid-test.js
georgegillams: Good point. I'm not sure why, but if I don't specify them one of the tests fails in the Austrailian TZ with the following snapshot difference:
`"focusedDate":"2010-03-01T00:00:00.000Z"`
`"focusedDate":"2010-02-28T00:00:00.000Z"`

georgegillams:
In my original approach I am specifying the hour of the day (as allowing it to default to 0 results in some timezones producing `"focusedDate":"2010-03-01T00:00:00.000Z"` whilst others produce `"focusedDate":"2010-02-28T00:00:00.000Z"`).


This is also what is being done in the solution @amolgupta proposed.
https://stackoverflow.com/questions/40935886/components-using-date-objects-produce-different-snapshots-in-different-timezones/46980825#46980825
The second part of this solution seems to have no effect on the success of our tests in different timezones (they pass either way, as long as the hour of the day is specified).

georgegillams:
I think I may have finally got to the bottom of why using UTC only works sometimes.

When we use certain date helper functions, they disregard the fact that the date is UTC, and return a newly constructed Date instance.

See how calling `toISOString` returns the UTC timestamp, whereas the result of `addMonths` returns the Locale timestamp:
screen shot 2018-08-09 at 15 07 57

PR #982, opened 2018-08-08
[BPK-1792] Update infinite scroll to load only the required data

tiagoengel:


packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -70,7 +70,9 @@ export class ArrayDataSource<T = any> extends DataSource<T> {
       if (n <= 0) {
         return resolve([]);
       }
-      return resolve(elements.slice(0, index + nElements));
+      const start = Math.max(index, 0);
+      const end = Math.max(start, start + nElements);
+      return resolve(elements.slice(start, end));
```
ShakMR: we can may be remove the upper part given that slice will return until the last element of the array even if we ask for more:
`[1].slice(0, 10)` => [1]
`[1].slice(10,10)` => []

dominikkrejcik:
Seems fine to me 👍 that API makes much more sense than always returning the full list.

packages/bpk-component-infinite-scroll/src/DataSource.js
tiagoengel: How do you mean? I don't think `slice(10, 10)` will ever happen here, since we use `start + nElements` so if someone calls `fetchItems(10, 1)` it will be `slice(10, 11)`. And `fetchItems(10, 0)` doesn't really make sense. It should return empty in this case.

tiagoengel: Just to clarify, the `max` check is just to "protect" to code against someone providing a negative number to `nElements`.

packages/bpk-component-infinite-scroll/README.md
```diff
@@ -68,19 +66,48 @@ export default () => (
...
+
+`fetchItems(index, nElements)`
+
+Called by the `InfiniteScroll` component every time new data is
+requested (by scrolling down) and should return the data starting from `index` plus `nElements` (number of elements). It should return a promise object.
+
+Example:
+```js
+fetchItems(0, 5) // should return 5 items starting from position 0
+fetchItems(5, 5) // shouls return 5 items starting from position 5
```
shaundon: should say `should`, not `shouls` 😝

packages/bpk-component-infinite-scroll/README.md
```diff
@@ -90,37 +117,46 @@ const SomeList = ({elements}) => (
...
   />
 )
 ```
 
-ArrayDataSource
+## ArrayDataSource
 
 `ArrayDataSource` is a class that extends from `DataSource`. Accepts an array
 as a parameter in the constructor and uses it as source for the infinite scroll.
 To see and example of this class [Usage](#Usage) .
```
shaundon: This would be better as:

> See [Usage](#Usage) for an example of this class in use.

packages/bpk-component-infinite-scroll/src/DataSource-test.js
```diff
@@ -16,34 +16,81 @@
  * limitations under the License.
  */
 
-import DataSource from './DataSource';
+import DataSource, { ArrayDataSource } from './DataSource';
 
-describe('DataSource', () => {
+const withCommonTests = (createNewInstace, extraTests) => {
```
shaundon: `createNewInstace` should be `createNewInstance`

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: sorry, I didn't explain myself, by upper part I meant from line 68 to 72

tiagoengel: Oh you mean removing the if? I think you're right, on the other hand it's probably clearer/faster to do keep it there instead of "slicing" it when we don't really need to.

packages/bpk-component-infinite-scroll/README.md
tiagoengel: damm, how do you do this? 😅

shaundon: 'tis both a gift and a curse 😂

packages/bpk-component-infinite-scroll/src/DataSource.js
ShakMR: yes, that's what I meant

PR #984, opened 2018-08-09
Added Scrollable calendar grid list item

boris-barath:
This component is the actual 'grid' of the BpkScrollableCalendar. You have to pass it `minDate` and `maxDate` and all the usual `BpkCalendarGrid` components and it will build the scrollable list in your given range of dates. 

scrollable grid list


boris-barath:
idea: when the year changes, the corresponding January label should be something like *January - 2019* ?

shaundon:
> idea: when the year changes, the corresponding January label should be something like January - 2019 ?

Sounds like a good idea to me!

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -0,0 +1,91 @@
...
+const BpkScrollableCalendarGridList = props => {
+  const { className, minDate, maxDate, focusedDate, ...rest } = props;
+
+  const classNames = getClassName(
+    'bpk-scrollable-calendar-grid-list',
+    className,
+  );
+  if (className) {
+    classNames.push(className);
+  }
```
shaundon: You can remove this `if` statement, the `getClassName` part is already doing this.

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -0,0 +1,91 @@
...
+    startDate,
+    ...Array.from(
+      Array(DateUtils.differenceInCalendarMonths(endDate, startDate)).keys(),
+    ).map(i => DateUtils.addMonths(startDate, i + 1)),
+  ];
+
+  return (
+    <div className={classNames}>
+      <div className={getClassName('bpk-scrollable-calendar-grid-list__strip')}>
+        {months.map((m, index) => (
```
shaundon: Could you change `m` to `month`? It just makes it a little more descriptive that way (same as how `index` is more descriptive than `i`).

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -0,0 +1,91 @@
...
+  maxDate: PropTypes.instanceOf(Date).isRequired,
+  focusedDate: PropTypes.instanceOf(Date),
+  ...BpkCalendarGridPropTypes,
+};
+
+BpkScrollableCalendarGridList.defaultProps = {
+  className: null,
+  // Subtract one day from today's date to make today selectable by default
+  minDate: DateUtils.addDays(new Date(), -1),
+  maxDate: DateUtils.addMonths(new Date(), 12),
```
shaundon: `minDate` and `maxDate` are required props, so they shouldn't have entries in `defaultProps`.

packages/bpk-component-scrollable-calendar/src/_variables.scss
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+$shadow-bottom: 0 $bpk-one-pixel-rem 0 0 $bpk-color-gray-100;
+$shadow-left: -$bpk-one-pixel-rem 0 0 0 $bpk-color-gray-100;
+$shadow-right: $bpk-one-pixel-rem 0 0 0 $bpk-color-gray-100;
+$calendar-width: 7 * ($bpk-calendar-day-size + $bpk-calendar-day-spacing);
+$calendar-height: 7 * ($bpk-calendar-day-size + $bpk-calendar-day-spacing);
```
shaundon: `$calendar-height` is the only one of these that's in use - do the others need to be here?

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
boris-barath: @shaundon I made the props not required, as it makes more sense to have one year by default than defining specific dates, what do you think?

shaundon: I think it's better to make customers be explicit (so have them required), as we don't know how accurate an assumption of one year is. Plus, if we change it for some reason in the future it'll be a breaking change. Better to remove the worry altogether.

packages/bpk-component-scrollable-calendar/src/_variables.scss
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+$calendar-height: 7 * ($bpk-calendar-day-size + $bpk-calendar-day-spacing);
```
georgegillams: As this is now the only variable in this file I would just do away with it and put the value in `bpk-scrollable-calendar-grid-list.scss`

PR #986, opened 2018-08-09
[BPK-1683] stories consistency

matteo-hertel:
This is the first of 3 PRs, it focuses on 2 main things:
1. only the story that will be added to the docs site have the prefix `docs:` and I've capitalised some names here and there
2. fix some docs issues, components not appearing in the sidebar and redirects


native/packages/react-native-bpk-component-banner-alert/stories.js
```diff
@@ -166,7 +166,7 @@ class BpkBannerAlertFadeDemo extends Component<
 
 storiesOf('react-native-bpk-component-banner-alert', module)
   .addDecorator(CenterDecorator)
-  .add('docs:banner-alerts', () => (
+  .add('Banner alerts', () => (
```
matthewdavidson: It looks like you are removing the `docs:default` example here but on further inspection you are not ... perhaps moving the `docs:` stories to the top will help?

packages/bpk-docs/src/constants/redirect-routes.js
```diff
@@ -60,7 +60,8 @@ export default {
   '/components/native/badge': ROUTES.BADGE,
   '/components/native/banner-alerts': ROUTES.BANNER_ALERT,
   '/components/native/button': ROUTES.BUTTON,
-  '/components/native/button-link': ROUTES.BUTTON_LINK,
```
matthewdavidson: Whats happening here?

native/packages/react-native-bpk-component-banner-alert/stories.js
matteo-hertel: yes absolutely, good catch!

on a side note, why is always banner alert creating problems? 😞

packages/bpk-docs/src/constants/redirect-routes.js
matteo-hertel: `/components/native/button-link` goes to `/components/link`
same for 
'/components/button-link' as we call em just link across the board, the component name in RN is misleading

shaundon:
Checked through and I think it's good. The build failed but looks like it was terminated, so I've restarted it. Once it passes I'll ✅

PR #991, opened 2018-08-13
Finishing touches for scrollable calendar component

boris-barath:


georgegillams:
Hey hey, sorry for the confusion.
I think the `BpkScrollableCalendar` component should still accept and pass down the props to the component, much like https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-calendar/src/BpkCalendarContainer.js does.

Then in the stories you'll probably still want a wrapper component to handle the state so that the stories are interactive, similar to:
https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-calendar/stories.js#L48-L80

PR #998, opened 2018-08-15
[NO-JIRA] Update url-parse dependency to fix CVE-2018-3774 vulnerability

georgegillams:


native/package-lock.json
```diff
@@ -11821,8 +11821,7 @@
     "requires-port": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/requires-port/-/requires-port-1.0.0.tgz",
-      "integrity": "sha1-kl0mAdOaxIXgkc8NpcbmlNw9yv8=",
-      "dev": true
```
shaundon: The loss of this line makes me think that this should be in `devDependencies`.

PR #1000, opened 2018-08-15
[BPK-1683] removing detox

matteo-hertel:
Removing detox 💔


shaundon:
If you're de-detoxing, does that mean we're toxing?

amolgupta:
There are also dependencies to detox in `native/android` and a test case

PR #1002, opened 2018-08-15
[NO-JIRA] Fix docs contributing link

tiagoengel:


packages/bpk-docs/src/pages/ContributingPage/ContributingPage.js
```diff
@@ -66,7 +66,7 @@ const components = [
         <Paragraph>
           Check out{' '}
           <BpkLink
-            href="https://github.com/Skyscanner/backpack/blob/master/contributing.md"
+            href="https://github.com/Skyscanner/backpack/blob/master/CONTRIBUTING.md"
             blank
           >
             contributing.md
```
georgegillams: Should this be capitalized too?

tiagoengel: Don't think so, this is part of a bigger text and it would be weird to have it capitalized.

PR #1003, opened 2018-08-16
[BPK-1807] Remove native code

shaundon:
* Remove the view controllers for the badge and panel native code (so our Storybook is like it was before with no navigation stack)
* Remove iOS parts of `react-native-backpack` package.
* Update Podspec, storybook etc accordingly.

I tested this works by deleting the repo, re-cloning it and following the setup instructions and it's all good.


georgegillams:
When I tried running this (having cleaned and rebuilt), `npm run android` produced the following:

screen shot 2018-08-16 at 13 59 26

georgegillams:
iOS app is looking good though 👍

native/packages/react-native-backpack/README.md
```diff
@@ -1,5 +1,5 @@
 # react-native-backpack
 
-This package contains a library of native Backpack components for iOS and Android along with React Native bindings for these.
+This package contains a library of native Backpack components for Android along with React Native bindings.
```
georgegillams: Good one putting this back in 👍

PR #1010, opened 2018-08-21
[BPK-1840] remove native native from android

matteo-hertel:
PR 1/2 

This removes all the native natiive components from the app and resets the default activity to RN

the build and whatnot is all working and so does `npm run android `


amolgupta:
Unused variables in `values/dimens.xml` to be removed.

matteo-hertel:
@amolgupta I've removed the file as after linting the project none of those vars were used

PR #1019, opened 2018-08-22
[BPK-1751] Upgrade react-router

tiagoengel:
More info #1008 (I've mistakenly closed it).

I reverted back the URL changes to update `components/badge?platform=native` to `components/badge/native` because that was proving to be quite difficult to do right with SSR.

I have an idea on how to implement in a way that plays nicely with our SSR config and will create a new card with all details.

Make sure to test the production build `npm run docs:dist` and check everything is still working. Here is a checklist of things to test:
- All pages render correctly
- Redirects still work. e.g `/components/native/badge` goes to `/components/badge?platform=native`
- SSR works correctly.


packages/bpk-docs/src/index.js
```diff
@@ -17,67 +17,93 @@
...
-    // Explicit check for null here due to odd behaviour with react router's match function
-    // It passes undefined in cases where matches are not found.
-    // So we use their error object if it is truthy, otherwise we create our own.
-    if (error !== null) {
-      return callback(
-        error ||
-          new Error(`React Router failed to match ${JSON.stringify(location)}`),
-      );
+      if (route.routes) {
+        return all.concat(flattenRoutes(route.routes));
```
matteo-hertel: maybe spread here rather than concat, it'll make us like the cool kids

tiagoengel: Hmm, I actually prefer `concat` here than `[...all, ...flattenRoutes(route.routes)]` because I'll have to `spread` twice. I generally go with spread operator only when I'm creating a new literal array and want to merge another one in, e.g. `[1, 2, 3, ...extra]`

matteo-hertel: but it's sooo ES5 😄

tiagoengel: And it turns out concat is much faster in this case :). https://jsperf.com/concat-vs-spread1
screen shot 2018-08-23 at 10 52 01

PR #1023, opened 2018-08-24
[NO-JIRA] Fix docs on IE 11

tiagoengel:
Turns out this code was not even being used anymore (since we're still using the query string method) 🤦‍♂️

### DIFF
--- a/packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
+++ b/packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
@@ -82,20 +82,6 @@ PlatformNav.propTypes = {
   disableWebTab: PropTypes.bool.isRequired,
 };
 
-const getBasePath = url => {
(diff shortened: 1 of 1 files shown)


packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -106,7 +92,7 @@ const DocsPageWrapper = props => {
     history,
     location,
   } = props;
-  const path = getBasePath(match.url);
+  const path = match.url;
```
k0nserv: Isn't this confusing? If it's a path why are we setting it to a URL? path != url

tiagoengel: it's actually the path, e.g. `/components/badge` and not the URL 🤷‍♂️ . `match.url` is from react router so not sure we can do something about it.

PR #1028, opened 2018-08-28
[BPK-1862] Removes onTouch event handler from rect elem

kbremner:


kbremner:
Thanks @matthewdavidson, RE not completely removing onTouch, should have looked at it a bit more before submitting, sorry!

In terms of `barProps`, the consumer can already specify a custom bar chart bar component via the `BarComponent` prop, so they could apply the props to a custom component directly rather than passing through `BpkBarchart`, seems like supporting 2 methods of doing the same thing could be a bit confusing?

To support arbitrary event handlers being passed to `BpkBarchartBar`, we could apply the `{...rest}` to the touchable area rect, instead of the inner rect? Saying that, from #25, it's not clear why `{...rest}` was switched from the other rect to the inner one...

matthewdavidson:
@kbremner, no worries!!

> In terms of barProps, the consumer can already specify a custom bar chart bar component via the BarComponent prop, so they could apply the props to a custom component directly rather than passing through BpkBarchart, seems like supporting 2 methods of doing the same thing could be a bit confusing?

They can but that forces the consumer to reimplement the style/markup/api of `BpkBarchartBar`, that is unless we chose to export it directly for the consumer to import. Your second point about applying the spread to the touchable `rect` applies if we go down this route.

Hey @tomcorke, do you have any opinion on this? You submitted the onTouch event handler but I am assuming you never used it (as it would have never worked)!

Perhaps the quickest fix for you Kyle at this stage would be to just remove it altogether given that it would have never worked in the first place. We can follow up in a separate PR to address the possibility to add arbitrary props to the bars (if we even require it!).

kbremner:
@matthewdavidson I've seen `BpkBarchartBar` imported in one project, however looking at that now it looks like it probably shouldn't be using it currently, as it's not actually exported, as you said. (it's used to support adding a tooltip to a bar).

> Your second point about applying the spread to the touchable rect applies if we go down this route.

We would still need to spread the props passed to `BpkBarchartBar` by `BpkBarchart`'s `barProps` prop on to something. The event handlers are currently added to the touchable `rect`, so I was suggesting that we'd need to change which rect we currently spread `rest` to in order to support arbitrary event handlers being supplied for a bar.

I could remove it, or I've changed it to `onTouchEnd` and `onTouchStart` in this PR. Probably better, as you say, to just remove it till a use case comes around. Let me know what you think and I can update the PR

matthewdavidson:
I know I initially said to fix it, but after sleeping on it and thinking about it more this morning... the fact that it would have never worked means that removing it is the easiest course of action here so lets do that 👍

UNRELEASED.md
```diff
@@ -1,3 +1,5 @@
 # Unreleased
 
-_Nothing yet..._
+**Breaking:**
```
matthewdavidson: This is a fix as opposed to breaking. Yes we are removing an api but it never worked so it cant possibly break anyones code right?

matthewdavidson:
I'll merge once outstanding question is resolved :)

UNRELEASED.md
kbremner: Thinking was that we're removing the `onBarTouch` event which some consumers have provided, but as you've said before it would have no effect anyways, so will change to `fixed` :+1:

PR #1030, opened 2018-08-28
[BPK-1864] Revert #1009

matthewdavidson:
It breaks `bpk-component-banner-alert` and causes a breaking change the `bpk-stylesheets`.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,9 +1,17 @@
 # Unreleased
 
+**Breaking:**
+
+- bpk-component-router-link:
+  - upgraded react router from 2.x to 4.x
+
 **Fixed:**
-- bpk-react-utils
+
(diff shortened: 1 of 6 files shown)


UNRELEASED.md
```diff
@@ -1,9 +1,17 @@
 # Unreleased
 
+**Breaking:**
+
+- bpk-component-router-link:
+  - upgraded react router from 2.x to 4.x
```
shaundon: Shouldn't this line be removed or changed to reflect that we're downgrading it?

matthewdavidson: I haven't touched the router link component. I just moved it to the top by convention i.e. breaking changes first, then adds, then fixes.

shaundon: Ahh - gotcha

PR #1035, opened 2018-08-31
[NO-JIRA] Fix Data table sorting

georgegillams:
To see this issue, open [any DataTable in Storybook](https://backpack.github.io/storybook/?knob-Copy=Lorem%20ipsum%20dolor%20sit%20amet%2C%20consectetuer%20adipiscing%20elit.%20Aenean%20commodo%0Aligula%20eget%20dolor.%20Aenean%20massa.%20Cum%20sociis%20natoque%20penatibus%20et%20magnis%20dis%0Aparturient%20montes%2C%20nascetur%20ridiculus%20mus.&knob-Button%20copy=Toggle%20height%21&selectedKind=bpk-component-datatable&selectedStory=Autowidth%20Example&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) and click on any of the headers

Before:

After:

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,5 +1,8 @@
 # Unreleased
 
 **Fixed:**
+- bpk-component-data-table
+  - Fixed an issue where sorting columns would cause an exception.
+
 - bpk-component-barchart
   - Removes `onBarTouch` prop as it ultimately had no effect
--- a/packages/bpk-component-datatable/src/BpkDataTable.js
+++ b/packages/bpk-component-datatable/src/BpkDataTable.js
@@ -113,11 +113,14 @@ class BpkDataTable extends Component {
       return;
     }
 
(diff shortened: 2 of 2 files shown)


k0nserv:
What does this fix? The code looks equal in my eyes

georgegillams:
I agree the code looks equal. The value of event.target is undefined where it's passed through to `getSortIconDirection`.

Don't ask me why - if there's a better solution to fixing this then I'm all ears, but at the moment it's definitely broken. I can show you this change working within another project where it was reported if you like?

k0nserv:
Can we figure out why this behaves like this? Maybe it's a read only getter and the point at which you reference the value matters? I'd like to include a comment in code or in the commit specifying why it's written this way

georgegillams:
I'll add a comment to say that it's a hack - and reference a BPK Jira ticket to follow up

PR #1039, opened 2018-09-03
[No JIRA] Update flow

k0nserv:
Supersedes #1036


k0nserv:
Nah, I've read the changelog and I guess we don't have anything that hits those cases

PR #1017, opened 2018-08-22
Bpk 1851 update top package json dependencies

georgegillams:

PR #1043, opened 2018-09-06
[BPK-1911] Fix RN ios icons

matthewdavidson:
Fixes a regression to the iOS icon component introduced in https://github.com/Skyscanner/backpack/pull/1017


shaundon:
Just sanity checking what you've done here - you've only downgraded `gulp-iconfont` to the previous version we were using, right?

k0nserv:
Ddi we figure out what changed in `gulp-iconfont` that caused this change? Or are we just doing a fast rollback for now and investigating staying on the latest version further?

matthewdavidson:
@shaundon correct.

@k0nserv, yeah fast rollback for now. gulp-iconfont doesn't have a changelog (see https://github.com/nfroidure/gulp-iconfont/issues/169) so it's hard to figure it out just now. We'll revisit when greenkeeper prompts in the future.

PR #1047, opened 2018-09-10
[NOJIRA] tidy up lerna stuff

matthewdavidson:


lerna.json
```diff
@@ -1,10 +1,10 @@
 {
-  "lerna": "2.4.0",
```
matthewdavidson: This is redundant information now, they removed this from lerna.json (see https://github.com/lerna/lerna/pull/1225).

matthewdavidson:
So due to https://github.com/Skyscanner/backpack/pull/1029#issuecomment-419863856 I have reintroduced `package-lock.json`'s to the mix.

This begs the question .. "what about the issue with lerna reporting that all packages have changed since last release?"

well i have also ignored package-lock.json in lernas publish config so that they dont contribute to lerna determining if something has changes since last release.

screen shot 2018-09-10 at 14 16 15

lerna.json
```diff
@@ -1,13 +1,12 @@
...
   "version": "independent",
   "packages": ["packages/*", "native/packages/*"],
   "command": {
     "bootstrap": {
       "ignore": ["bpk-tether"]
     },
     "publish": {
-      "ignoreChanges": ["bpk-tether"],
+      "ignoreChanges": ["bpk-tether", "package-lock.json"],
```
shaundon: So, does this look at *all* `package-lock.json` files and not just the one in the root?

PR #1049, opened 2018-09-11
[BPK-1895] Remove contribution page from docs site

shaundon:


georgegillams:
I agree with @matteo-hertel - I think a 404 should suffice as we actually don't have anything to replace this

matthewdavidson:
I'm @matthewdavidson ... 😂

PR #1052, opened 2018-09-12
WIP: [BPK-1832] Add native submodules

tiagoengel:


CONTRIBUTING.md
```diff
@@ -322,6 +322,15 @@ When a component is released for the first time on npm, remember to add the comp
 
 </details>
 
+## Submodules
+
+`backpack-android` and `backpack-ios` folders are git submodules used solely for documentation. They should't be directly used for anything else.
```
shaundon: `should't` should be `shouldn't`

CONTRIBUTING.md
```diff
@@ -322,6 +322,15 @@ When a component is released for the first time on npm, remember to add the comp
...
+
+The documentation build will ensure the local submodules are up to date before using it so there is no need to do any git command directly. That being
+said, from time to time is good to update the submodules to point to a newer commit so less changes will be pulled before each doc build.
```
shaundon: `is` should be `it's`

shaundon: `less` should be `fewer`

.eslintignore
```diff
@@ -6,3 +6,5 @@ coverage
 /packages/bpk-component-icon/*/
 !/packages/bpk-component-icon/src
 /packages/bpk-tokens/tokens
+backpack-android
+backpack-ios
```
matthewdavidson: is there a difference between `/backpack-android` and `backpack-android` ?

Just pointing out the inconsistency with the ignores above the lines you'e added.

webpack.config.babel.js
```diff
@@ -189,6 +189,13 @@ const config = {
       index: 'index.html',
     },
   },
+
+  resolve: {
+    alias: {
+      'backpack-android': path.resolve(__dirname, 'backpack-android'),
+      'backpack-ios': path.resolve(__dirname, 'backpack-ios'),
```
matthewdavidson: To be clear, this change allows us to do the following yeah?
```
import nativeIosReadme from 'backpack-ios/Backpack/Panel/README.md';
```

matthewdavidson:
When adding an import like `import readme from 'backpack-ios/Backpack/Panel/README.md';` i get the following eslitn error:
```
[eslint] Unable to resolve path to module 'backpack-ios/Backpack/Panel/README.md'. (import/no-unresolved)
```

.eslintignore
tiagoengel: not sure, but it's not really inconsistent with the rest, everything except packages is not using the forward slash at the start 🤷‍♂️

webpack.config.babel.js
tiagoengel: Yeah, but I reverted it because of the eslint problem.

tiagoengel:
I've removed the alias for now @matthewdavidson. Making eslint work with it was proving to be quite annoying.

PR #1053, opened 2018-09-12
[nojira] Small tweaks to travis config

matthewdavidson:
see comments inline

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
 language: node_js
-node_js:
-  - lts/carbon
 env:
   global:
     - NPM_CONFIG_LOGLEVEL=warn
@@ -24,6 +22,7 @@ script:
 cache:
   directories:
     - node_modules
+    - native/node_modules
 branches:
   only:
     - master


.travis.yml
```diff
@@ -24,6 +22,7 @@ script:
 cache:
   directories:
     - node_modules
+    - native/node_modules
```
matthewdavidson: Something i noticed, there is no reason why we cant cache the native `node_modules` too

.travis.yml
```diff
@@ -1,6 +1,4 @@
 language: node_js
-node_js:
```
matthewdavidson: travis automatically picks up the version from `.nvmrc` - no need to over specify nd give ourselves more work. We do this in other repos too btw.

.travis.yml
shaundon: Yep good idea!

.travis.yml
matthewdavidson: See https://travis-ci.org/Skyscanner/backpack/builds/427659336#L432

PR #1056, opened 2018-09-13
[BPK-1833] Add Android and iOS tabs to platform switcher in docs

shaundon:
Preference order for the default platform to display goes:

1. Android
2. iOS
3. React Native
4. Web

We don't have any Android or iOS pages yet, but here's some screenshots to demonstrate how it would look with various states:

screen shot 2018-09-13 at 13 05 39
screen shot 2018-09-13 at 13 06 11
screen shot 2018-09-13 at 13 06 21
screen shot 2018-09-13 at 13 06 36

**What does this PR not contain?**
- The side nav filter is still just 'web' and 'native' right now (out of scope and we can still use 'native' filter for iOS and Android for the moment)
- I've not renamed `nativeSubpage` to `reactNativeSubpage` - YAGNI?

### DIFF
--- a/packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
+++ b/packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
@@ -23,9 +23,6 @@ import BpkContentContainer from 'bpk-component-content-container';
 import BpkHorizontalNav, {
   BpkHorizontalNavItem,
 } from 'bpk-component-horizontal-nav';
-import BpkSmallMobileIcon from 'bpk-component-icon/sm/mobile';
-import BpkSmallWindowIcon from 'bpk-component-icon/sm/window';
-import { withButtonAlignment } from 'bpk-component-icon';
 import { cssModules } from 'bpk-react-utils';
 
 import Heading from '../Heading';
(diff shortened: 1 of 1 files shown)


packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -34,106 +31,143 @@ import STYLES from './DocsPageWrapper.scss';
...
+    initiallyRenderedSubpage = androidSubpage;
+  } else if (iosSubpage) {
+    initiallySelectedPlatform = 'ios';
+    initiallyRenderedSubpage = iosSubpage;
+  } else if (nativeSubpage) {
+    initiallySelectedPlatform = 'native';
+    initiallyRenderedSubpage = nativeSubpage;
+  }
+
+  const platformQueryParamRegex = /platform=([a-zA-Z]+)/.exec(location.search);
```
k0nserv: Strictly speaking this is not a regex, it's a potential match. We could also create the regex once as a constant at the top of the file.

georgegillams:
```
I've not renamed nativeSubpage to reactNativeSubpage - YAGNI?
```

I agree this is probably YAGNI - except if we have a new person in Backpack / contributor from outside the squad then it could be quite confusing. Probably too high cost:benefit ratio but without investigating I don't know 🤷‍♀️

packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
shaundon: Good idea

shaundon: Have updated

packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -34,106 +31,146 @@ import STYLES from './DocsPageWrapper.scss';
...
+    initiallySelectedPlatform = 'native';
+    initiallyRenderedSubpage = nativeSubpage;
+  }
+
+  const platformQueryParamMatches = platformQueryParamRegex.exec(
+    location.search,
+  );
+  if (
+    platformQueryParamMatches &&
+    platformQueryParamMatches[1] &&
```
tiagoengel: Is this necessary? The group is not optional so if the regex matches it should be always there. Also instead of verifying if the platform name is valid here you could add it to the regex. _`/platform=(android|ios|native|web)/`; and make the regex case insensitive

packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -34,106 +31,146 @@ import STYLES from './DocsPageWrapper.scss';
...
+    platforms[platformQueryParamMatches[1]]
```
tiagoengel: There's a potential problem here as the regex is matching any upper or lower case string but the object has only lower case keys

packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -34,106 +31,146 @@ import STYLES from './DocsPageWrapper.scss';
...
+  ) {
+    const platformQueryParam = platformQueryParamMatches[1];
+    if (platforms[platformQueryParam]) {
```
tiagoengel: This and the conditions above will not be necessary if we put the platforms in the regex

packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
shaundon: Good idea, I'll do that

packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
shaundon: Good catch, I'll make it lowercase only.

PR #1063, opened 2018-09-17
[HNT-539]: Add Custom Image Component

RCiesielczuk:
In order to share the image cache between Native and React Native, we're looking at passing in a custom image component into `BpkImage` to accommodate that. This is done by passing in an `imageComponent` prop, which is `Animated.Image` by default.


native/packages/react-native-bpk-component-image/src/BpkImage-test.common.js
```diff
@@ -21,6 +21,7 @@ import { StyleSheet } from 'react-native';
 import renderer from 'react-test-renderer';
 
 import BpkImage from './BpkImage';
+import CustomImage from './CustomImage';
```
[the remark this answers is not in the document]
RCiesielczuk: Cheers for the review @k0nserv.

`FastImage` is animatable, so what we'll be doing for the Android side is that we'll wrap `FastImage` via [createAnimatedComponent](https://facebook.github.io/react-native/docs/animated#createanimatedcomponent) and pass that in as the `imageComponent` prop in the `SkyBpkImage` wrapper.

Now, if anyone is planning on using the `BpkImageWithLoading` component, they need to define it like in the documentation:

```js
const BpkImageWithLoading = withLoadingBehaviour(BpkImage);
```

and similarly pass in the `imageComponent` prop.

```js
<BpkImageWithLoading
  imageComponent={SkyBpkImage}
  source={this.image}
  style={this.style}
  rounded={this.props.rounded}
  alt={this.props.alt}
/>
```

The only issue that arrises in this instance, is that if the image passed into `BpkImageWithLoading` is **not** animatable, then that'll not work. Will need to find a good way of determining whether the passed in component is animatable or not.

native/packages/react-native-bpk-component-image/src/BpkImage.js
```diff
@@ -132,8 +136,16 @@ class BpkImage extends Component<Props> {
...
+    const {
+      style: userStyle,
+      inView,
+      rounded,
+      loaded,
+      imageComponent,
+      ...rest
+    } = this.props;
+
+    const ImageComponent = imageComponent;
```
k0nserv: This can be done inline in the destructing as 

```
imageComponent: ImageComponent
```

native/packages/react-native-bpk-component-image/stories.js
```diff
@@ -58,6 +59,7 @@ storiesOf('react-native-bpk-component-image', module)
   ))
   .add('With Loading Behaviour', () => (
     <BpkImageWithLoading
+      imageComponent={CustomImage}
```
georgegillams: I think this test should remain the same - this functionality is now tested on line 91.

native/packages/react-native-bpk-component-image/README.md
```diff
@@ -66,16 +66,50 @@ export default () => (
...
+| Property        | PropType                               | Required  | Default Value   |
+| --------------- | -------------------------------------- | --------- | --------------- |
+| alt             | string                                 | true      | -               |
+| source          | oneOf(number, object, arrayOf(object)) | true      | -               |
+| inView          | boolean                                | false     | true            |
+| loaded          | boolean                                | false     | true            |
+| onLoad          | function                               | false     | null            |
+| rounded         | boolean                                | false     | true            |
+| style           | style                                  | false     | null            |
+| imageComponent  | function                               | false     | Animated.Image  |
```
georgegillams: Can we put these in alphabetical order (required first, then non-required)

native/packages/react-native-bpk-component-image/README.md
```diff
@@ -66,16 +66,50 @@ export default () => (
...
+  image: {
+    width: '100%',
+    height: spacingLg * 10,
+  },
+});
+
+export default () => (
+  <View style={styles.container}>
+    <BpkImage
+      imageComponent={CustomImage}
```
georgegillams: We usually don't include optional props in the readme example.
If we do, it tends to be a separate example

PR #1068, opened 2018-09-19
[BPK-1839] Add api doc links for android and ios

tiagoengel:
This is ready the only thing missing is replacing the little icon before the link.


packages/bpk-docs/src/components/DocsPageBuilder/DocsPageBuilder.js
```diff
@@ -62,18 +62,11 @@ const markdownToHTML = (readmeString, headerPrefix = '') =>
...
-);
-
-toSassdocLink.propTypes = {
-  sassdocId: PropTypes.string.isRequired,
-  category: PropTypes.string.isRequired,
-};
+const toSassdocLink = createFromType('sass', { transparentBackground: true });
+const toIosdocLink = createFromType('ios', { transparentBackground: true });
+const toAndroiddocLink = createFromType('android', {
+  transparentBackground: true,
```
shaundon: It looks like `transparentBackground` is always true - maybe we should just remove that prop?

packages/bpk-docs/src/components/DocsPageBuilder/DocsPageBuilder.js
```diff
@@ -137,6 +130,20 @@ const ComponentExample = (component, registerPlatformSwitchingContent) => {
       })
     : null;
 
+  const androiddocLink = component.androiddocId
```
shaundon: I think these should be `androidDocLink` and `iosDocLink` (uppercase `D`). The sassdoc one is also wrong but I think it's ok to leave that as it would be a lot to change.

shaundon:
screen shot 2018-09-20 at 10 38 24

So this is how it looks atm. I think we should:

- Remove the sassdoc badge. I know you want a new badge to replace it, but if we merged this in its current state it'll look really weird, so I think it's better to not display any badge for iOS and Android for now.
- I think the wording can be improved. How about something like `Check out [Badge on Backpack's iOS docs] for more.`

PR #1069, opened 2018-09-21
[NO-JIRA] Adjust button to use min height for accessibility

emmajam:
Swap the hardcoded height value in the button component for minHeight instead. :)


emmajam:
synk sure ain't happy 😂

k0nserv:
Yeah the integration is borked 🤷‍♂️

k0nserv:
Thanks a bunch Emma

PR #1073, opened 2018-09-26
[NO-JIRA] Fix accessibiliy dynamic scaling issues

emmajam:
https://gojira.skyscanner.net/browse/BPK-1721
https://gojira.skyscanner.net/browse/BPK-1720
https://gojira.skyscanner.net/browse/BPK-1718
https://gojira.skyscanner.net/browse/BPK-1713
https://gojira.skyscanner.net/browse/BPK-1715
https://gojira.skyscanner.net/browse/BPK-1717

PR #1074, opened 2018-09-27
[NOJIRA] Update screenshots from master

matthewdavidson:
- badges
- button-link
- icon
- navigation-bar
- text-inputs

I noticed these differences in https://github.com/Skyscanner/backpack/pull/1073 but turned out these were on master too.


emmajam:
💯

PR #1075, opened 2018-10-01
Default autoComplete to off on BpkAutosuggest input

Andy-Rennie:


matthewdavidson:
Cheers @Andy-Rennie ! Can we have a test (snapshot is fine) and an entry in the UNRELEASED.md please 🙏

Andy-Rennie:
Added tests and updated UNRELEASED, noticed that snapshot was already there with autoComplete="off" maybe defaulting in tests for some reason? So added a test to check the value is defined and can be overridden

PR #1076, opened 2018-10-01
[NO-JIRA] Fix font scaling from all links to picker only

emmajam:
- remove font scaling from all button text links
- add it only for iOS picker menu button link


native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js
```diff
@@ -126,6 +126,7 @@ const BpkPickerMenu = (props: PickerMenuProps) => {
             return (
               <BpkPickerItem
                 {...item}
+                allowFontScaling={false}
```
matthewdavidson: I think this is in the wrong place, no? It should be on the button link component.

native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.ios.js
```diff
@@ -92,7 +92,11 @@ const BpkPickerMenu = (props: Props) => {
       </TouchableWithoutFeedback>
       <View style={styles.modal}>
         <View style={styles.modalHeader}>
-          <BpkButtonLink title={doneLabel} onPress={onClose} />
+          <BpkButtonLink
+            allowFontScaling={false}
```
matthewdavidson: Yup this is now in the correct place, however we now face the issue that BpkButtonLink doesn't pass through `allowFontScaling` to the underlying `BpkText` 😬

native/packages/react-native-bpk-component-button-link/src/BpkButtonLink.android.js
```diff
@@ -40,6 +40,7 @@ export type Props = {
   ...$Exact<CommonProps>,
   borderlessBackground: boolean,
   style: ?any,
+  allowFontScaling: boolean,
```
matthewdavidson: Actually, i should have thought about this further. Rather than exposing individual props to the underlying `BpkText`, we should expose a `textProps` prop that we spread on to the component:

```
<BpkButtonLink
  title={doneLabel}
  onPress={onClose}
  textProps={{
    allowFontScaling={false}
  }}
/>
```

This solves the use case and prevents us from having to make changes going forward. I.e. there are [lots more props ](https://facebook.github.io/react-native/docs/text.html#props)that we might want to pass.

PR #1081, opened 2018-10-02
[NO-JIRA] Remove flex from button view stye

emmajam:


emmajam:
@matthewdavidson I've tested:
- hotels direct booking
- shields up 
- returning user checkout 
- rails d booking

and all looks good from the app side ✅

matthewdavidson:
We have a problem with secondary large buttons:
secondary large buttons

matthewdavidson:
I have raised the following PR to fix the above issues https://github.com/Skyscanner/backpack/pull/1083/files

PR #1082, opened 2018-10-03
[BPK-1977] Update RN button screenshots from master

matthewdavidson:
Before I make changes to the RN button component I want to make sure the screenshots are up to date.

PR #1083, opened 2018-10-03
Bpk 1977 fix button flex woes

matthewdavidson:
~Blocked on #1082.~

I have removed all concept of height from the buttons. Layout is now determined by lineheight of text + padding.


native/packages/react-native-bpk-component-button/src/styles.js
```diff
@@ -28,79 +28,52 @@ import {
...
   colorPink600,
-  buttonHeight,
-  buttonPaddingVertical,
-  buttonPaddingHorizontal,
   buttonBorderWidth,
-  buttonLineHeightLarge,
-  lineHeightXs,
   borderRadiusPill,
   spacingSm,
+  spacingMd,
```
georgegillams: If we are no longer using the button token values, should they in fact be removed?

matthewdavidson: That would be a breaking change that I want to avoid just now.

matthewdavidson: The idea of component specific tokens were well intentioned. But in practice they end up making the code complex and in reality nobody uses them.

georgegillams:
Oh and `unreleased` wasn't updated

https://github.com/Skyscanner/backpack/pull/1083#issuecomment-426553445

matthewdavidson:
@georgegillams it was  -> https://github.com/Skyscanner/backpack/pull/1083/files#diff-1f9a24a2e063326eff13af43500f8b74

PR #1090, opened 2018-10-04
[No JIRA] Generate duplicate copy of icon font

k0nserv:
We need to generate a second copy of the icon font with a different font
name to allow two copies of the font to be loaded in the same app. The
reason we need two copies is one is for Backpack React Native and the other is
for Backpack iOS. Crucially the name of the font as recorded in the
`name` table must be different to allow UIKit to load the font twice.


georgegillams:
Just out of interest why rename this one instead of the one in the new ios codebase?

native/ios/native/Info.plist
```diff
@@ -41,6 +41,7 @@
 	<array>
 		<string>BpkIcon.ttf</string>
 		<string>BpkIcon.woff</string>
+		<string>BpkIconIOS.ttf</string>
```
matthewdavidson: We generate `eot`, `ttf` and `woff` but only add the `ttf` in the Xcode project files? For comparison, the existing `BpkIcon` is included as `ttf` and `woff` ...

packages/bpk-svgs/gulpfile.babel.js
```diff
@@ -295,28 +295,47 @@ gulp.task('icons-lg', ['icons-common'], () => {
...
+          timestamp: 1436442578, // A static timestamp to prevent changes showing up in git, backpack's first commit!
+        }),
+      );
+
+  const fontStream = generateFont();
+  const saveFonts = [
+    fontStream.pipe(clone()).pipe(gulp.dest('dist/font')),
+    generateFont('BpkIconIOS')
+      .pipe(clone())
+      .pipe(rename({ basename: 'BpkIconIOS' }))
```
matthewdavidson: Why is a rename required when we already pass in the name?

packages/bpk-svgs/gulpfile.babel.js
```diff
@@ -339,7 +358,7 @@ gulp.task('icons-font', ['icons-common'], () => {
     mappingStream.end();
   });
 
-  return merge(saveFont, saveMapping);
+  return merge.apply(null, [...saveFonts, saveMapping]);
```
matthewdavidson: Why not just `merge(...saveFonts, saveMapping)`?

packages/bpk-svgs/gulpfile.babel.js
k0nserv: Good point

packages/bpk-svgs/gulpfile.babel.js
k0nserv: I had doubts about that method, but never tried it

native/ios/native/Info.plist
k0nserv: Dunno, don't care

k0nserv:
> Just out of interest why rename this one instead of the one in the new ios codebase?

As the comment in the code states renaming is not enough. The value for index 1 in the font's name table is what needs to be changed. Fonts are binary files and modifying this after the font is created is substantially more difficult than just generating two fonts

PR #1105, opened 2018-10-11
FIX: float precision error in IE11

ShakMR:


packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -183,7 +183,7 @@ const withInfiniteScroll = (
     handleIntersection = (entries: Array<IntersectionObserverEntry>) => {
       const { onScroll } = this.props;
       const entry = entries[0];
-      if (entry.intersectionRatio >= 1) {
+      if (entry.isIntersecting) {
```
ShakMR: MS Edge (> v15) already support `isIntersecting`. MS Edge 15 usage is <0.05 Globally.

ShakMR: this is handled by the polyfill in IE11

matthewdavidson: FYI: We do support Edge, we also still support IE11

shaundon: For future reference, there's an internal page named "Website Browser Support" where you can find our current browser support - I won't link to it because this is public but you know where to look 😉 

Matt is right though - IE11 is the lowest we go.

ShakMR: In that page I saw that we don't support Edge 15, so could make this change.

PR #1107, opened 2018-10-12
[BPK-1987] update typeramp tokens

georgegillams:


packages/bpk-tokens/formatters/adjust-typography.js
```diff
@@ -0,0 +1,110 @@
...
+  '53': 6,
+  '56': 5,
+  '60': 4,
+  '65': 3,
+  '69': 2,
+  '79': 0,
+  '80': 0,
+};
+
+const correspondingFontSizeTokens = {
```
k0nserv: This seems very unfortuante, can't we get this from the tokens themselves somehow instead of duplicating the data here?

georgegillams: Yeah - that's part of the refactoring that is needed ;)

packages/bpk-tokens/formatters/adjust-typography.js
```diff
@@ -0,0 +1,127 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import androidTokenAliases from '../src/android/base/aliases.json';
+import iosTokenAliases from '../src/ios/base/aliases.json';
+
+const sfProTracking = {
```
k0nserv: Can we include a comment about the source of this data?

packages/bpk-tokens/formatters/adjust-typography.js
```diff
@@ -0,0 +1,127 @@
...
+
+const getCorrespondingFontSizeTokensAndroid = letterSpacingKey =>
+  getCorrespondingFontSizeTokens(androidTokenAliases, letterSpacingKey);
+
+const getCorrespondingFontSizeTokensIos = letterSpacingKey =>
+  getCorrespondingFontSizeTokens(iosTokenAliases, letterSpacingKey);
+
+const adjustTypographyAndroid = prop => {
+  let adjustedValue = prop.value;
+  if (prop.type === 'letter-spacing' || prop.category === 'letter-spacings') {
```
k0nserv: Why do we need to check both category and type?

packages/bpk-tokens/formatters/adjust-typography.js
```diff
@@ -0,0 +1,127 @@
...
+    const correspondingFontSize = getCorrespondingFontSizeTokensIos(prop.name);
+    if (correspondingFontSize === undefined) {
+      throw new Error(
+        `A suitable adjustment for token ${
+          prop.name
+        } could not be found as no corresponding font-size exists`,
+      );
+    }
+    const sfProTrackingForFont =
```
k0nserv: Why are we using string keys as opposed to a `Map<Number, Number>`?

packages/bpk-tokens/src/android/base/typography.json
```diff
@@ -13,16 +11,30 @@
...
+      "meta": {
+        "deprecated": true
+      }
     },
     "FONT_SIZE_SM": {
       "value": "{!FONT_SIZE_SM}",
       "type": "font-size",
       "category": "typesettings"
     },
+    "FONT_SIZE_MD": {
```
k0nserv: I thought we agreed to map these to the old styles and adding `caps` and `xxxl`. i.e.

+ _ -> caps **New**
+ s -> xs
+ m -> s
+ base -> base
+ l -> l
+ xl -> xl
+ xxl -> xxl
+ _ -> xxxl **new**

k0nserv: I am not expeciting `MD` to be introduced here

packages/bpk-tokens/formatters/adjust-typography.js
```diff
@@ -0,0 +1,127 @@
...
+  '19': -26,
+  '20': 19,
+  '21': 17,
+  '22': 16,
+  '24': 15,
+  '25': 14,
+  '27': 13,
+  '30': 12,
+  '33': 11,
+  '36': 11,
```
k0nserv: Where is this value from? It's not in the table on https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/

packages/bpk-tokens/gulpfile.babel.js
```diff
@@ -24,32 +25,42 @@ import gulpTheo from 'gulp-theo';
...
+    { format: 'IOS_react.native.es6.js', nest: true },
+    { format: 'IOS_react.native.common.js', nest: true },
   ],
   android: [
     'android.xml',
     'raw.android.json',
     'react.native.android.js',
-    { format: 'react.native.es6.js', nest: true },
-    { format: 'react.native.common.js', nest: true },
+    { format: 'ANDROID_react.native.es6.js', nest: true },
```
k0nserv: Can you explain these two changes a bit? I.e. the prefixes

packages/bpk-tokens/formatters/adjust-typography.js
georgegillams: We don't 🙈

packages/bpk-tokens/src/android/base/typography.json
georgegillams: I remember something like this coming up in discussions. So was the consensus that we should overrule Will's naming and provide consumers with xs,s,base,lg,xl,xxl,etc or did we decide that this mapping should exist internally so that consumers have the new names exposed to them?

k0nserv: We decided where possible we will map to the old names and only introduce CAPS and xxxl as new sizes. cc @willhg

packages/bpk-tokens/gulpfile.babel.js
georgegillams: Basically, I needed a way to register different variants of the `react.native.es6.js` and `react.native.common.js` formatters to `gulp-theo`. The only way I could see to do this was to add something to the identifiers to differentiate them.

I know this solution ain't pretty, but theo seems pretty closed in the way it invokes the formatters. If there's a better way I'm all ears 👍

packages/bpk-tokens/src/android/base/typography.json
georgegillams: Cool 👍

packages/bpk-tokens/formatters/adjust-typography.js
```diff
@@ -0,0 +1,128 @@
...
+  19: -26,
+  20: 19,
+  21: 17,
+  22: 16,
+  24: 15,
+  25: 14,
+  27: 13,
+  30: 12,
+  33: 11,
+  36: 11,
```
k0nserv: Can we explain how we have picked these values that aren't listed on https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography#font-usage-and-tracking?

packages/bpk-tokens/formatters/adjust-typography.js
```diff
@@ -0,0 +1,128 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import androidTokenAliases from '../src/android/base/aliases.json';
+import iosTokenAliases from '../src/ios/base/aliases.json';
+
+// These are the default letter tracking values used by SF Pro and SF Pro Display.
+// They can be found at https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography#font-usage-and-tracking
```
k0nserv: Worth also mentioning the unit here which is 1/1000em

packages/bpk-tokens/formatters/adjust-typography.js
```diff
@@ -0,0 +1,128 @@
...
+    if (sfProTrackingForFont === undefined) {
+      throw new Error(
+        `A suitable adjustment for token ${
+          prop.name
+        } could not be found as no corresponding SF Pro tracking value exists`,
+      );
+    }
+    adjustedValue =
+      parseFloat(prop.value) -
+      (sfProTrackingForFont * correspondingFontSize) / 1000;
```
k0nserv: Can we make the intent clearer here by changing it to 

```
      parseFloat(prop.value) -
      ((sfProTrackingForFont * correspondingFontSize) / 1000);
```

I know it's the same with order of operations, but this makes it clear that it's intentional

packages/bpk-tokens/formatters/adjust-typography.js
georgegillams: NGL I made it up based on the closest values 🙈

packages/bpk-tokens/formatters/adjust-typography.js
georgegillams: We can, but the linter will complain!

georgegillams: We could do
```const adjustment = (sfProTrackingForFont * correspondingFontSize) / 1000;
adjustedValue =  parseFloat(prop.value) - adjustment;```

File: packages/bpk-tokens/src/android/base/typography.json (hunk 5 above)
Comment (georgegillams) [reply packages/bpk-tokens/src/android/base/typography.json]: spoken to @willhg about this and he's happy to crack on with caps, xs, s, base, ...

File: packages/bpk-tokens/formatters/adjust-typography.js (hunk 10 above)
Comment (k0nserv) [reply packages/bpk-tokens/formatters/adjust-typography.js]: Do that

File: packages/bpk-tokens/package.json [hunk 11]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bpk-tokens",
-  "version": "27.1.0",
+  "version": "27.1.0-beta.4",
```
k0nserv: This should be reverted back to `27.1.0`

PR #1114, opened 2018-10-15
[BPK-1754] Upgrade React Native to 0.57.3

shaundon:
* Upgraded RN based on [this guide](https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#updating-to-this-version). 
* Changed Storybook command based on [this issue](https://github.com/storybooks/storybook/issues/4179).

**Notes**

We got Jest to run with the following things:

* Change `.babelrc` to `babel.config.js` (weird I know).
* Add `@babel/core` and `babel-core` and `babel-jest` to deps, based on the RN upgrade guide.
* **Added Jest as a dependency in `native`. This isn't desirable long term as it diverges from the version of Jest used in the root. However I think this is better *for now* to keep this PR small. After this I plan to do another PR to upgrade Jest for web, then I can remove Jest from `native`.**
* Upgrade React to 16.6. I wanted to avoid this but it's unavoidable, the tests won't work without it. It's specified in the upgrade guide here: https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#0573.
* I had to add `jest-cli` to deps to get rid of `environment.dispose is not a function` errors. How is this even a thing I really don't know


native/.babelrc
```diff
@@ -1,3 +1,3 @@
 {
-  "presets": ["react-native"]
+  "presets": ["module:metro-react-native-babel-preset"]
```
shaundon: As mandated here: https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#updating-to-this-version

native/ios/Podfile.lock
```diff
@@ -2,52 +2,52 @@ PODS:
   - boost-for-react-native (1.63.0)
```
shaundon: This update is because I did `pod update` to bump RN. I presume the other changes are ok though, seems to work fine.

native/package.json
```diff
@@ -11,20 +11,21 @@
     "android:build": "react-native run-android",
     "android": "npm run android:emulator & npm run android:build",
     "test": "jest --coverage",
-    "storybook": "storybook start -p 7007 --root ../"
+    "storybook": "storybook start -p 7007 --skip-packager | react-native start --watchFolders $PWD/.."
```
shaundon: Storybook used to run the RN packager for us but their usage doesn't work with the new RN (see https://github.com/storybooks/storybook/issues/4179).

So, now we run the packager ourselves. `$PWD/..` is because Watchman won't accept a relative path (i.e. `..`), so this was my ~hack~ workaround.

native/package.json
```diff
@@ -11,20 +11,21 @@
...
-    "react-dom": "16.3.1",
-    "react-native": "0.55.3",
-    "react-test-renderer": "16.3.1",
+    "@storybook/react-native": "^3.4.11",
+    "babel-core": "7.0.0-bridge.0",
+    "metro-react-native-babel-preset": "^0.45.0",
+    "react": "16.5.1",
+    "react-dom": "16.5.1",
+    "react-native": "0.57.1",
+    "react-test-renderer": "16.5.1",
```
shaundon: * Updated Storybook to match the version in the root `package.json`. We should keep these in sync.
* Updated React as mandated in RN upgrade guide: https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#updating-to-this-version

jmikulka:
@shaundon just by comparing our `package.json` and yours, I can see few differences:
* you need both `"@babel/core": "^7.0.0"` and `"babel-core": "7.0.0-bridge.0"`
* `"metro-react-native-babel-preset": "^0.45.6"`
* `"jest": "23.6.0"`

In jest config, we have
```
    "transform": {
      "^.+\\.jsx?$": "<rootDir>/node_modules/react-native/jest/preprocessor.js"
    },
```

`.babelrc`:
```
{
  "plugins": ["codegen"]
}
```

`skyscanner-app/react-native/scripts/metro.config.js`
```
const path = require('path');

module.exports = {
  transformer: {
    babelTransformerPath: require.resolve('./transformer'), // NOTE: we use react-native-obfuscating-transformer
  },
  projectRoot: path.resolve(__dirname, '..'),
};
```

shaundon:
https://github.com/facebook/react-native/issues/21150#issuecomment-429305215

shaundon:
https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#0573

matthewdavidson:
Closing since we have moved the react-native code to https://github.com/Skyscanner/backpack-react-native

PR #1115, opened 2018-10-15
[No-JIRA] Add tracking docs for android

amolgupta:
Adds tracking docs for Android


decisions/tracking.md
```diff
@@ -0,0 +1,28 @@
+# Support for Tracking
+
+Type-ramp system has following values for tracking
```
k0nserv: How about "The type-ramp uses the following tracking values for tracking:"?

decisions/tracking.md
```diff
@@ -0,0 +1,28 @@
...
+
+font size | letter Spacing
```
k0nserv: We should mention that this is the letter spacing as defined in Sketch. We could also calcualted the em tracking which is what Android uses and add it as a column. Tools like lillustrator/photoshop use thousands of em which can also be calculated.

decisions/tracking.md
```diff
@@ -0,0 +1,28 @@
+# Support for Tracking
+
+The type-ramp system uses the following values for tracking:
```
shaundon: Two words 'type-ramp' please 😇

decisions/tracking.md
```diff
@@ -0,0 +1,28 @@
...
+30	| -0.8|-0.02666
+36	| -1.0|-0.02778
+42	| -1.2|-0.02857
+52  | -1.4|-0.02692
+62  | -1.6|-0.02580
+74  | -1.8|-0.02432
+
+## Android
+
+For android the value for `letterSpacing` is calculated as
```
shaundon: Capitalised Android 😇

shaundon:
Otherwise it's looking 💯

PR #1123, opened 2018-10-24
[BPK-2019] Add new icons

tiagoengel:
screen shot 2018-10-24 at 11 45 34


UNRELEASED.md
```diff
@@ -1,3 +1,13 @@
...
+
+- bpk-svgs
+  - New icons added:
+  - bws--call-back
+  - bws--end-call
+  - bws--keypad
+  - bws--mute
+  - bws--phone-call
+  - bws--send-message
+  - bws--un-mute
```
shaundon: This affects quite a lot of packages so we usually list those too in the changelog - see here for an example: https://github.com/Skyscanner/backpack/blob/master/CHANGELOG.md#2018-07-12---new-icons-and-better-close-link

tiagoengel: done 👍

shaundon:
Just want to confirm that `bws--send-message` is correct as the grey area doesn't fit our usual icon styles, then I'm happy to ✅

tiagoengel:
The icon was wrong @shaundon, good catch! I've updated it.

PR #1124, opened 2018-10-24
[NO-JIRA] Add package lock version check to script

georgegillams:
Added checks to the bpk dependency version checking script, to now also check `package-lock.json` versions match the `package.json`.
If they don't, an error will be displayed for each mismatching one.
If the `--fix` flag is used, they will be updated to the latest version automatically 👍

### DIFF
--- a/scripts/npm/check-bpk-dependencies.js
+++ b/scripts/npm/check-bpk-dependencies.js
@@ -21,7 +21,8 @@
 const fs = require('fs');
 const { execSync } = require('child_process');
 
-const errors = [];
+const dependencyErrors = [];
+const packageLockVersionErrors = [];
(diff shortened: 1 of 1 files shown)


scripts/npm/check-bpk-dependencies.js
```diff
@@ -48,7 +49,7 @@ const findReplace = (file, findReplaces) => {
 
 const fixDependencyErrors = packageFiles => {
   const findReplaces = [];
-  errors.forEach(error => {
+  dependencyErrors.forEach(error => {
```
matteo-hertel: forEach and then pushing to an empty array, why don't we use `.map`?

PR #1122, opened 2018-10-24
Added initial element number to infinite scroll

boris-barath:
Added the possibility of specifying how many elements to load initially.

PR #1125, opened 2018-10-24
Scrollable calendar - replace react-virtualized with bpk-infinite-scroll

boris-barath:
Replacing react-virtualized with bpk-component-infinite-scroll since virtualized is a very large dependency.

This PR needs #1122


packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -39,8 +41,7 @@ class BpkScrollableCalendarGridList extends React.Component {
   constructor(props) {
     super(props);
 
-    this.rowRenderer = this.rowRenderer.bind(this);
-    this.renderList = this.renderList.bind(this);
+    this.ref = React.createRef();
```
k0nserv: This conflicts with the `peerDependencies` of the package(this API is only available in 16.3.0 and will thus fail for `^15.0.0` which should be supported per the `peerDependencies`

boris-barath:
Re-opened as #1141 because my rebase broke it all.

PR #1139, opened 2018-10-30
[BPK-ABORT] let sf pro display do its own thing

georgegillams:
Due to issues with the look of SF Pro Display when custom tracking is applied, we have decided for now to drop the custom tracking and let the font dictate this.

There were a number of options for how to achieve this:
 - Remove the relevant letter-spacing token values
 - Override them in the typography adjustment phase (make them undefined)


packages/bpk-tokens/tokens/ios/base.react.native.es6.js
```diff
@@ -150,27 +150,27 @@ export const textCapsLetterSpacing = 0.28;
...
 export const textXsLineHeight = 14;
 export const textXxlFontSize = 36;
 export const textXxlFontWeight = "400";
-export const textXxlLetterSpacing = -1.396;
+export const textXxlLetterSpacing = NaN;
 export const textXxlLineHeight = 43;
 export const textXxxlFontSize = 42;
 export const textXxxlFontWeight = "400";
-export const textXxxlLetterSpacing = -1.6199999999999999;
+export const textXxxlLetterSpacing = NaN;
```
shaundon: We probably shouldn't have `NaN` in our tokens. Can we have `null` for these?

shaundon:
Snapshots need updating too.

packages/bpk-tokens/tokens/base.ios.json
```diff
@@ -932,7 +932,7 @@
       "name": "textLgFontWeight"
     },
     {
-      "value": -0.78,
+      "value": "null",
```
k0nserv: This makes me sad, can we make these actual `null` values rather than strings?

PR #1142, opened 2018-11-01
[BPK-2010] Add validation props to RN BpkSelect

georgegillams:


UNRELEASED.md
```diff
@@ -1,3 +1,6 @@
 # Unreleased
 
-_Nothing yet!___
+**Added:**
+
+- react-native-bpk-component-select:
+  - Added validation options consistent with those for text inputs.
```
shaundon: Can we make this a bit more explicit? I like the note about consistency but it'd be more helpful to users if we specifically stated which props are new. No harm in popping a link to the page on the docs site too.

native/packages/react-native-bpk-component-select/src/BpkSelect.js
```diff
@@ -72,18 +81,30 @@ const TouchablePlatformComponent = Platform.select({
 });
 
 type Props = {
-  onPress: () => mixed,
   disabled: boolean,
   label: ?(string | Element),
+  onPress: () => mixed,
```
shaundon: Tiny thing but can `onPress` go above `label`? For Flow types we do non-nullable ones first, nullable ones afterwards.

PR #1147, opened 2018-11-02
[NO-JIRA] Update components that are beastly under the new type scale

georgegillams:


UNRELEASED.md
```diff
@@ -1,3 +1,8 @@
 # Unreleased
 
-_Nothing yet!___
+**Fixed:**
+- react-native-bpk-component-flat-list:
+- react-native-bpk-component-navigation-bar:
+- react-native-bpk-component-phone-input:
+- react-native-bpk-component-section-list:
+  - Following the new type scale release, we decided that some Backpack components (iOS only) should use a different text style.
```
shaundon: I think we should be more specific here - remember that this is user-facing documentation, so should make sure it tells them everything they need to know about changes.

How about something like:

```
- react-native-bpk-component-flat-list:
- react-native-bpk-component-phone-input:
- react-native-bpk-component-section-list:
  - Changed table views to use `base` text size, as the new type scale (see below) made these slightly too large.

- react-native-bpk-component-navigation-bar:
  - Changed all text to use `base` text size, as the new type scale (see below) made these slightly too large.
```

georgegillams: Good shout!
See commit `discussion_r230658119`

PR #1150, opened 2018-11-06
Added "onDataFetchFailed" prop to "withInfiniteScroll" HOC

sogame:
We want to show a "Refresh results" button at the bottom of infinite scroll (like the "see more" one) when the request to fetch more items fails.

For this reason, we have added the `onDataFetchFailed` prop, to be able to render on our side this new button.

### DIFF
--- a/packages/bpk-component-infinite-scroll/README.md
+++ b/packages/bpk-component-infinite-scroll/README.md
@@ -167,6 +167,7 @@ Updates the internal array and triggers all listeners.
 | dataSource              | instanceOf(DataSource) | true     | -             |
 | elementsPerScroll       | number                 | false    | 5             |
 | initiallyLoadedElements | number                 | false    | 5             |
+| onDataFetchFailed       | func                   | false    | null          |
 | onScroll                | func                   | false    | null          |
 | onScrollFinished        | func                   | false    | null          |
 | renderLoadingComponent  | func                   | false    | null          |
--- a/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
+++ b/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
@@ -43,6 +43,7 @@ export type Props = {
   initiallyLoadedElements: number,
   elementsPerScroll: number,
   dataSource: DataSource<any>,
+  onDataFetchFailed: ?(event: SyntheticEvent<>) => mixed,
   onScroll: ?(o: ScrollEvent) => void,
   onScrollFinished: ?(o: ScrollFinishedEvent) => void,
   renderLoadingComponent: ?() => Element<any>,
@@ -67,6 +68,7 @@ const propTypes = {
   initiallyLoadedElements: PropTypes.number,
   elementsPerScroll: PropTypes.number,
   dataSource: PropTypes.instanceOf(DataSource).isRequired,
+  onDataFetchFailed: PropTypes.func,
   onScroll: PropTypes.func,
   onScrollFinished: PropTypes.func,
   renderLoadingComponent: PropTypes.func,
@@ -77,6 +79,7 @@ const propTypes = {
 const defaultProps = {
   initiallyLoadedElements: 5,
   elementsPerScroll: 5,
+  onDataFetchFailed: null,
   onScroll: null,
   onScrollFinished: null,
   renderLoadingComponent: null,
@@ -121,9 +124,11 @@ const withInfiniteScroll = (
     componentDidMount() {
       this.fetchItems({
         elementsPerScroll: this.props.initiallyLoadedElements,
-      }).then(newState => {
-        this.setState(newState);
-      });
+      })
+        .then(newState => {
+          this.setState(newState);
+        })
+        .catch(this.handleFetchItemsFailed);
     }
 
     componentDidUpdate(prevProps) {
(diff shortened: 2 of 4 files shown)


packages/bpk-component-infinite-scroll/README.md
```diff
@@ -172,3 +172,4 @@ Updates the internal array and triggers all listeners.
 | renderLoadingComponent  | func                   | false    | null          |
 | renderSeeMoreComponent  | func                   | false    | null          |
 | seeMoreAfter            | number                 | false    | null          |
+| onDataFetchFailed       | func                   | false    | null          |
```
shaundon: Could you move this up to be in alphabetical order?

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -121,9 +124,13 @@ const withInfiniteScroll = (
...
       this.fetchItems({
         elementsPerScroll: this.props.initiallyLoadedElements,
-      }).then(newState => {
-        this.setState(newState);
-      });
+      })
+        .then(newState => {
+          this.setState(newState);
+        })
+        .catch(() => {
```
shaundon:
```suggestion
        .catch(this.handleFetchItemsFailed);
```
Bit shorter and has the same effect.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -138,32 +145,56 @@ const withInfiniteScroll = (
...
-          this.setState({
-            ...newState,
+        })
+          .then(newState => {
+            this.setState({
+              ...newState,
+            });
+          })
+          .catch(() => {
+            this.handleFetchItemsFailed();
```
shaundon:
```suggestion
        .catch(this.handleFetchItemsFailed);
```
Bit shorter and has the same effect.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -138,32 +145,56 @@ const withInfiniteScroll = (
...
+      const { dataSource } = this.props;
+      dataSource
+        .fetchItems(0, index)
+        .then(updatedElements => {
+          this.setState({
+            elementsToRender: updatedElements,
+          });
+        })
+        .catch(() => {
+          this.handleFetchItemsFailed();
```
shaundon:
```suggestion
        .catch(this.handleFetchItemsFailed);
```
Bit shorter and has the same effect.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -182,7 +213,6 @@ const withInfiniteScroll = (
               index: nextIndex,
               elementsToRender: (elementsToRender || []).concat(nextElements),
               showSeeMore: seeMoreAfter === index / elementsPerScroll,
-              isListFinished: false,
```
shaundon: I don't really understand why this was removed, could you explain it?

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -193,30 +223,42 @@ const withInfiniteScroll = (
...
         }
-        return this.fetchItems().then(newState => {
-          this.setState(newState);
-        });
+        return this.fetchItems()
+          .then(newState => {
+            this.setState(newState);
+          })
+          .catch(() => {
+            this.handleFetchItemsFailed();
```
shaundon:
```suggestion
        .catch(this.handleFetchItemsFailed);
```
Bit shorter and has the same effect.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -193,30 +223,42 @@ const withInfiniteScroll = (
...
     handleSeeMoreClick = () => {
-      this.fetchItems().then(newState => {
-        this.setState(newState);
-      });
+      this.fetchItems()
+        .then(newState => {
+          this.setState(newState);
+        })
+        .catch(() => {
+          this.handleFetchItemsFailed();
```
shaundon:
```suggestion
        .catch(this.handleFetchItemsFailed);
```
Bit shorter and has the same effect.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
sogame: Because this is not needed: it's already initialised to `false`, and it's only changed to `true` when the results finish or when the request fails, so there's no need to set it to `false` again every time more results are fetch.

Not a big deal, but makes the code slightly cleaner :-P

shaundon: Sounds good to me, thanks for explaining.

shaundon:
Only other thing is it would be good to have a test and storybook entry of this new prop too, if possible.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
tiagoengel: I don't think this can be removed. If the datasource changes and the previous list has been finished `isListFinished` would have been set to false.

tiagoengel:
Also, I'd like to suggest a different alternative. All the fetching logic is handled by the DataSource class being used so we can do something like this.

```javascript
class MyDSWithFailure extends ArrayDataSource {
  constructor(data, onFailed) {
    super(data);
    this.onFailed = onFailed;
  }

  fetchItems(index, nElements) {
    return super.fetchItems(index, nElements).catch(this.onFailed);
  }
}

....

<InfiniteList
      dataSource={new MyDSWithFailure(elementsArray, () => {
        this.setSate({ showRefreshButton: true })
      })}
      onScrollFinished={action('scroll finished')}
      onScroll={action('onScroll')}
    />
```

sogame:
Hi @tiagoengel ,

You're totally right, your proposal makes total sense :-)

I started with a similar approach but found some issue (I don't remember what now) and then I switched to the one in this pr, but it can totally be done the way you're proposed, many thanks!!

I'll close this PR, sorry for the noise :-)

PR #1154, opened 2018-11-08
[NoJira][Backpack love] Update to node 10

matteo-hertel:
The change it's easier than it looks 😅 

I've done a `npm run clean` beforehand that's why the package lock changes


matteo-hertel:
Closing until node-sass plays well with node10 💔

PR #1161, opened 2018-11-14
[BPK-2040] Update send-message icon

tiagoengel:


UNRELEASED.md
```diff
@@ -1,3 +1,5 @@
 # Unreleased
 
-_Nothing yet!_
+**Changed:**
```
k0nserv: This should be Added right?

PR #1162, opened 2018-11-14
TER-140 Adding `tabIndex` property to bpk-popover component.

anakincho:
Our squad noticed a small bug where the `Close` button was not being focused by default once a popover was opened if other links existed within the container body. After investigating and communicating with the backpack team, they advised me to just make the change and add a contribution. I have followed the contribution guides, updated the docs and added a test (also played around with the stories to see if it actually works). 

The change itself allows users to set the tabIndex (as a prop to the `<BpkPopover>` React component) if they need to, or by default it will be set to "1" and is not a mandatory prop.

### DIFF
--- a/packages/bpk-component-popover/README.md
+++ b/packages/bpk-component-popover/README.md
@@ -53,6 +53,7 @@ class App extends Component {
           renderTarget={() =>
             document.getElementById('popover-container')
           }
+          tabIndex="0"
         >
           <BpkText>My popover content</BpkText>
         </BpkPopover>
@@ -84,6 +85,7 @@ class App extends Component {
 | portalClassName       | string                                    | false    | null          |
 | renderTarget          | func                                      | false    | null          |
 | popperModifiers       | object                                    | false    | null          |
+| tabIndex              | string                                    | false    | "1"           |
 
 In order to attach the popover to a regular DOM element, provide a function which returns it to `target`:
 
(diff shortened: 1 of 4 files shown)


packages/bpk-component-popover/README.md
```diff
@@ -84,6 +85,7 @@ class App extends Component {
 | portalClassName       | string                                    | false    | null          |
 | renderTarget          | func                                      | false    | null          |
 | popperModifiers       | object                                    | false    | null          |
+| tabIndex              | string                                    | false    | "1"           |
```
georgegillams: This should have type `number`

georgegillams: Also I think this name could be interpreted as the tabIndex for the popover itself. Could we rename it to something like `closeButtonTabIndex`?

matthewdavidson: I actually think we should avoid cherry picking specific props and instead add a `closeButtonProps` prop to allow the consumer to spread anything on that button. use case: adding an event for analytics purposes...

georgegillams:
@anakincho if any of the comments above don't make sense feel free to reach out to us 👍

georgegillams:
@anakincho I'm going to close this PR as it's had no activity for a while. We are going to implement this functionality in another PR. We'll let you know when it's released 👍

PR #1169, opened 2018-11-21
[BPK-2078] Add documentation for Android and iOS dialog component

shaundon:
I've maintained the status quo of the folder structure for now:

```
WebDialogPage
  WebDialogPage.js
  index.js
NativeDialogPage
  NativeDialogPage.js
  index.js
AndroidDialogPage
  AndroidDialogPage.js
  index.js
IOSDialogPage
  IOSDialogPage.js
  index.js
DialogPage
  DialogPage.js <-- Imports the other pages and exports a DocsPageWrapper 
  index.js
```

But I think in the future we should migrate **all** pages to be like this:

```
DialogPage
  WebDialogPage.js
  NativeDialogPage.js
  AndroidDialogPage.js
  IOSDialogPage.js
  DialogPage.js <-- Imports the other pages and exports a DocsPageWrapper 
  index.js
```

This would reduce complexity and visual noise in the filesystem a lot. But we should do it all at once in one PR.

**Android**
0 0 0 0_8080_components_dialog_platform android

**iOS**
0 0 0 0_8080_components_dialog_platform ios

**Note**
The link in the Android readme that says 'check the main Readme' is broken as it's a relative link. I'm going to fix that now in the Android repo and I'll update the submodule later. I don't think it needs to block this PR though.

### DIFF
--- a/packages/bpk-docs/src/components/DocsPageBuilder/ComponentScreenshots.js
+++ b/packages/bpk-docs/src/components/DocsPageBuilder/ComponentScreenshots.js
@@ -49,7 +49,9 @@ const ComponentScreenshot = ({ title, subText, ...rest }) => {
       className={getClassName('bpkdocs-component-screenshots__item')}
     >
       <dt className={getClassName('bpkdocs-component-screenshots__item-title')}>
-        <BpkText tagName="h3">{title}</BpkText>
+        <BpkText tagName="h3" textStyle="xs">
+          {title}
+        </BpkText>
       </dt>
       <dd
         className={getClassName(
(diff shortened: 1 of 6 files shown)


packages/bpk-docs/src/components/DocsPageBuilder/ComponentScreenshots.js
```diff
@@ -49,7 +49,9 @@ const ComponentScreenshot = ({ title, subText, ...rest }) => {
       className={getClassName('bpkdocs-component-screenshots__item')}
     >
       <dt className={getClassName('bpkdocs-component-screenshots__item-title')}>
-        <BpkText tagName="h3">{title}</BpkText>
+        <BpkText tagName="h3" textStyle="xs">
```
shaundon: Agreed with Will to do this as the blurb was using the same text style as the screenshot title and it looked weird in this PR.

tiagoengel:
LGTM. I'll pull and test. Just one thing, the submodules update was not supposed to happened but I just realised even though we are ignoring it (not showing as changed locally) it is still added when we run `git add .`  

Not something we should look into in this PR though.

shaundon:
Ah yeah I did wonder why it didn't appear locally but ended up in the PR.

PR #1170, opened 2018-11-21
[NO-JIRA] Move deprecated prop function to react-utils

georgegillams:


native/packages/react-native-bpk-component-text/src/common-types.js
```diff
@@ -88,6 +89,7 @@ export const weightPropType = (
   componentName: string,
 ) => {
   if (!isWeightValid(props[propName], props.textStyle)) {
+    // eslint-disable-next-line no-console
```
georgegillams: I don't know how CI was working without this before 🤷‍♂️

UNRELEASED.md
```diff
@@ -1 +1,5 @@
 # Unreleased
+
+**Added:**
+ - bpk-react-utils:
+   - Added `depreciated` function that accepts a prop type and adds a suitable console warning when the prop is used.
```
```suggestion
   - Added `deprecated` function that accepts a prop type and adds a suitable console warning when the prop is used.
```

_Depreciation_ is when an asset loses value over time :trollface:

native/packages/react-native-bpk-component-text/src/common-types.js
shaundon: It only gives an eslint warning, not an error, I think. Nice one for spotting it though!

UNRELEASED.md
georgegillams: I think in my previous life I must have been a banker, and I still have some old muscle memory 🤦‍♂️

PR #1173, opened 2018-11-22
[BPK-2074] Add accessory view to Badge

georgegillams:


willhg:
LGTM

native/packages/react-native-bpk-component-badge/src/BpkBadge-test.common.js
```diff
@@ -19,14 +19,26 @@
...
 
+const style = StyleSheet.create({
+  iconDestructive: {
+    color: colorWhite,
+  },
+  iconOutline: {
+    color: colorWhite,
+  },
+});
+
```
shaundon: These styles will be unnecessary once we change it to automatically colour the accessory view.

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -38,6 +38,9 @@ import { View, ViewPropTypes, StyleSheet } from 'react-native';
 
 const styles = StyleSheet.create({
   viewBase: {
+    display: 'flex',
```
shaundon: Isn't `display: 'flex'` the default? You may be able to remove this line and it'll still work - worth a go!

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -97,6 +100,9 @@ const styles = StyleSheet.create({
     borderBottomLeftRadius: borderRadiusSm,
     borderBottomRightRadius: 0,
   },
+  accessoryView: {
+    marginRight: spacingSm,
```
shaundon: `marginEnd` is better - works with RTL then.

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -157,6 +164,9 @@ const BpkBadge = (props: Props) => {
 
   return (
     <View style={viewStyle}>
+      {accessoryView && (
+        <View style={styles.accessoryView}>{accessoryView}</View>
+      )}
```
shaundon: This wrapper view is only being used to add the margin style. Could we avoid an extra view by conditionally applying the margin to the text instead, when the accessory view is supplied?

native/packages/react-native-bpk-component-badge/stories.js
```diff
@@ -64,13 +67,22 @@ const style = StyleSheet.create({
   inverse: {
     backgroundColor: colorGray700,
   },
+  iconDestructive: {
+    color: colorWhite,
+  },
+  iconOutline: {
+    color: colorWhite,
+  },
```
shaundon: These styles will be unnecessary once we change it to automatically colour the accessory view.

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
shaundon: I spoke to @willhg and we agreed that users shouldn't have to (or indeed be able to) colour the accessory view themselves, so the badge component should do this for them. 

We can do this by cloning the children and colouring any `Text` or `BpkIcon` components to match the badge's text colour.

Here's some examples of where we've iterated over or cloned children before that you could base it off:

1. [Cloning images passed to flat list items to add a margin](https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-flat-list/src/BpkFlatListItem.android.js#L74)
1. [Iterating over children passed into pickers to grab some data](https://github.com/Skyscanner/backpack/blob/master/native/packages/react-native-bpk-component-picker/src/BpkPickerMenu.android.js#L76)

packages/bpk-docs/src/pages/NativeBadgePage/NativeBadgePage.js
```diff
@@ -51,6 +53,28 @@ const components = [
       },
     ],
   },
+  {
+    id: 'withAccessoryView',
```
shaundon: Convention for IDs is kebab case: `with-accessory-view`.

shaundon:
Oh just noticed, `lodash` isn't in the dependencies of this package. Although I think you'll end up removing the lodash-using parts anyway, so.. 🤷‍♂️

packages/bpk-docs/src/pages/NativeBadgePage/NativeBadgePage.js
shaundon: The link in `unreleased.md` will need to change accordingly too.

georgegillams: Great catch for `UNRELEASED.md`

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
georgegillams: Strangely it seemed to work correctly with `marginRight`, but I'll use RTL friendly ones just to be sure

screenshot 2018-11-22 at 17 37 29

georgegillams:
I've addressed your feedback now @shaundon 

screenshot 2018-11-23 at 07 12 21
screenshot 2018-11-23 at 07 12 04

Within the badge component itself, I've applied styling to the `accessoryView` as suggested.
This works only for 1 `accessoryView` component, however - and I'm not sure how we should go about adjusting multiple `accessoryView` components.
 - I thought about recursively iterating through the `accessoryView`'s children and apply the styling to each descendant - but this seems like quite hacky behaviour, and will probably be surprising to consumers.
 - Alternatively I could change the API to accept an `accessoryViews` array, and apply the styling to each element in there

Let me know what you think is best

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -97,6 +100,15 @@ const styles = StyleSheet.create({
...
   },
+  accessoryView: {
+    marginEnd: spacingSm,
+  },
+  accessoryViewDestructive: {
+    color: colorWhite,
+  },
+  accessoryViewOutline: {
+    color: colorWhite,
+  },
```
shaundon: Can't we reuse `textBase`, `textDestructive` and `textOutline` styles for the accessory view?

UNRELEASED.md
```diff
@@ -1,5 +1,10 @@
 # Unreleased
 
+**Added:**
+
+ - react-native-bpk-component-badge:
+     - Badges now accept an optional `accessoryView` that is displayed on the leading side, and can contain any content including icons. See https://backpack.github.io/components/badge/?platform=native#with-multiple-icons
+
```
shaundon: This'll need to change as the props have changed too.

native/packages/react-native-bpk-component-badge/README.md
```diff
@@ -44,8 +45,10 @@ export default class App extends Component {
 ```
 ## Props
 
-| Property              | PropType                                                                 | Required | Default Value |
+| Property              | PropType                                                                  | Required | Default Value |
 | --------------------- | --------------------------------------------------------------------------| -------- | ------------- |
 | message               | string                                                                    | true     | -             |
 | docked                | oneOf('start', 'end')                                                     | false    | null          |
+| iconSeparator         | string                                                                    | false    | null          |
```
shaundon: Let's make sure this is well documented. I was thinking we could have a heading underneath this table, something like:

```
## iconSeparator

`iconSeparator` is an optional string to place between icons when there are more than one, for example when you want to place a "+" between the icons.
```

native/packages/react-native-bpk-component-badge/README.md
```diff
@@ -44,8 +45,10 @@ export default class App extends Component {
...
+| icons                 | arrayOf(string)                                                           | false    | []            |
```
shaundon: Can we make this more clear that it needs to be an icon string specifically? `arrayOf(BpkIcon.icons)` or something?

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -166,15 +194,19 @@ const BpkBadge = (props: Props) => {
...
   docked: PropTypes.oneOf(Object.keys(BADGE_DOCKED_TYPES)),
+  icons: PropTypes.arrayOf(PropTypes.oneOf(Object.keys(bpkIcons))),
+  iconSeparator: PropTypes.string,
   style: ViewPropTypes.style,
+  type: PropTypes.oneOf(Object.keys(BADGE_TYPES)),
 };
 
 BpkBadge.defaultProps = {
-  type: BADGE_TYPES.warning,
+  icons: [],
```
shaundon: I think the default should be `null` - no need to use an array if it isn't needed (remember to update prop types too if you do this).

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -157,6 +169,22 @@ const BpkBadge = (props: Props) => {
 
   return (
     <View style={viewStyle}>
+      {icons &&
+        icons.map((i, n) => (
```
shaundon: One-letter variable names can be a bit obtuse - can we use `iconName` and `index` instead?

native/packages/react-native-bpk-component-badge/stories.js
```diff
@@ -21,11 +21,13 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react-native';
 import { View, StyleSheet } from 'react-native';
+import { icons } from 'react-native-bpk-component-icon';
 import {
   colorGray50,
   colorGray700,
   spacingSm,
 } from 'bpk-tokens/tokens/base.react.native';
+import { capitalize } from 'lodash';
```
shaundon: Why are we adding an import instead of doing it in a one-liner that we already had?

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
georgegillams: Yo - I tried this but Flow starts complaining at me 😢

packages/bpk-docs/src/pages/NativeBadgePage/NativeBadgePage.js
```diff
@@ -51,6 +55,50 @@ const components = [
...
+        height: 1920,
+        src: `/${androidScreenshotWithIcon}`,
+        altText: 'Android Badge Component with a single icon',
+        subText: '(Google Pixel emulator)',
+      },
+    ],
+  },
+  {
+    id: 'with-icons',
+    title: 'With multiple icons',
```
shaundon: The icon separator is optional, so let's make it clear it's been supplied here.
```suggestion
    title: 'With multiple icons and an icon separator',
```

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
georgegillams: I agree - I hate it when people write code like this. Goddamnit!

native/packages/react-native-bpk-component-badge/stories.js
georgegillams: Why write a one-liner that we don't already have when we could just import it from a library we do already have 😆

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -157,6 +169,26 @@ const BpkBadge = (props: Props) => {
 
   return (
     <View style={viewStyle}>
+      {icons &&
+        icons.map((icon, index) => (
+          <Fragment key={`${icon}`}>
+            <BpkIcon
+              small
+              icon={icon}
+              style={JSON.parse(JSON.stringify(textStyle))}
```
georgegillams: For some reason, this fixes the flow issue I was seeing.
Don't know why - this isn't really a solution!

native/packages/react-native-bpk-component-badge/stories.js
shaundon: Counter-counter question/statement: You **don't** already have Lodash; it isn't in this package's dependencies 😉

georgegillams: Counter-counter-counter statement: Yeah but this is stories 🤷‍♂️ 
So we do have Lodash

shaundon: Counter-counter-counter-counter statement: It's only working because you have Lodash installed elsewhere - by that logic why do we add anything at all to `devDependencies` 😛

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -155,8 +159,13 @@ const BpkBadge = (props: Props) => {
     viewStyle.push(userStyle);
   }
 
+  const adjustAccessoryView = React.cloneElement(accessoryView, {
```
georgegillams: Will break if `accessoryView === null`

native/packages/react-native-bpk-component-badge/src/BpkBadge.js
```diff
@@ -115,6 +118,7 @@ export const BADGE_DOCKED_TYPES = {
 
 export type Props = {
   message: string,
+  accessoryView: ?Element<BpkBadgeIcons>,
```
georgegillams: Should this be less limited?

native/packages/react-native-bpk-component-badge/stories.js
georgegillams: So what d'yall suggest then?

UNRELEASED.md
```diff
@@ -1 +1,6 @@
 # Unreleased
+
+**Added:**
+ - react-native-bpk-component-badge:
+     - Badges now accept an optional `accessoryView` that allows icons, or other content, to be placed alongside the text inside the bade. See https://backpack.github.io/components/badge/?platform=native#with-multiple-icons
```
shaundon: You changed the ID in the docs to `with-icons` so the link will need to change too.

georgegillams: 🤦‍♂️

native/packages/react-native-bpk-component-badge/README.md
```diff
@@ -34,18 +35,36 @@ export default class App extends Component {
...
-| Property              | PropType                                                                 | Required | Default Value |
+| Property              | PropType                                                                  | Required | Default Value |
 | --------------------- | --------------------------------------------------------------------------| -------- | ------------- |
 | message               | string                                                                    | true     | -             |
+| accessoryView         | node                                                                      | false    | null          |
 | docked                | oneOf('start', 'end')                                                     | false    | null          |
 | type                  | oneOf('success', 'warning', 'destructive', 'light', 'inverse', 'outline') | false    | warning       |
+
+
+## Accessory view
```
shaundon:
```suggestion
## `accessoryView`
 The accessory view allows for icons to be placed in front of the text inside the Badge component in conjunction with `BpkBadgeIcons`.

## `BpkBadgeIcons`

Renders icons suitable to be placed inside a Badge.

### Props

<props table here for BpkBadgeIcons here>
```

I based this off of https://backpack.github.io/components/navigation-bar?platform=native

native/packages/react-native-bpk-component-badge/src/BpkBadgeIcons.js
```diff
@@ -0,0 +1,123 @@
...
+    justifyContent: 'center',
+  },
+  item: {
+    color: colorGray700,
+    includeFontPadding: false,
+    marginEnd: spacingSm,
+  },
+});
+
+export const BADGE_TYPES = {
```
shaundon: Unused - remove.

native/packages/react-native-bpk-component-badge/src/BpkBadgeIcons.js
```diff
@@ -0,0 +1,123 @@
...
+  success: 'success',
+  warning: 'warning',
+  destructive: 'destructive',
+  light: 'light',
+  inverse: 'inverse',
+  outline: 'outline',
+};
+
+export const BADGE_DOCKED_TYPES = {
```
shaundon: Unused - remove.

native/packages/react-native-bpk-component-badge/src/BpkBadgeIcons.js
```diff
@@ -0,0 +1,123 @@
...
+import { View, ViewPropTypes, Text, StyleSheet } from 'react-native';
+
+const styles = StyleSheet.create({
+  outerView: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  item: {
+    color: colorGray700,
```
shaundon: Default colour not required.

UNRELEASED.md
```diff
@@ -1 +1,6 @@
 # Unreleased
+
+**Added:**
+ - react-native-bpk-component-badge:
+     - Badges now accept an optional `accessoryView` that allows icons, or other content, to be placed alongside the text inside the bade. See https://backpack.github.io/components/badge/?platform=native#with-icons
```
```suggestion
     - Badges now accept an optional `accessoryView` that allows icons, or other content, to be placed alongside the text inside the badge. See https://backpack.github.io/components/badge/?platform=native#with-icons
```

native/packages/react-native-bpk-component-badge/README.md
```diff
@@ -34,18 +35,36 @@ export default class App extends Component {
...
 }
 ```
+
 ## Props
 
-| Property              | PropType                                                                 | Required | Default Value |
+| Property              | PropType                                                                  | Required | Default Value |
 | --------------------- | --------------------------------------------------------------------------| -------- | ------------- |
 | message               | string                                                                    | true     | -             |
+| accessoryView         | node                                                                      | false    | null          |
```
shaundon: Should be `element`.

native/packages/react-native-bpk-component-badge/src/BpkBadgeIcons.js
```diff
@@ -0,0 +1,92 @@
...
+    includeFontPadding: false,
+    marginEnd: spacingSm,
+  },
+});
+
+export type Props = {
+  icons: Array<Element<typeof BpkIcon>>,
+  itemStyle: ?(Object | Array<Object>),
+  separator: ?string,
+  style: ?(Object | Array<Object>),
```
shaundon: `style` should be removed - it's not applied anywhere.

native/packages/react-native-bpk-component-badge/README.md
```diff
@@ -34,18 +35,36 @@ export default class App extends Component {
...
 | docked                | oneOf('start', 'end')                                                     | false    | null          |
 | type                  | oneOf('success', 'warning', 'destructive', 'light', 'inverse', 'outline') | false    | warning       |
+
+
+## `accessoryView`
```
shaundon: https://github.com/Skyscanner/backpack/pull/1173#discussion_r236237279

native/packages/react-native-bpk-component-badge/README.md
```diff
@@ -34,18 +35,46 @@ export default class App extends Component {
...
+
+## `BpkBadgeIcons`
+
+Renders icons suitable to be placed inside a Badge.
+
+### Props
+
+| Property  | PropType                | Required | Default Value |
+| --------- | ------------------------| -------- | ------------- |
+| icons     | arrayOf(oneOf(BpkIcon)) | true     | -             |
```
shaundon:
```suggestion
| icons     | arrayOf(BpkIcon) | true     | -             |
```

PR #1174, opened 2018-11-22
[BPK-2073] Constrain icons used inside badges to 16x16

shaundon:
screen shot 2018-11-22 at 11 42 58

**Summary**
When a badge contains an icon, we constrain it to 16x16, add some margin to separate it from the text and increase the badge height to 20px.

**Details**
As our icons don't have a class name and use an inline style attribute, I've had to write some horrible CSS to target all SVGs and adds width and height using `!important`.

The only other way I can think of doing this is to add logic inside the badge component to find all children that are SVGs and then manually add a style prop to override the existing one. But it feels a bit complex when CSS can do the job.

Sidenote: Should we add a data attribute to our icon SVGs so that they can be targeted easier in the future?

### DIFF
--- a/packages/bpk-component-badge/BadgeLayout.js
+++ b/packages/bpk-component-badge/BadgeLayout.js
@@ -19,7 +19,7 @@
 /* @flow */
 
 import PropTypes from 'prop-types';
-import React, { type Element, Children } from 'react';
+import React, { type Node, Children } from 'react';
 import { cssModules } from 'bpk-react-utils';
 
 import { BADGE_TYPES } from './index';
@@ -35,7 +35,7 @@ const LIGHT_BADGES = [
 
 export type Props = {
   docked: ?string,
-  children: Element<*>,
+  children: Node,
 };
 
 const BadgeLayout = (props: Props) => {
--- a/packages/bpk-mixins/src/mixins/_badges.scss
+++ b/packages/bpk-mixins/src/mixins/_badges.scss
@@ -31,13 +31,29 @@
 ///   }
 
 @mixin bpk-badge {
-  display: inline-block;
+  display: inline-flex;
   padding: $bpk-badge-padding-y $bpk-badge-padding-x;
+  align-items: center;
   background-color: $bpk-badge-background-color;
 
   @include bpk-border-radius-xs;
   @include bpk-text;
   @include bpk-text--sm;
+
+  // // HACK: Icons have an inline style attribute setting the height and width.
+  // // so the only way to override this in CSS is to use !important.
+  > svg {
+    /* stylelint-disable declaration-no-important */
+    width: 16 * $bpk-one-pixel-rem !important;
+    height: 16 * $bpk-one-pixel-rem !important;
+    /* stylelint-enable declaration-no-important */
+    margin-top: $bpk-one-pixel-rem * 2;
+    margin-bottom: $bpk-one-pixel-rem * 2;
+
+    &:last-of-type {
+      @include bpk-margin-trailing($bpk-badge-padding-x);
+    }
+  }
 }
 /// Centered badge. Modifies the bpk-badge mixin.
 ///
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -3,3 +3,8 @@
 **Added:**
  - bpk-react-utils:
    - Added `deprecated` function that accepts a prop type and adds a suitable console warning when the prop is used.
+
+**Changed**
+
+- bpk-component-badge:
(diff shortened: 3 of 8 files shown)


packages/bpk-component-badge/BadgeLayout.js
```diff
@@ -35,7 +35,7 @@ const LIGHT_BADGES = [
 
 export type Props = {
   docked: ?string,
-  children: Element<*>,
+  children: Node,
```
shaundon: Flow was complaining when I put an array of `div`s inside a `BadgeLayout`, this fixes that.

packages/bpk-mixins/src/mixins/_badges.scss
```diff
@@ -31,13 +31,29 @@
...
+
+  // // HACK: Icons have an inline style attribute setting the height and width.
+  // // so the only way to override this in CSS is to use !important.
+  > svg {
+    /* stylelint-disable declaration-no-important */
+    width: 16 * $bpk-one-pixel-rem !important;
+    height: 16 * $bpk-one-pixel-rem !important;
+    /* stylelint-enable declaration-no-important */
+    margin-top: $bpk-one-pixel-rem * 2;
+    margin-bottom: $bpk-one-pixel-rem * 2;
```
shaundon: Adding margin here to increase the overall badge size. As CSS still doesn't have a `:parent` selector I think it makes more sense to do it using the SVG instead of messing about with logic to apply it to the badge.

packages/bpk-component-badge/BadgeLayout.js
georgegillams: Yeah I think this is the correct way of doing it!

packages/bpk-mixins/src/mixins/_badges.scss
```diff
@@ -31,13 +31,29 @@
 ///   }
 
 @mixin bpk-badge {
-  display: inline-block;
+  display: inline-flex;
```
matthewdavidson: I'm a little worried that this change can break layout of badges alongside other elements. Specifically, I worry that switching from `display: inline-block` to `display: inline-flex` would make the container have a similar display property as `display: inline`.

package-lock.json
```diff
@@ -29710,7 +29710,7 @@
         },
         "uglify-js": {
           "version": "2.4.24",
-          "resolved": "http://registry.npmjs.org/uglify-js/-/uglify-js-2.4.24.tgz",
+          "resolved": "https://registry.npmjs.org/uglify-js/-/uglify-js-2.4.24.tgz",
```
georgegillams: Pretty sure these shouldn't have all changed

packages/bpk-mixins/src/mixins/_badges.scss
matthewdavidson: It might be ok https://codesandbox.io/s/l2118o538l

packages/bpk-mixins/src/mixins/_badges.scss
```diff
@@ -31,13 +31,29 @@
...
+  > svg {
+    /* stylelint-disable declaration-no-important */
+    width: 16 * $bpk-one-pixel-rem !important;
+    height: 16 * $bpk-one-pixel-rem !important;
+    /* stylelint-enable declaration-no-important */
+    margin-top: $bpk-one-pixel-rem * 2;
+    margin-bottom: $bpk-one-pixel-rem * 2;
+
+    &:last-of-type {
+      @include bpk-margin-trailing($bpk-badge-padding-x);
```
georgegillams: Shouldn't we leave margins up to userland? What's the rationale for this margin over any other?

packages/bpk-mixins/src/mixins/_badges.scss
shaundon: Yeah I did some googling before I made the change and I think it's ok. 

From what I can tell, `inline-flex` is to `flex` as `inline-block` is to `block`. So it should behave the same. https://stackoverflow.com/a/46864624/1011161

shaundon: Also your example wasn't really applicable to this scenario IMO as you put block-level elements inside the wrappers. This one makes more sense, I think: https://codesandbox.io/s/8yk138z9jj

shaundon: Updated the example to add `margin-top: 30px` to each element. `inline` doesn't respect top and bottom margins, but you can see that `inline-flex` does. So I think it's safe.

PR #1176, opened 2018-11-23
[BPK-2087] Adjust XXL and XXXL tokens

shaundon:
Also updated RN BpkText.


matthewdavidson:
This change probably fixes the issue, but to be sure: which android emulator are you using? Notice the image diffs... aside from the status bar colour issue (which is something else we need to fix) the stats bar and bottom bar heights are different...

UNRELEASED.md
```diff
@@ -1 +1,12 @@
 # Unreleased
+
+**Breaking:**
+
+- react-native-bpk-component-text:
+  - Now consuming token values for the new type scale (only XXL and XXXL styles have changed).
+
+**Added**
+
+- bpk-tokens:
```
matthewdavidson: Aren't these all breaking?

matthewdavidson:
This change probably fixes the issue, but to be sure: which android emulator are you using? Notice the image diffs... aside from the status bar colour issue (which is something else we need to fix) the stats bar and bottom bar heights are different...

UNRELEASED.md
shaundon: I went off what was used last time: https://github.com/Skyscanner/backpack/blob/master/CHANGELOG.md#2018-10-31---new-type-scale

shaundon:
> which android emulator are you using? 

I'm using bpk-avd but for some reason it keeps doing this for me. I've tried deleting and rebuilding the emulator, making sure my a11y settings are the same, I just can't figure out why it's different.

matthewdavidson:
@amolgupta / @tiagoengel  - any ideas?

PR #1179, opened 2018-11-26
[NO-JIRA] Support touchable overlay with lg border radius

georgegillams:
We need touchable overlay to support `lg` border radius in order to achieve [BPK-2058](https://gojira.skyscanner.net/browse/BPK-2058)


shaundon:
Also this is a 💯 example of a discrete PR - nice one 👍

PR #1181, opened 2018-11-28
Revert "Use BpkInfiniteScroll instead of react-virtualized"

MoOmEeN:
what: revert scrollable calendar back to using react-virtualized,
why: new implementation has a bug of not automatically scrolling to the selected day. also the experience is not good

we are working on improving the component here: 
https://gojira.skyscanner.net/browse/TSS-4044
https://gojira.skyscanner.net/browse/BPK-2083

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,2 +1,6 @@
 # Unreleased
 
+**Fixed:**
+   
+ - bpk-component-scrollable-calendar:
+   - Revert to use react-virtualized instead of bpk-component-infinite-scroll
--- a/packages/bpk-component-scrollable-calendar/package-lock.json
+++ b/packages/bpk-component-scrollable-calendar/package-lock.json
@@ -4,6 +4,30 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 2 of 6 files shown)


matthewdavidson:
Hey - thanks for this. If we refer back to the original PR to introduce `bpk-component-infinite-scroll` (https://github.com/Skyscanner/backpack/pull/1141/files), there is a change that moves the following prop assignment from `DateComponent` to `DateContainer`:
```
isEmptyCell={!isSameMonth(date, month) && ignoreOutsideDate}
```

(See https://github.com/Skyscanner/backpack/pull/1141/files#diff-b09d1b032e5b98fd4593b1ffd3b0ff36R148)

I'm not close to this work so I've not looked into the details ... any idea why this wasn't also reverted? oversight perhaps?

matthewdavidson:
On closer inspection, `isEmptyCell` is defined on `DateContainer` so this must have just been fixed in the original PR.

matthewdavidson:
I rebased it for you :)

MoOmEeN:
thanks @matthewdavidson, yep, I left this as this property is required on `DateContainer`, not `BpkCalendarDate`

PR #1185, opened 2018-11-30
Fix RN nav bar title ellipsis behaviour

franrios:


native/packages/react-native-bpk-component-navigation-bar/src/BpkNavigationBar.android.js
```diff
@@ -75,9 +75,8 @@ const styles = StyleSheet.create({
     justifyContent: 'flex-end',
   },
   titleString: {
-    position: 'absolute',
-    start: 72, // eslint-disable-line backpack/use-tokens
-    bottom: 15, // eslint-disable-line backpack/use-tokens
+    flex: 1,
+    paddingHorizontal: 24, // eslint-disable-line backpack/use-tokens
```
matthewdavidson: I see the direction your headed here ... but the existing hard coded values were picked to aligned directly to the underlying native component. Where did 24 come from?

PR #1186, opened 2018-11-30
Implement BpkBackgroundImage in bpk-component-image

kirisu:
* Implemented BpkBackgroundImage to match BpkImage interface
* Moved some reusable scss into mixins to be shared between BpkImage and BpkBackgroundImage
* Fixed up the old tests that were lying around
* Added storybook entries for BpkBackgroundImage

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -4,6 +4,8 @@
 
 - bpk-component-popover:
   - Added `closeButtonProps` so that consumers can set props on the close button - for example, `tabIndex`.
+- bpk-component-image
+  - Introducing `BpkBackgroundImage` for rendering background images. See https://backpack.github.io/components/image?platform=web#background-image.
 
 **Fixed:**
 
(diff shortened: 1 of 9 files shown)


shaundon:
This looks really great 😄

I've tested it in Chrome (Mac and Android) and IE11 (Windows) and it works in them all.

There are two small documentation things left to do and then we're good to go 🎉 

1) We need a quick entry in `UNRELEASED.md`, along the lines of:

```
**Added**

- bpk-component-image
  - Introducing `BpkBackgroundImage` for rendering background images. See https://backpack.github.io/components/image?platform=web#background-image.
```

You can also look at `CHANGELOG.md` for more examples of this.

2) We need some documentation for backpack.github.io. This is where the existing image documentation lives, I think we can just add the background image as some additional examples alongside it (like you've done in `stories.js`).

The component's `README.md` will also need to be updated to include the prop type table and an example for the background image component. This then gets automatically included in the page on the docs site.

PR #1187, opened 2018-11-30
[BPK-2079] Activate the bagcheck link

tiagoengel:
I've added a new page in `/bag-check-bookmarklet` that looks like this:

0 0 0 0_8080_bag-check-bookmarklet

I've implemented this a bit different from the original implementation, which was using the raw file loader, because that meant we wouldn't be able to update the code that downloads the bookmarklet since it would be inlined there. With the current implementation we can update both the bag check lib and the code that loads it, and it will be picked up automatically by everyone that has the link.

### DIFF
--- a/packages/bpk-docs/src/pages/BagCheckPage/BagCheckPage.js
+++ b/packages/bpk-docs/src/pages/BagCheckPage/BagCheckPage.js
@@ -40,7 +40,7 @@ const Page = () => {
   return (
     <div className={getClassName('bpk-docs-bag-check-page')}>
       <BpkText tagName="h1" textStyle="lg">
-        Drag and drop the link bellow to your bookmarks toolbar
+        Drag and drop the link below to your bookmarks toolbar.
       </BpkText>
       <BpkLink
         href={script}
--- a/packages/bpk-docs/src/pages/BagCheckPage/BagCheckPage.js
+++ b/packages/bpk-docs/src/pages/BagCheckPage/BagCheckPage.js
@@ -1,0 +1,55 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+import React from 'react';
+import BpkLink from 'bpk-component-link';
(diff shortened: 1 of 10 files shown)


packages/bpk-docs/src/template.js
```diff
@@ -39,6 +39,14 @@ export default ({ head = {}, html = '', assets = {} }) => `<!doctype html>
   ${html}
 </div>
 
+<script type="text/javascript">
+  window.BPK_DOC_GLOBALS = window.BPK_DOC_GLOBALS || {};
+  window.BPK_DOC_GLOBALS.entries = {
+    docs: '${assets.docs.js}',
```
matthewdavidson: Why is docs there? we don't use that do we?

webpack.config.babel.js
```diff
@@ -57,6 +57,7 @@ const sassOptions = {
 const config = {
   entry: {
     docs: './packages/bpk-docs/src/index.js',
+    bagCheck: './packages/bpk-docs/src/bag-check-bookmarklet/index.js',
```
matthewdavidson: My only issue with this is that this puts us down a path where we can't use tools like `create-react-app` further down the line...

k0nserv: Why aren't we using a raw-loader solution as we did in 41b7cf5bf?

tiagoengel: I've explained in more details the PR description, but basically because with that we can't update the page that downloads the bookmarklet without having to ask everyone to delete and re add it. This approach will make changing things like changing it to open source transparent to users.

tiagoengel: As for `create-react-app` we can't do it right now, even without this. The custom static site generator we use is not supported.

packages/bpk-docs/src/template.js
tiagoengel: Just adding all entries there, not really required but I don't see any harm in having it there.

webpack.config.babel.js
matthewdavidson: > As for create-react-app we can't do it right now, even without this. The custom static site generator we use is not supported.

We can - check this out -> https://facebook.github.io/create-react-app/docs/pre-rendering-into-static-html-files

packages/bpk-docs/src/template.js
matthewdavidson: I'd prefer to remove it, but don't block this PR on me.

packages/bpk-docs/src/pages/BagCheckPage/BagCheckPage.js
```diff
@@ -0,0 +1,55 @@
...
+
+  // NOTE: this is updated with the proper hash during the prod build.
+  const bagCheckFile = getGlobal('entries.bagCheck', 'bagCheck.js');
+  const link = `${window.location.origin}/${bagCheckFile}`;
+  const script = `javascript:(function(){var t=document.createElement("script");t.type="application/javascript",t.src="${link}",document.getElementsByTagName("body")[0].appendChild(t)}());`;
+
+  return (
+    <div className={getClassName('bpk-docs-bag-check-page')}>
+      <BpkText tagName="h1" textStyle="lg">
+        Drag and drop the link bellow to your bookmarks toolbar
```
shaundon:
```suggestion
        Drag and drop the link below to your bookmarks toolbar.
```

PR #1188, opened 2018-11-30
Use flex-box to vertically size scrollable calendar

detzgk:
This significantly simplifies the code we have that uses the scrollable calendar since it now resizes to fill whatever space is available to it (our current code is full of `calc()`ed height properties that seem rather fragile to me). If it's not constrained, it will behave as it did before.

### DIFF
--- a/packages/bpk-component-scrollable-calendar/src/bpk-scrollable-calendar.scss
+++ b/packages/bpk-component-scrollable-calendar/src/bpk-scrollable-calendar.scss
@@ -20,4 +20,9 @@
 
 .bpk-scrollable-calendar {
   display: flex;
+  flex-direction: column;
+
+  &__header {
+    flex-shrink: 0;
+  }
 }
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,3 +1,5 @@
 # Unreleased
 
+- bpk-component-scrollable-calendar:
+  - Use flex-box to let components resize vertically to any height in an elegant, unhacky way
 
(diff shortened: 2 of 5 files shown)


detzgk:
I'm not really sure how to make flow happy with that without repeating the BpkCalendar.propTypes (it's hidden inside another HoC so I can't just refer to it...)

packages/bpk-component-scrollable-calendar/src/bpk-scrollable-calendar.scss
```diff
@@ -20,4 +20,9 @@
 
 .bpk-scrollable-calendar {
   display: flex;
+  flex-direction: column;
```
matthewdavidson: This has been renamed but no reference to it has been updated ... was this a phantom file? 😱

detzgk: Yep. We snuck it past review last time somehow. :)

matthewdavidson:
@detzgk, thats cool and it sounds reasonable. My predicament is that if i remove those styles ... nothing happens 😅  i.e. it's not clear what those styles are achieving in the scrollable calendar when I view in its storybook.

detzgk:
Let me make a new story to show what it does and to make sure I've got the CSS right (spoiler: I rarely do).

detzgk:
Never mind. These changes aren't necessary with the version that uses react-virtualized. Closing this one and opening one to deal with a small cosmetic issue and clean up the stray .scss file.

PR #1224, opened 2018-12-11
Set the calendar date background to transparent instead of pure white

detzgk:
This avoids a problem with the scrollable calendar where the background of the dates is drawn above the horizontal line under the weekday names.

Also remove an unused SCSS file and fixed up some bad CSS and JavaScript in the scrollable calendar.


packages/bpk-component-calendar/src/bpk-calendar-date.scss
```diff
@@ -24,7 +24,7 @@
   padding: 0;
   border: none;
   border-radius: $bpk-border-radius-pill;
-  background-color: $bpk-calendar-day-background-color;
+  background-color: transparent;
```
detzgk: Should I remove this variable from bpk-tokens? It's not used anywhere.

matthewdavidson: Yup its a good shout - what we have done previously is mark that token as deprecated and we'll remove it in a future release of `bpk-tokens`.

(FYI for other readers here is a link to the values of the tokens above: https://github.com/Skyscanner/backpack/blob/3df5ea4c52386a2e51acb428c0c716a965f9ed33/packages/bpk-tokens/src/web/base/calendar.json#L25)

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -88,13 +88,11 @@ class BpkScrollableCalendarGridList extends React.Component {
     );
   }
 
-  renderList({ width, height }) {
```
matthewdavidson: What's happening here?

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -88,13 +88,11 @@ class BpkScrollableCalendarGridList extends React.Component {
...
+  renderList(width, height) {
     return (
       <List
         extraData={this.props}
-        style={
-          this.getHtmlElement().dir === 'rtl' ? { direction: 'rtl' } : null
-        }
+        style={this.getHtmlElement().dir === 'rtl' ? { direction: 'rtl' } : {}}
```
matthewdavidson: IMO `null` is a more appropriate value than and empty object as it signals that there is no style whereas an empty object is ambiguous. Any particular reason to make this change?

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -129,7 +127,7 @@ class BpkScrollableCalendarGridList extends React.Component {
           className={getClassName('bpk-scrollable-calendar-grid-list__strip')}
         >
           <AutoSizer>
-            {(width, height) => this.renderList(width, height)}
+            {({ width, height }) => this.renderList(width, height)}
```
matthewdavidson: Similar to change above, whats happening here?

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
detzgk: AutoSizer calls its child with a single object `{ width, height }`.  We were accepting `(width, height)` without destructuring it which means:
```js
width = { width: xxx, height: xxx };
height = undefined;
```
So that wasn't really correct. Now it's passing width and height properly to renderList. It's a local method so there's not much reason to wrap it in an object, if we need to change the interface we can do it right in that file.

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
detzgk: We get warnings in the console if we pass a null. It's declared in react-virtualized as `PropTypes.object.isRequired` with a default of `{}`.

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
matthewdavidson: So it was previously broken yeah? 😱 

Whats the before & after behaviour?

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
matthewdavidson: Ok doke - nice one 👍

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
detzgk: It was broken, yeah, but I didn't see any change in behaviour. It might change something about how it decides whether to render an item in the grid or not.

matthewdavidson: Ok doke 😅

packages/bpk-component-calendar/src/bpk-calendar-date.scss
detzgk: I added the deprecation markers.

matthewdavidson: Unfortunately you didn't run `npm run build` before committing :P

detzgk: But I did realize it before you wrote that comment. So partial credit? :)

matthewdavidson: Absolutely 😄

PR #1228, opened 2018-12-13
[BPK-2117] Fix RN text component prop type warning

matthewdavidson:
Removes the following warning:
```
Warning: Failed prop type: Calling PropTypes validators directly is not supported by the `prop-types` package. Use `PropTypes.checkPropTypes()` to call them. Read more at http://fb.me/use-check-prop-types
        in BpkText
```
Doing a browser search in the build log yields:

_Before:_
image

_After:_
image

### DIFF
--- a/native/packages/react-native-bpk-component-text/src/common-types.js
+++ b/native/packages/react-native-bpk-component-text/src/common-types.js
@@ -87,6 +87,7 @@ export const weightPropType = (
   props: Props,
   propName: string,
   componentName: string,
+  ...rest: Array<any>
 ) => {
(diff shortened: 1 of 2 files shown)


native/packages/react-native-bpk-component-text/src/common-types.js
```diff
@@ -87,6 +87,7 @@ export const weightPropType = (
   props: Props,
   propName: string,
   componentName: string,
+  ...rest: Array<any>
```
matthewdavidson: React passes a secret argument to prop types to identify the source of the caller. We just need to spread the arguments through in custom validators.

PR #1230, opened 2018-12-13
[No JIRA] Remove redundant warning

shaundon:
CI covers this and it's visual noise.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -84,24 +84,6 @@ if (packagesModified && !unreleasedModified && !declaredTrivial) {
   );
 }
 
-// If source files have changed, the snapshots should have been updated.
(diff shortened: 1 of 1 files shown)


k0nserv:
CI doesn't strictly speaking cover this. This is reminding you that if the component change snapshots should fail. If they don't maybe that's an indication you need to refactor?

shaundon:
Hmm yeah fair point. Maybe we can rephrase the message to make it more relevant.

k0nserv:
Yeah I am not opposed to removing it if the majority are in favour of that, but we should do so for the right reasons

shaundon:
Gonna close for now, it's very minor and we don't have a consensus.

PR #1231, opened 2018-12-13
[No JIRA] Add release script for consistency

shaundon:
So we can use `npm run release` across monorepo, Android and iOS repos.


shaundon:
> And I will hate this change because I will use the old script name.

That's why I did it as an alias to preserve backwards compatibility for now 😄

PR #1232, opened 2018-12-13
Remove the list--strip element from the scrollable calendar

detzgk:
Why:
1. It's unnecessary.
2. It has no className property, making it impossible to customize
   with additional CSS classes.

### DIFF
--- a/packages/bpk-component-scrollable-calendar/stories.js
+++ b/packages/bpk-component-scrollable-calendar/stories.js
@@ -18,6 +18,8 @@
 
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import startOfMonth from 'date-fns/start_of_month';
+import endOfMonth from 'date-fns/end_of_month';
 import { storiesOf } from '@storybook/react';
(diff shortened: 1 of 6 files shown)


packages/bpk-component-scrollable-calendar/stories.js
```diff
@@ -94,6 +96,38 @@ storiesOf('bpk-component-scrollable-calendar', module)
...
+        DateComponent={BpkScrollableCalendarDate}
+        showWeekendSeparator
+        selectTodaysDate
+        // Subtract one day from today's date to make today selectable by default
+        minDate={DateUtils.addDays(new Date(), -1)}
+        maxDate={DateUtils.addMonths(new Date(), 12)}
+      />
+    </div>
+  ))
+  .add('Scrollable Calendar with a single month', () => (
```
matthewdavidson: Whats the motivation for this story?

detzgk: To show that if there's more space than the component needs, it will align to the top of the space.

PR #1269, opened 2018-12-14
[No JIRA] Bye native

shaundon:
Now we have a separate repo for React Native, I've done the following:

* Deleted the `native` folder.
* Added `backpack-react-native` as a submodule and updated the docs to handle it correctly.
* Updated docs and such to not care about RN anymore.

I've run the docs manually and it still works, and I've also done a patch release of RN components from the RN repo to prove we can still do it.


.flowconfig
```diff
@@ -1,6 +1,6 @@
 [ignore]
+<PROJECT_ROOT>/backpack-react-native/.*
```
shaundon: Have to ignore it to stop Flow trying to run on the RN code.

.flowconfig
```diff
@@ -12,17 +12,10 @@
 [options]
 module.name_mapper.extension='scss' -> '<PROJECT_ROOT>/scripts/stubs/styleStub.js'
 module.name_mapper.extension='md' -> '<PROJECT_ROOT>/scripts/stubs/fileStub.js'
-module.name_mapper='\(react-native\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
-module.name_mapper='\(react-native-linear-gradient\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
-module.name_mapper='\(@storybook/react-native\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
-module.name_mapper='\(@skyscanner/react-native-transitiongroup\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
-module.name_mapper='\(react-native-maps\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
-module.name_mapper='\(recyclerlistview\)' -> '<PROJECT_ROOT>/scripts/stubs/jsModuleStub.js'
+module.name_mapper.extension='png' -> '<PROJECT_ROOT>/scripts/stubs/fileStub.js'
```
shaundon: And because we're ignoring the RN code (see above comment) we need to stub this out now or Flow fails.

PR #1276, opened 2018-12-31
[BPK-2083] calendar performance

georgegillams:
This PR doesn't actually address anything performance related, but it does replace `react-virtualised` with `react-window` (a new, cut-down version written from scratch).

The improvement in prod bundle size is approximately 64.5kb => 2.8kb!

From testing so far, performance seems to at least be similar to before. See profiling screenshots below.

Note that profiling is only available in React 16.5 - so I created [this repo](https://github.com/georgegillams/bpk-calendar-performance-react-16.5) for profiling the component.

## API differences:
The `react-window` API is somewhat different to `react-virtualized`. It requires the height to display at (in pixels) as well as the heights of each item.
To get the item heights, I have added code to the constructor that calculates the heights of the grids based on the number of days etc.
In order to get the component height, I am measuring the outer height using `componentDidMount` and listeners on the window size.
These are two necessary evils if we want to use `react-window`.
It also requires the initial offset to be specified in pixels. For this, I am using the pre-calculated `itemHeight` values to convert the item index to pixels.

## Testing differences:
With `react-virtualized`, the first render doesn't include any children. The subsequent render then renders children as needed.
`react-window` is more clever. Because it knows the height and itemHeights at first render, it can populate the children at the same time.
As a result, the snapshot tests are much larger (as they include the children)
I have therefore employed shallow rendering.

## Profiling before and after:
screenshot 2019-01-08 at 09 50 30

react-virtualized

react-window

### DIFF
--- a/packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendar-test.js
+++ b/packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendar-test.js
@@ -18,7 +18,7 @@
 /* @flow */
 
 import React from 'react';
-import renderer from 'react-test-renderer';
+import Shallow from 'react-test-renderer/shallow';
 import { DateUtils } from 'bpk-component-calendar';
 
 import BpkScrollableCalendar from './BpkScrollableCalendar';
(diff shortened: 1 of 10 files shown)


packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendar-test.js
```diff
@@ -18,7 +18,7 @@
 /* @flow */
 
 import React from 'react';
-import renderer from 'react-test-renderer';
+import Shallow from 'react-test-renderer/shallow';
```
shaundon: https://www.youtube.com/watch?v=bo_efYhYU2A

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -22,25 +22,27 @@ import React from 'react';
...
 import STYLES from './bpk-scrollable-calendar-grid-list.scss';
 import BpkScrollableCalendarGrid from './BpkScrollableCalendarGrid';
 import getMonthsArray from './utils';
 
 const getClassName = cssModules(STYLES);
 
+// The `react-window` API requires the height in pixels to be specified.
+// These constants are here to facilitate calculating the height.
+const ROW_HEIGHT = 42;
+const GRID_HEIGHT_WITHOUT_ROWS = 60;
```
shaundon: Could you quickly explain what these mean a little more? Particular `GRID_HEIGHT_WITHOUT_ROWS`.

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -49,71 +51,74 @@ class BpkScrollableCalendarGridList extends React.Component {
...
+    for (let i = 0; i < months.length; i += 1) {
+      const firstDayOffset = (months[i].getDay() + 7 - props.weekStartsOn) % 7;
+      const monthLength = DateUtils.daysInMonth(
+        months[i].getYear(),
+        months[i].getMonth(),
+      );
+      const calendarGridSpaces = firstDayOffset + monthLength;
+      const rowCount = Math.ceil(calendarGridSpaces / COLUMN_COUNT);
+      monthItemHeights[i] = GRID_HEIGHT_WITHOUT_ROWS + ROW_HEIGHT * rowCount;
+    }
```
shaundon: There may be a better-looking way to do this functionally, something like:

```
const monthItemHeights = months.map(month => {
  // Logic here
});
```

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -49,71 +51,74 @@ class BpkScrollableCalendarGridList extends React.Component {
...
+      outerHeight: ESTIMATED_MONTH_ITEM_HEIGHT,
     };
   }
 
+  componentDidMount = () => {
+    this.setComponentHeight();
+    if (typeof window !== 'undefined') {
+      document.addEventListener('resize', this.setComponentHeight);
+      document.addEventListener('orientationchange', this.setComponentHeight);
+      document.addEventListener('fullscreenchange', this.setComponentHeight);
```
shaundon: It feels weird that we check `window` is defined but then add event listeners to `document`. Should these listeners be added to the window? Or should the check be that `document` is defined?

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -49,71 +51,74 @@ class BpkScrollableCalendarGridList extends React.Component {
...
-            focusedDate={this.props.focusedDate}
-            preventKeyboardFocus={this.props.preventKeyboardFocus}
-            aria-hidden={index !== 1}
-            className={getClassName('bpk-scrollable-calendar-grid-list__item')}
-          />
-        </div>
-      </CellMeasurer>
-    );
-  }
+  getItemSize = index => this.state.monthItemHeights[index];
```
shaundon: Wondering if there should be a check for `undefined` here so we can pass `0` or something safer through.

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendar-test.js
georgegillams: Wow 😲

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -49,71 +52,74 @@ class BpkScrollableCalendarGridList extends React.Component {
...
+  );
+
+  calculateOffsetInPixels = numberOfMonths => {
+    // The `react-window` API requires the scroll offset to be provided in pixels.
+    // Here we use the pre-calculated item heights to find the correct pixel offset
+    let result = 0;
+    for (let i = 0; i < numberOfMonths; i += 1) {
+      result += this.getItemSize(i);
+    }
+    return result;
```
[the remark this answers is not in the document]
georgegillams: That code is gonna be less readable, so sticking with this for now 😄

PR #1278, opened 2019-01-03
Add range selection support to Week component & classes for date cells

detzgk:
These are not completely related but both are necessary for the in-progress date range selector we're working on.

1. It allows adding a class to the cells created by the Week component so we can remove the margin from it and turn it into padding on the date component itself (so we can colour them with background properties).
2. It adds logic to the Week component to figure out of a week needs to update itself based on a selectionStart and selectionEnd property.

Since this logic was ugly and complicated, I first added a bunch of unit tests to make sure I didn't break existing update logic in the Week component.


detzgk:
No new components, @backpackbot. Just some unit tests.

PR #1289, opened 2019-01-09
[BPK-2144] Replace babel-preset-es2015 with babel-preset-env

shaundon:
This is the result of attempting to upgrade to Babel 7. That wasn't successful, but I was able to replace this one deprecated dependency and I think it has some value on its own.

This also adds Danger to the Greenkeeper ignored packages as we can't upgrade it without upgrading Babel.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [
-    "es2015",
+    "env",
     "react"
   ],
   "plugins": [
--- a/package-lock.json
+++ b/package-lock.json
@@ -4860,43 +4860,11 @@
           }
         },
         "electron-to-chromium": {
-          "version": "1.3.73",
-          "resolved": "https://registry.npmjs.org/electron-to-chromium/-/electron-to-chromium-1.3.73.tgz",
-          "integrity": "sha512-6PIg7v9zRoVGh6EheRF8h6Plti+3Yo/qtHobS4/Htyt53DNHmKKGFqSae1AIk0k1S4gCQvt7I2WgpbuZNcDY+g==",
+          "version": "1.3.100",
+          "resolved": "https://registry.npmjs.org/electron-to-chromium/-/electron-to-chromium-1.3.100.tgz",
+          "integrity": "sha512-cEUzis2g/RatrVf8x26L8lK5VEls1AGnLHk6msluBUg/NTB4wcXzExTsGscFq+Vs4WBBU2zbLLySvD4C0C3hwg==",
           "dev": true
         }
-      }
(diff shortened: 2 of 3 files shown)


matthewdavidson:
I think this is a good intermediary step. However I am certain that you need to configure `babel-preset-env` somewhat to let it know which browsers we support, no?

See https://babeljs.io/docs/en/babel-preset-env.

Btw the mechanism that our consumer use relies on [`babel-preset-react-app`](https://www.npmjs.com/package/babel-preset-react-app)...

shaundon:
> However I am certain that you need to configure babel-preset-env somewhat to let it know which browsers we support, no?

```
Sidenote, if no targets are specified, @babel/preset-env behaves exactly
the same as @babel/preset-es2015, @babel/preset-es2016 and @babel/preset-es2017 
together (or the deprecated babel-preset-latest).

We don't recommend using preset-env this way because it doesn't take advantage of 
its ability to target specific browsers.
```

So it looks like it will work but it's best practice to add something. I guess I'll set it to any browsers with >0.25% market, plus IE11?

> Btw the mechanism that our consumer use relies on babel-preset-react-app...

I'm not really sure what you mean here, could you clarify?

matthewdavidson:
> So it looks like it will work but it's best practice to add something. I guess I'll set it to any browsers with >0.25% market, plus IE11?

Ah I didn't know that - happy to proceed given that it'll be a like for like.

> I'm not really sure what you mean here, could you clarify?

So our consumers use `backpack-react-scripts`. `backpack-react-scripts` uses [`babel-preset-react-app`](https://www.npmjs.com/package/babel-preset-react-app) for it's babel config. Maybe we could just use that?

shaundon:
> Maybe we could just use that? If it's not easy then scrap it and we'll do it later...

I just tried it and it's not looking easy, for some reason 😖 

Literally all I've done is remove `babel-preset-env` and `babel-preset-es2015` and replaced them with `babel-preset-react-app`, then changed `.babelrc` to `{"presets": ["react-app"]}` and I get:

```
> bpk-tokens@27.4.5 tokens /Users/shaundonnelly/projects/backpack/packages/bpk-tokens
> gulp

[15:30:59] Requiring external module babel-register

lerna ERR! npm run tokens stderr:
[15:30:59] Failed to load external module @babel/register
/Users/shaundonnelly/projects/backpack/packages/bpk-tokens/gulpfile.babel.js:19
import fs from 'fs';
^^^^^^

SyntaxError: Unexpected token import
```

¯\\_(ツ)_/¯

PR #1290, opened 2019-01-10
[No JIRA] Change spacingNone from 0rem to 0

shaundon:


shaundon:
> does this fix the "null" representation on the docs site too?

Even better - it fixes that issue and also displays it in the correct order now!

screen shot 2019-01-10 at 10 09 29

PR #1291, opened 2019-01-10
add a different percentage of intersection prop for infinite scroll loading component

arnaugm:
Current implementation of the infinite scroll does not allow using big elements, like an animated card placeholder, as a loading component. The reason is that the pagination is only triggered if the loading component is visible at 0.99%. In small devices that may never be the case if the element is bigger than the viewport which results in a broken infinite scroll.
This PR introduces a prop to the component to require visibility of smaller parts of the element to trigger the pagination.
Default value keeps the current behaviour to avoid any potential conflict with existing usages of the component.

### DIFF
--- a/packages/bpk-component-infinite-scroll/README.md
+++ b/packages/bpk-component-infinite-scroll/README.md
@@ -162,13 +162,14 @@ Updates the internal array and triggers all listeners.
 
 ## Props
 
-| Property                | PropType               | Required | Default Value |
-| ----------------------- | ---------------------- | -------- | ------------- |
-| dataSource              | instanceOf(DataSource) | true     | -             |
-| elementsPerScroll       | number                 | false    | 5             |
-| initiallyLoadedElements | number                 | false    | 5             |
-| onScroll                | func                   | false    | null          |
-| onScrollFinished        | func                   | false    | null          |
-| renderLoadingComponent  | func                   | false    | null          |
-| renderSeeMoreComponent  | func                   | false    | null          |
-| seeMoreAfter            | number                 | false    | null          |
+| Property                | PropType                         | Required | Default Value |
+| ----------------------- | -------------------------------- | -------- | ------------- |
+| dataSource              | instanceOf(DataSource)           | true     | -             |
+| elementsPerScroll       | number                           | false    | 5             |
+| initiallyLoadedElements | number                           | false    | 5             |
+| loaderMinDisplay        | oneOf(['small', 'half', 'full']) | false    | 'full'        |
+| onScroll                | func                             | false    | null          |
+| onScrollFinished        | func                             | false    | null          |
+| renderLoadingComponent  | func                             | false    | null          |
+| renderSeeMoreComponent  | func                             | false    | null          |
+| seeMoreAfter            | number                           | false    | null          |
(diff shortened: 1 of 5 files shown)


packages/bpk-component-infinite-scroll/README.md
```diff
@@ -162,13 +162,14 @@ Updates the internal array and triggers all listeners.
...
-| onScrollFinished        | func                   | false    | null          |
-| renderLoadingComponent  | func                   | false    | null          |
-| renderSeeMoreComponent  | func                   | false    | null          |
-| seeMoreAfter            | number                 | false    | null          |
+| Property                | PropType                         | Required | Default Value |
+| ----------------------- | -------------------------------- | -------- | ------------- |
+| dataSource              | instanceOf(DataSource)           | true     | -             |
+| elementsPerScroll       | number                           | false    | 5             |
+| initiallyLoadedElements | number                           | false    | 5             |
+| loaderMinDisplay        | oneOf(['small', 'half', 'full']) | false    | 'full'        |
```
tiagoengel: Not a big fan of the name. It doesn't really indicate what it does IMO. Can we maybe use something like `loaderIntersectionTrigger`? with values like initial, half, all/full?

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -113,8 +115,15 @@ const withInfiniteScroll = (
...
+
+      const thresholds = {
+        small: 0.01,
+        half: 0.5,
+        full: 0.99, // using 0.99 instead of 1 to avoid problems with float precision in IE11
+      };
+      const displaySize = this.props.loaderMinDisplay || 'full';
       this.observer = new IntersectionObserver(this.handleIntersection, {
-        threshold: 0.99, // using 0.99 instead of 1 to avoid problems with float precision in IE11
+        threshold: thresholds[displaySize],
```
tiagoengel: It's worth adding a default here, the prop type will only trigger a warning if the wrong value is passed and in that case the component will break. Can be something like `thresholds[displaySize] || thresholds .full`.

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -67,6 +68,7 @@ const propTypes = {
   initiallyLoadedElements: PropTypes.number,
   elementsPerScroll: PropTypes.number,
   dataSource: PropTypes.instanceOf(DataSource).isRequired,
+  loaderMinDisplay: PropTypes.oneOf(['small', 'half', 'full']),
```
tiagoengel: I think we need a defaultValue for this prop too, no? Should be `full` by default.

packages/bpk-component-infinite-scroll/README.md
arnaugm: I was not very proud of the name either, `loaderIntersectionTrigger` sounds better.
`initial` is not accurate because the intersector uses percentages and if for some reason you start displaying the element from below it will trigger as well, that's why small instead

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
arnaugm: true after several changes I missed the case where the prop is not empty but invalid

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
arnaugm: true

arnaugm:
Comments amended and tests added

arnaugm:
`UNRELEASED` entry added

PR #1298, opened 2019-01-16
[BPK-2168] Allow docs to remember selected platform

georgegillams:
Behaviour:
 - If a user navigates to a component page with no preference already set, it will show the first platform available. Their preference will remain unset.
 - If a user navigates to a component page with a preference already set, it will show their prefered platform (iff it exists), falling back to the first one available. Their preference will not change either way.
 - If a user navigates to a component page with the platform specified as a URL parameter, then the URL parameter will take precedence. Their current preference will remain, but won't be used in favour of the URL parameter.
 - If a user selects a platform from the horizontal nav, their platform preference will be set/updated
 - If a user selects a platform from the nav filter (Web or Native), their platform preference will be set/updated
 - If a user selects `all` from the nav filter, their platform preference will remain the same.
 - If a user navigates to a component page with a preference already set, the nav list filter will be set to `all` regardless

This should provide a better experience for most consumers.
Those only interested in Android may notice no improvement, as they are already used to Android being selected by default!

### DIFF
--- a/packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
+++ b/packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
@@ -24,6 +24,10 @@ import BpkHorizontalNav, {
   BpkHorizontalNavItem,
 } from 'bpk-component-horizontal-nav';
 import { cssModules } from 'bpk-react-utils';
+import {
+  setPlatformInLocalStorage,
+  getPlatformFromLocalStorage,
+} from '../../helpers/storage-helper';
 
 import Heading from '../Heading';
 import Blurb from './Blurb';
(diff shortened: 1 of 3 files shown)


packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -116,17 +120,28 @@ const DocsPageWrapper = props => {
     web: webSubpage,
   };
 
+  const platformPreferenceIsValid = platformPreference => {
```
georgegillams: Could probably think of a better name

packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
```diff
@@ -135,12 +150,14 @@ const DocsPageWrapper = props => {
   if (platformQueryParamMatches && platforms[platformQueryParamMatches[1]]) {
     const platformQueryParam = platformQueryParamMatches[1];
     initiallySelectedPlatform = platformQueryParam;
-    initiallyRenderedSubpage = platforms[platformQueryParam];
   } else {
     history.replace(`${path}?platform=${initiallySelectedPlatform}`);
   }
 
+  const initiallyRenderedSubpage = platforms[initiallySelectedPlatform];
```
georgegillams: Just refactored this as it didn't need to be handled by every `if` statement

PR #1300, opened 2019-01-17
Shift first and last date by 1ms to deal with edge cases

detzgk:
My last commit had a bug in it. areRangesOverlapping is not inclusive
(uses strict less-than comparison) and so right now changing the selection
from the very edge doesn't work properly.

To fix it I just enlarged the range we compare against by 1ms in either
direction.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,2 +1,7 @@
 # Unreleased
 
+**Fixed:**
+
+- bpk-component-calendar
+  - Fix edge cases with Week component
+
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -3,5 +3,5 @@
 **Fixed:**
 
 - bpk-component-calendar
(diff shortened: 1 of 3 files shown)


UNRELEASED.md
```diff
@@ -1,2 +1,7 @@
 # Unreleased
 
+**Fixed:**
+
+- bpk-component-calendar
+  - Fix edge cases with Week component
```
georgegillams: One tiny thing - might be worth mentioning that this is specific to range-selection

detzgk:
So the thing about range selection is that nothing in backpack currently uses it. We're developing the date range selection component in our own repo for now (the code is ugly, tightly coupled with stuff specific to our component and has no useful desktop version yet), but we still need to make sure the Week knows when to update.

So visually there aren't any changes to show in a storybook. If (when?) we finish the date range selector, we'll see if it should get pushed upstream to backpack.

PR #1305, opened 2019-01-28
[BPK-2212] New social-like icon

matthewdavidson:
cc @chrisbehr 
image


chucklehound:
Had a look at the SVG and it appears to be misaligned in the artboard - it's up at the top left of the 24px space rather than centered.  Can this be fixed and resubmitted?

matthewdavidson:
@chrisbehr fixed!

PR #1318, opened 2019-02-05
[BPK-2225] Smoothly scroll in selected platform

georgegillams:


georgegillams:
The latest version of Safair supports `scrollIntoView`, but does not support `scrollIntoViewOptions`. As a result the behaviour is undesirable, so closing this PR.

Safari beta does support this, so I plan to reopen when that's released.

PR #1323, opened 2019-02-06
[BPK-2225] Janky scrolly thingy

georgegillams:
~This doesn't work on Chrome unless we scroll the nav into view. This is not ideal, given that it may scroll users back up the page if following a # link.~
Does work on Chrome now that I've removed the "smooth" aspect of the scrolling behaviour.

Still disabled on IE and Edge but no one uses those on mobile anyway 😆 

[Archive.zip](https://github.com/Skyscanner/backpack/files/2836710/Archive.zip)

### DIFF
--- a/packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
+++ b/packages/bpk-docs/src/components/DocsPageWrapper/DocsPageWrapper.js
@@ -20,9 +20,6 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { withRouter } from 'react-router-dom';
 import BpkContentContainer from 'bpk-component-content-container';
-import BpkHorizontalNav, {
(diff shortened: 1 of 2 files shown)


matthewdavidson:
So it only works on Safari and Firefox?

georgegillams:
@matthewdavidson pretty much 😞

matthewdavidson:
We can implement something similar using [scrollLeft](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft) (this would need appropriate RTL consideration too).

georgegillams:
> We can implement something similar using [scrollLeft](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft) (this would need appropriate RTL consideration too).

That's gonna involve changing the MobileScrollContainer component 😢

PR #1326, opened 2019-02-08
Fix overflowing docs pages on narrow devices

georgegillams:
I've replaced `marked` with `react-markdown`, which is a react wrapper around `marked`.
This gives us more control over how MD is rendered on our docs and information pages, and saves us the headache of dangerously setting inner HTML.

Wherever we allow the renderer to do it's default things, our bareHtml metal styles are applied. So there should be no visual difference due to the swap 👍


packages/bpk-mixins/src/mixins/_typography.scss
```diff
@@ -591,6 +591,7 @@
 
 @mixin bpk-table__cell {
   padding: $bpk-spacing-sm;
+  overflow: hidden;
```
georgegillams: obviously we would need to assess the blast radius of this change - as the mixin could be used in other components...

shaundon:
@georgegillams is this still valid? Might be worth having a before/after video or gif to illustrate the change if so

georgegillams:
https://github.com/Skyscanner/backpack/pull/1326#issuecomment-467446142
Yeah - sorry I'd kinda forgotten about this.
There are actually videos in the attached zip 👍

packages/bpk-mixins/src/mixins/_typography.scss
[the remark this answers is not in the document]
georgegillams: That's a great shout!

UNRELEASED.md
```diff
@@ -1,2 +1,6 @@
 # Unreleased
 
+**Fixed:**
+  - bpk-mixins:
+    - Table cell mixin will now prevent content from overflowing into the next cell in webkit browsers, providing scroll behaviour instead.
```
```suggestion
    - Table cell mixin will now prevent content from overflowing into the next cell in WebKit browsers, providing scroll behaviour instead.
```

packages/bpk-mixins/src/mixins/_typography.scss
```diff
@@ -591,6 +591,11 @@
 
 @mixin bpk-table__cell {
   padding: $bpk-spacing-sm;
+  overflow: auto;
+
+  &::-webkit-scrollbar {
+    display: none;
```
shaundon: Why are we hiding the scrollbar?

georgegillams: Good question - probably one for @matthewdavidson...

georgegillams: It doesn't seem to actually have any effect...

shaundon: Did you try it in Safari?

georgegillams: Ah - only tried it on mobile safari before, and it makes no difference there.
On desktop it does, but tbh I'm not sure it's the right thing to do. Maybe we should just leave it be? Scrollbars in MacOS are hardly invasive...

shaundon: Yeah I think leave it in - I'm concerned it'll harm a11y to remove them.

matthewdavidson: Can we get design input into this please? We need to choose an option that is the best for all browsers / platforms. I.e. what does this look like on chrome for windows?

georgegillams: AFAI can tell the `webkit-scrollbar` selector has no effect in edge.
Hiding the scroll bars in Safari and Chrome seems a little pointless as they are so unobtrusive anyway (ie they are invisible unless actively scrolling)

matthewdavidson: > (ie they are invisible unless actively scrolling)

Have you checked this on windows?

shaundon: Yeah invisible scrollbars is a macOS / iOS thing afaik

georgegillams: tbh - the more I think about this the more I think that the whole table should be scrolling, not the individual cell's content
Then we don't have to screw with scrollbars as the user / system setting will be approriate

matthewdavidson: Perhaps we could use the mobile scroll container ^^

georgegillams: Agreed (this was just a really quick gif I made in dev tools 😉)

georgegillams: Dogfooding yo!
https://www.reddit.com/r/funny/comments/b0ksok/pure_betrayal/?utm_source=share&utm_medium=ios_app

shaundon:
Happy to approve once conflicts are resolved.

PR #1327, opened 2019-02-11
Add BpkMapMarker component

sogame:
Hi,

This is the initial work to add markers to the map component, to get your feedback about the best approach.

I've added the component `BpkMapMarker` inside `bpk-component-map`, as I think this doesn't make sense without a map. Are you ok with this, or should I move it into its own component package (like `bpk-component-map-marker`)?

To change the marker color there's a prop (named `themeColor`), as this is the way we're currently doing it, but I guess it should be replaced with theming, right? I might need a bit of help with this, I've never build a themeable component 😅 

Please, let me know any other concern or wrong approach :-)

--------------

Edits from Shaun:

* Added example to docs site
* Add to the readme
* Replace theme colour with primary/secondary type
* Add support for theming via theming API
* Remove `children`, add required `icon` prop


georgegillams:
Hey @sogame - I think you're right. If at some point consumers want to use it elsewhere we can refactor it into a separate package, but for now it can live with the map 👍

packages/bpk-component-map/src/BpkMapMarker.js
```diff
@@ -0,0 +1,105 @@
...
+  const classNames = [STYLES['bpk-map-marker']];
+  if (onClick) {
+    classNames.push(STYLES['bpk-map-marker--dynamic']);
+  }
+  if (large) {
+    classNames.push(STYLES['bpk-map-marker--large']);
+  }
+  if (className) {
+    classNames.push(className);
+  }
```
shaundon: We've got a simpler way of defining this logic now:

```
const classNames = [
  STYLES['bpk-map-marker'],
  onClick && STYLES['bpk-map-marker--dynamic'],
  large && STYLES['bpk-map-marker--large'],
  className,
];
```

shaundon:
Looks good from the diff - just gonna pull it down to run locally now. 

I can put together a PoC to show how it can work with the theming API too.

packages/bpk-component-map/src/BpkMapMarker.js
sogame: Nice! 👏

sogame:
Awesome, thanks @shaundon! I just migrated the code as we have it now to get initial feedback, I haven't checked any existing component to see how to implement theming 😅 

But maybe some documentation could be added in `contributing.md` or some other place? It would be useful, if new components should use it 😝

packages/bpk-component-map/src/BpkMapMarker.js
sogame: Updated :-)

shaundon:
I feel like the `large` prop is a bit strange right now - as it's effectively a modifier, but it's enabled by default. In other components, when there's a `large` prop it's disabled by default, or we invert the logic and make it a `small` prop instead.

So I propose that here we do one of the following:

1) Make it a `small` prop (and change the underlying stylesheet so that `--small` is a classname that gets applied instead of `--large`).

2) Make `large` default to `false`.

What do you think?

packages/bpk-component-map/src/BpkMapMarker.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  &--dynamic {
+    cursor: pointer;
+  }
+
+  &__wrapper {
+    position: relative;
+    display: flex;
+    padding-bottom: 5px; // stylelint-disable-line unit-blacklist
```
shaundon: I understand why you need to opt out of the rule here - could you add a quick, one-line comment to explain it's because of the arrow positioning.

Our convention for eslint is to do this every time we opt out, I'll change the convention to include stylelint too: https://github.com/Skyscanner/backpack/blob/master/decisions/eslint-ignores.md

sogame:
Yes, I also noticed this when migrating the component, but I forgot to mention it when submitting the pr 😅 

Both options would work fine, so, which one do you think follows better the Backpack conventions?

packages/bpk-component-map/src/BpkMapMarker.scss
sogame: Sure, makes total sense 👍

sogame: Added

shaundon:
I'd go with option 2 - make `large` default to `false` :)

sogame:
Updated to make `large` false by default

packages/bpk-component-map/src/BpkMapMarker.js
```diff
@@ -0,0 +1,98 @@
...
+    themeColor,
+    className,
+    arrowClassName,
+    large,
+    onClick,
+    ...rest
+  } = props;
+
+  const classNames = [
+    STYLES['bpk-map-marker'],
```
sogame: @shaundon Should I use the `cssModules` function instead of accessing `STYLES` directly?

shaundon: Yes please! Here's an example file that demonstrates it: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-horizontal-nav/src/BpkHorizontalNavItem.js#L48

packages/bpk-component-map/package.json
```diff
@@ -13,6 +13,7 @@
     "react": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
+    "bpk-component-icon": "^3.26.2",
```
shaundon: Should be a dev dependency as it's only needed in the dev environment.

packages/bpk-component-map/src/BpkMapMarker.js
```diff
@@ -0,0 +1,105 @@
...
+    onClick,
+    type,
+    ...rest
+  } = props;
+
+  const classNames = getClassName(
+    'bpk-map-marker',
+    `bpk-map-marker--${type}`,
+    onClick && 'bpk-map-marker--dynamic',
+    large ? 'bpk-map-marker--large' : null,
```
shaundon:
```suggestion
    large && 'bpk-map-marker--large',
```

PR #1328, opened 2019-02-12
[NO-JIRA] Fix submodule PR noise

georgegillams:
I'd like to propose a change to how we handle submodules. At the moment, `npm run docs` causes a git submodules `pull`. This causes the submodules to be updated to the latest version of each project, and if this is checked in it adds noise to PR's.

Instead, if we use a git submodule `init` instead of `pulling`, the submodules content won't change and we won't get PR noise! Getting the latest version of submodules would require manual contributor action.

To ensure that submodules are always up to date for docs builds, Travis would need to pull submodules before building.


.travis.yml
```diff
@@ -14,6 +14,7 @@ script:
   - ./scripts/check-pristine-state package-lock.json
   - npm run danger
   - npm test
+  - npm run submodules:pull
```
matthewdavidson: This change means we are giving up the ability to control which changes we push to the docs.

Previously we were 100% in control (perhaps at a burden). I want to see others opinions on this.

tiagoengel: My only problem with this is we run the risk of testing locally with an outdated version and maybe find a problem only after it is published with the latest version of the submodules. At the moment we always test with the latest version locally and this ensures we don't run into this problem

package.json
```diff
@@ -44,6 +44,7 @@
     "danger": "danger ci",
     "prettier": "prettier --config .prettierrc --write \"**/*.js\"",
     "create-component": "node scripts/npm/create-component.js",
+    "submodules:init": "git submodule update --init --recursive",
     "submodules:pull": "git submodule foreach git pull origin master",
     "submodules:update": "npm run submodules:unlock && npm run submodules:pull && git add backpack-android && git add backpack-ios && git commit -m\"[skip ci] Update submodules\" && npm run submodules:lock",
```
tiagoengel: Unrelated but `backpack-react-native` is missing here

package.json
```diff
@@ -30,8 +30,8 @@
     "storybook:dist": "build-storybook -c .storybook -o dist/storybook",
     "storybook": "start-storybook -p 9001",
     "sassdoc": "sassdoc {packages/bpk-mixins/src/**/*,packages/bpk-svgs/dist/*,packages/bpk-tokens/tokens/base.default}.scss -d dist/sassdoc -v --strict",
-    "docs": "npm run submodules:pull && BPK_BUILT_AT=$( date -u +%s ) webpack-dev-server --open",
-    "docs:dist": "npm run submodules:pull && npm run clean:dist && NODE_ENV=production BPK_BUILT_AT=$( date -u +%s ) webpack --progress --display=minimal --bail && npm run sassdoc",
+    "docs": "npm run submodules:init && BPK_BUILT_AT=$( date -u +%s ) webpack-dev-server --open",
```
tiagoengel: If I understand correctly this will make it impossible to test with the latest version of the submodules, unless we first update and merge the submodules, and rebase the branch. Is that right?

.travis.yml
georgegillams: @matthewdavidson previously travis would have always pulled the latest version anyway...

package.json
georgegillams: So you could still test with the latest ones by running `npm run submodules:pull`

georgegillams: Happy to close this though if we don't think it improves the current process

tiagoengel: But if the docs command runs init again it will revert back to the current version no? I'm ok with the change if we can somehow still test with the latest locally

georgegillams: I see what you mean... 🤔

package.json
```diff
@@ -44,6 +44,7 @@
     "danger": "danger ci",
     "prettier": "prettier --config .prettierrc --write \"**/*.js\"",
     "create-component": "node scripts/npm/create-component.js",
+    "submodules:init": "git submodule update --init --recursive",
```
georgegillams: So in hind-sight, @matthewdavidson and @tiagoengel, I think it's best we leave the scripts as they are. However, I am going to propose adding this script to make it easy for us to undo accidentally commited submodule changes.

PR #1331, opened 2019-02-12
[NO-JIRA] Add helpful npm script

georgegillams:
With this change, `npm run submodules:reset` will undo any accidentally committed changes to submodules, allowing us to keep our PR's cleaner 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,8 +44,9 @@
     "danger": "danger ci",
     "prettier": "prettier --config .prettierrc --write \"**/*.js\"",
     "create-component": "node scripts/npm/create-component.js",
+    "submodules:reset": "git checkout origin/master backpack-android backpack-ios backpack-react-native && git submodule update --init --recursive",
     "submodules:pull": "git submodule foreach git pull origin master",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -44,8 +44,9 @@
     "danger": "danger ci",
     "prettier": "prettier --config .prettierrc --write \"**/*.js\"",
     "create-component": "node scripts/npm/create-component.js",
+    "submodules:reset": "git checkout origin/master backpack-android backpack-ios backpack-react-native && git submodule update --init --recursive",
     "submodules:pull": "git submodule foreach git pull origin master",
-    "submodules:update": "npm run submodules:unlock && npm run submodules:pull && git add backpack-android && git add backpack-ios && git add backpack-react-native && git commit -m\"[skip ci] Update submodules\" && npm run submodules:lock",
+    "submodules:update": "npm run submodules:unlock && npm run submodules:pull && git add backpack-android backpack-ios backpack-react-native && git commit -m\"[skip ci] Update submodules\" && npm run submodules:lock",
```
georgegillams: Also, `git add` accepts any number of path args, so no need to call the command 3 times.

PR #1337, opened 2019-02-18
[BPK-2250] Document primary text buttons for nav bar

shaundon:
screen shot 2019-02-18 at 11 35 55


packages/bpk-docs/src/pages/NativeNavigationBarPage/NativeNavigationBarPage.js
```diff
@@ -78,6 +79,21 @@ const components = [
       },
     ],
   },
+  {
+    id: 'primary-text-buttons',
+    title: 'Primary text buttons',
+    blurb: 'On iOS, text buttons can have an alternate colour.',
```
matthewdavidson: It's not really alternate is it? It's "primary".

matthewdavidson:
Oops that was supposed to be a "comment" as opposed to an "approval"

packages/bpk-docs/src/pages/NativeNavigationBarPage/NativeNavigationBarPage.js
shaundon: Yeah fair, I'll change.

PR #1338, opened 2019-02-18
[No JIRA] Add mixin for primary color

k0nserv:


UNRELEASED.md
```diff
@@ -5,3 +5,13 @@
 
 - bpk-component-map:
   - Added the component `BpkMapMarker`.
+
+- bpk-mixins:
+  - Add a mixin for primary color use `bpk-primary-color-property`.
```
shaundon: Language here feels a bit weird..

```suggestion
  - Added a mixin for primary colour. Use `bpk-primary-color-property`:
```

maybe?

UNRELEASED.md
```diff
@@ -5,3 +5,13 @@
...
+    ```scss
+    @import '~bpk-mixins';
+
+    .selector {
+      @include bpk-primary-color-property(color);
```
shaundon: Might be worth making it clear that this applies it to `color` too. We could have something like:

```
.selector {
  // Instead of
  // color: $bpk-color-blue-500;
  include bpk-primary-color-property(color);
}
```

shaundon: Sorry, my last comment wasn't clear. What I meant is that I didn't understand at a glance that the `(color)` argument means it applies the mixin to the CSS color property, which is why I recommended an inline comment to clarify.

k0nserv: Yeah makes sense, I've changed it as requested. Also happy to change the name of the mixin if you have a better one in mind

shaundon: I can't really think of a better name either.. maybe something like `bpk-apply-primary-color-to(property)` but then it's a bit Objective-C sounding, it wouldn't really fit with the other mixins.

matthewdavidson: What about having two mixins? One for `color` and one for `background-color`?

matthewdavidson: Specifically: `@include bpk-primary-color;` and `@include bpk-primary-background-color;`

shaundon: Yeah not a bad idea Matt. We could always leave this mixin in to cover any edge cases we can't think of (e.g. `border-color`).

matthewdavidson: Scrap that, there are too many variants. Equally `bpk-primary-color-property($property-name)` doesn't fit well with shorthand properties i.e. `border: $width, $style, $color;`

k0nserv: I think Shaun's suggestion is pretty good

```sass
@include bpk-apply-primary-color-to(color);

@include bpk-apply-primary-color-to(background);

@include bpk-apply-primary-color-to(fill);

@include bpk-apply-primary-color-to(background-color);

@include bpk-apply-primary-color-to(border-color);
```

shaundon: @k0nserv happy to go with that.

PR #1339, opened 2019-02-21
[No JIRA] Clearer ignore logic

shaundon:


decisions/README.md
```diff
@@ -1,18 +1,3 @@
...
-* [Adding types](adding-types.md)
-* [Backpack bot access token used by Danger](backpack-bot-api-token-for-danger.md)
-* [Component SCSS filenames](component-scss-filenames.md)
-* [How we ignore `eslint` rules](eslint-ignores.md)
-* [Naming JS files in components](js-filenames.md)
-* [Naming convention for theme attributes](theme-props-naming.md)
-* [Parallel builds](parallel-builds.md)
-* [Type Ramp](tracking.md)
-* [We don't set line heights for native typography](native-line-heights.md)
-* [Writing docs](writing-docs.md)
```
shaundon: We're always forgetting to update this so I think we should lose it and just use the directory itself as a table of contents.

PR #1357, opened 2019-03-12
Add labels to Phone Input component

chiraggupta:
As discussed with @georgegillams and @k0nserv this is a breaking change to `BpkPhoneInput`. It adds labels to both Dialing code `BpkSelect` and Telephone number `BpkInput`.

phone_input_with_labels

With long labels:
Screenshot 2019-03-13 at 15 44 28

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,2 +1,4 @@
 # Unreleased
 
+- bpk-component-phone-input:
+  - Add labels for dialing code and telephone number.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,4 +1,6 @@
 # Unreleased
 
+**Breaking:**
+
 - bpk-component-phone-input:
-  - Add labels for dialing code and telephone number.
+  - Add labels for dialing code and telephone number as they are preferred over placeholders per UX guidelines.
--- a/packages/bpk-component-phone-input/README.md
+++ b/packages/bpk-component-phone-input/README.md
@@ -47,7 +47,6 @@ export default class extends Component {
         id="phone-input-id"
         name="Telephone input"
         label="Telephone number"
-        placeholder="Telephone number"
         onChange={this.onChange}
         onDialingCodeChange={this.onDialingCodeChange}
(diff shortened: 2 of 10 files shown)


packages/bpk-component-phone-input/README.md
```diff
@@ -30,23 +30,23 @@ const getFlag = dialingCode => {
...
-
+  };
 
   render() {
     return (
       <BpkPhoneInput
         id="phone-input-id"
         name="Telephone input"
+        label="Telephone number"
         placeholder="Telephone number"
```
georgegillams: Now that the label is displayed, the placeholder is unnecessary. Could we remove it from the stories, readme, docs and tests?

UNRELEASED.md
```diff
@@ -1,2 +1,4 @@
 # Unreleased
 
+- bpk-component-phone-input:
+  - Add labels for dialing code and telephone number.
```
georgegillams: It may be worth mentioning that these make placeholder redundant.

Also, could we add a `**Breaking:**` title, and explicitly call out that labels are now required.

georgegillams:
Sorry - only just caught this.
The behaviour when the viewport becomes narrower has changed slightly. Previously the dialing code selection was a fixed width, but now it's being squashed

Before:


After:

chiraggupta:
> Sorry - only just caught this.
> The behaviour when the viewport becomes narrower has changed slightly. Previously the dialing code selection was a fixed width, but now it's being squashed
> 
> Before:
> 
> 
> After:
> 

Ah interesting, I'm glad you noticed! Let me take a look.

k0nserv:
Worth consdering a few translations of `Dialing code` or just a double length pseudo string of `Dialing code Dialing Code`.

chiraggupta:
@georgegillams i changed the behaviour so it's similar to old one.
phone_input_shrink

chiraggupta:
New behaviour for small screens

Screenshot 2019-03-14 at 11 08 36

Screenshot 2019-03-14 at 11 09 50

PR #1358, opened 2019-03-15
[No JIRA] Fix spelling

shaundon:
No, **you're** beginning to regret doing this because it's a pain to set up (jokes aside the Android and iOS repos should be much easier as they have far fewer .md files).


packages/bpk-component-barchart/README.md
```diff
@@ -49,7 +49,7 @@ export default () => (
 | className                               | string                                | false    | null             |
 | leadingScrollIndicatorClassName         | string                                | false    | null             |
 | trailingScrollIndicatorClassName        | string                                | false    | null             |
-| [outlierPercentage](#outlierpercentage) | number                                | false    | null             |
+| [`outlierPercentage`](#outlierpercentage) | number                                | false    | null             |
```
k0nserv: Why should these have code style when those that aren't likes don't

shaundon: Ideally, *all* references to code should be in backticks and then we wouldn't need that huge list of terms in `.spelling` because it ignores anything in backticks. In this specific case I was using the interactive CLI to escape other uses of this and it offers to change others to the same format, which I must have accidentally pressed. I'll change them back.

PR #1363, opened 2019-03-18
Add "buttonProps" prop to "BpkMapMarker"

sogame:
Like in other components, it might be useful to be able to send arbitrary props to another dom element rather than the root one.

In our case, we would need to set some attribute to run some acceptance tests.

### DIFF
--- a/packages/bpk-component-map/src/BpkMapMarker.js
+++ b/packages/bpk-component-map/src/BpkMapMarker.js
@@ -44,6 +44,7 @@ type Props = {
   arrowClassName: ?string,
   large: ?boolean,
   onClick: ?(event: SyntheticEvent<>) => mixed,
+  buttonProps: ?{ [string]: any },
 };
 
 const BpkMapMarker = (props: Props) => {
@@ -55,6 +56,7 @@ const BpkMapMarker = (props: Props) => {
     large,
     onClick,
     type,
+    buttonProps,
     ...rest
   } = props;
 
@@ -74,6 +76,7 @@ const BpkMapMarker = (props: Props) => {
         type="button"
         className={getClassName('bpk-map-marker__wrapper')}
         onClick={onClick}
+        {...buttonProps}
       >
         <div className={classNames}>{icon}</div>
         <div className={arrowClassNames}>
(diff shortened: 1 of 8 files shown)


packages/bpk-component-map/src/BpkMapMarker.js
```diff
@@ -55,6 +56,12 @@ const BpkMapMarker = (props: Props) => {
     large,
     onClick,
     type,
+    buttonProps: {
+      type: buttonPropsType,
+      className: buttonPropsClassName,
+      onClick: buttonPropsOnClick,
+      ...restButton
```
sogame: Following the approach used in other components (like [BpkDatepicker](https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-datepicker/src/BpkDatepicker.js#L107) or [BpkBreadcrumbItem](https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-breadcrumb/src/BpkBreadcrumbItem.js#L52)), `buttonProps` would be send directly to `button`, but I think it makes sense to extract the props that are already set in the component, to avoid overwriting them.

It's just a proposal, so let me know if you think it's best not to do it and follow the same approach from other components :-)

shaundon: I'm having a difficult time picturing it - so if we went with your proposal, what would the proptypes look like?

sogame: My idea was to keep it just as an object (like the other examples), as in fact I wouldn't expect any of this "props" to be defined in `buttonProps`. This is just a "safety" measure to make sure users don't overwrite the props defined in the component.

But maybe it's not needed to be that "safe" 😅 

I see that Flow has complained, so not sure what's best: trying to make the proptype more complex or simply keep it as an object and trust the users.

What do you think?

shaundon: Yeah I think it might be best to just make it looser and keep it as an object, it'll make for an simpler API and if users overwrite things, it's on them 😅

sogame: Makes sense, I've changed it to send all props like in the other components 😅

packages/bpk-component-map/src/BpkMapMarker.js
```diff
@@ -100,6 +104,7 @@ BpkMapMarker.defaultProps = {
   large: false,
   onClick: null,
   type: MARKER_TYPES.primary,
+  buttonProps: {},
```
shaundon: Nitpick: This should be `null` by default as the Flow type indicates it's nullable. Having `{}` as the default means it can never be null.

sogame: Updated :-)

PR #1375, opened 2019-03-28
Add a default font color to BpkBadge

sogame:
I noticed a possible "bug" in `BpkBadge`: currently the color is not set (only for `destructive` and `outline`) so the font color relies on the one defined in the parent. This can lead to the component not being rendered as expected if the base font color has been changed:

image

I think the font color should be defined in the component, to avoid side-effects. What do you think?

The change submitted is just a proposal to clearly show what I mean. Not sure if the correct implementation should define a new token for the color, use theming...

### DIFF
--- a/packages/bpk-mixins/src/mixins/_badges.scss
+++ b/packages/bpk-mixins/src/mixins/_badges.scss
@@ -35,6 +35,7 @@
   padding: $bpk-badge-padding-y $bpk-badge-padding-x;
   align-items: center;
   background-color: $bpk-badge-background-color;
+  color: $bpk-font-color-base;
 
   @include bpk-border-radius-xs;
   @include bpk-text;


matthewdavidson:
This is an interesting one @sogame  - i want more opinions on it.

shaundon:
I'm happy with it. Any side effects where colours are cascading from parent elements is almost certainly undesirable from a design point of view. We wouldn't allow people to customise the text colour of a button, this is the same thing IMO.

tiagoengel:
I think is good idea. We can always add a theme variable if we actually want to change the text colour in the future.

shaundon:
Needs an entry in `UNRELEASED.md` but I can do that after it's merged 😄

sogame:
Thanks @shaundon! I didn't added any entry in `UNRELEASED.md` because I was not sure about the approach used 😅

I was going to add it now, but I'm not sure if I should put it under `bpk-mixins` (as it's the component I've changed) or under `bpk-component-badge` (as it's the component the change impacts).

Also, a note for when you publish it: the change is just a "fix", but given some consumer might be using it to change the text color, should this be published as a major change?

shaundon:
> given some consumer might be using it to change the text color, should this be published as a major change?

We think patch should be fine - any consumers changing the font colour currently are using unsupported behaviour so it doesn't count as breaking.

PR #1381, opened 2019-04-01
Add altText prop to BpkBackGroundImage

DewarM:
As described in the [documentation](https://backpack.github.io/components/image?platform=web#default) `altText` is a required prop on the `BackgroundImage` component.

`alt` was not being applied correctly to the div. This PR fixes this issue.
A test has been included and other test cases updated to include this required prop.

### DIFF
--- a/packages/bpk-component-image/src/BpkBackgroundImage-test.js
+++ b/packages/bpk-component-image/src/BpkBackgroundImage-test.js
@@ -34,6 +34,7 @@ describe('BpkBackgroundImage', () => {
             height: '20rem',
           }}
           src="./path/to/image.jpg"
+          altText="Default alt text"
         >
           <div
             style={{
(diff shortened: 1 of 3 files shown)


shaundon:
I've done some research and it seems like alt text for background images [generally isn't a good idea](http://davidmacd.com/blog/alternate-text-for-css-background-images.html).

From the CSS spec:

>  The CSS background-image property was designed for decorative purposes and it is not possible to associate text alternatives with images that are included via CSS. Text alternatives are necessary for people who cannot see images that convey important information. Therefore, it is a failure to use this property to add images to convey important information.

So I think we should probably close this PR - is that ok?

DewarM:
Hey makes sense! Wasn't aware of this. Close away! We should probably update the docs to reflect that this prop is not required.

shaundon:
> We should probably update the docs to reflect that this prop is not required.

Yep definitely, feel free to make another PR for that, or we can do it 😇

PR #1384, opened 2019-04-03
[BPK-2350] Text input validation

ojcurtis:
Updated the mixins to change the styling of input fields to highlight red when an invalid input is entered into the fields.

The field will add a red border and the background will change red when incorrect entry is input, please see below screenshot as result of the change.

Screenshot 2019-04-03 at 17 07 43

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,7 +9,9 @@
   - Added a default font colour to badges to prevent the badge colour changing due to cascading styles.
 
 **Breaking:**
-- bpk-mixins, bpk-component-input, bpk-component-fieldset:
+- bpk-mixins:
+- bpk-component-input: 
+- bpk-component-fieldset:
     - `_forms.scss`:
         - Added `border-color` to `bpk-input--invalid` mixin for input fields to highlight red when there is an error on the input.
         - Updated `background` to `bpk-input--invalid` mixin for fields to change the color from white to red.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,3 +15,5 @@
     - `_forms.scss`:
         - Added `border-color` to `bpk-input--invalid` mixin for input fields to highlight red when there is an error on the input.
         - Updated `background` to `bpk-input--invalid` mixin for fields to change the color from white to red.
+
+**Note**: The API hasn't changed, it's only breaking because it's a major visual change.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,7 +9,7 @@
   - Added a default font colour to badges to prevent the badge colour changing due to cascading styles.
 
 **Breaking:**
-- bpk-mixins:
-    - Added `border-color` to input fields to highlight red when there is an error on the input
-    - Updated `background` to change the color from white to red
-    - Added `background-color` to set background to red and also add transparency to the field
+- bpk-mixins, bpk-component-input, bpk-component-fieldset:
+    - `_forms.scss`:
+        - Added `border-color` to `bpk-input--invalid` mixin for input fields to highlight red when there is an error on the input.
+        - Updated `background` to `bpk-input--invalid` mixin for fields to change the color from white to red.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -7,3 +7,9 @@
 **Fixed:**
 - bpk-component-badge:
   - Added a default font colour to badges to prevent the badge colour changing due to cascading styles.
+
+**Breaking:**
+- bpk-mixins:
+    - Added `border-color` to input fields to highlight red when there is an error on the input
+    - Updated `background` to change the color from white to red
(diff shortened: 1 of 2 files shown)


UNRELEASED.md
```diff
@@ -7,3 +7,9 @@
 **Fixed:**
 - bpk-component-badge:
   - Added a default font colour to badges to prevent the badge colour changing due to cascading styles.
+
+**Breaking:**
+- bpk-mixins:
+    - Added `border-color` to input fields to highlight red when there is an error on the input
```
k0nserv: Could all of these three be more specific about which state of the input and which mixins specifically that is changing? Can we also end each line with a period?

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -108,7 +108,9 @@
 
 @mixin bpk-input--invalid {
   padding-right: $bpk-spacing-base;
-  background: $bpk-color-white url(map-get($bpk-icons-sm, exclamation-circle-red-500)) no-repeat right $bpk-spacing-sm center;
+  border-color: $bpk-color-red-500;
+  background: $bpk-color-red-500 url(map-get($bpk-icons-sm, exclamation-circle-red-500)) no-repeat right $bpk-spacing-sm center;
+  background-color: rgba($bpk-color-red-500, .1);
```
matthewdavidson: Shouldn't line 113 be merged with 112:
```
  background: rgba($bpk-color-red-500, .1) url(map-get($bpk-icons-sm, exclamation-circle-red-500)) no-repeat right $bpk-spacing-sm center;

```

UNRELEASED.md
matthewdavidson: Also `bpk-component-input` should be listed under this breaking change.

matthewdavidson: And I guess also `bpk-component-fieldset`? 🤔

UNRELEASED.md
```diff
@@ -7,3 +7,9 @@
 **Fixed:**
 - bpk-component-badge:
   - Added a default font colour to badges to prevent the badge colour changing due to cascading styles.
+
+**Breaking:**
+- bpk-mixins, bpk-component-input, bpk-component-fieldset:
```
k0nserv: We usually don't list these like this instead we use

```
- bpk-mixins:
- bpk-component-input:
- bpk-component-fieldset:
```

matthewdavidson: LGTM otherwise

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -108,7 +108,8 @@
 
 @mixin bpk-input--invalid {
   padding-right: $bpk-spacing-base;
-  background: $bpk-color-white url(map-get($bpk-icons-sm, exclamation-circle-red-500)) no-repeat right $bpk-spacing-sm center;
+  border-color: $bpk-color-red-500;
+  background: rgba($bpk-color-red-500, .1) url(map-get($bpk-icons-sm, exclamation-circle-red-500)) no-repeat right $bpk-spacing-sm center;
```
georgegillams: I'm not sure we should be actually making this transparent - I think it could have strange side-effects. Could we use red50 or red100 so that it's not actually transparent?
@matthewdavidson @k0nserv wdyt?

shaundon: That's my bad - I assumed transparent would work but I was wrong for the reasons you said above. Good catch 😄

ojcurtis: Sure I can change it to that :) I also based it off the designs in zeplin that had it at red-500 with .1 opacity, but we can sure try the approach from @georgegillams

UNRELEASED.md
```diff
@@ -7,3 +7,11 @@
...
 - bpk-component-badge:
   - Added a default font colour to badges to prevent the badge colour changing due to cascading styles.
+
+**Breaking:**
+- bpk-mixins:
+- bpk-component-input: 
+- bpk-component-fieldset:
+    - `_forms.scss`:
+        - Added `border-color` to `bpk-input--invalid` mixin for input fields to highlight red when there is an error on the input.
+        - Updated `background` to `bpk-input--invalid` mixin for fields to change the color from white to red.
```
shaundon: I think we should probably add another quick note here just to make it clear to consumers about what they need to do. This the the place they come to find out what's changed, how to migrate etc, so something along the lines of:

> Note: The API hasn't changed, it's only breaking because it's a major visual change.

Would be good to have

ojcurtis: Where would you advise this to go? Would I add the note underneath or would you have a recommendation on where this should be?

shaundon: Yeah just pop it on a line underneath, we don't have a hard-and-fast rule about it, so long as it's somewhere people will see it.

PR #1389, opened 2019-04-05
[NO-JIRA] Fix readme table borders

georgegillams:
_Note the screenshots below use an extra-wide border to make things more visible..._

Currently the borders around our MD tables are cut off by the mobile scroll container. This is due to us uising `box-shadow` to apply borders, which are not included in width calculations.

Screenshot 2019-04-05 at 09 27 10

This PR adds an actual `border` to the table elements, so that the browser includes the border width when calculating the size of the element.

Screenshot 2019-04-05 at 09 28 15

### DIFF
--- a/packages/bpk-docs/src/components/DocsPageBuilder/BpkMarkdownRenderer.scss
+++ b/packages/bpk-docs/src/components/DocsPageBuilder/BpkMarkdownRenderer.scss
@@ -20,9 +20,16 @@
 
 .bpkdocs-markdown-renderer {
   &__table {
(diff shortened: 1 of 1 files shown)


georgegillams:
> Looks good, the only thing is that the border of the other "boxes" on the readme are a bit darker, or thicker. Can we make this one the same?

Good catch. Fixed the color. Unfortunately it's hard to make the border radius the same because of this in the specification:
```
The border-radius property does not apply to table elements when border-collapse is collapse.
```

PR #1390, opened 2019-04-05
When calling `onClear`, provide input `name`

sogame:
We're building a dynamic form and we want to use clearable input fields. It would be good to use the same callback for all the fields:

```
  onClear = evt => {
    const { onChange } = this.props;
    const { name } = evt.target;
     onChange(name, null);
  };
```

but the event target is the svg (not the input), so `evt.target.name` doesn't contain the field name.

This PR contains two possible ways of achieving this:
* Set `evt.target.name` to the input name
* Provide a second parameter to `onClear` with the input name

Would you be fine with any of these two proposals? Is there a better way to achieve this?

### DIFF
--- a/packages/bpk-component-input/src/BpkInput-test.js
+++ b/packages/bpk-component-input/src/BpkInput-test.js
@@ -91,12 +91,7 @@ describe('BpkInput', () => {
     jest.spyOn(console, 'error').mockImplementation(() => jest.fn());
     const tree = renderer
       .create(
-        <BpkInput
-          id="test"
-          name="test"
-          value=""
-          clearButtonMode="whileEditing"
-        />,
+        <BpkInput id="test" name="test" value="" clearButtonMode="always" />,
       )
       .toJSON();
     expect(tree).toMatchSnapshot();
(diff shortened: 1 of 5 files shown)


packages/bpk-component-input/src/BpkInput-test.js
```diff
@@ -91,12 +91,7 @@ describe('BpkInput', () => {
     jest.spyOn(console, 'error').mockImplementation(() => jest.fn());
     const tree = renderer
       .create(
-        <BpkInput
-          id="test"
-          name="test"
-          value=""
-          clearButtonMode="whileEditing"
-        />,
+        <BpkInput id="test" name="test" value="" clearButtonMode="always" />,
```
sogame: Not related to the PR, but I noticed `clearButtonMode` is wrong according to the test name 😅

shaundon: Good catch!

shaundon:
I like the first approach, but I'm not sure if there's any best practice guidance about mutating native browser events, it might not be a best practice. I'm going to do some research and see if I can find out.

shaundon:
I can't seem to find anything to say we shouldn't mutate the native event so maybe this is fine!

sogame:
Conflicts in `unreleased.md` fixed :-)

PR #1419, opened 2019-04-22
[BPK-2377]: Updating styling for BpkPhoneInput

ojcurtis:
This is to allow for max-content on smaller devices so that the description field is not cut off and all content can be displayed

Updated stories to show examples of the longer description in fields


ojcurtis:
@tiagoengel - I noticed this is also the same on large screens too as it looks like the field sets based on any content in the field that could be large

PR #1421, opened 2019-04-22
[NO-JIRA] Fix flow for infinite scroll

tiagoengel:
I've introduced a test file for flow here, the idea is to test if required and default props are working as expected.

Lines with the `$ExpectError` annotation will suppress the error in the next line, if any, or show a warning saying there are no errors in the next line, which means the flow typing is wrong as we are expecting an error.

Because it only throw a warning in this case I have also changed the build to fail if any warnings are shown, so if we break the flow typing in any way the build will fail.

### DIFF
--- a/packages/bpk-component-infinite-scroll/src/withInfiniteScroll-test.flow.js
+++ b/packages/bpk-component-infinite-scroll/src/withInfiniteScroll-test.flow.js
@@ -1,0 +1,86 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2019 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow strict */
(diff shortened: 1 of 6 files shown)


packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -18,7 +18,12 @@
 
 /* @flow */
```
ojcurtis: Can you make this `/* @flow strict */`?

packages/bpk-component-infinite-scroll/stories.js
```diff
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+/* @flow */
```
ojcurtis: Can you make this /* @flow strict */?

packages/bpk-component-infinite-scroll/src/withInfiniteScroll-test.flow.js
```diff
@@ -0,0 +1,86 @@
...
+    <InfiniteList aria-label="infinite list" />
+
+    {/* $ExpectError (incompatible type) */}
+    <InfiniteList dataSource={new ArrayDataSource([])} aria-label={null} />
+
+    {/* Test default props */}
+    <InfiniteList
+      dataSource={new ArrayDataSource([])}
+      aria-label="infinite list"
+    />
```
shaundon: I'm a bit confused as to how this tests default props. I get that when this is rendered the default props will be added, but what are we doing to verify that in this test file?

tiagoengel: The idea is, if default props are not working we will get a flow error here, so as long as we don't get any error here it means flow is considering default props correctly, or so I hope 😅

shaundon: Ahh gotcha

PR #1422, opened 2019-04-22
[BPKR-29] Flow strict

dominicfraser:
Suggestion:

If a backpack consumer has
```
[strict]
nonstrict-import
```
in their `flowconfig` then Backpack imports will result in an error, as Backpack does not have `strict` Flow annotations. 

This results in either consumers adding Backpack modules to be `[ignored]`'d in their Flow config, so not benefitting from Backpack types, or not being able to use `nonstrict-import`.

It would be great to support `@flow strict` so that consumers were able to easily benefit from Backpack's own use of Flow. 

Currently Backpack has no `strict` rules so I do not think this would have any effect on Backpack.

I added the [unsafe-getters-setters](https://flow.org/en/docs/linting/rule-reference/#toc-unsafe-getters-setters) rule as an example of a strict rule, that is currently being followed anyway.

What do you think?

------

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -19,3 +19,4 @@ module.file_ext=.js
 module.file_ext=.json
 
 [strict]
+unsafe-getters-setters
--- a/packages/bpk-component-badge/BadgeLayout.js
+++ b/packages/bpk-component-badge/BadgeLayout.js
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-/* @flow */
+/* @flow strict */
 
 import PropTypes from 'prop-types';
 import React, { type Node, Children } from 'react';
--- a/packages/bpk-component-badge/index.js
+++ b/packages/bpk-component-badge/index.js
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-/* @flow */
+/* @flow strict */
 
 import component, { BADGE_TYPES, type Props } from './src/BpkBadge';
 
--- a/packages/bpk-component-badge/src/BpkBadge.js
+++ b/packages/bpk-component-badge/src/BpkBadge.js
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-/* @flow */
+/* @flow strict */
 
 import PropTypes from 'prop-types';
 import React from 'react';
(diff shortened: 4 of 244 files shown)


ojcurtis:
LGTM, however can you review the following warning from the Backpack bot?

> Warnings
> ⚠️	
> One or more packages have changed, but `UNRELEASED.md` wasn't updated.

shaundon:
@dominicfraser I'm reading the docs for Flow strict (I wasn't aware of this before your PR) and this came up:

> We recommend you enable all your desired rules from the beginning, then adopt Flow Strict file-by-file. This works better than enabling a single rule, adding @flow strict to many files, and then adding more rules to the config.

Do you think we should do that instead, or do it the way the PR is already doing it?

dominicfraser:
@ojcurtis: I thought that was only for Backpack to update? https://github.com/Skyscanner/backpack/blob/master/CONTRIBUTING.md#L230 For past (mini) PRs I have not touched it and I think Backpack squad members updated it (as under How to -> Publish packages)

@shaundon I think it depends on your flow strict needs and immediate intentions. 

(1) To meet the needs as a consumer I would prefer that they are all made strict now, as that solves my use case of wanting to keep my strict rule of `nonstrict-import` turned on as soon as possible. 

If you are not planning on adding more flow strict rules then this method should not be a problem. As you were not using it before then I would suspect this is the case?

(2) If you are wanting to add more of your own flow strict rules then it is more of a debate. 

Adding `flow strict` incrementally will block me (as I cannot turn on `nonstrict-import` on a per file basis), but be easier to do over time for you. I suspect the recommendation as you quote is as adding a new rule when all files already use Flow strict will mean that you have to fix it in all places before adding. For example I did not turn on `unclear-type` as it would mean fixing the use of `Object` in multiple files. This makes it slightly harder to add more strict rules, as it requires it being followed (or $FlowFixMe'd) everywhere at once. 

I would prefer (1), but can understand that (2) is a consideration. What do you think? Does Backpack plan on adding more `strict` rules, and would it be a problem to do so if at the point of adding them Backpack had to follow them in every file?

shaundon:
In light of that, I'm happy to merge this in its current state 😄

PR #1433, opened 2019-04-26
[No JIRA] Add 'selected' prop to BpkMapMarker

shaundon:
Kapture 2019-04-26 at 13 02 07

This was already being done by consumers so this pulls it into Backpack. I've also updated the docs page to be stateful so it works as an example (see GIF above).

### DIFF
--- a/packages/bpk-component-map/README.md
+++ b/packages/bpk-component-map/README.md
@@ -36,6 +36,14 @@ export default () => (
       onClick={() => {}}
       large
     />
+    <BpkMapMarker
+      icon={<AlignedLandmarkIconLg />}
+      type={MARKER_TYPES.primary}
+      position={{ latitude: 27.9881, longitude: 86.925 }}
+      onClick={() => {}}
+      large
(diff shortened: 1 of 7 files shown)


packages/bpk-component-map/README.md
```diff
@@ -111,6 +119,7 @@ When using `withGoogleMapsScript`, some additional props are available:
 | className        | string                                       | false                    | null                 |
 | large            | bool                                         | false                    | false                |
 | onClick          | func                                         | false                    | null                 |
+| selected         | bool                                         | false                    | false
```
sogame: Nitpicking: the last `|` is missing :-P

On a personal project I use prettier to format `md` files in a pre-commit hook, maybe you would be interested in adding something like that:
https://github.com/sogame/danger-plugin-toolbox/blob/master/package.json#L35-L38

shaundon: Ooh thanks, good tip!

PR #1437, opened 2019-04-30
Gracefully closes popover when render target is no longer present

kbremner:

I have a component that renders a BpkPopover component, with a render target that exists within the component rendering the popover and is accessed via a reference:

```js
<div ref={this.renderTargetRef}>
    <div ref={this.targetRef}>
        ...
    </div>
    <BpkPopover
        ...
        target={() => this.targetRef.current}
        renderTarget={() => this.renderTargetRef.current}
        ...
    />
</div>
```

(_both refs are created using `React.createRef`, as mentioned here: https://reactjs.org/docs/refs-and-the-dom.html_)

While implementing tests for this component, using the `react-testing-library`, I kept getting the following error:
`Cannot read property 'removeChild' of null`

The stacktrace pointed towards the close method of `Portal.js`. It appears that, in this case, when the component under test is unmounted, the render target div is removed before `Portal.componentWillUnmount` is called.

It seemed that we should gracefully handle the case where the render target is no longer present, as at that point, the `portalElement` will no longer exist anyway.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -4,6 +4,11 @@
 
 - `bpk-component-map`:
   - Added `selected` prop for when markers are selected or highlighted. See https://backpack.github.io/components/map?platform=web#map-markers.
+- `bpk-component-autosuggest`:
+- `bpk-component-datepicker`:
+- `bpk-component-popover`:
+- `bpk-react-utils`:
+  - Fixed issue that caused tests to error if a portal's render target is removed from the DOM before the portal is unmounted
 
 ## How to write a good changelog entry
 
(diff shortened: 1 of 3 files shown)


georgegillams:
Hey @kbremner this is so great!
Would you mind adding something to `UNRELEASED.md`, and then it should be good to merge 🎉

kbremner:
> Hey @kbremner this is so great!
> Would you mind adding something to `UNRELEASED.md`, and then it should be good to merge 🎉
> 
> Something like this would be fine:
> 
> ```
> **Fixed:**
> 
> - `bpk-component-autosuggest`:
> - `bpk-component-datepicker`:
> - `bpk-component-popover`:
> - `bpk-react-utils`:
>   - Fixed issue that caused tests to error if a portal's render target has already been removed.
> ```

done!

PR #1439, opened 2019-05-01
Fixes types for render and renderTarget to accept refs

kbremner:

The type for `React.createRef` is `() => null | T`, with `T` being a type that inherits from `HTMLElement`.

This PR updates the types for the `target` and `renderTarget` props of `bpk-component-popover` to accept this type

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -224,7 +224,6 @@ PropType
 PropTypes
 ref
 refs
-createRef
 SectionList
 SyntheticEvent
 TextInput
--- a/.spelling
+++ b/.spelling
@@ -224,6 +224,7 @@ PropType
 PropTypes
 ref
 refs
+createRef
 SectionList
 SyntheticEvent
 TextInput
--- a/packages/bpk-component-popover/src/BpkPopoverPortal.js
+++ b/packages/bpk-component-popover/src/BpkPopoverPortal.js
@@ -37,12 +37,12 @@ const getClassName = cssModules(STYLES);
 
 export type Props = {
   ...$Exact<PopoverProps>,
-  target: (() => HTMLElement) | Node,
+  target: (() => null | HTMLElement) | Node,
   isOpen: boolean,
   placement: ?('top' | 'right' | 'bottom' | 'left'),
   portalStyle: ?Object,
   portalClassName: ?string,
-  renderTarget: ?() => HTMLElement,
+  renderTarget: ?() => null | HTMLElement,
   popperModifiers: ?Object,
 };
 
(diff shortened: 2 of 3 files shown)


k0nserv:
Thanks for the PR 👍 

My concern is that if we make this change we need to trace all usage of `target` and `renderTarget` and ensure those locations are correctly handling the possibility for `target` and `renderTarget` to be `null`

packages/bpk-component-popover/src/BpkPopoverPortal.js
```diff
@@ -37,12 +37,12 @@ const getClassName = cssModules(STYLES);
 
 export type Props = {
   ...$Exact<PopoverProps>,
-  target: (() => HTMLElement) | Node,
+  target: (() => null | HTMLElement) | Node,
```
k0nserv: Is there a specific reason that this is not `() => ?HTMLElement`?

.spelling
```diff
@@ -224,6 +224,7 @@ PropType
 PropTypes
 ref
 refs
+createRef
```
georgegillams: This isn't actually needed as the spellchecker ignores code 👍

kbremner: It kept flagging it up cause I use createRef in `UNRELEASED.md` :)

packages/bpk-component-popover/src/BpkPopoverPortal.js
kbremner: Just my lack of experience with flow :) will change :+1:

kbremner:
> Thanks for the PR 👍
> 
> My concern is that if we make this change we need to trace all usage of `target` and `renderTarget` and ensure those locations are correctly handling the possibility for `target` and `renderTarget` to be `null`

Absolutely, I felt it was better to get this up to discuss that before diving in to the weeds

My concern with the current typing is that we can never rely on the renderTarget always being present - if we were to get the renderTarget via some other means, like `document.getElementById`, that could still return null.

If we support accepting a function to resolve the render target, it seems that we need to accept that at some point it could be null

Any thoughts on that?

kbremner:
This change would also match the types used by `BpkModal`:
https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-modal/src/BpkModal.js#L52

.spelling
georgegillams: Sorry - I wasn't clear enough. It ignores code within markdown files. So it won't flag it because you wrote ``createRef`` in `UNRELEASED.md`

kbremner: hmm, weird - the pre-commit hook was complaining about spelling, specifically talking about the use of createRef in the markdown file, but I can't replicate it now 🤷‍♂ - removed!

PR #1441, opened 2019-05-02
Fix test/linter warnings

sogame:
When running linters and tests there're a bunch of warnings. These changes fix them.

I've also added a [Danger check](https://github.com/sogame/danger-plugin-toolbox/blob/master/docs/validations.md#commonfilewarnings) to show warnings from the build, so these can be spotted when submitting a PR.


packages/bpk-component-datatable/src/utils.js
```diff
@@ -21,3 +21,13 @@ export const hasClassName = (node, className) => {
   const nodeClassName = node.getAttribute('class');
   return nodeClassName && nodeClassName.split(' ').indexOf(className) !== -1;
 };
+
+export const remToPx = value => {
```
sogame: I've copied this function from [bpk-component-barchart](https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-barchart/src/utils.js#L29), not sure if this should be placed somewhere else and reused 😅

packages/bpk-component-infinite-scroll/src/withInfiniteScroll-test.js
```diff
@@ -204,18 +204,6 @@ describe('withInfiniteScroll', () => {
     expect(currentOptions).toEqual({ threshold: 0.99 });
   });
 
-  it('should create the IntersectionObserver with a 0.99 threshold if the prop is set with an invalid value', () => {
```
sogame: Not sure about this: on one side it's weird to add a test to check the behaviour when a prop is invalid, but on the other it seems safer 😂

.travis.yml
```diff
@@ -12,8 +12,8 @@ install:
 script:
   - set -e
   - ./scripts/check-pristine-state package-lock.json
+  - (set -o pipefail; npm test |& tee test.log);
```
sogame: * Keep test output in a file so Danger can parse it.
* Run tests before Danger so it has access the the test logs

dangerfile.js
```diff
@@ -133,3 +134,8 @@ markdown.forEach(path => {
       });
     });
 });
+
+commonFileWarnings('test.log', {
+  msg:
+    'The build logs contain these warnings (check the [Travis](https://travis-ci.org/Skyscanner/backpack) build output for more details):',
+});
```
sogame: New Danger rule to show warnings from the build: https://github.com/sogame/danger-plugin-toolbox/blob/master/docs/validations.md#commonfilewarnings

This is an example output:
warnings

sogame:
I just noticed that there's a false positive in the Danger check, and it detects one of the tests of BpkBadge as a warning 😅: `should render correctly with type="warning"`.

I guess it'll be better to remove the rule for now, until I can fix it.

sogame:
Fixed the false positive, this is ready to be reviewed now

packages/bpk-component-datatable/src/utils.js
```diff
@@ -21,3 +21,13 @@ export const hasClassName = (node, className) => {
...
+  let parsed = null;
+
+  if (/rem$/.test(value)) {
+    parsed = parseFloat(value.replace(/rem/, '')) * 16;
```
k0nserv: This does not actually work because there is no guarantee that the root font size is 16. This will break when zooming for example. [This PR](https://github.com/Skyscanner/backpack/pull/877/files) is an example where we fixed an issue based on an assumption like this

packages/bpk-component-barchart/src/BpkChartMargin-test.js
```diff
@@ -20,13 +20,15 @@ import React from 'react';
...
+import { remToPx } from './utils';
 import BpkChartMargin from './BpkChartMargin';
 
+const spacing = remToPx(spacingSm);
 const margin = {
-  top: spacingSm,
-  right: spacingSm,
-  bottom: spacingSm,
-  left: spacingSm,
+  top: spacing,
```
k0nserv: I think we are better of changing the `propTypesS` for `BpkCharMargin` from

```javascript

BpkChartMargin.propTypes = {
  children: PropTypes.node.isRequired,
  margin: PropTypes.shape({
    top: PropTypes.number,
    bottom: PropTypes.number,
    left: PropTypes.number,
    right: PropTypes.number,
  }).isRequired,
};
```

to 

```javascript
BpkChartMargin.propTypes = {
  children: PropTypes.node.isRequired,
  margin: PropTypes.shape({
    top: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
    bottom: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
    left: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
    right: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
  }).isRequired,
};
```

sogame: Makes sense, changed it 👍

packages/bpk-component-datatable/src/utils.js
sogame: Given this function was used only in the test, I think it makes more sense to just use a number instead of the `spacingSm` token, so I've removed this function again and replaced the token in the test.

What do you think @k0nserv ?

.travis.yml
georgegillams: Just to confirm, `|&` is used to merge std err so that `tee` gets that as well as the std output, right?

sogame: Exactly, this way both output messages (std out and std err) are send to the log file and will still be also in the build output. Also the exit code is preserved, so if the test fail for example, the build will fail anyway.

packages/bpk-component-datatable/src/BpkDataTableColumn-test.js
```diff
@@ -18,12 +18,15 @@
...
-import { spacingSm } from 'bpk-tokens/tokens/base.es6';
 
 import BpkDataTableColumn from './BpkDataTableColumn';
 import bpkHeaderRenderer from './bpkHeaderRenderer';
 
-const defaultProps = { label: 'Name', dataKey: 'name', width: spacingSm };
+const defaultProps = {
+  label: 'Name',
+  dataKey: 'name',
+  width: 12, // eslint-disable-line backpack/use-tokens
```
georgegillams: why can't this still be `spacingSm`?

sogame: Because `width` is defined as `number` in `react-virtualized` (https://github.com/bvaughn/react-virtualized/blob/master/source/Table/Column.js#L81), and using `spacingSm` throws a proptypes warning when running the tests :-(

PR #1445, opened 2019-05-06
[BPK-2377]: Phone input cuts description

ojcurtis:
+ [ ] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack/blob/master/CONTRIBUTING.md)


packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -83,6 +83,17 @@ const BpkPhoneInput = (props: Props) => {
...
+  let sizeMultiplier = 5;
+  if (large) sizeMultiplier = 7;
+
+  // This function get the size value of the dialingCode to resize the field correctly
+  function dcChange() {
+    const dialogCode = dialingCodes.find(e => e.code === dialingCode);
+    if (dialogCode) return sizeMultiplier * dialogCode.description.length + 100;
+    return sizeMultiplier + 100;
+  }
+
```
shaundon: Few small things here:

1. Try to give the function name as descriptive a name as possible for when it's called. How about `widthForDialingCode`? Then when it's used it'll go: `width: ${widthForDialingCode()}px` which is a bit easier for a reader to parse.

2. We try to use arrow functions wherever we can for consistency:

```
const nameOfFunction = () => { ...
```

3. `dialogCode` feels like a typo?

4. This function relies on variables set outside of itself (`dialingCode`. `sizeMultiplier`, `dialingCodes`). How about making it a pure function by having it take all the data it needs as arguments? Then it can live outside of the component code too further up the file, to make the component code a bit leaner.

5. We try to avoid having `if` statements all on one line because they can be less readable.

Putting all those together, how about something like:

```
const widthForDialingCode = (large, dialingCodes, selectedDialingCode) => {
  const sizeMultiplier = large ? 7 : 5;
  const foundDialingCode = dialingCodes.find(dialingCode => dialingCode.code === selectedDialingCode);
  if (foundDialingCode && foundDialingCode.description) {
    return sizeMultiplier * foundDialingCode.description.length + 100;
  }
  return sizeMultiplier + 100;
};
```

The only other thing here is I'm a bit wary of all the magic numbers (e.g. 5 and 7, the +100) because it's not obvious what they're for and why they're set to those particular numbers. Is there a way to make it clearer for future devs?

ojcurtis:
@tiagoengel - It is strange you are seeing it like this, as mine its resizing as expected. For the calculation we are using the length of the content to dynamically display

Input

packages/bpk-component-phone-input/src/BpkPhoneInput.js
tiagoengel: Another thing about the +100, this will likely fail with a different font size, or zooming in on the page

ojcurtis: The 5 and 7 are assumed values based on average letter size, whereas the 100 is an assumed padding and caret addition to the field also. As this value is almost impossible to calculate due to the variations of these types of fields and worldwide people do different things (e.g. just flag, just number or both but allow cutoff or oversized field)

I have had a look online and there are no real best ways of doing this so input would be appreciated 😄

ojcurtis: Testing on zoom and also with larger font size (20px in this case) shows it can still work but we need to maybe see if there is an improved solution overall for the approach to this field (i.e. the style in which that dropdown is used)

packages/bpk-docs/src/pages/PhoneInputPage/PhoneInputPage.js
```diff
@@ -29,6 +29,8 @@ import IntroBlurb from '../../components/IntroBlurb';
 import NativePhoneInput from '../NativePhoneInputPage';
 
 const DIALING_CODE_TO_ID_MAP = {
+  '1': 'us',
```
georgegillams: These changes will need to be moved to the [new docs repo](http://github.com/skyscanner/backpack-docs)

tiagoengel:
It looks good for me, my only concern is still with long descriptions "Double length label" example:

Screenshot 2019-05-31 at 09 58 36

It's not breaking down the description like it was before. If @shaundon, @georgegillams are ok with it I don't mind keeping it like this, but I'm also not sure it will cause problems somewhere 🤷‍♂

packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -58,6 +58,20 @@ type CommonProps = {
   valid: ?boolean,
 };
 
+// This function get the size value of the dialingCode to resize the field correctly
+const widthForDialingCode = (large, dialingCodes, selectedDialingCode) => {
+  // This sizeConstant is an average character size for each size of field
+  const averageLetterSize = large ? 8 : 6;
+  const foundDialogCode = dialingCodes.find(
```
shaundon: Typo in the variable name 😇 

```suggestion
  const foundDialingCode = dialingCodes.find(
```

PR #1458, opened 2019-05-09
[No JIRA] Add `greedyGestureHandling` prop to BpkMap

shaundon:
Previously we set [`gestureHandling`](https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions.gestureHandling) to `auto`. This created an issue when maps were placed inside modals, as they _should_ have been scrollable with one finger but instead get the 'use two fingers' overlay which was a bad experience for travellers. 

This PR adds a new prop to override the gesture handling so that in situations like this, developers can manually override to get the more desirable behaviour.

Here's a video to illustrate the difference:

ezgif-1-0625a5b6702b

Please note that this only applies to touchscreens. In the above video I'm emulating a touchscreen in Chrome.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -710,3 +710,4 @@ onBarHover
 onBarFocus
 getBarLabel
 getBarSelection
+greedyGestureHandling
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Added:**
+- bpk-component-map:
+  - New `greedyGestureHandling` prop for enforcing one-finger pan and zoom on touch devices.
+
 
 ## How to write a good changelog entry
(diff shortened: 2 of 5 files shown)


k0nserv:
Do we have any concerns about mapping a boolean value to essentially an enum in Gmaps?

shaundon:
> Do we have any concerns about mapping a boolean value to essentially an enum in Gmaps?

If `panEnabled` wasn't already an option I'd have mapped it to the enum directly, but I made it a boolean because we already have that one and I thought this was better than a breaking change. I

Maybe it's better to make the breaking change instead to be more direct. What do you reckon?

k0nserv:
I have no strong oppinion, I just wanted to make sure we had considered it and made a deliberate decision

PR #1459, opened 2019-05-09
Bring map marker to front when selected

arnaugm:
Selected map markers are displayed on top of non selected by increasing their z-index. This is useful when markers overlap.
Added a stateful component in Storybook to reflect that, not sure if is the desired approach, let me know any concern and I'll change it.


shaundon:
Could you make the example have two overlapping pins, so we have a good example of the `z-index` change?

Also, I think we should have a couple more examples of normal/selected markers *without* state too, just so it's really clear to consumers that they need to add the state and it isn't built in. I can do that once this is merged if you don't want to do it though, as it's a bit out of scope.

PR #1463, opened 2019-05-15
[BPK-2477] - Gray theming utility addition

ojcurtis:


packages/bpk-theming/src/BpkThemeProvider.js
```diff
@@ -39,10 +39,9 @@ const createStyle = (theme, themeAttributes) => {
   const missingThemeAttributes = [];
   flattenedThemeAttributes.forEach(attribute => {
     if (theme[attribute]) {
-      const cssName = attribute.replace(
-        /([A-Z])/g,
-        variable => `-${variable.toLowerCase()}`,
-      );
+      const cssName = attribute
+        .replace(/([A-Z])/g, variable => `-${variable.toLowerCase()}`)
+        .replace(/([0-9])/, variable => `-${variable.toLowerCase()}`);
```
ojcurtis: We are adding the second replace so that we can also create cssNames if numbers are used in the attribute too which would allow the following `color-gray-100` instead of `color-gray100`

PR #1467, opened 2019-05-20
[No JIRA] Update react-svg-loader

shaundon:
Closes #1442.

Tested icons in Chrome and IE11 and all looks ok.


packages/bpk-svgs/dist/js/icons/lg/account.js
```diff
@@ -1,14 +1,8 @@
...
-  }} {...props}><path d="M3 21v-3.1c0-2.8 2.2-5 5-5h8c2.8 0 5 2.2 5 5V21H3zM16 7c0-2.2-1.8-4-4-4S8 4.8 8 7s1.8 4 4 4 4-1.8 4-4z" /></svg>;
-});
\ No newline at end of file
+export default (({
+  styles = {},
+  ...props
+}) => <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" style={{
+  width: "1.5rem",
+  height: "1.5rem"
+}} {...props}><path d="M3 21v-3.1c0-2.8 2.2-5 5-5h8c2.8 0 5 2.2 5 5V21H3zM16 7c0-2.2-1.8-4-4-4S8 4.8 8 7s1.8 4 4 4 4-1.8 4-4z" /></svg>);
```
georgegillams: Do we know why this is no longer being transpiled?

packages/bpk-component-breadcrumb/src/__snapshots__/BpkBreadcrumbItem-test.js.snap
```diff
@@ -68,7 +67,6 @@ exports[`BpkBreadcrumbItem should render correctly with a custom class name 1`]
         "width": "1.5rem",
       }
     }
-    viewBox="0 0 24 24"
```
georgegillams: I don't really know what viewBox does, but I think removing this will affect the default / max width and height... That might be enforced through our styling anyway 🤷‍♂

shaundon:
We already set a width and height on our generated icons so it's probably ok:

```
width="24" height="24" style={{
  width: "1.5rem",
  height: "1.5rem"
}```

packages/bpk-svgs/dist/js/icons/lg/account.js
shaundon: They dropped support for Node 6 so it might be just be using more modern syntax (https://github.com/boopathi/react-svg-loader/blob/master/CHANGELOG.md#300)

It was never transpiled earlier as it was using `export` which isn't valid ES5.

PR #1474, opened 2019-05-27
[NOJIRA] Add "sideEffects": false to (mostly) all components

matthewdavidson:
Official documentation on `"sideEffects": false,` is available [here](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free).

> A "side effect" is defined as code that performs a special behavior when imported, other than exposing one or more exports. An example of this are polyfills, which affect the global scope and usually do not provide an export.

AFAIK no Backpack component's have side effects.

Notice in the below example of a fresh `backpack-react-scripts` project that `react-transition-group` has been removed from the bundle as it's not used.

| Before | After |
| --- | --- |
| Screenshot 2019-05-27 at 16 37 30 | Screenshot 2019-05-27 at 16 37 35 |

TODO:
- [x] Provide minimal reproducible example 


matthewdavidson:
I removed the following packages from the `"sideEffects": false,` treatment - key one being `bpk-stylesheets` because it does some funky stuff i.e. it has side effects.

matthewdavidson:
image

This is interesting - I think I might also remove `bpk-component-infinite-scroll` from this PR too.

PR #1489, opened 2019-06-05
Infinite Scroll Use fetch items in update data

EndersbyD:
This will allow infinite scroll to be initialise with an empty data source
As I am not creating any new functionality tests have not been added, only one snapshot
Tested locally

### DIFF
--- a/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
+++ b/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
@@ -170,10 +170,12 @@ const withInfiniteScroll = <T: ExtendedProps>(
     }
 
     updateData = () => {
-      const { index } = this.state;
-      this.props.dataSource.fetchItems(0, index).then(updatedElements => {
+      this.fetchItems({
+        index: this.state.index,
+        elementsPerScroll: this.props.elementsPerScroll,
+      }).then(newState => {
         this.setState({
-          elementsToRender: updatedElements,
+          ...newState,
         });
       });
     };
--- a/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
+++ b/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
@@ -170,9 +170,13 @@ const withInfiniteScroll = <T: ExtendedProps>(
     }
 
     updateData = () => {
+      const { index } = this.state;
       this.fetchItems({
-        index: this.state.index,
-        elementsPerScroll: this.props.elementsPerScroll,
+        index: 0,
+        elementsPerScroll:
+          index > this.props.elementsPerScroll
+            ? index
+            : this.props.elementsPerScroll,
       }).then(newState => {
         this.setState({
           ...newState,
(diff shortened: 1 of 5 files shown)


packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -170,10 +170,12 @@ const withInfiniteScroll = <T: ExtendedProps>(
     }
 
     updateData = () => {
-      const { index } = this.state;
-      this.props.dataSource.fetchItems(0, index).then(updatedElements => {
+      this.fetchItems({
+        index: this.state.index,
```
tiagoengel: Can you explain better why are you doing this/what is the problem you are trying to solve? `updateData` is called when the DataSource's date changes, e.g. when new data comes back from the server in the background, and should call `fetchItems(0, currentIndex)` to load all new data. 

I'm afraid what you are doing here will break this functionality.

EndersbyD: So this an attempt to solve a bug I found when you initialise an ArrayDataSource with an empty array and then call updateData method on that data source, infinite scroll does not render this updated data. We know this is caused by the fetch items method in ArrayDataSource on line 169 as the index value is set to 0 which causes no elements to be returned. Here is the example I have made to prove this https://github.skyscannertools.net/davidendersby/backpack-testing
So I thought I would use the fetch items method in infinite scroll, as it allows config setup and returns fetch items from the datasource

EndersbyD: so as currentIndex on the first render is 0 no new data is return, but this doesn't update when updateData on the arraydatasource is called

tiagoengel: I see, unfortunately this will cause other problems. From what you are describing this seems more of a misunderstanding of how the infinite scroll/data source work. The `updateData` is not supposed to be called to add new data, it is supposed to be called to update the current data in the data source. You can have a look at the `Force update data` example in our stories.

Another problem is that the `ArrayDataSource` is supposed to be a `static` data source, meaning that initialising it with an empty array and then later on calling `updateData` with more data is not supported.

In this case you need to provide a new instance of the data source to your component. Alternatively you can implement your own data source to apply any custom logic you need.

DewarM: Hey, we are struggling to come up with a use case for `ArrayDataSource`. My understanding is that to use `ArrayDataSource` as it is currently implemented you need to know the length of the array when you create it. Once it has been created you cannot change the length of that `ArrayDataSource` - you can only modify values within in it. Is that correct?

tiagoengel: Thinking more about this I believe we can make this change but it should be ```
 this.fetchItems({
    index: 0,
    elementsPerScroll: index > this.props.elementsPerScroll ? index : this.props.elementsPerScroll
})

We need to add a test for this too, that will call `updateData` in the data source when index is zero and make sure it does the right thing

PR #1493, opened 2019-06-06
[WOM-119] - Country Code Select Display Flag-only and Phone Input Code Display 

Andi24:
In Accommodations we would like to improve how we prompt users for their phone number. 

User feedback suggests that people get confused by us displaying the name of the country truncated next to the country code. People also get confused as to whether they still need to add the country code to their phone number or not. This PR is aimed at addressing those ambiguities.

To solve this, we decided to only display the country flag and add the country code in the phone input. Unfortunately this behaviour is not supported by the native `select` component, so we had to go for something more inventive. This PR adds a prop applying a style that makes the select box display only the provided flag (in effect hiding the selection text; the only way to hide it otherwise was to change the options list to not display country names, but that is undesirable from a UI point of view). 

Changes in this PR:
- prop for phone input masking (i.e. the country code will be prepended to the phone number when this prop is active)
Screenshot 2019-06-11 at 10 31 07

- prop for displaying only the flag in the country code selector
Screenshot 2019-06-11 at 10 31 30

Below is what the component looks when both props are active
Screenshot 2019-06-06 at 13 44 25
Screenshot 2019-06-06 at 13 44 33
Screenshot 2019-06-06 at 13 44 41


packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -97,6 +101,17 @@ const BpkPhoneInput = (props: Props) => {
     disabled: !!disabled,
   };
 
+  let phoneDisplayValue;
+
+  if (countryCodeMask) {
```
ojcurtis: Can this code be run through prettier? Or set to be consisten of brace style rather than a mix of brace and brace less

packages/bpk-component-select/src/BpkSelect.scss
```diff
@@ -21,6 +21,10 @@
 .bpk-select {
   @include bpk-select;
 
+  &--image-only {
+    width: 1rem;
```
ojcurtis: Is there a bpk token that can support this value?

packages/bpk-component-phone-input/src/BpkPhoneInput.js
Andi24: addressed

packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -97,6 +101,20 @@ const BpkPhoneInput = (props: Props) => {
...
+    if (value.startsWith('+')) {
+      if (value.startsWith(`+${dialingCode}`)) {
+        phoneDisplayValue = value;
+      } else {
+        phoneDisplayValue = `+${dialingCode} ${value.split(' ')[1]}`;
+      }
+    } else {
+      phoneDisplayValue = `+${dialingCode} ${value}`;
+    }
+  } else phoneDisplayValue = value;
```
shaundon:
```suggestion
  } else {
    phoneDisplayValue = value;
}
```

This bit is still brace-less 😇

packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -145,8 +172,8 @@ const BpkPhoneInput = (props: Props) => {
         {...rest}
         id={id}
         name={name}
-        value={value}
-        type={INPUT_TYPES.number}
+        value={phoneDisplayValue}
+        type={INPUT_TYPES.tel}
```
shaundon: Good catch! 🙌

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -72,6 +74,7 @@ const BpkSelect = (props: Props) => {
         image && 'bpk-select--with-image',
         image && large && 'bpk-select--with-image-large',
         isInvalid && 'bpk-select--invalid',
+        imageOnly && 'bpk-select--image-only',
```
shaundon: I'm wondering if this should be changed to only apply if `image !== null`, that way it avoids a potential scenario where somebody sets `imageOnly` but not `image` and it looks off. What do you think?

```suggestion
        image && imageOnly && 'bpk-select--image-only',
```

packages/bpk-component-phone-input/src/BpkPhoneInput.js
Andi24: whoa, you can give commit suggestions 😮 so cool! will accept this but will make another commit updating snapshots too

packages/bpk-component-select/src/BpkSelect.js
Andi24: yup, great catch - adjusting

packages/bpk-component-phone-input/src/BpkPhoneInput.js
Andi24: done

packages/bpk-component-select/src/BpkSelect.js
Andi24: done

ojcurtis:
Code is looking good however in testing I came across this bug when using the `input mask on phone input` story 
Phone Input

packages/bpk-component-phone-input/src/BpkPhoneInput.js
shaundon: > whoa, you can give commit suggestions 😮

Haha yeah it only works on the open-source GitHub I think. It's a handy feature!

PR #1453, opened 2019-05-08
Upgrade Storybook amongst other things

shaundon:
Webpack 3 ➡️ 4
- Had to change stylesheets build accordingly as the extract text plugin doesn't work, need to use the shiny new CSS one.

Storybook 3 ➡️ 5
- This requires Webpack 4, which is why we had to upgrade that too.
- **This part is broken**, details below.

I get these errors when doing `npm run storybook`:
I think it's trying to transpile deps, not sure if it even should be doing that. Needs further investigation.

<details>
<summary>Stack trace</summary>

```
ERROR in ./packages/bpk-component-datatable/node_modules/react-virtualized/dist/es/index.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: Couldn't find preset "./.babelrc.js" relative to directory "/Users/shaundonnelly/projects/backpack/packages/bpk-component-datatable/node_modules/react-virtualized"
    at /Users/shaundonnelly/projects/backpack/node_modules/babel-core/lib/transformation/file/options/option-manager.js:293:19
    at Array.map (<anonymous>)
    at OptionManager.resolvePresets (/Users/shaundonnelly/projects/backpack/node_modules/babel-core/lib/transformation/file/options/option-manager.js:275:20)
    at OptionManager.mergePresets (/Users/shaundonnelly/projects/backpack/node_modules/babel-core/lib/transformation/file/options/option-manager.js:264:10)
    at OptionManager.mergeOptions (/Users/shaundonnelly/projects/backpack/node_modules/babel-core/lib/transformation/file/options/option-manager.js:249:14)
    at OptionManager.init (/Users/shaundonnelly/projects/backpack/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
    at File.initOptions (/Users/shaundonnelly/projects/backpack/node_modules/babel-core/lib/transformation/file/index.js:212:65)
    at new File (/Users/shaundonnelly/projects/backpack/node_modules/babel-core/lib/transformation/file/index.js:135:24)
    at Pipeline.transform (/Users/shaundonnelly/projects/backpack/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at transpile (/Users/shaundonnelly/projects/backpack/node_modules/babel-loader/lib/index.js:50:20)
    at /Users/shaundonnelly/projects/backpack/node_modules/babel-loader/lib/fs-cache.js:118:18
[...]

PR #1496, opened 2019-06-10
[NO JIRA]: Upgrade storybook

ojcurtis:
This PR is based on an original PR opened by Shaun: https://github.com/Skyscanner/backpack/pull/1453 

On top of the original PR I have done the following: 

- Added new .babelrc file to .storybook as detailed in storybook 5 docs where root config can cause storybook issues: https://storybook.js.org/docs/configurations/custom-babel-config/
- Added new package create-react-class to `bpk-component-slider` which was causing build issues with webpack and storybook upgrades


package.json
```diff
@@ -120,6 +120,7 @@
     "markdown-spellcheck": "^1.3.1",
     "merge-stream": "^2.0.0",
     "merge2": "^1.2.3",
+    "mini-css-extract-plugin": "^0.6.0",
```
shaundon: Was this added because webpack 4 requires it?

packages/bpk-component-slider/package.json
```diff
@@ -20,7 +20,8 @@
     "react-slider": "^0.11.2"
   },
   "devDependencies": {
-    "bpk-component-rtl-toggle": "^1.1.88"
+    "bpk-component-rtl-toggle": "^1.1.88",
+    "create-react-class": "^15.6.3"
```
shaundon: I think this might not be quite right - looking at the docs for this module says it's intended for those using React without ES6, which isn't us: https://reactjs.org/docs/react-without-es6.html

What was the issue you saw before you added this dependency?

packages/bpk-stylesheets/base.js
```diff
@@ -59,58 +79,34 @@
...
+  \********************/
+/*! no static exports found */
+/***/ (function(module, exports, __webpack_require__) {
+
+eval("// extracted by mini-css-extract-plugin\n\n//# sourceURL=webpack:///./index.scss?");
 
 /***/ })
-/******/ ]);
\ No newline at end of file
+
```
shaundon: Always a bit nervous when the base stylesheet changes, do we have a good way to test nothing is broken? cc @matthewdavidson

webpack.config.babel.js
```diff
@@ -1,230 +0,0 @@
...
-      compress: {
-        unused: true,
-        dead_code: true,
-        warnings: false,
-      },
-    }),
-  );
-}
-
-export default config;
```
shaundon: I guess this is no longer needed at all because it was used by the docs site?

packages/bpk-stylesheets/base.js
matthewdavidson: The only functionality of the javascript part of the base stylesheet is to append `bpk-no-touch-support` or ` bpk-touch-support` based on the browsers ability to support touch interactions. If hovers styles still work then 👍

ojcurtis: @shaundon - This was based on your changes that you had made when we moved from webpack version upgrade

ojcurtis: Will test using the recommendation from @matthewdavidson above

webpack.config.babel.js
ojcurtis: I believe so, as you had also removed this and looking at the way builds are done storybook uses the version in the .storybook folder

package.json
ojcurtis: By the looks of it yes as for the `webpack.config.babel.js` file moves from `ExtractTextPlugin` to use `MiniCssExtractPlugin` plugin instead

packages/bpk-stylesheets/base.js
shaundon: Thanks both 😄

shaundon:
Only outstanding question is this one, I think: https://github.com/Skyscanner/backpack/pull/1496#discussion_r292053974

packages/bpk-component-slider/package.json
matthewdavidson: Yeah this shouldn't be required at all.

packages/bpk-stylesheets/base.js
ojcurtis: Works as normal when doing the recommendation above 😃

packages/bpk-component-slider/package.json
ojcurtis: Will have a look with it removed and see why slider was failing. It was giving some errors to do with compilation but will double check

ojcurtis: So had a look at react-slider too itself and seems the issue exists there where it currently isn't supporting react 16: https://github.com/mpowaga/react-slider/issues/120 

`create-react-class` is listed as a peer dependency for `react-slider` so we need to add it to the project to run it, but can be a devDependency as it is only being thrown by storybook when running.

What do you think?

matthewdavidson: Oooh - ok so we _should_ have this listed as a dependency then so that we handle it for consumers.

Going forward the usage of this third party module should be reconsidered.

shaundon:
Some warnings when running this:

```
DeprecationWarning: Extend-mode configuration is deprecated, please use full-control mode instead.

See https://storybook.js.org/docs/configurations/custom-webpack-config/#full-control-mode
 67% building 1576/1649 modules 73 active ..._modules/date-fns/end_of_month/index.jsBrowserslist: caniuse-lite is outdated. Please run next command `npm update caniuse-lite browserslist`
```

These aren't blocking, I'm just making a note of them as we'll need to address them at some point.

PR #1501, opened 2019-06-13
[No JIRA] A bit of housekeeping on Danger

shaundon:
I think we can utilise Danger more than we currently are to improve the contribution process and remind us/consumers to do certain things, like cross-browser testing.

This PR:

- Moves the 'have you read the contribution guide' message to a Danger rule that is only shown to contributors.
- Sends a nice motivational message as well as a random gif to external contributors.
- Prints a checklist of browsers to check in when components have changed.

**This is a first step.** We can put more things in here later, but I thought this would be a good way to get the ball rolling.

There's a comment below from Backpack bot showing what users would see. I changed my github username in `meta.json` so Danger would think I was external, but ordinarily only external squad members would see this.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -20,13 +20,17 @@
 
 // See http://danger.systems/js if you're not sure what this is.
 
+// @flow
+
 import fs from 'fs';
 
 import { includes } from 'lodash';
-import { danger, fail, warn, message } from 'danger';
+import { danger, fail, warn, message, markdown } from 'danger';
 import { commonFileWarnings } from 'danger-plugin-toolbox';
 
 import * as meta from './meta.json';
+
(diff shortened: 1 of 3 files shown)


shaundon:
This is what consumers will see (gif is randomly chosen from list of 11)

Screen Shot 2019-06-13 at 21 51 00

dangerfile.js
```diff
@@ -60,6 +90,25 @@ if (componentIntroduced) {
   );
 }
 
+const componentChangedOrCreated = fileChanges.some(filePath =>
+  filePath.match(/packages\/bpk-component.+\/src\/.+\.js/),
```
sogame: This regex could be extracted, so it's not duplicated (also used a few lines before to get `componentIntroduced`

PR #1502, opened 2019-06-14
Messing about with Danger

shaundon:


shaundon:
Alright so we can't do this until we get on Danger 4+ which requires Babel 7

shaundon:
Closing, will keep branch around.

ojcurtis:
@shaundon - do we still require this to be around?

ojcurtis:
@shaundon do we still need this PR open?

PR #1511, opened 2019-06-19
Upgrade to Babel 7

shaundon:
As well as Babel, this also upgrades Jest and Danger to the latest versions.

## Getting `npm run build` to work

[Babel 7 doesn't work the same way for monorepos as Babel 6 did](https://babeljs.io/docs/en/config-files#6x-vs-7x-babelrc-loading). 
* In v6, Babel would traverse upwards through directories until it found a config file.
* In v7, it stops traversing upwards once it find a `package.json`.

As every folder in `packages/` has a `package.json` this means Babel no longer works. Babel 7.1+ provides an option here, named `rootMode: 'upward'` which reinstates the previous behaviour. To get this to work, I've added a new `gulpfile.js` to every Gulp-using package that registers Babel with this mode enabled and then requires the ES6 gulpfile.

For `bpk-stylesheets` that uses Webpack, I had to do a similar thing except with another option specified. This is detailed in the `build.js` file in `bpk-stylesheets`.

## Getting `npm run storybook` to work

Removed `/storybook/.babelrc`. Now inherits from the root config.

## Getting `npm test` to work

Done. Some snapshots have updated as I also updated to the latest Jest and babel-jest.

## Checking `base.js` is ok

`bpk-stylesheets/base.js` was updated. To check this won't break things, I checked:

✅  The `bpk-no-touch-support` class is added to non-touch browsers (tested in desktop Chrome)
✅ The `bpk-touch-support` class is added to touch browsers (tested in mobile Safari)
## Helpful links
* https://github.com/babel/babel/issues/8309


package.json
```diff
@@ -70,20 +71,23 @@
...
     "autoprefixer": "^9.4.7",
-    "babel-core": "^6.26.3",
-    "babel-jest": "^23.6.0",
-    "babel-loader": "^7.1.5",
-    "babel-plugin-transform-class-properties": "^6.24.1",
-    "babel-plugin-transform-object-rest-spread": "^6.20.2",
-    "babel-preset-env": "^1.7.0",
-    "babel-preset-react": "^6.16.0",
+    "babel-core": "^7.0.0-bridge.0",
+    "babel-jest": "^23.4.2",
```
shaundon: `npx babel-upgrade` downgraded Jest, we should revert this.

packages/bpk-stylesheets/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Backpack's stylesheets.",
   "main": "index.js",
   "scripts": {
-    "build": "webpack"
+    "build": "node build.js"
```
shaundon: We have to manually call `webpack` from `build.js` to avoid this error:

Screen Shot 2019-06-26 at 16 37 12

package.json
```diff
@@ -19,9 +19,9 @@
     "lint:js": "eslint . .storybook --ext .js,.jsx",
     "lint:js:fix": "eslint . .storybook --ext .js,.jsx --fix",
     "lint": "npm run lint:js && npm run lint:scss",
-    "jest:update": "TZ=Europe/London jest --updateSnapshot",
-    "jest:watch": "TZ=Europe/London jest --watch",
-    "jest": "TZ=Europe/London jest --coverage",
+    "jest:update": "TZ=Etc/UTC jest --updateSnapshot",
+    "jest:watch": "TZ=Etc/UTC jest --watch",
+    "jest": "TZ=Etc/UTC jest --coverage",
```
shaundon: One of the snapshots changed to use BST. I expect this is because something internally changed in Jest, so I set the timezone to UTC here.

packages/bpk-component-infinite-scroll/src/DataSource.js
```diff
@@ -49,9 +49,9 @@ class DataSource<T = any> {
     return false;
   }
 
-  triggerListeners(...args: Array<any>): void {
+  triggerListeners = (...args: Array<any>): void => {
     this.listeners.forEach(cb => cb(...args));
-  }
+  };
```
shaundon: This function stopped working without this, so I made it use arrow syntax.

packages/bpk-component-map/src/__snapshots__/BpkMapMarker-test.js.snap
```diff
@@ -24,7 +24,7 @@ exports[`BpkMapMarker should render correctly with a "arrowClassName" attribute
     <div
       className="bpk-map-marker__arrow custom-class-1 custom-class-2"
     >
-      <Component />
+      <_default />
```
shaundon: I'm struggling to find documentation for why this would change. It seems ok though so I'm not concerned. My gut feeling is that this is happening for components that don't have a name. In this case it's an imported icon.

georgegillams: Yeah I would read that as "The default import from the package"

PR #1515, opened 2019-06-25
[BPKR-43]: New and updated icons

ojcurtis:


ojcurtis:
There is an updated star-outline that needs to come in from Chris so waiting for this before the merge can happen

ojcurtis:
Screenshot 2019-06-25 at 10 36 28

PR #1518, opened 2019-06-27
[BPK-2642] Infinite scroll story bugs

shaundon:
Fixed some small issues with the storybook, diff should be self evident.

### DIFF
--- a/packages/bpk-component-infinite-scroll/stories.js
+++ b/packages/bpk-component-infinite-scroll/stories.js
@@ -173,7 +173,7 @@ storiesOf('bpk-component-infinite-scroll', module)
           onClick={() => {
             const newElements = [];
             const k = Math.floor(Math.random() * 10);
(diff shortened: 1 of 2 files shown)


packages/bpk-component-infinite-scroll/stories.js
```diff
@@ -172,7 +173,7 @@ storiesOf('bpk-component-infinite-scroll', module)
           onClick={() => {
             const newElements = [];
             const k = Math.floor(Math.random() * 10);
-            for (let i = 0; i < 100; i = 1) {
+            for (let i = 0; i < 100; i += 1) {
```
ojcurtis: Is this a best practise or does `i++` work/look better?

shaundon: eslint discourages using unary operators

ojcurtis: No problem! In which case 👍

PR #1519, opened 2019-06-27
WOM-119 - Bpk onChange mask removal

Andi24:
Integrating the component into the codebase uncovered that we were passing back out from the component the user's phone number together with the country code. This PR cleans up this behaviour so that the input only returns the phon number. This change also greatly simplifies the masking logic as the mask always needs to be added (i.e. the `value` of the input will never contain the country code mask)


packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -111,28 +111,30 @@ const BpkPhoneInput = (props: Props) => {
...
+    let { value: phoneValueWithCode } = formFieldEvt.target;
+    const { name: fieldPath } = formFieldEvt.target;
+
+    if (phoneValueWithCode.charAt(`+${dialingCodeText}`.length) !== ' ') {
+      const number = phoneValueWithCode.slice(`+${dialingCodeText}`.length);
+      phoneValueWithCode = `+${dialingCodeText} ${number}`;
+    }
+
+    const phoneNumber = phoneValueWithCode.split(' ')[1] || '';
+    const passThroughEvent = {
```
tiagoengel: I'm just afraid we won't be able to merge it back to master like this. This is removing all other properties from the event, like `preventDefault` that people might be using. What do you think of just mutating the event object and returning the original event instead?

Andi24: great shout - feedback applied

tiagoengel: this will still result in an error @Andi24. 

```
const a = new Event('MouseEvent')
a.preventDefault()
// undefined
const b = { ...a }
b.preventDefault()
// TypeError: preventDefault is not a function
```

Spreading the object will only copy the properties not apply the prototype, I think the simplest solution is to just mutate the original event.


```formFieldEvt.target.value = newValue```

Andi24: the issue with mutating the original event object is getting an ESlint error - do you think it's more desirable to make an eslint exception block on this param reassing than losing the prototype?

tiagoengel: This is a case where disabling the eslint rule is acceptable, it's better than losing the event. Another solution would be to call `onChange` with two arguments.

`onChange(originalEvent, phoneNumber)`. This will be a breaking change but it seems to me this is the best solution.

What do you think?

Andi24: Refactoring the `onChange` to take parameters would mean adding an exception to read the `value` from the parameter rather than the received value in a method that is generically used across multiple fields in the context where the phone input is used. I would rather add an eslint exception in this case.

Andi24: disabled eslint rule

PR #1526, opened 2019-06-28
[BPK-2630] Fix mobile dialog scrim transition bug

georgegillams:


packages/bpk-scrim-utils/src/bpk-scrim.scss
```diff
@@ -47,9 +47,5 @@
 
   &--appear-active {
     opacity: $bpk-scrim-opacity;
-
-    @include bpk-breakpoint-mobile {
-      opacity: $bpk-scrim-mobile-opacity;
```
georgegillams: This value was 1, which is why, on mobile, it was causing it to reach full opacity befoore returning to the final opacity value of 0.7

UNRELEASED.yaml
```diff
@@ -1,6 +1,12 @@
 # Unreleased
 
 > Place your changes below this line.
+FIXED:
```
ojcurtis: Can you change this to **Fixed:**

georgegillams: I'll leave it as is for now as it's what the release script expects!

georgegillams:
> Can an entry be added to UNRELEASED

Lol - it was but I didn't update it when it became UNRELEASED.yaml 🤦‍♂

PR #1534, opened 2019-07-05
[BPK-2668] Fix behaviour when data is changed to an empty array

tiagoengel:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
+++ b/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
@@ -154,11 +154,7 @@ const withInfiniteScroll = <T: ExtendedProps>(
           index: 0,
           elementsPerScroll: this.props.elementsPerScroll,
           elementsToRender: [],
-        }).then(newState => {
-          this.setState({
-            ...newState,
-          });
-        });
+        }).then(newState => this.setStateAfterDsUpdate(newState));
       }
     }
 
@@ -167,6 +163,15 @@ const withInfiniteScroll = <T: ExtendedProps>(
       if (this.sentinel) {
         this.observer.unobserve(this.sentinel);
       }
+    }
+
+    setStateAfterDsUpdate(newState: State) {
+      const { isListFinished } = newState;
+      this.setState({
+        ...newState,
+        elementsToRender: isListFinished ? [] : newState.elementsToRender,
+        index: isListFinished ? 0 : newState.index,
+      });
     }
 
     updateData = () => {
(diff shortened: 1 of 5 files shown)


packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -169,6 +165,15 @@ const withInfiniteScroll = <T: ExtendedProps>(
       }
     }
 
+    setStateAfterDsUpdate(newState: State) {
+      const { isListFinished } = newState;
+      this.setState({
+        ...newState,
+        elementsToRender: isListFinished ? [] : newState.elementsToRender,
```
georgegillams: I'm a little confused about setting `elementsToRender: []` and `index: 0`.
Maybe it's just the state key names that makes this seem a little counter-intuitive

tiagoengel: So the way it works is basically like this:
- Normal behaviour of `fetchItems` (when the user is scrolling)
  - Called when user hits the end of the page
  - Calls dataSource.fetchItems and fetch new items from `index` (last item) + `elementsToRender`
  - If the result is empty it means the list is finished, so the function returns `isListFinished` = true and keeps the `elementsToRender` and `index` as it is.

- Called via `updateData` or changing the dataSource prop.
  - This case is done programatically and is supposed to be used in response to changes in the "source" of the data. Think of flight search when you have preview (or cached) values and after a while the server updates you with the real data.
  - This triggers a call the `fetchItems` with the intention to replace the whole data set, and re render the whole list.
  - If we get `isListFinished` in response to that, it means the datasource is empty and we should "reset" the list, and we do that by setting `elementsToRender` to `[]` and `index` to `0`.

Hope that helps clarifying a bit :)

georgegillams: Yeah that makes sense 👍 

Would it not be possible for `fetchItems` to ensure that if `isListFInished` then `elementsToRender` and `index` are also appropriately set to `[]` and `0`?

tiagoengel: Not really because of the first usage, in that case it should not be set to zero or reset the list was we want to keep the current data visible.

It would be possible to put all this logic in the `fetchItems` function but it is already complex and big enough, I don't think adding yet another param and more conditionals there is a good solution

georgegillams: Ok - I guess I dislike the way we're inferring that the dataset is empty based on `isListFinished`. Seems hard to understand. Could we add a comment above line 170 to explain so that in the future it doesn't confuse someone?

tiagoengel: We can yes, but that is actually the only way to infer that. If the data set returns an empty array when called with start position `0` then is safe to assume it's empty.

tiagoengel: done @georgegillams

PR #1536, opened 2019-07-10
Add `yAxisDomain` to barchart component

hintofbasil:
The y axis domain is currently hardcoded to 0 and the height of the
largest bar.  This results in the inability to accurately show
percentage based data or small variations in large values.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-component-barchart/README.md
+++ b/packages/bpk-component-barchart/README.md
@@ -25,7 +25,7 @@ const priceData = [
 export default () => (
   <BpkBarchart
     xAxisLabel="Weekday"
-    yAxislabel="Price (£)"
+    yAxisLabel="Price (£)"
     xScaleDataKey="day"
     yScaleDataKey="price"
     initialWidth={500}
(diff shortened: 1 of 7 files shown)


packages/bpk-component-barchart/README.md
```diff
@@ -25,7 +25,7 @@ const priceData = [
 export default () => (
   <BpkBarchart
     xAxisLabel="Weekday"
-    yAxislabel="Price (£)"
+    yAxisLabel="Price (£)"
```
georgegillams: Great catch 😛

packages/bpk-component-barchart/README.md
```diff
@@ -58,6 +58,7 @@ export default () => (
 | yAxisMargin                             | number                                | false    | 2.625            |
 | yAxisTickValue                          | func                                  | false    | identity         |
 | yAxisNumTicks                           | number                                | false    | null             |
+| yAxisDomain                             | array                                 | false    | See prop details |
```
georgegillams: I think it would be clearer if we call out that the component will calculate these if not provided

```suggestion
| yAxisDomain                             | array                                 | false    | calculated by component |
```

hintofbasil: Fixed.  Had to update the padding for the whole table.

packages/bpk-component-barchart/README.md
hintofbasil: The joys of copy paste!

PR #1553, opened 2019-07-26
[BPK-2731]: New configurable nudger component

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


UNRELEASED.yaml
```diff
@@ -24,3 +24,7 @@
 #   FIXED:
 #     - bpk-component-horcrux:
 #       - Fixed issue where `BpkHorcrux` would occasionally possess the living.
+
+ADDED:
+  - bpk-component-nudger:
+    - Created new `BpkConfigurableNudger` component that allow the nudger to be configured to different types such as dates, text, etc.
```
georgegillams:
```suggestion
    - Created new `BpkConfigurableNudger` component that allow the nudger to be configured to use different types such as dates, text, etc.
```

shaundon:
```suggestion
    - Added `BpkConfigurableNudger`, which is more flexible than the existing `BpkNudger` and works with any data.
```

packages/bpk-component-nudger/src/BpkNudger.js
```diff
@@ -44,82 +47,34 @@ const BpkNudger = props => {
...
-      </BpkButton>
-    </div>
+    <BpkConfigurableNudger
+      id={id}
+      min={min}
+      max={max}
+      value={value}
+      onChange={onChange}
+      className={className}
+      inputClassName={getClassName('bpk-nudger__input--numeric')}
```
georgegillams: nice 😁

packages/bpk-component-nudger/README.md
```diff
@@ -10,6 +10,8 @@ npm install bpk-component-nudger --save-dev
 
 ## Usage
 
+### Default `BpkNudger`
```
shaundon:
```suggestion
### `BpkNudger`
```

The `default` bit feels redundant.

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,104 @@ class App extends Component {
 }
 ```
 
+### Configurable `BpkConfigurableNudger`
```
shaundon:
```suggestion
### `BpkConfigurableNudger`
```

'Configurable' feels redundant.

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,104 @@ class App extends Component {
...
+      value: 1,
+    };
+  }
+
+  handleChange = value => {
+    this.setState({ value });
+  };
+
+  compareValues = (a, b): number => {
+    const options = ['economy', 'premium', 'business', 'first'];
```
shaundon: Can we abstract `options` out seeing as it's defined three times?

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,104 @@ class App extends Component {
...
+    this.state = {
+      value: 1,
+    };
+  }
+
+  handleChange = value => {
+    this.setState({ value });
+  };
+
+  compareValues = (a, b): number => {
```
shaundon:
```suggestion
  compareValues = (a, b): => {
```

I _think_ we usually keep Flow types out of the example code in readmes.

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,104 @@ class App extends Component {
...
+          max="first"
+          value={this.state.value}
+          onChange={this.handleChange}
+          decreaseButtonLabel="Decrease"
+          increaseButtonLabel="Increase"
+          compareValues={this.compareValues}
+          incrementValue={this.incrementValue}
+          decrementValue={this.decrementValue}
+          formatValue={this.formatValue}
+          inputClassName={getClassName('bpk-nudger-configurable')}
```
shaundon: I'd just remove this line, `getClassName` isn't defined and we generally only document the most obvious props and not the more 'advanced' stuff like passing in class names.

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,104 @@ class App extends Component {
...
+    this.setState({ value });
+  };
+
+  compareValues = (a, b): number => {
+    const options = ['economy', 'premium', 'business', 'first'];
+    const [aIndex, bIndex] = [options.indexOf(a), options.indexOf(b)];
+    return aIndex - bIndex;
+  };
+
+  incrementValue = (a): string => {
```
shaundon: Can we make this a bit more explicit? We should aim to avoid variables with undescriptive names

```suggestion
  incrementValue = (currentValue): string => {
```

packages/bpk-component-nudger/src/BpkNudger-test.js
```diff
@@ -57,6 +58,27 @@ describe('BpkNudger', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('should render as a configurable nudger correctly', () => {
+    const tree = renderer
+      .create(
+        <BpkConfigurableNudger
+          id="nudger"
+          min={1}
+          max={9}
```
shaundon: Could we make this test do something other than numbers? Just because it's not really that different from the other tests otherwise. String should be good

packages/bpk-component-nudger/src/BpkNudger.js
```diff
@@ -44,82 +47,34 @@ const BpkNudger = props => {
...
+      increaseButtonLabel={increaseButtonLabel}
+      decreaseButtonLabel={decreaseButtonLabel}
+      buttonType={buttonType}
+      compareValues={compareValues}
+      incrementValue={incrementValue}
+      decrementValue={decrementValue}
+      formatValue={formatValue}
```
shaundon: If you're not doing any logic on `id`, `min`, `max` and so on, you can just pass them through with `...rest` to save some code

packages/bpk-component-nudger/README.md
k0nserv: None of these use `this` so they probably shouldn't be instance methods and they definitely don't need to be bound to `this`

packages/bpk-component-nudger/src/BpkConfigurableNudger.js
```diff
@@ -0,0 +1,145 @@
...
+  COMMON_DEFAULT_PROPS,
+} from './common-types';
+import STYLES from './BpkNudger.scss';
+
+const getClassName = cssModules(STYLES);
+
+const AlignedMinusIcon = withButtonAlignment(MinusIcon);
+const AlignedPlusIcon = withButtonAlignment(PlusIcon);
+
+type Props = {
```
k0nserv: We should make this generic over `T` instead this way we can introduce more type safety than `any`. For example currently the following is valid

```js
const badIncrement = (current: number): string => {
  return "hello";
}
```

Here's a full [example](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDieAdngE4CWAxgArlw4DOAPACoB8YAvGAN6oAkJRLVyeALakMANQCGMAK54AXGAAUbAJQ8ubADSoAvgG5MuAvUZMe-ISLGTp8pao1ySWHdy4esh03RqOBImDDAAIzkAEwBJUXEpEnDedWpFcnFktRJFCQiKLTUwqhIAc107QXEMDJIwACIACzwYeAazQNRg0PCy0goaK2Y1YjIqOgZmFlz8ii5eATAwQWBgMAASADF4BC3KAA8AWTx7BKdkl2U1KLjzpIwAs26QsLAcKaY1YZtFs8cHlc3Ld4gDpE9UEA)

packages/bpk-component-nudger/README.md
ojcurtis: Sure thing, I guess this was more of a good example people could just copy and run with straight away?

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,104 @@ class App extends Component {
...
+| Property            | PropType                      | Required | Default Value |
+| ------------------- | ----------------------------- | -------- | ------------- |
+| id                  | string                        | true     | -             |
+| decreaseButtonLabel | string                        | true     | -             |
+| increaseButtonLabel | string                        | true     | -             |
+| max                 | number                        | true     | -             |
+| min                 | number                        | true     | -             |
+| onChange            | func                          | true     | -             |
+| value               | number                        | true     | -             |
+| compareValues       | func                          | true     | -             |
```
k0nserv: We need to better document these functions. Especially `compareValues`

packages/bpk-component-nudger/src/BpkConfigurableNudger.js
ojcurtis: Ah this might be the thing me and shaun were looking for earlier! As the BpkNudger was throwing errors when passing the function through to here which is why we used any as the Generics didn't work

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,104 @@ class App extends Component {
...
+  };
+
+  compareValues = (a, b): number => {
+    const options = ['economy', 'premium', 'business', 'first'];
+    const [aIndex, bIndex] = [options.indexOf(a), options.indexOf(b)];
+    return aIndex - bIndex;
+  };
+
+  incrementValue = (a): string => {
+    const options = ['economy', 'premium', 'business', 'first'];
```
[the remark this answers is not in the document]
ojcurtis: Could you provide some more detail about this suggestion?

packages/bpk-component-nudger/README.md
ojcurtis: So they shouldn't be inside the class function?

packages/bpk-component-nudger/src/BpkConfigurableNudger.js
k0nserv: This should definitely work

packages/bpk-component-nudger/README.md
shaundon: Yeah totally! But that makes it even more of a reason to follow best practice right? If people are going to copy and use this in their own code we should make sure it's 💯

packages/bpk-component-nudger/README.md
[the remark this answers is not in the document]
shaundon: As in preventing accessing things outside the bound of the array, so like in this example if they didn't set a maximum but kept clicking the + button they'll eventually get `undefined` because they'll go off the end of the array. So we could put an `if` in here to stop that happening.

packages/bpk-component-nudger/src/BpkConfigurableNudger.js
shaundon: @k0nserv we ran into issues and got stuck. I'll send you the code and see if you have any ideas

packages/bpk-component-nudger/README.md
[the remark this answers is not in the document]
ojcurtis: I don't think that scenario would ever happen as max and min are required props so it would never get over/undefined due to there always being the limit value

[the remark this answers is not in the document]
ojcurtis: I am having issues seeing how a user could get the function to go out of bounds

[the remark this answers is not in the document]
shaundon: That's a good point, I didn't realise max and min were required. I'm going to defer to @k0nserv in that case

packages/bpk-component-nudger/README.md
ojcurtis: Done

packages/bpk-component-nudger/README.md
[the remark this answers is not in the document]
k0nserv: What if there's a bug in our code 😱? It's fairly reasonable behaviour to guard it with an out of bounds check imo

packages/bpk-component-nudger/README.md
k0nserv: They could either be top level functions or static methods

packages/bpk-component-nudger/src/BpkConfigurableNudger.js
k0nserv: We resolved this right?

shaundon: We did :)

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,116 @@ class App extends Component {
...
+| formatValue         | func                          | true     | -             |
+| className           | string                        | false    | null          |
+| inputClassName      | string                        | false    | null          |
+| buttonType          | oneOf('secondary', 'outline') | false    | secondary     |
+
+### `compareValues`
+
+A comparator function similar to [Array.prototype.sort()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description) for comparing two values in your data set.
+
+The function is used to take either the min or max value provided and compare it to the current value to determine if the value can be incremented or decremented.
```
k0nserv: Can we mention explicitly when people should return a value less than 0, equal to zero, and greater than zero here? Saves people going to look at `Array.prototype.sort`.

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,116 @@ class App extends Component {
...
+| incrementValue      | func                          | true     | -             |
+| decrementValue      | func                          | true     | -             |
+| formatValue         | func                          | true     | -             |
+| className           | string                        | false    | null          |
+| inputClassName      | string                        | false    | null          |
+| buttonType          | oneOf('secondary', 'outline') | false    | secondary     |
+
+### `compareValues`
+
+A comparator function similar to [Array.prototype.sort()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description) for comparing two values in your data set.
```
k0nserv: By convention you usually don't include the `()` when writing out methods on builtin prototypes. So `Array.prototype.sort` not `Array.prototype.sort()`

packages/bpk-component-nudger/src/BpkConfigurableNudger.js
```diff
@@ -0,0 +1,148 @@
...
+const getClassName = cssModules(STYLES);
+
+const AlignedMinusIcon = withButtonAlignment(MinusIcon);
+const AlignedPlusIcon = withButtonAlignment(PlusIcon);
+
+type Props<T> = {
+  ...$Exact<CommonProps<T>>,
+  inputClassName: ?string,
+  formatValue: T => string,
+  incrementValue: T => mixed,
```
k0nserv: Why are we using `mixed` and not `T` for the return value? `mixed` is intended for use when the type is not known, but we infact know that the type here should be `T`

packages/bpk-component-nudger/src/BpkNudger.js
```diff
@@ -16,110 +16,53 @@
...
-  }
+const BpkNudger = (props: Props) => {
+  const { ...rest } = props;
 
-  const inputStyles = [getClassName('bpk-nudger__input')];
-  if (buttonType === 'outline') {
-    inputStyles.push(getClassName('bpk-nudger__input--outline'));
-  }
+  const compareValues = (a: number, b: number): number => a - b;
+  const incrementValue = (currentValue: number): number => currentValue + 1;
```
k0nserv: These don't need to be declared inside the `render` function. Doing so means they are needlessly redelcared on each render and might also confuse the reader by making it look like they are closures over some value which they are not

packages/bpk-component-nudger/src/common-types.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+/* @flow strict */
+import PropTypes from 'prop-types';
+
+export type CommonProps<T> = {
+  id: string,
+  min: T,
+  max: T,
+  value: T,
+  onChange: <T>(T) => void,
```
k0nserv: I don't think we need `<T>` here. `T` is already in scope so to say

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,116 @@ class App extends Component {
...
+  }
+
+  handleChange = value => {
+    this.setState({ value });
+  };
+
+  render() {
+    return (
+      <div>
+        <BpkLabel htmlFor="my-nudger">Number of passengers</BpkLabel>
```
shaundon:
```suggestion
        <BpkLabel htmlFor="nudger">Number of passengers</BpkLabel>
```

Wrong ID

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,116 @@ class App extends Component {
...
+
+The function is used to take either the min or max value provided and compare it to the current value to determine if the value can be incremented or decremented.
+
+### `incrementValue` & `decrementValue`
+
+Functions that handle the incrementing or decrementing of the current selected value.
+
+### `formatValue`
+
+A simple function that will allow you to set the format of the display value e.g. local dates or times.
```
shaundon: These descriptions are brilliant, nice one 😄

packages/bpk-component-nudger/stories.js
```diff
@@ -59,6 +61,63 @@ class NudgerContainer extends Component {
...
+  }
+
+  handleChange = value => {
+    this.setState({ value });
+  };
+
+  render() {
+    return (
+      <div>
+        <BpkLabel htmlFor="my-nudger">Traveller Class</BpkLabel>
```
shaundon:
```suggestion
        <BpkLabel htmlFor="nudger">Traveller Class</BpkLabel>
```

packages/bpk-component-nudger/src/common-types.js
shaundon: Ah good point

packages/bpk-component-nudger/src/common-types.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+/* @flow strict */
+import PropTypes from 'prop-types';
+
+export type CommonProps<T> = {
+  id: string,
+  min: T,
+  max: T,
+  value: T,
+  onChange: T => void,
```
k0nserv: This should be `onChange: T => mixed` imo since people might write a single line function that implicitly returns some value that we don't care about, but which will cause a flow error

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,122 @@ class App extends Component {
...
+| inputClassName      | string                        | false    | null          |
+| buttonType          | oneOf('secondary', 'outline') | false    | secondary     |
+
+### `compareValues`
+
+A comparator function similar to [Array.prototype.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description) for comparing two values in your data set.
+
+The function is used to take either the min or max value provided and compare it to the current value to determine if the value can be incremented or decremented.
+
+The `compareValues` requires two input values `currentValue` and `compareValue` (which will be the `min` or `max` value provided) function should return the following given each scenario.
```
shaundon: The wording feels a bit clunky to me, how about:

```suggestion 
`compareValues` requires two input values: `currentValue` and `compareValue` (which will be the `min` or `max` value provided). It so return one of the following given each scenario:

* If `currentValue < compareValue`, return a negative number.
* If `currentValue == compareValue`, return zero.
* If `currentValue > compareValue`, return a positive number.
```

packages/bpk-component-nudger/README.md
```diff
@@ -47,19 +49,126 @@ class App extends Component {
...
+A comparator function similar to [Array.prototype.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description) for comparing two values in your data set.
+
+The function is used to take either the min or max value provided and compare it to the current value to determine if the value can be incremented or decremented.
+
+`compareValues` requires two input values: `currentValue` and `compareValue` (which will be the `min` or `max` value provided). It so return one of the following given each scenario:
+
+* If `currentValue < compareValue`, return a negative number.
+* If `currentValue == compareValue`, return zero.
+* If `currentValue > compareValue`, return a positive number.
+Given `currentValue` and `compareValue`:
```
k0nserv: These examples are only correct at the extremes i.e. `max` and `min`

packages/bpk-component-nudger/stories.js
```diff
@@ -59,6 +61,68 @@ class NudgerContainer extends Component {
   }
 }
 
+const options = ['economy', 'premium', 'business', 'first'];
+const compareValues = (a, b): number => {
```
k0nserv: Why haven't we typed the arguments in these four functions? Only the return value is typed

ojcurtis: Good catch!

PR #1554, opened 2019-07-26
[BPK-2174] Add rtl metadata

tiagoengel:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/bpk-svgs/tasks/metadata/automirror.js
```diff
@@ -0,0 +1,17 @@
+const fs = require('fs');
```
ojcurtis: Do we need to add the licence header here?

tiagoengel: yep, good catch

PR #1561, opened 2019-08-01
[BPK-2784]: Fix for babel 7 to support IE 11 and Edge 16+

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -23,6 +23,7 @@ module.exports = {
       {
         targets: {
           node: 'current',
+          browsers: ['ie >= 11'],
         },
       },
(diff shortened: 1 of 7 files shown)


ojcurtis:
This PR solves the issue of the default list Babel 7 uses https://babeljs.io/docs/en/babel-preset-env#browserslist-integration which the default set is based on `default` list that is defined here: https://github.com/browserslist/browserslist

We are essentially adding more support on top of the default list by defining the IE >= 11 in the babel file

PR #1567, opened 2019-08-08
[BPK-2784] Add IE11 support to Storybook webpack config

shaundon:
Storybook wasn't working in IE11. I changed the `env` setting in the Webpack config for Storybook and it seems to be ok now 😄 

Screen Shot 2019-08-08 at 12 37 58

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -13,8 +13,18 @@ module.exports = {
     rules: [
       {
         test: /\.jsx?$/,
-        use: ['babel-loader'],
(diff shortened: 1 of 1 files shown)


georgegillams:
Doesn't this break the base stylesheet tho - that whole issue @ojcurtis was having?

shaundon:
> Doesn't this break the base stylesheet tho - that whole issue @ojcurtis was having?

@georgegillams I don't believe so - this change is only inside the Storybook folder so it doesn't affect our components themselves. I ran the build process twice and nothing changed.

PR #1568, opened 2019-08-08
[BPK-2781]: Added ability to apply borderRadius to BpkImage

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-component-image/README.md
+++ b/packages/bpk-component-image/README.md
@@ -111,19 +111,19 @@ export default () => (
 
 ## Props
 
-| Property                     | PropType             | Required | Default Value       |
-| ---------------------------- | -------------------- | -------- | ------------------- |
-| altText                      | string               | true     | -                   |
-| height                       | number               | true     | -                   |
-| src                          | string               | true     | -                   |
-| width                        | number               | true     | -                   |
-| borderRadiusStyle            | oneOf(CORNER_STYLES) | false    | null                |
-| className                    | string               | false    | null                |
-| inView                       | bool                 | false    | true                |
-| loading                      | bool                 | false    | false               |
-| onLoad                       | func                 | false    | null                |
-| style                        | object               | false    | {}                  |
-| suppressHydrationWarnings    | bool                 | false    | false               |
+| Property                     | PropType                         | Required | Default Value             |
+| ---------------------------- | -------------------------------- | -------- | ------------------------- |
+| altText                      | string                           | true     | -                         |
+| height                       | number                           | true     | -                         |
+| src                          | string                           | true     | -                         |
+| width                        | number                           | true     | -                         |
+| borderRadiusStyle            | oneOf(BORDER_RADIUS_STYLES)      | false    | BORDER_RADIUS_STYLES.none |
+| className                    | string                           | false    | null                      |
+| inView                       | bool                             | false    | true                      |
+| loading                      | bool                             | false    | false                     |
+| onLoad                       | func                             | false    | null                      |
+| style                        | object                           | false    | {}                        |
+| suppressHydrationWarnings    | bool                             | false    | false                     |
 
 ## BpkBackgroundImage
 
--- a/packages/bpk-component-image/README.md
+++ b/packages/bpk-component-image/README.md
@@ -111,18 +111,19 @@ export default () => (
 
 ## Props
 
-| Property                     | PropType  | Required | Default Value       |
-| ---------------------------- | --------- | -------- | ------------------- |
-| altText                      | string    | true     | -                   |
-| height                       | number    | true     | -                   |
-| src                          | string    | true     | -                   |
-| width                        | number    | true     | -                   |
-| className                    | string    | false    | null                |
-| inView                       | bool      | false    | true                |
-| loading                      | bool      | false    | false               |
-| onLoad                       | func      | false    | null                |
-| style                        | object    | false    | {}                  |
-| suppressHydrationWarnings    | bool      | false    | false               |
+| Property                     | PropType             | Required | Default Value       |
+| ---------------------------- | -------------------- | -------- | ------------------- |
+| altText                      | string               | true     | -                   |
+| height                       | number               | true     | -                   |
+| src                          | string               | true     | -                   |
+| width                        | number               | true     | -                   |
+| borderRadiusStyle            | oneOf(CORNER_STYLES) | false    | null                |
+| className                    | string               | false    | null                |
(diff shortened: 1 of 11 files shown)


ojcurtis:
Before:

Screenshot 2019-08-08 at 15 32 29

After:

Screenshot 2019-08-08 at 15 32 10

packages/bpk-component-image/src/BpkImageCornerStyles.js
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+const CORNER_STYLES = { sm: 'sm' };
```
shaundon: As the prop itself is named `borderRadiusStyles`, I think we should name this `BORDER_RADIUS_STYLES` and update usages of it for more consistency.

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -206,6 +218,7 @@ BpkImage.propTypes = {
 };
 
 BpkImage.defaultProps = {
+  borderRadiusStyle: null,
```
shaundon: I think it makes more sense to have a `BORDER_RADIUS_STYLES.none` option which is the default, instead of `null`. It feels like a cleaner API to me if the border radius is always set to one of `BORDER_RADIUS_STYLES`.

packages/bpk-component-image/src/BpkImage.scss
```diff
@@ -83,4 +83,8 @@
   &--no-background {
     @include bpk-image--no-background;
   }
+
+  &--borderRadius-sm {
```
shaundon: We use kebab case in the rest of this file so we should preserve it.

```suggestion
  &--border-radius-sm {
```

packages/bpk-component-image/src/BpkImageCornerStyles.js
ojcurtis: Done :)

packages/bpk-component-image/src/BpkImage.js
ojcurtis: Done :)

packages/bpk-component-image/src/BpkImage.scss
ojcurtis: Done :)

packages/bpk-component-image/README.md
```diff
@@ -111,18 +111,19 @@ export default () => (
...
-| onLoad                       | func      | false    | null                |
-| style                        | object    | false    | {}                  |
-| suppressHydrationWarnings    | bool      | false    | false               |
+| Property                     | PropType                         | Required | Default Value             |
+| ---------------------------- | -------------------------------- | -------- | ------------------------- |
+| altText                      | string                           | true     | -                         |
+| height                       | number                           | true     | -                         |
+| src                          | string                           | true     | -                         |
+| width                        | number                           | true     | -                         |
+| borderRadiusStyle            | oneOf(BORDER_RADIUS_STYLES)      | false    | BORDER_RADIUS_STYLES.none |
```
shaundon: Sorry to be a pain, but:

```suggestion
| borderRadiusStyle            | oneOf(BORDER_RADIUS_STYLES.none, BORDER_RADIUS_STYLES.sm)      | false    | BORDER_RADIUS_STYLES.none |
```

That's how the [RN card one](https://backpack.github.io/components/card?platform=native) does it so it's consistent then.

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -35,6 +36,7 @@ type BpkImageProps = {
   loading: boolean,
   src: string,
   width: number,
+  borderRadiusStyle: ?$Keys<typeof CORNER_STYLES>,
```
shaundon:
```suggestion
  borderRadiusStyle: ?$Keys<typeof BORDER_RADIUS_STYLES>,
```

packages/bpk-component-image/src/BpkImage.js
```diff
@@ -134,6 +137,14 @@ class BpkImage extends Component<BpkImageProps> {
       classNames.push(getClassName('bpk-image--no-background'));
     }
 
+    if (borderRadiusStyle) {
```
shaundon: Now that `borderRadiusStyle` defaults to `none`, this `if` statement will always be matched. To preserve the logic you'd need to change it:

```suggestion
    if (borderRadiusStyle !== BORDER_RADIUS_STYLES.none) {
```

packages/bpk-component-image/src/BpkImageCornerStyles.js
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+const BORDER_RADIUS_STYLES = { none: null, sm: 'sm' };
+
+export default BORDER_RADIUS_STYLES;
```
shaundon: Looks good but we should also rename the file from `BpkImageCornerStyles.js` to `BpkImageBorderRadiusStyles.js`.

packages/bpk-component-image/src/BpkImageCornerStyles.js
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+const BORDER_RADIUS_STYLES = { none: null, sm: 'sm' };
```
shaundon: Sorry I just noticed something else. This shouldn't evaluate to `null`:

```suggestion
const BORDER_RADIUS_STYLES = { none: 'none', sm: 'sm' };
```

packages/bpk-component-image/src/BpkImage.js
ojcurtis: Good spot!

PR #1574, opened 2019-08-13
[BPK-2813]: Updating gray colors

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/bpk-tokens/src/common/base/colors.json
```diff
@@ -11,16 +11,16 @@
...
-    "GRAY_300": "#B2AEBD",
-    "GRAY_400": "#9A95A7",
-    "GRAY_500": "#817B8F",
+    "GRAY_50": "#F1F2F8",
+    "GRAY_100": "#DDDDE5",
+    "GRAY_200": "#CDCDD7",
+    "GRAY_300": "#B2B2BF",
+    "GRAY_400": "#8f90a0",
+    "GRAY_500": "#68697F",
     "GRAY_600": "#696179",
```
georgegillams: I'm actually now thinking maybe we should make this the average of 500 and 700 🤔 
```suggestion
     "GRAY_600": "#565770",
```

packages/bpk-tokens/src/common/base/colors.json
```diff
@@ -11,16 +11,16 @@
...
-    "GRAY_700": "#524C61",
+    "GRAY_700": "#444560",
     "GRAY_800": "#3B344B",
```
georgegillams: As above
```suggestion
     "GRAY_800": "#2B2C4B",
```

packages/bpk-tokens/src/common/base/colors.json
ojcurtis: I would suggest we leave it as they are, as this should hopefully mean consumers are forced to change the usage from deprecated values to non deprecated values 😄

packages/bpk-tokens/src/common/base/colors.json
ojcurtis: I would suggest we leave it as they are, as this should hopefully mean consumers are forced to change the usage from deprecated values to non deprecated values 😄

PR #1576, opened 2019-08-15
[BPK-2708]: Added new `bpk-component-rating`

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/bpk-component-rating/src/BpkRating.scss
```diff
@@ -0,0 +1,49 @@
...
+  grid-auto-flow: column;
+  grid-column-gap: $bpk-spacing-sm;
+
+  &--base-rating {
+    width: $bpk-text-base-line-height * 2;
+    height: $bpk-text-base-line-height * 2;
+    padding: $bpk-spacing-sm;
+    grid-column: 1;
+    grid-row: 1 / 3;
+    border-radius: $bpk-text-xl-line-height;
```
shaundon: There should be a `border-radius-pill` token we can use here. If not, we can add one ourselves. I know RN has one but maybe it doesn't exist for web. What border radius are buttons using? Those are rounded so we should be able to mirror whatever's done there.

packages/bpk-component-rating/src/BpkRating.scss
```diff
@@ -0,0 +1,49 @@
...
+    color: $bpk-color-white;
+    text-align: center;
+    vertical-align: middle;
+
+    @include bpk-themeable-property( background-color, --bpk-rating-green, $bpk-color-green-600 );
```
shaundon: Can we name the theme props 'high' 'medium' and 'low', so that they aren't tied to specific colours? Imagine if in future we made the high rating blue by default, now the variable name would be weird.

packages/bpk-component-rating/src/BpkRatingSizes.js
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow */
+
+const RATING_SIZES = { base: 'base', lg: 'lg', sm: 'sm' };
```
shaundon: I know we spoke about this yesterday but we shouldn't be including this if it doesn't do anything yet. 

Imagine if we released this right now - if a consumer looks at the source code they'll see that they can pass any 'base', 'lg' or 'sm' into the component, but nothing will happen if they do provide one of these values, which is a poor experience for them and violates the principle of least surprise.

My advice would be to:
- Not export `RATING_SIZES` from the main package until it's useful for consumers
- Comment out the prop type and flow type from the code for now with a quick `// TODO make use of this` comment next to it

That way we don't lose the code you've written but we don't confuse consumers either

UNRELEASED.yaml
```diff
@@ -24,3 +24,7 @@
 #   FIXED:
 #     - bpk-component-horcrux:
 #       - Fixed issue where `BpkHorcrux` would occasionally possess the living.
+
+ADDED:
+  - bpk-component-rating:
+    - Introducing the web rating component.
```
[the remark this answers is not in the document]
ojcurtis: No problem! I had copied it from a different changelog 😄

packages/bpk-component-rating/src/BpkRatingSizes.js
ojcurtis: That is true but in this case as it will be coming really soon we keep this as is? The README.md does not mention this prop type so unless someone digs into the core code this won't be known 😄

shaundon: If their IDE suggest props based on Flow types they'd still see it..

packages/bpk-component-rating/src/BpkRating.scss
ojcurtis: So I tried with radius-pill and it produced the following output. I had originally hardcoded with a specific value and then found this one was close enough. So in this case I will create new tokens specific to this component 😄 

Screenshot 2019-08-15 at 10 18 19

shaundon: Ok cool, sounds good!

packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -0,0 +1,85 @@
...
+  } else if (value >= 6 && value <= 7.9) {
+    scoreStyles.push(getClassName('bpk-rating--medium-rating'));
+  }
+
+  return (
+    <div>
+      <div className={classNames} aria-label={ariaLabel}>
+        <BpkText
+          textStyle="base"
+          tagName="p"
```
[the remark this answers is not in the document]
ojcurtis: This may change to a div as the value type might change to a node once I get clarification from Rhys around the spec

packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -0,0 +1,85 @@
...
+  size: string,
+  value: number,
+  className: ?string,
+};
+const BpkRating = (props: Props) => {
+  const { ariaLabel, title, subtitle, value, className, size } = props;
+  const classNames = getClassName('bpk-rating', className);
+  const scoreStyles = [getClassName(`bpk-rating--${size}-rating`)];
+
+  if (value <= 5.9) {
```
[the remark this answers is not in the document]
ojcurtis: That sounds good to me!

packages/bpk-component-rating/src/BpkRatingSizes.js
georgegillams: I agree with @shaundon that we shouldn't release code that isn't functional - however we won't be releasing this because of `"private": true` in the `package.json`.

As a general rule though - I also think including non-functional code bloats PR's - so in the interest of smaller PR's I'm also not a fan

ojcurtis: Sure thing! Will kill it out for now and then when I do the ticket to add other sizes after this PR is merged will just copt and paste it back in

ojcurtis: Commented out the code so it exists but has TODOs around it 😄

packages/bpk-component-rating/src/BpkRating.scss
ojcurtis: Updated the colours as per the suggestion so will need to add it to definitions and also to themeAttributes

packages/bpk-component-rating/src/BpkRating.js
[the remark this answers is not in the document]
ojcurtis: Having a look at what is done on the hotels flows for the existing non Backpack component and it seems that span has been used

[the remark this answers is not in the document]
shaundon: Yeah perhaps span is the most appropriate

packages/bpk-component-rating/src/BpkRating.scss
ojcurtis: Created new token to handle this

packages/bpk-component-rating/src/BpkRating.js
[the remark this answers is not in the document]
ojcurtis: Used span and seems better now

packages/bpk-component-rating/src/BpkRating-test.js
```diff
@@ -0,0 +1,54 @@
...
+      .create(
+        <BpkRating
+          ariaLabel="6.7 Average might recommend"
+          title="Average"
+          subtitle="Might recommend"
+          value={6.7}
+        />,
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
shaundon: We should have some more tests here, I'm thinking:

- One for each type of rating (low, medium, high) to test the logic of setting the right class depending on the value.
- A couple of 'edge case' ones - for example what happens if you put `-1`, or `15`, or `4.1964` to test that the value is correctly sanitised.

packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -0,0 +1,92 @@
...
+  if (value >= HIGH_RATING_THRESHOLD) {
+    scoreStyles.push(getClassName('bpk-rating--high-rating'));
+  } else if (value >= MEDIUM_RATING_THRESHOLD) {
+    scoreStyles.push(getClassName('bpk-rating--medium-rating'));
+  } else {
+    scoreStyles.push(getClassName('bpk-rating--low-rating'));
+  }
+
+  return (
+    <div>
```
shaundon: This outermost `div` doesn't seem to have a purpose - it doesn't have any attributes associated with it. Can it be removed?

packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -0,0 +1,92 @@
...
+      <div className={classNames} aria-label={ariaLabel}>
+        <BpkText
+          textStyle="base"
+          tagName="span"
+          className={scoreStyles.join(' ')}
+          aria-hidden="true"
+        >
+          <strong>{value}</strong>
+        </BpkText>
+        <div className={getClassName('bpk-rating--text')}>
```
shaundon: This isn't quite right in terms of [BEM naming](http://getbem.com/naming/) - it's block, element, modifier and the text is an element here, not a modifier. So ` bpk-rating__text` makes more sense. 

Above where you have `bpk-rating--high-rating` makes sense because the `high-rating` part is a modifier.

packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -0,0 +1,92 @@
...
+          <BpkText textStyle="base" tagName="p" aria-hidden="true">
```
shaundon: As with above, not sure if these make sense as `tagName="p"` as it's not a paragraph of text.

packages/bpk-component-rating/src/BpkRating.scss
```diff
@@ -0,0 +1,49 @@
...
+    text-align: center;
+    vertical-align: middle;
+  }
+
+  &--text {
+    padding-left: $bpk-spacing-sm;
+  }
+
+  &--high-rating {
+    @include bpk-themeable-property( background-color, --bpk-rating-high-color, $bpk-color-green-600 );
```
shaundon: Nitpick - we should use consistent spacing here (sadly we can't run this through Prettier as with JS code).

```suggestion
    @include bpk-themeable-property(background-color, --bpk-rating-high-color, $bpk-color-green-600);
```

packages/bpk-tokens/src/web/base/ratings.json
```diff
@@ -0,0 +1,14 @@
+{
+  "imports": [
+    "./aliases.json"
+  ],
+  "global": {
+    "category": "ratings"
+  },
+  "props": {
+    "RATINGS_BASE_BORDER_RADIUS": {
+      "value": "2.625rem",
```
shaundon: Where did this value come from? We typically have an extra layer of indirection here. For example:

In `buttons.json` we use `BORDER_RADIUS_PILL_LG` when defining `BUTTON_BORDER_RADIUS_LG`: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-tokens/src/web/base/buttons.json#L26)

Then, the source of truth for `BORDER_RADIUS_PILL_LG` is here: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-tokens/src/web/base/aliases.json#L24

So in this case we should be able to do a similar thing where you define a value in `aliases.json` and then use it in this file as a variable.

As an aside, is it not possible to use one of the existing border radius tokens in here instead of defining your own? I imagine `BORDER_RADIUS_PILL_LG` might work for your use case.

packages/bpk-component-rating/src/BpkRating.scss
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins';
+
+.bpk-rating {
+  display: flex;
+
+  &--base-rating {
+    width: $bpk-spacing-sm * 4;
```
shaundon: As we're already adding new tokens, why don't we define these there too so that they all share one source of truth?

E.g. in `ratings.json`:

```
    "RATING_WIDTH": {
      "value": "{!SPACING_SM} * 4",
      "type": "color"
    },
```

Then in this file you can just do:

```
width: $bpk-rating-width;
```

We don't need to do these for all things, e.g. `text-align` but where there are numbers or colours involved it's generally a good thing to do.

packages/bpk-tokens/src/web/base/ratings.json
```diff
@@ -0,0 +1,14 @@
+{
+  "imports": [
+    "./aliases.json"
+  ],
+  "global": {
+    "category": "ratings"
+  },
+  "props": {
+    "RATINGS_BASE_BORDER_RADIUS": {
```
shaundon: We typically define these in singular, not plural.

```suggestion
    "RATING_BASE_BORDER_RADIUS": {
```

packages/bpk-component-rating/src/BpkRating-test.js
ojcurtis: Ah yes good catch! Was meant to do that 😄

packages/bpk-component-rating/package-lock.json
```diff
@@ -0,0 +1,279 @@
+{
+	"name": "bpk-component-rating",
+	"version": "0.0.20",
```
georgegillams: Remove this line or it will just cause headaches later on 😆

ojcurtis: Good spot!

packages/bpk-component-rating/src/BpkRating.js
georgegillams: Also the outermost `div` should have `{...rest}` spread onto it

shaundon: Ah yeah, good catch

packages/bpk-component-rating/src/BpkRating.js
ojcurtis: So changing this to span instead causes this issue even when flex is applied to it, so will either need some more styling to do it or should be a paragraph. I will look at what the components that are live are doing

Screenshot 2019-08-16 at 10 37 42

packages/bpk-tokens/src/web/base/ratings.json
ojcurtis: Resolved through use of `border-radius: 50%`

packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -0,0 +1,100 @@
...
+import PropTypes from 'prop-types';
+import clamp from 'lodash.clamp';
+import { cssModules } from 'bpk-react-utils';
+import BpkText from 'bpk-component-text';
+
+import STYLES from './BpkRating.scss';
+// import RATING_SIZES from './BpkRatingSizes'; TODO: make use of this
+
+const getClassName = cssModules(STYLES);
+const MEDIUM_RATING_THRESHOLD = 6;
```
georgegillams: I think these should be configurable... Good to have sensible defaults, but ultimately the consumer should be able to specify their own ranges IMO.

packages/bpk-component-rating/src/BpkRating.scss
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins';
+
+.bpk-rating {
+  display: flex;
+
+  &--base-rating {
```
georgegillams:
```suggestion
  &__rating {
```

packages/bpk-component-rating/src/BpkRating.js
ojcurtis: So for these they will never changed as confirmed by design. Whenever used they need to be consistent so letting user land set these will cause them to be inconsistent

packages/bpk-component-rating/src/BpkRating.scss
ojcurtis: This is due to the version being implemented being the base version which is specific to a certain size. I want to keep this here as once this is merged I will be working on the other sizes so this will change back anyway as to dynamically pick a size based on userland size prop

packages/bpk-component-rating/src/BpkRating-test.js
```diff
@@ -0,0 +1,99 @@
...
+          title="Amazing"
+          subtitle="Brilliant"
+          value={9}
+          className="custom-classname"
+        />,
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+});
```
shaundon: Can we rename these tests a bit to make it clearer what they're testing? I can't really tell from the descriptions. The things we should be testing are:

- Low score
- Medium score
- High score
- Correctly handling values lower than 0
- Correctly handling values higher than 10
- Custom classname

PR #1585, opened 2019-08-23
[No JIRA] Fix semantic text names

k0nserv:
I messed up the names in the last PRs. Didn't follow the naming
convention properly.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-tokens/src/android/base/typography.json
+++ b/packages/bpk-tokens/src/android/base/typography.json
@@ -20,10 +20,22 @@
     },
     "TEXT_COLOR_PRIMARY_LIGHT": {
       "value": "{!GRAY_900}",
+      "deprecated": true,
       "type": "color",
       "category": "text-colors"
     },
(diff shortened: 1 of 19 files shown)


packages/bpk-tokens/src/android/base/typography.json
```diff
@@ -20,10 +20,22 @@
     },
     "TEXT_COLOR_PRIMARY_LIGHT": {
       "value": "{!GRAY_900}",
+      "deprecated": true,
```
ojcurtis: Given it was so short lived do we assume a consumer has started to use these values? As I would rather we delete than just deprecate

But if not we will remove it in the future anyway

k0nserv: Let's remove in next breaking

ojcurtis: Sounds good to me 👍

PR #1589, opened 2019-08-28
OC-693 - Invalid state radio buttons now show as red

ronan-skyscanner:
This change adds:
 * A new property to the BpkRadio component to define a valid/invalid state
 * A new radio-invalid element similar to the radio-normal but red (500)

image

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/bpk-component-radio/src/BpkRadio.scss
```diff
@@ -31,6 +31,12 @@ $bpk-radio-circle-size: $bpk-one-pixel-rem * 10;
     @include bpk-radio--disabled;
   }
 
+  &--invalid {
+    input {
+      background: url(map-get($bpk-elements, radio-invalid)) no-repeat;
```
georgegillams: This should be provided as a mixin too.

packages/bpk-component-radio/src/BpkRadio.js
```diff
@@ -46,6 +64,7 @@ const BpkRadio = props => {
         name={name}
         disabled={disabled}
         aria-label={ariaLabel || label}
+        aria-invalid={isInvalid}
```
georgegillams: This should only be set if valid is defined. With this code, if `valid` is undefined or null this will set `aria-invalid="false"` which isn't necessarily true.

packages/bpk-component-radio/src/BpkRadio.js
```diff
@@ -26,14 +26,32 @@ const getClassName = cssModules(STYLES);
...
+  const isInvalid = valid === false;
 
   if (white) {
     classNames.push(getClassName('bpk-radio--white'));
   }
   if (disabled) {
     classNames.push(getClassName('bpk-radio--disabled'));
   }
+  if (valid) {
+    classNames.push(getClassName('bpk-radio--valid'));
```
georgegillams: I don't think this style is defined...

UNRELEASED.yaml
```diff
@@ -1,25 +1,4 @@
-# UNRELEASED:
```
georgegillams: Undelete this please

georgegillams:
Hey @ronan-skyscanner thanks for the contribution! I've got a couple of comments about the code, but I should warn you that this can't be accepted until we have consesnsus from a design POV.

packages/bpk-component-radio/src/BpkRadio.js
ronan-skyscanner: @georgegillams I think this is protected against by having the explicit check above (https://github.com/Skyscanner/backpack/pull/1589/files/72f1851552272704d0ebb3e082fd8d1e15fad6fc#diff-ff05dd3a3cbb09216a69f345e8cc07ccR40)
This is taken from the BpkInput code.

packages/bpk-component-radio/src/BpkRadio.js
ronan-skyscanner: Removing this

UNRELEASED.yaml
ronan-skyscanner: Will restore that one line.

packages/bpk-component-radio/src/BpkRadio.js
georgegillams: Just checked the logic and it's 💯  👍

PR #1591, opened 2019-08-29
OC-693 - Reworking the look of inputs in invalid state

ronan-skyscanner:
This is a rework of the invalid state for various inputs:
  * Checkboxes should have red outline when invalid
image

  * Label should be red when the associated field is invalid
image
image

  * Radio buttons should have red outline when invalid
image

  * Textarea, input and select should have white background when invalid
image
image
image

[x] Storybooks updated
[x] Snapshot tests added 
[x] Readme updated to add relevant properties
Will review docs once these changes are approved


UNRELEASED.yaml
```diff
@@ -1,25 +1,16 @@
-# UNRELEASED:
```
ojcurtis: Can you re add this back and your changes are appended to the bottom

ronan-skyscanner: Fixed now?

ronan-skyscanner:
> ## Browser support
> 
> Please complete this list of browsers that you've checked this works in.
> 
>     * [ ]  IE11 - checkbox and radio not supported, fails to default
> 
>     * [x]  Edge
> 
>     * [x]  Safari (iOS)
> 
>     * [x]  Chrome (Android)
> 
>     * [x]  Chrome (Desktop)
> 
>     * [x]  Firefox (Desktop)
>     * [x]  Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack/blob/master/CONTRIBUTING.md)

ronan-skyscanner:
> ## Browser support
> 
> Please complete this list of browsers that you've checked this works in.
> 
>     * [ ]  IE11 - checkbox and radio not supported, fails to default
> 
>     * [x]  Edge
> 
>     * [x]  Safari (iOS)
> 
>     * [x]  Chrome (Android)
> 
>     * [x]  Chrome (Desktop)
> 
>     * [x]  Firefox (Desktop)
>     * [x]  Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack/blob/master/CONTRIBUTING.md)

packages/bpk-component-label/src/BpkLabel.js
```diff
@@ -31,10 +31,19 @@ export type Props = {
   disabled: boolean,
   required: boolean,
   white: boolean,
+  invalid: boolean,
```
georgegillams: A label cannot have a concept of valid/invalid.
The field-set should be handling setting the colour as it's not a concern of the label.

The red colour used should also be darker to maintain good contrast.

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -689,6 +721,20 @@
...
+///
+/// @require {mixin} bpk-label
+///
+/// @example scss
+///   .selector {
+///     @include bpk-label();
+///     @include bpk-label--invalid();
+///   }
+
+@mixin bpk-label--invalid {
```
georgegillams: As per my previous comment - this style should not exist...

packages/bpk-component-label/src/BpkLabel.js
ronan-skyscanner: @georgegillams How can a label have a concept of `required`, or `disabled`, both present in the props, but not a concept of `invalid`?

ronan-skyscanner: Now moved to the fieldset

georgegillams:
Hey @ronan-skyscanner 
This is all looking good. The only thing that we aren't sure about is the red labels above the fields. We think that red around the field and in the error message is enough without the label being red too. Rhys said that they weren't crucial. wdyt?

ronan-skyscanner:
Hi @georgegillams,
Happy either way really. On a personal note, I feel like having the labels in red as well makes it clearer and the eye is drawn to the invalid fields. To me it feels more consistent and less "jumpy".
First option:
image

Second option:
image

Your call, let us knwo what you want me to go with

packages/bpk-component-fieldset/src/BpkFieldset.scss
```diff
@@ -29,4 +29,10 @@
     @include bpk-apply-gray-500-to(color);
     @include bpk-text--xs;
   }
+
+  &--invalid {
+    label {
```
georgegillams: This isn't a pattern we generally use. Instead could we create the following style:
```
&__label {
  &--invalid {
    color: $bpk-form-validation-color;
  }
}
```

And then apply the className `bpk-fieldset__label--invalid` to the `BpkLabel` element?

PR #1598, opened 2019-09-02
Bpk 2900 pagination to always be circular

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/UNRELEASED.yaml
+++ b/UNRELEASED.yaml
@@ -23,3 +23,7 @@
 #   FIXED:
 #     - bpk-component-horcrux:
 #       - Fixed issue where `BpkHorcrux` would occasionally possess the living.
+
+FIXED:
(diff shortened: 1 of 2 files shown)


UNRELEASED.yaml
```diff
@@ -23,3 +23,7 @@
 #   FIXED:
 #     - bpk-component-horcrux:
 #       - Fixed issue where `BpkHorcrux` would occasionally possess the living.
+
+FIXED:
+- bpk-component-pagination
+  - Fixed pagination buttons to no longer apply borderRadius from `BpkButton` and are always circular.
```
shaundon:
```suggestion
  - Pagination buttons are now always circular regardless of theming.
```

PR #1602, opened 2019-09-03
Bpk 2889 theming for inputs

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-component-input/README.md
+++ b/packages/bpk-component-input/README.md
@@ -47,13 +47,6 @@ export default () => (
 | onClear          | func                       | if clearable={true} | null                     |
 | valid            | bool                       | false               | null                     |
 
-## Theme Props
-
-* `inputValidIconColor`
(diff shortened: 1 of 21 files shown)


packages/bpk-component-input/README.md
```diff
@@ -47,6 +47,13 @@ export default () => (
 | onClear          | func                       | if clearable={true} | null                     |
 | valid            | bool                       | false               | null                     |
 
+## Theme Props
+
+* `inputValidIconColor`
+* `inputInvalidIconColor`
+* `inputInvalidBorderColor`
+* `inputInvalidBackgroundColor`
```
shaundon: Looks like a few of these lines are left in from when I originally was trying to theme the icon colour, should probably be removed.

ojcurtis: Updated this

PR #1608, opened 2019-09-06
[BPK-2811]: BpkRating multiple size support

ojcurtis:
Small rating
Screenshot 2019-09-06 at 15 55 51

Large rating
Screenshot 2019-09-06 at 15 56 04


packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -72,10 +74,18 @@ const BpkRating = (props: Props) => {
...
+        <BpkText
+          textStyle={RATING_SIZES[size]}
+          tagName="span"
+          aria-hidden="true"
+        >
           <strong>{title}</strong>
         </BpkText>
-        <BpkText textStyle="xs" tagName="span" aria-hidden="true">
+        <BpkText
+          textStyle={size === RATING_SIZES.lg ? 'base' : 'sm'}
```
shaundon: Previously, `textStyle` was `xs` but now it's `base` or `sm`. How come it's changed?

packages/bpk-component-rating/src/BpkRating.scss
```diff
@@ -21,20 +21,38 @@
 .bpk-rating {
   display: flex;
 
+  &--component {
```
shaundon:
```suggestion
  &__component {
```
In BEM `--` is for modifiers, this should be an element so `__`.

packages/bpk-component-rating/src/BpkRating.js
ojcurtis: It is due to `xs` and `sm` being the same values so didn't make a difference so thought it would fit the sizings better rather than missing `sm`

shaundon: Ah ok 👍

packages/bpk-component-rating/src/BpkRating.scss
ojcurtis: Done

PR #1622, opened 2019-09-09
[BPK-2830]: Removed gray theming from all components

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/bpk-component-map/src/BpkMapMarker.scss
```diff
@@ -55,18 +55,14 @@
   }
 
   &--plain {
-    background-color: $bpk-color-gray-700; // For IE
```
georgegillams: Surely IE still needs this?

ojcurtis: Nope as this was added due to the lack of support for `var()` but now it has been removed it is no longer needed.

See this PR https://github.com/Skyscanner/backpack/pull/1463

ojcurtis: Nope as this was added due to the lack of support for `var()` so we had to add this in order for the fallback.

See this PR wheere it was added https://github.com/Skyscanner/backpack/pull/1463

ojcurtis:
For reference it is reverting the work that was completed in this PR: https://github.com/Skyscanner/backpack/pull/1463

PR #1630, opened 2019-09-13
[BPK-2812]: Support vertical layout for rating component

ojcurtis:
Screenshot 2019-09-13 at 10 47 12

### DIFF
--- a/packages/bpk-component-rating/src/BpkRating.js
+++ b/packages/bpk-component-rating/src/BpkRating.js
@@ -91,7 +91,7 @@ const BpkRating = (props: Props) => {
       >
         <strong>{adjustedValue}</strong>
       </BpkText>
-      <div className={getClassName('bpk-rating__text-container')}>
+      <div className={getClassName('bpk-rating__text-wrapper')}>
         <BpkText
           className={textStyles.join(' ')}
           textStyle={RATING_SIZES[size]}
--- a/packages/bpk-component-rating/src/BpkRating.js
+++ b/packages/bpk-component-rating/src/BpkRating.js
@@ -91,7 +91,7 @@ const BpkRating = (props: Props) => {
       >
         <strong>{adjustedValue}</strong>
       </BpkText>
-      <div className={getClassName('bpk-rating__text--container')}>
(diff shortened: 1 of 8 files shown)


packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -73,15 +91,17 @@ const BpkRating = (props: Props) => {
       >
         <strong>{adjustedValue}</strong>
       </BpkText>
-      <div className={getClassName('bpk-rating__text')}>
+      <div className={getClassName('bpk-rating__text--container')}>
```
shaundon: Nitpick but in the past we've tried to avoid using 'container' for things like this and instead say 'wrapper'. I can't remember the exact reason but I think 'container' has some sort of meaning in Reactland. I'll see if I can find any more info

packages/bpk-component-rating/src/BpkRating.scss
```diff
@@ -48,23 +52,39 @@
     font-size: $bpk-font-size-sm;
   }
 
-  &__text {
+  &__text--container {
```
georgegillams:
```suggestion
  &__text-container {
```

packages/bpk-component-rating/src/BpkRating.js
shaundon: https://medium.com/@learnreact/container-components-c0e67432e005

Also should it be `__container` because it's an element, not a modifier?

georgegillams:
```suggestion
      <div className={getClassName('bpk-rating__text-container')}>
```

georgegillams: Actually @shaundon I think it should be `bpk-rating__text-container` because it's not normal to have multiple layers of elements

georgegillams: Actually @shaundon I think it should be `bpk-rating__text-container` because it's not normal to have multiple layers of elements

shaundon: Ah yeah you're right George

PR #1650, opened 2019-09-25
[BPK-3002]: Update-components-to-use-new-brand-p-to-z

ojcurtis:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/bpk-component-progress/src/BpkProgress.scss
```diff
@@ -41,7 +41,7 @@
     transition: width $bpk-duration-base ease-in-out;
     border-radius: $bpk-border-radius-sm;
 
-    @include bpk-themeable-property(background-color, --bpk-progress-bar-fill-color, $bpk-color-green-500);
+    @include bpk-themeable-property(background-color, --bpk-progress-bar-fill-color, $bpk-color-sky-blue);
```
georgegillams: I would have thought the new brand uses a darker blue... 🤔

georgegillams: Ignore me - this is correct!

ojcurtis: No according to the theming 🤔 might check the designs though. I think it keeps in line with the spinner colour

PR #1678, opened 2019-10-07
[ABS-2408]: Add 'roundRating' prop to Star Ratings

sbennett117:
Adds an optional prop to bpk-component-star-rating, `roundRating`. This defaults to false - if true, the rating is rounded to the nearest half star, instead of always being rounded down.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -741,3 +741,6 @@ greedyGestureHandling
 ariaLabel
 props.label
 semver.org
+ - packages/bpk-component-star-rating/README.md
+ROUNDING_TYPES
+ROUNDING_TYPES.down
--- a/.spelling
+++ b/.spelling
@@ -741,6 +741,5 @@ greedyGestureHandling
 ariaLabel
 props.label
 semver.org
- - packages/bpk-component-star-rating/README.md
 ROUNDING_TYPES
 ROUNDING_TYPES.down
--- a/packages/bpk-component-star-rating/README.md
+++ b/packages/bpk-component-star-rating/README.md
@@ -51,7 +51,7 @@ export default () => (
 | large          | bool                    | false    | false         |
 | maxRating      | number                  | false    | 5             |
 | rating         | number                  | false    | 0             |
-| roundRating    | bool                    | false    | false         |
+| rounding       | func                    | false    | n => n        |
 
 ### BpkStar
 
--- a/packages/bpk-component-star-rating/README.md
+++ b/packages/bpk-component-star-rating/README.md
@@ -44,14 +44,14 @@ export default () => (
 
 ### BpkStarRating
 
-| Property       | PropType                | Required | Default Value |
-| -------------- | ----------------------- | -------- | ------------- |
-| ratingLabel    | oneOfType(string, func) | true     | -             |
-| className      | string                  | false    | null          |
-| large          | bool                    | false    | false         |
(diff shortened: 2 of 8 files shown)


packages/bpk-component-star-rating/README.md
```diff
@@ -51,6 +51,7 @@ export default () => (
 | large          | bool                    | false    | false         |
 | maxRating      | number                  | false    | 5             |
 | rating         | number                  | false    | 0             |
+| roundRating    | bool                    | false    | false         |
```
georgegillams: Can this be a function of type `number => number`

```suggestion
| roundRating    | func                    | false    | false         |
```

sbennett117: It could be, the bigger problem I have right now is that something seems to think `roundRating` is a spelling error judging by the failed build 🤔

sbennett117: If it's going to be a function instead of a bool I could call it `rounding` instead?

georgegillams: Yeah that sounds like a better name 💯

sbennett117: So I made the change, but looking over it again I feel like this goes against the spirit of the change we were trying to make, i.e. adding a secondary behaviour 🤔

sbennett117:
So I only just saw your overall review, I see what you're saying now. Ignore my earlier comment!

.spelling
```diff
@@ -741,3 +741,6 @@ greedyGestureHandling
 ariaLabel
 props.label
 semver.org
+ - packages/bpk-component-star-rating/README.md
```
georgegillams: This shouldn't be needed

packages/bpk-component-star-rating/src/BpkStarRating.js
```diff
@@ -77,13 +91,19 @@ BpkStarRating.propTypes = {
   large: PropTypes.bool,
   maxRating: PropTypes.number,
   rating: PropTypes.number,
+  rounding: PropTypes.oneOf([
+    ROUNDING_TYPES.down,
+    ROUNDING_TYPES.up,
+    ROUNDING_TYPES.nearest,
+  ]),
```
georgegillams:
```suggestion
  rounding: PropTypes.oneOf(Object.keys(ROUNDING_TYPES)),
```

sbennett117: I changed this to `Object.values`, if I use `Object.keys` it expects `[down, up, nearest]`

georgegillams: Ah yeah of course it does 🤦‍♂ 
Ignore me 😆

georgegillams: `Object.values` isn't supported by all browsers, so best to roll with what you had before
https://caniuse.com/#search=Object.values

sbennett117: Good catch! Will do 👍

georgegillams:
Merged - will release once I have a stable wifi connection 🎉

sbennett117:
Cheers mate! 🎉

PR #1686, opened 2019-10-11
a11y fixes for accordion

metalix2:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

Browser support
Please complete this list of browsers that you've checked this works in.

+ [ ] IE11
+ [ ] Edge
+ [x] Safari (iOS)
+ [x] Chrome (Android)
+ [x] Chrome (Desktop)
+ [x] Firefox (Desktop)

### No Violations

BpkAccordionA11y

### Before with Violations
BpkAccordionOld

1) ARIA role must be appropriate for the element

2) ARIA attributes must conform to valid values

3) <dl> elements must only directly contain properly-ordered <dt> and <dd> groups, <script> or <template> elements

Also Tested on Android with Talkback, functionality is exact same.

### DIFF
--- a/packages/bpk-component-accordion/src/BpkAccordionItem.js
+++ b/packages/bpk-component-accordion/src/BpkAccordionItem.js
@@ -45,14 +45,13 @@ const BpkAccordionItem = props => {
     );
   }
 
-  const titleId = `${id}_title`;
   const contentId = `${id}_content`;
 
   return (
     <div id={id} {...rest}>
       <dt
-        role="heading"
         aria-level="3"
+        aria-labelledby={id}
         className={getClassName('bpk-accordion__title')}
       >
         <button
@@ -76,16 +75,15 @@ const BpkAccordionItem = props => {
           </span>
         </button>
       </dt>
-      <AnimateHeight duration={200} height={expanded ? 'auto' : 0}>
-        <dd
-          id={contentId}
-          role="region"
-          aria-labelledby={titleId}
-          className={getClassName('bpk-accordion__content-container')}
-        >
+      <dd
+        id={contentId}
+        aria-labelledby={contentId}
+        className={getClassName('bpk-accordion__content-container')}
+      >
+        <AnimateHeight duration={200} height={expanded ? 'auto' : 0}>
           {children}
-        </dd>
-      </AnimateHeight>
+        </AnimateHeight>
+      </dd>
     </div>
   );
 };
--- a/packages/bpk-component-accordion/src/BpkAccordionItem.js
+++ b/packages/bpk-component-accordion/src/BpkAccordionItem.js
@@ -32,7 +32,16 @@ const ExpandIcon = withButtonAlignment(ChevronDownIcon);
 
 const BpkAccordionItem = props => {
   const iconClassNames = [getClassName('bpk-accordion__item-expand-icon')];
-  const { id, title, children, expanded, onClick, tagName, ...rest } = props;
+  const {
+    id,
+    title,
+    children,
+    expanded,
+    onClick,
+    tagName,
+    textStyle,
+    ...rest
+  } = props;
 
   // if this component is passed initiallyExpanded, this makes sure it doesn't
   // end up on the node. Not ideal as our container component shouldn't be passing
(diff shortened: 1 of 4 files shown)


packages/bpk-component-accordion/src/BpkAccordionItem.js
```diff
@@ -45,14 +45,13 @@ const BpkAccordionItem = props => {
     );
   }
 
-  const titleId = `${id}_title`;
   const contentId = `${id}_content`;
 
   return (
     <div id={id} {...rest}>
       <dt
-        role="heading"
```
metalix2: ARIA role must be appropriate for the element
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role
This Element type cannot support the Role `Heading`

packages/bpk-component-accordion/src/BpkAccordionItem.js
```diff
@@ -45,14 +45,13 @@ const BpkAccordionItem = props => {
...
         aria-level="3"
+        aria-labelledby={id}
```
metalix2: ARIA attributes must conform to valid values
`aria-labelledby` must match a valid element id
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

packages/bpk-component-accordion/src/BpkAccordionItem.js
```diff
@@ -45,14 +45,13 @@ const BpkAccordionItem = props => {
     );
   }
 
-  const titleId = `${id}_title`;
```
metalix2: no longer need.

packages/bpk-component-accordion/src/BpkAccordionItem.js
```diff
@@ -76,16 +75,15 @@ const BpkAccordionItem = props => {
           </span>
         </button>
       </dt>
-      <AnimateHeight duration={200} height={expanded ? 'auto' : 0}>
```
metalix2: <dl> elements must only directly contain properly-ordered <dt> and <dd> groups, <script> or <template> elements

`<AnimateHeight>` which is a `<div>` must not be a child of `<dl>` only `<dt>`, `<dd>` are allowed to be children of `<dl>`

Functionally still works the same.

packages/bpk-component-accordion/src/BpkAccordionItem.js
georgegillams: In my mind, it does seem appropriate for it to be a heading though...  Maybe the `dt` tag should be replaced with `h3`? wdyt?

metalix2: We could set a child element of the `<dt/>` to have the role set on it but as it stands the dt tag roles are limited to Dialog roles only https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role and maybe some others but not heading unfortunately.

packages/bpk-component-accordion/src/BpkAccordionItem.js
```diff
@@ -76,16 +75,15 @@ const BpkAccordionItem = props => {
...
-        <dd
-          id={contentId}
-          role="region"
-          aria-labelledby={titleId}
-          className={getClassName('bpk-accordion__content-container')}
-        >
+      <dd
+        id={contentId}
+        aria-labelledby={contentId}
```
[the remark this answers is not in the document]
metalix2: Well We're already using the `id` for the div want this aria-labelled-by to link to that same one? and remove contentId?

packages/bpk-component-accordion/src/BpkAccordionItem.js
metalix2: Sorry not dialog - definition roles see.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute Example 6

georgegillams:
Just tested on IE and Edge, and all good there too 💯

PR #1607, opened 2019-09-06
[BPK-2653] Investigate shipping CSS

georgegillams:
Remember to include the following changes:
+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests
+ [ ] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)

PR #1687, opened 2019-10-12
[BPK-2653] Investigate shipping CSS (#1607)

georgegillams:
Using tags alone to distinguish which packages ship with CSS is a bad idea. A NPM package tag is a 1-1 mapping from a tag name to a specific version. This means that every time we publish a package with the tag "css" we untag the previous css version. This means that there will be no way to distinguish between versions shipping with css and not.

Instead we should ensure that the version is updated to `X.Y.Z-css.0` and then it is published with a `CSS` tag.

image

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/scripts/publish-process/util.js
+++ b/scripts/publish-process/util.js
@@ -18,6 +18,7 @@
 
 /* eslint-disable no-console */
 
+const path = require('path');
 const fs = require('fs');
 const { exec, execSync } = require('child_process');
 const readline = require('readline');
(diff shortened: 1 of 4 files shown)


scripts/publish-process/util.js
```diff
@@ -429,7 +456,7 @@ const performPublish = async (changes, changeSummary) => {
   if (!testing) {
     disableKrypton();
     execSync(`git add . && git commit -m "${commitMessage}" --no-verify`);
-    await publishPackagesToNPM(changes);
+    await publishPackagesToNPM(changes, null);
```
ojcurtis: so you don't need to pass null here you could modify the function declaration to be `tag = null` for the default case

PR #1691, opened 2019-10-15
[NO-JIRA] Fix typo in textStyle default value

georgegillams:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/bpk-component-accordion/src/__snapshots__/BpkAccordionItem-test.js.snap
```diff
@@ -20,7 +20,7 @@ exports[`BpkAccordionItem should not render an "initiallyExpanded" attribute on
         className="bpk-accordion__flex-container"
       >
         <span
-          className="bpk-text  bpk-accordion__title-text"
+          className="bpk-text bpk-text--base bpk-accordion__title-text"
```
georgegillams: The `bpk-text--base` classname wasn't being applied before as the textStyle was `bsae` instead of `base`. This is now correctly applying the `--base` class.

PR #1694, opened 2019-10-17
[BPK-3098] Enable dynamic color output

georgegillams:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-tokens/gulpfile.babel.js
+++ b/packages/bpk-tokens/gulpfile.babel.js
@@ -27,6 +27,7 @@ import { flatten } from 'lodash';
 import gulpMerge from 'merge2';
 import jsonLint from 'gulp-jsonlint';
 
+import transformDarkValues from './src/transformDarkValues';
 import bpkScss from './formatters/bpk.scss';
 import bpkEs6Js from './formatters/bpk.es6.js';
 import bpkIosJson from './formatters/bpk.ios.json';
@@ -122,6 +123,8 @@ const createTokens = done => {
           format: { type: format },
         }),
       )
+      .on('error', done)
+      .pipe(transformDarkValues())
       .on('error', done)
       .pipe(gulp.dest(path.resolve(__dirname, outputPath)))
       .on('error', done)
--- a/packages/bpk-tokens/src/transformDarkValues.js
+++ b/packages/bpk-tokens/src/transformDarkValues.js
@@ -1,0 +1,154 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2019 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this buffer except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


packages/bpk-tokens/gulpfile.babel.js
```diff
@@ -123,6 +124,7 @@ const createTokens = done => {
         }),
       )
       .on('error', done)
+      .pipe(transformDarkValues())
```
georgegillams: ??????
```suggestion
      .pipe(transformDarkValues()).on("error", done)
```

packages/bpk-tokens/src/transformDarkValues.js
```diff
@@ -0,0 +1,154 @@
...
+import through from 'through2';
+
+const calculatePotentialMerges = properties => {
+  const merges = {};
+
+  if (!properties) {
+    return merges;
+  }
+
+  properties.forEach(p => {
```
k0nserv: Wouldn't a category specific for dyanmic colours be much nice than using the name as we have here?

packages/bpk-tokens/src/transformDarkValues.js
```diff
@@ -0,0 +1,154 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import through from 'through2';
+
+const calculatePotentialMerges = properties => {
```
k0nserv: This can be written with lodash to avoid explicit for loops and building state over time. This is how it's [done](https://github.com/Skyscanner/backpack-ios/blob/master/gulpfile.js#L207-L219) in Backpack iOS

packages/bpk-tokens/gulpfile.babel.js
```diff
@@ -123,6 +124,8 @@ const createTokens = done => {
...
```
k0nserv: It seems like the wrong level of abstraction to do this after the files have been serialized to JSON rather than as part of the process in Theo. The responsbility is now spread in different places instead of being in Theo.

georgegillams: Yeah - we know this isn't ideal but theo doesn't support this. I tried to write a transformer but theo only allows you to handle a single value in a transformer 😢

k0nserv: I would rather have this logic in the app repos, which has similar logic given that limitation in Theo.

packages/bpk-tokens/src/transformDarkValues.js
georgegillams: 👍

packages/bpk-tokens/src/transformDarkValues.js
georgegillams: 👍

packages/bpk-tokens/gulpfile.babel.js
georgegillams: Yeah - at the time it seemed nicer to keep it in one place, but I dunno...

georgegillams: I don't like the idea of having 3 repos doing the same task when it could just be one..

PR #1699, opened 2019-10-22
[No JIRA] Upgrade to Node Erbium LTS

k0nserv:


.nvmrc
```diff
@@ -1 +1 @@
-lts/carbon
```
ojcurtis: Do we still need this file given we are setting the node version in `package.json`?

k0nserv: The file is still here, just changed it to say `lts/erbium` instead

ojcurtis: Yeah - but my question is do we still need it at all in the codebase?

k0nserv: Yes to ensure we all get the right version by doing `nvm use`

PR #1701, opened 2019-10-23
[BPK-3098] Add background color variants for ios

georgegillams:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


packages/bpk-tokens/tokens/base.ios.json
```diff
@@ -35,6 +35,48 @@
       "originalValue": "{!WHITE}",
       "name": "backgroundLightColor"
     },
+    {
+      "type": "color",
+      "category": "colors",
+      "value": "#000000ff",
+      "originalValue": "{!BLACK}",
+      "name": "backgroundPrimaryDarkColor"
```
imchristian: I think that for consistency with Apple we should call the `backgroundPrimaryColor` just `backgroundColor`
Ref: https://developer.apple.com/documentation/uikit/uicolor/ui_element_colors

georgegillams: Ok cool.
What do you think about `textPrimaryColor`?

georgegillams: Also just thinking... What happens in swift if someone uses `Color.backgroundColor`. Would they get Apple's, or ours?

georgegillams: Apple's colours are named differently, so no clash would occur!!
https://developer.apple.com/documentation/uikit/uicolor/ui_element_colors?language=objc

PR #1713, opened 2019-11-04
[BPK-3176] add rn semantic tokens

tiagoengel:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-tokens/formatters/react-native-value-template.js
+++ b/packages/bpk-tokens/formatters/react-native-value-template.js
@@ -1,6 +1,6 @@
 import tinycolor from 'tinycolor2';
 
-export default function(value, type) {
+export default function formatValue(value, type) {
   let formattedValue = value;
 
   switch (type) {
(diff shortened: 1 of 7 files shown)


packages/bpk-tokens/formatters/react-native-value-template.js
```diff
@@ -14,9 +14,18 @@ export default function(value, type) {
...
+    case 'string': {
       formattedValue = `"${value.replace(/"/g, '\\"')}"`;
       break;
+    }
+    case 'semantic': {
+      const parsedValues = Object.keys(value).reduce((parsed, tokenKey) => {
+        const { value: currValue, type: currType } = value[tokenKey];
+        return `${parsed} ${tokenKey}: ${formatValue(currValue, currType)},\n`;
+      }, '');
+      formattedValue = `{\n${parsedValues}};`;
```
georgegillams: The `;` inside the string here is causing an exteraneous `;` to appear in the output files
```suggestion
      formattedValue = `{\n${parsedValues}}`;
```

tiagoengel:
Not adding  `@flow` to the files now because this will require us to address some behaviours in our tokens, like the fact we export some extra keys per platform and that will cause flow to fail. We need to export these as `nullable` and have the same exports for both platforms.
This will likely cause a breaking change in the app though.

PR #1719, opened 2019-11-11
[No Jira] Remove unused 'chalk' dependency

shaundon:
Removes the need for https://github.com/Skyscanner/backpack/pull/1717 as `chalk` isn't actually being used (probably a holdover from `backpack-docs`).


shaundon:
Closes #1717

PR #1723, opened 2019-11-21
Add "dismissible" prop to BpkChip

sogame:
To allow using `BpkChip` in a read-only way.
When `dismissible` is set to `false`, the close button is hidden.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,7 +4,7 @@ Please include a description of the changes you are introducing and some screens
 -->
 
 Remember to include the following changes:
-+ [x] `UNRELEASED.md`
++ [x] `UNRELEASED.yaml`
 + [x] `README.md`
 + [x] Tests
 + [x] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)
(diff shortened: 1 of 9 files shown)


.github/pull_request_template.md
```diff
@@ -4,7 +4,7 @@ Please include a description of the changes you are introducing and some screens
 -->
 
 Remember to include the following changes:
-+ [x] `UNRELEASED.md`
++ [x] `UNRELEASED.yaml`
```
ojcurtis: Great spot!

CONTRIBUTING.md
```diff
@@ -50,14 +50,14 @@ Conventions and squad decisions are kept in the [decisions folder](/decisions).
 
 Backpack is developed using Node, using the following versions:
 
-* `LTS/Carbon` (Node ^8.12.0)
-* `^6.4.1` (npm)
+* `LTS/Erbium` (Node ^12.13.0)
```
ojcurtis: Great spot!

packages/bpk-component-map/stories.js
```diff
@@ -149,7 +149,7 @@ storiesOf('bpk-component-map', module)
   .add('With onTilesLoaded', () => (
     <StoryMap
       center={{ latitude: 55.944357, longitude: -3.1967116 }}
-      onTilesLoaded={() => console.log('Tiles loaded')}
+      onTilesLoaded={() => console.log('Tiles loaded')} // eslint-disable-line no-console
```
sogame: Danger was complaining about this, so better to silent the warning in this case to avoid noise in all pr's 😅 

image

PR #1729, opened 2019-11-25
Adding new carousel

ojcurtis:
Outstanding issues:

- Add carousel indicator
- ~Need to stop full cycle of items~
- ~Add animation functionality~


packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,169 @@
...
+const AlignedArrowLeft = withButtonAlignment(withRtlSupport(ArrowLeftIcon));
+const AlignedArrowRight = withButtonAlignment(withRtlSupport(ArrowRightIcon));
+
+const getClassName = cssModules(STYLES);
+
+export type Props = {
+  className: ?string,
+  wrapperClassName: ?string,
+  itemClassName: ?string,
+  children: ?any,
```
tiagoengel: This should be `Node` (import from react) and doesn't need to be nullable and `Node` can be null or undefined.

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,169 @@
...
+      ...rest
+    } = this.props;
+    const { prevArrowDisabled, nextArrowDisabled } = this.state;
+    const classNames = getClassName('bpk-carousel', className);
+    const wrapperClassNames = getClassName(
+      'bpk-carousel__wrapper',
+      wrapperClassName,
+    );
+    const itemClassNames = getClassName('bpk-carousel__item', itemClassName);
+    const normalisedChildren = Array.isArray(children) ? children : [children];
```
tiagoengel: This should use `React.Children.toArray(children)`

packages/bpk-component-carousel/src/BpkCarousel.scss
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-carousel {
+  display: flex;
+  justify-content: space-evenly;
+
+  &__wrapper {
+    width: 64.9375rem;
```
tiagoengel: nit: would be more readable if we extract those magic numbers to variables

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,169 @@
...
+      position: 0,
+      nextArrowDisabled: false,
+      prevArrowDisabled: true,
+    };
+  }
+
+  getOrder(itemIndex) {
+    const { position } = this.state;
+    const { children } = this.props;
+    const numItems = children.length;
```
tiagoengel: use React.Children.count(children)

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,169 @@
...
+    if (itemIndex - position < 0) {
+      return numItems - Math.abs(itemIndex - position);
+    }
+    return itemIndex - position;
+  }
+
+  nextSlide = () => {
+    const { position } = this.state;
+    const { children } = this.props;
+    const numItems = children.length || 1;
```
tiagoengel: React.Children.count(children)

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,169 @@
...
+    this.doSliding(position === numItems - 1 ? 0 : position + 1);
+  };
+
+  prevSlide = () => {
+    const { position } = this.state;
+    const { children } = this.props;
+    const numItems = children.length || 1;
```
tiagoengel: React.Children.count(children)

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,169 @@
...
+          >
+            <AlignedArrowLeft />
+          </BpkButton>
+        </div>
+        <div className={wrapperClassNames}>
+          <div className={getClassName('bpk-carousel__container')}>
+            {normalisedChildren.map((child, index) => (
+              <div
+                className={itemClassNames}
+                id={index}
```
tiagoengel: If you use `React.Children.toArray(children)` this is most likely not necessary

packages/bpk-component-carousel/src/BpkCarousel.scss
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-carousel {
+  display: flex;
+  justify-content: space-evenly;
```
tiagoengel: this is not supported in IE11

packages/bpk-component-carousel/src/BpkCarousel.scss
ojcurtis: This value is probably going to get removed

packages/bpk-component-carousel/src/BpkCarousel.scss
ojcurtis: Will have a look into an alternative but I think there are other Backpack components with this same issue

packages/bpk-component-carousel/src/BpkCarousel.js
ojcurtis: Which part is not necessary?

packages/bpk-component-carousel/src/BpkCarousel.js
ojcurtis: Yep this is on the list to fix, flow was throwing a panic at the time

packages/bpk-component-carousel/src/BpkCarousel.js
tiagoengel: Sorry I read `key` instead of `id`. Key would not be necessary :)

packages/bpk-component-carousel/src/BpkCarousel.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  &__wrapper {
+    overflow: hidden;
+  }
+
+  &__container {
+    display: flex;
+    justify-content: space-between;
+    flex: 1 0;
+    transition: margin-left $bpk-duration-base;
```
tiagoengel: Have we tested this in IE11? If I remember correctly transition + calc does not work :/.
Also a suggestion, maybe not for this iteration but using transition like this is not very performant, translateX would give a smother result.

ojcurtis: Yes has been verified on IE and Edge, works as expected 😄

ojcurtis: It has been tried in IE and the only problem is that it doesn't do the smooth animation which might not be awful but something to look at for the future

tiagoengel: 👍

packages/bpk-component-carousel/carousel.scss
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index';
+
+/* stylelint-disable unit-blacklist */
+.bpk-carousel-story__item {
+  width: 333px;
```
shaundon: We should use Backpack tokens here e.g. `$bpk-spacing-lg * 10`

packages/bpk-component-carousel/README.md
```diff
@@ -0,0 +1,31 @@
...
+```js
+import React from 'react';
+import BpkCarousel from 'bpk-component-carousel';
+
+export default () => (
+  <BpkCarousel>
+    <BpkCard>Item 1</BpkCard>
+    <BpkCard>Item 2</BpkCard>
+  </BpkCarousel>
+);
```
shaundon: We should add a bit more guidance here to let consumers know that all items should be the same width.

packages/bpk-component-carousel/src/BpkCarousel-test.js
```diff
@@ -0,0 +1,50 @@
...
+      .create(
+        <BpkCarousel className="custom-classname">
+          <div>Item 2</div>
+          <div>Item 3</div>
+          <div>Item 4</div>
+        </BpkCarousel>,
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
```
shaundon: We should add some more tests to cover cases like there only being one item.

packages/bpk-component-carousel/carousel.scss
shaundon: If this file is only used in stories we should follow convention and call it `stories.scss`.

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,195 @@
...
+};
+
+type State = {
+  position: number,
+  prevArrowDisabled: boolean,
+  nextArrowDisabled: boolean,
+};
+
+class BpkCarousel extends Component<Props, State> {
+  wrapperRef: Object;
```
shaundon:
```suggestion
  wrapperRef: ?HTMLDivElement;
```

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,195 @@
...
+
+  firstItemRef: Object;
```
shaundon:
```suggestion
  firstItemRef: ?HTMLElement;
```

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,195 @@
...
+        ? this.firstItemRef.current.offsetWidth
+        : 0;
+    const itemsShown = Math.floor(this.wrapperWidth / this.itemWidth);
+    this.setState({
+      nextArrowDisabled:
+        position === Children.count(children) - itemsShown ||
+        Children.count(children) === itemsShown,
+      prevArrowDisabled: position === 0,
+    });
+  }
```
shaundon: Let's refactor this as it's a bit tricky to read

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,195 @@
...
+
+  // Function to do the animation based on the item size that is moving
+  getTranslate = () => {
+    const { position } = this.state;
+    return `calc((-${this.itemWidth}px) * ${position})`;
+  };
+
+  // Function that will handle moving to the next slide
```
shaundon: Good to have comments but in this case the function names should be descriptive so as to self document.

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,195 @@
...
+    const classNames = getClassName('bpk-carousel', className);
+    const wrapperClassNames = getClassName(
+      'bpk-carousel__wrapper',
+      wrapperClassName,
+    );
+    const itemClassNames = getClassName('bpk-carousel__item', itemClassName);
+
+    return (
+      <div className={classNames} {...rest}>
+        <div>
```
shaundon: Do we need this extra div?

packages/bpk-component-carousel/src/BpkCarousel.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  &__item {
+    margin-right: 1.25rem;
```
shaundon: Can we make this use Backpack tokens?

packages/bpk-component-carousel/src/BpkCarousel.js
ojcurtis: This was flows suggestion as that is what element is created and not HTMLDivElement. Changing to your suggestion causes flow to fail.

packages/bpk-component-carousel/src/BpkCarousel.js
ojcurtis: This was flows suggestion as that is what element is created and not HTMLDivElement. Changing to your suggestion causes flow to fail.

packages/bpk-component-carousel/src/BpkCarousel.js
ojcurtis: Yes otherwise the buttons stretch the full height of the carousel so this keeps it at the correct size

packages/bpk-component-carousel/src/BpkCarousel.scss
ojcurtis: We don't have tokens that would make this value hence the hardcoded here

packages/bpk-component-carousel/src/BpkCarousel-test.js
ojcurtis: Sounds good

packages/bpk-component-carousel/carousel.scss
ojcurtis: We don't have tokens that would make this value hence the hardcoded here so it follows the example designs - plus some Backpack components have values in px hardcoded too as the actual running component so not sure if as an example component this would be ok

PR #1733, opened 2019-11-29
stop redundant datasource call

iamcdonald:
infer list has ended if less than requested elements are returned from data source

+ [] `UNRELEASED.yaml`
+ [] `README.md`
+ [x] Tests
+ [] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)

### DIFF
--- a/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
+++ b/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
@@ -193,48 +193,6 @@ const withInfiniteScroll = <T: ExtendedProps>(
       }).then(newState => this.setStateAfterDsUpdate(newState));
     };
 
-    fetchItemsX(config): Promise<$Shape<State>> {
-      const { onScrollFinished, seeMoreAfter } = this.props;
-      const {
-        index,
-        elementsPerScroll,
-        elementsToRender,
-        computeShowSeeMore,
-      } = extend(
-        {
-          index: this.state.index,
-          elementsPerScroll: this.props.elementsPerScroll,
-          elementsToRender: this.state.elementsToRender,
-          computeShowSeeMore: true,
-        },
-        config,
-      );
-
-      return this.props.dataSource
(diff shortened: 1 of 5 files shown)


packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -193,7 +193,7 @@ const withInfiniteScroll = <T: ExtendedProps>(
       }).then(newState => this.setStateAfterDsUpdate(newState));
     };
 
-    fetchItems(config): Promise<$Shape<State>> {
+    fetchItemsX(config): Promise<$Shape<State>> {
```
tiagoengel: This should be removed

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -170,11 +170,11 @@ const withInfiniteScroll = <T: ExtendedProps>(
       // all visible data is fetched again (from 0 to current index) to update the list with the new data.
       // If after this call `isListFinished` is true, it means the new data source has no items and we need to
       // reset the list, which we do by setting `elementsToRender` to `[]` and `index` to `0`
-      const { isListFinished } = newState;
+      const { elementsToRender, index } = newState;
```
tiagoengel: Any specific reason for this change other than code style? I prefer the previous version as `isListFinished` is easier to read and understand.

iamcdonald: It wouldn't actually work the way it's currently written.
Fetch items now sets `isListFinished` to true when less than the requested elements are returned. The old code would have replaced those elements with an empty array because it saw `isListFinished` is true.

tiagoengel: Makes sense, can we update the inline comment to reflect this then, it is a bit confusing now.

PR #1741, opened 2019-12-03
Rolling relative through Backpack

ojcurtis:
This PR sets up the use of relative in bpk-stylesheets and has a fallback to old fonts

Before
Screenshot 2019-12-03 at 14 25 22

After
Screenshot 2019-12-03 at 14 25 36


packages/bpk-stylesheets/fonts.scss
```diff
@@ -0,0 +1,104 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* stylelint-disable font-family-no-missing-generic-family-keyword */
+@mixin bpk-define-font-stack($baseURL, $fallback-font-stack) {
+  :global(body) {
+    font-family: 'Skyscanner Relative', $fallback-font-stack;
```
k0nserv: This can come from the token we already have for font family i.e. `$bpk-font-family-base` right? Given that we add Relative to that token, which we now can.

ojcurtis: Totally agree!

packages/bpk-stylesheets/fonts.scss
```diff
@@ -0,0 +1,104 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* stylelint-disable font-family-no-missing-generic-family-keyword */
+@mixin bpk-define-font-stack($baseURL, $font-family-stack) {
+  :global(body) {
+    font-family: $font-family-stack;
```
k0nserv: Why not just leavel the existing definition that defined this in place instead of doing it via the mixin? In fact do we even need the mixin anymore?

packages/bpk-stylesheets/index.scss
```diff
@@ -18,6 +18,7 @@
 
 @import '~normalize.css';
 @import '~bpk-mixins/index';
+@import './_fonts.scss'; /* stylelint-disable-line scss/at-import-no-partial-leading-underscore */
```
k0nserv: This can just be `@import 'fonts'` SASS is smart enough to figure this out

ojcurtis: Well that is clever! Learnt something new today

PR #1743, opened 2019-12-05
[BPK-3340] Changed textEmphasizedFontWeight to 700

shaundon:


tiagoengel:
Does it make RN use the correct font variation?

shaundon:
@tiagoengel I think so - RN reads this token so this should filter down

tiagoengel:
> @tiagoengel I think so - RN reads this token so this should filter down

What I meant is, will setting the font weight to 700 result in using `relative-bold`?

shaundon:
Ah sorry. I manually changed the font weight to 700 in the app and it did make it look fine. I'll DM you a link

PR #1744, opened 2019-12-06
[BPK-3327]: Removed border radius them attribute

ojcurtis:
+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests


UNRELEASED.yaml
```diff
@@ -28,3 +28,9 @@ FIXED:
 #   FIXED:
 #     - bpk-component-horcrux:
 #       - Fixed issue where `BpkHorcrux` would occasionally possess the living.
+
+FIXED:
```
tiagoengel: Shouldn't this be a breaking change?

ojcurtis: It follows the same change we made for RN which was marked as not breaking

ojcurtis: Visually it will be breaking if someone has decided not to use the standard squared button

PR #1750, opened 2019-12-12
Adopting new typescale for components

olliecurtis:
The purpose of this PR is to adopt the new typescale and align to the Native platforms.

This PR covers off changing the values only - there will need to be more work in order to remove usages of line-height and letter-spacing as these are baked into the font already


packages/bpk-component-drawer/src/BpkDrawerContent.scss
```diff
@@ -70,7 +70,7 @@
 
   &__header {
     display: flex;
-    min-height: #{2 * $bpk-spacing-sm + $bpk-text-base-line-height}; // keep height the same if visually hidden title
+    min-height: #{2 * $bpk-spacing-sm + $bpk-spacing-base}; // keep height the same if visually hidden title
```
ojcurtis: Prep for when we remove line-height

ojcurtis: Prep for when we remove line-height

packages/bpk-component-text/src/BpkText.js
```diff
@@ -35,7 +45,17 @@ TEXT_STYLES.forEach(textStyle => {
 
 type Props = {
   children: Node,
-  textStyle: 'xs' | 'sm' | 'base' | 'lg' | 'xl' | 'xxl',
+  // eslint-disable-next-line flowtype/space-after-type-colon
```
ojcurtis: Had to add this lint rule here as the list is too long, which causes `prettier` to format it as below but then causes a `lint` error which also tries to format the code which then violates a `prettier` rule - meaning there is a circular error loop of rules conflicting

georgegillams:
This reminds me of the first time I saw a universal remote.
Made me think; “this changes everything”!

packages/bpk-mixins/src/mixins/_typography.scss
```diff
@@ -32,9 +32,12 @@
 ///
 /// @access private
 
-@mixin _bpk-text-factory($font-size, $line-height, $font-weight: false, $letter-spacing: false) {
+@mixin _bpk-text-factory($font-size, $font-weight: false, $line-height: false, $letter-spacing: false) {
```
ojcurtis: Swapped these around as was testing the effects or removing line-heights and letter-spacing.

Also made line-height optional as the new larger sizes don't have line-heights due to the font

packages/bpk-component-checkbox/src/BpkCheckbox.scss
```diff
@@ -68,7 +68,7 @@
 
   &__icon {
     position: absolute;
-    top: ($bpk-spacing-base - $bpk-spacing-sm) / 2;
+    top: $bpk-spacing-xs;
```
k0nserv: I feel like the old value was better because it captured more intent even though it was more complex. From it it was clear that this an attempt to center, now it's just an "arbitrary" value

packages/bpk-component-drawer/src/BpkDrawerContent.scss
k0nserv: Won't this be too tall now? It's correct that `$bpk-text-base-line-height` and `$bpk-spacing-base` are both 1.5rem however the correct line height here is actually `$bpk-text-base-font-size * 1.3` which is `1.3rem`

packages/bpk-component-nudger/src/BpkNudger.scss
```diff
@@ -27,8 +27,8 @@
     display: inline-block;
     padding: $bpk-button-padding-y 0;
     border: none;
-    font-size: $bpk-font-size-lg;
-    line-height: $bpk-button-line-height;
+    font-size: $bpk-font-size-xl;
```
k0nserv: Text mixin?

packages/bpk-component-radio/src/BpkRadio.scss
```diff
@@ -57,7 +57,7 @@ $bpk-radio-circle-size: $bpk-one-pixel-rem * 10;
 
   &__circle {
     position: absolute;
-    top: ($bpk-spacing-base - $bpk-radio-circle-size) / 2;
+    top: $bpk-spacing-xs;
```
k0nserv: I have the same concern as above about intent

packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -132,7 +132,7 @@ BpkRating.propTypes = {
 
 BpkRating.defaultProps = {
   className: null,
-  size: RATING_SIZES.base,
+  size: 'sm',
```
k0nserv: Does this need to change? The Rating Size is not the same as the text size so we can keep the current interface while changing the text styles used?

packages/bpk-component-text/src/BpkText.js
```diff
@@ -26,7 +26,17 @@ import STYLES from './BpkText.scss';
...
+const TEXT_STYLES = [
+  'xs',
+  'sm',
+  'base',
+  'lg',
+  'xl',
+  'xxl',
+  'xxxl',
+  'xxxxl',
+  'xxxxxl',
```
k0nserv: iOS stops at xxxl. Should we and native be aligned?

packages/bpk-mixins/src/mixins/_badges.scss
```diff
@@ -32,14 +32,14 @@
 
 @mixin bpk-badge {
   display: inline-flex;
-  padding: $bpk-badge-padding-y $bpk-badge-padding-x;
+  padding: .175rem $bpk-badge-padding-x;
```
k0nserv: This isn't a token value now, why?

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -554,7 +554,7 @@
 
 @mixin bpk-checkbox__input {
   position: absolute;
-  top: ($bpk-spacing-base - $bpk-spacing-md) / 2;
+  top: .175rem;
```
k0nserv: Intent gone again?

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -660,7 +660,7 @@
 
 @mixin bpk-radio__input {
   position: absolute;
-  top: ($bpk-spacing-base - $bpk-spacing-md) / 2;
+  top: $bpk-spacing-sm / 5;
```
k0nserv: Same as above

packages/bpk-component-checkbox/src/BpkCheckbox.scss
ojcurtis: Using this old value meant that the alignment was off and the box looked all funky this now brings it back to align with the text and the box itself

packages/bpk-component-radio/src/BpkRadio.scss
ojcurtis: Using this old value meant that the alignment was off and the box looked all funky this now brings it back to align with the text and the box itself

packages/bpk-component-text/src/BpkText.js
ojcurtis: So these are new ones that Will has defined that are to be used for web use cases as part of the the web type ramp but Will is also reviewing about having these new ones on Native platforms as well

packages/bpk-mixins/src/mixins/_badges.scss
ojcurtis: It is for the correct spacing that matches the Sketch libraries and we don't have a token value here for this one

packages/bpk-mixins/src/mixins/_forms.scss
ojcurtis: It is for the correct spacing that matches the Sketch libraries and we don't have a token value here for this one

packages/bpk-mixins/src/mixins/_forms.scss
ojcurtis: It is for the correct spacing that matches the Sketch libraries that Will has done as part of the type ramp changes

packages/bpk-component-checkbox/src/BpkCheckbox.scss
ojcurtis: I can add a comment here to explain what this variable is actually doing

packages/bpk-component-rating/src/BpkRating.js
ojcurtis: Yes because of the font sizes needed to change but I wanted to make sure that it didn't break the existing API for users, otherwise it meant adding additional JS logic to this file to accomplish the same result

packages/bpk-component-calendar/src/BpkCalendarGridHeader.scss
```diff
@@ -30,8 +30,7 @@
 
   &__week {
     display: table-row;
-    font-size: $bpk-font-size-sm;
-    line-height: $bpk-line-height-sm;
+    font-size: $bpk-font-size-xs;
```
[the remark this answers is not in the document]
ojcurtis: Sounds reasonable just tried to make slow and steady changes

packages/bpk-component-nudger/src/BpkNudger.scss
ojcurtis: Sounds reasonable

packages/bpk-component-drawer/src/BpkDrawerContent.scss
ojcurtis: This is something that was pre existing and not aware of how that works but changing it from using the old line height to not keeps it the exact same as before all that changed is the naming

packages/bpk-component-calendar/src/BpkCalendarGridHeader.scss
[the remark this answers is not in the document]
k0nserv: Let's keep it as is

packages/bpk-component-drawer/src/BpkDrawerContent.scss
k0nserv: I'm unsure if this is even needed. Quick test shows removing it has no effect regardless of title visibility. Need to be explored for other browsers though. Let's keep it for now

packages/bpk-mixins/src/mixins/_typography.scss
k0nserv: Can we remove the two optional parameters since they are not used.

k0nserv: We should also include in the comms for this that you have to update the base stylesheet to get the correct line-height

packages/bpk-component-checkbox/src/BpkCheckbox.scss
k0nserv: It shouldn't the math works out to the same value

packages/bpk-mixins/src/mixins/_badges.scss
k0nserv: Hmmmmmmmmmmmmmmmmmmmmmmmmmmmmm.

packages/bpk-component-rating/src/BpkRatingSizes.js
```diff
@@ -18,6 +18,6 @@
 
 /* @flow */
 
-const RATING_SIZES = { base: 'base', lg: 'lg', sm: 'sm' };
+const RATING_SIZES = { sm: 'sm', base: 'base', lg: 'lg' };
```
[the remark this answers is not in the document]
olliecurtis: I just swapped the order here as it was bugging me

packages/bpk-component-rating/src/BpkRating.scss
```diff
@@ -42,14 +42,12 @@
     display: flex;
     width: $bpk-rating-lg-width;
     height: $bpk-rating-lg-height;
-    font-size: $bpk-font-size-lg;
```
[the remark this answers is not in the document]
olliecurtis: Here we were setting the font size but now we are just letting BpkText do this for us

packages/bpk-component-rating/src/BpkRatingSizes.js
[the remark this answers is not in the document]
k0nserv: A shit, I didn't read properly :|

PR #1769, opened 2020-01-17
move infinite scroll to use Fragment rather than div as root

iamcdonald:
We've had issues embedding an InfiniteScroll component within a table, this has been at least partially due to the `<div>` root element. This PR moves that use a `<React.Fragment>`.

+ [x] `UNRELEASED.yaml` - added `BpkText-component-infinite-scroll` and description
+ [x] Tests - updated snapshots

### DIFF
--- a/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
+++ b/packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
@@ -20,6 +20,7 @@
 
 import React, {
   Component,
+  Fragment,
   type Element,
   type Config,
   type AbstractComponent,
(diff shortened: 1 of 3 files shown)


packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -290,10 +291,11 @@ const withInfiniteScroll = <T: ExtendedProps>(
       }
 
       return (
-        <div>
+        <Fragment>
+          HIYA
```
olliecurtis: Was this just left in from testing?

iamcdonald: 🤕

shaundon:
Closing this as Iain found a workaround.

PR #1773, opened 2020-01-21
[BPK-3470]: Added iOS elevation tokens

olliecurtis:
+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests


packages/bpk-tokens/src/react-native-output-test.js
```diff
@@ -44,6 +44,15 @@ const ANDROID_ONLY_PROPS = [
 ];
 
 const IOS_ONLY_PROPS = [
+  'BACKGROUND_ELEVATION_01_DARK_COLOR',
```
georgegillams: These aren't IOS_ONLY.
Instead of adding them here, remove them from ANDROID_ONLY above

PR #1775, opened 2020-01-23
Add support for using different weekDay formats

paolo-chiabrera:
Add support for changing the week day name displayed on the calendar header, currently it's set to use only `nameAbbr`, but there're more options available such as `nameShort` and `nameNarrow`.

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests
+ [ ] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)


packages/bpk-component-calendar/src/BpkCalendarGridHeader.js
```diff
@@ -54,24 +60,30 @@ const WeekDay = props => {
 
   return (
     <Element className={classNames.join(' ')} title={weekDay.name}>
-      <span aria-hidden="true">{weekDay.nameAbbr}</span>
+      <span aria-hidden="true">{weekDay[weekDayKey || 'nameAbbr']}</span>
```
tiagoengel: nit: `|| 'nameAbbr'` shouldn't be necessary anymore as this is the default prop

packages/bpk-component-calendar/README.md
```diff
@@ -174,6 +174,7 @@ withCalendarState(composeCalendar(
 | headerProps           | object               | false               | null             |
 | gridProps             | object               | false               | null             |
 | dateProps             | object               | false               | null             |
+| weekDayKey            | string               | false               | nameAbbr         |
```
georgegillams: Could we list the valid options in the #weekDayKey section, and then link to it like this:

```suggestion
| weekDayKey            | see [weekDayKey](#weekDayKey)               | false               | nameAbbr         |
```

packages/bpk-component-calendar/src/BpkCalendarGrid-test.js
```diff
@@ -149,4 +149,22 @@ describe('BpkCalendarGrid', () => {
...
+      .create(
+        <BpkCalendarGrid
+          month={new Date('2016-10')}
+          formatMonth={formatMonth}
+          formatDateFull={formatDateFull}
+          DateComponent={BpkCalendarDate}
+          daysOfWeek={weekDays}
+          weekStartsOn={0}
+          showWeekendSeparator={false}
+          weekDayKey="nameNarrow"
```
georgegillams: Could the WEEK_DAY_KEY be exposed as an enum, so that we can use it like
```suggestion
          weekDayKey={WEEK_DAY_KEY.nameNarrow}
```

packages/bpk-component-calendar/src/BpkCalendarGridHeader.js
paolo-chiabrera: Fixed

PR #1779, opened 2020-01-28
Modify a couple of "import" in readme.md

sogame:
* Add missing import for `withNavigationStackState`
* Use `BpkNavigationBarButtonLink` instead of `BpkButtonLink`

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -111,8 +111,8 @@ if (componentChangedOrCreated) {
 
 // If any of the packages have changed, the UNRELEASED log should have been updated.
 const unreleasedModified = includes(modifiedFiles, 'UNRELEASED.yaml');
-const packagesModified = fileChanges.some(filePath =>
-  filePath.startsWith('packages/'),
+const packagesModified = fileChanges.some(
+  filePath => filePath.startsWith('packages/') && !filePath.endsWith('.md'),
 );
 if (packagesModified && !unreleasedModified && !declaredTrivial) {
   warn(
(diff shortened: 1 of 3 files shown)


dangerfile.js
```diff
@@ -111,8 +111,8 @@ if (componentChangedOrCreated) {
 
 // If any of the packages have changed, the UNRELEASED log should have been updated.
 const unreleasedModified = includes(modifiedFiles, 'UNRELEASED.yaml');
-const packagesModified = fileChanges.some(filePath =>
-  filePath.startsWith('packages/'),
+const packagesModified = fileChanges.some(
+  filePath => filePath.startsWith('packages/') && !filePath.endsWith('.md'),
```
sogame: Happy to undo this change, but I think it makes sense to avoid leaving the message `One or more packages have changed, but UNRELEASED.yaml wasn't updated.` when only `*.md` files have been modified :-)

packages/bpk-component-navigation-stack/README.md
```diff
@@ -34,7 +34,7 @@ that can be used to push and pop views from the stack component.
 
 ```js
 import React from 'react';
-import BpkNavigationStack from 'bpk-component-navigation-stack';
+import BpkNavigationStack, { withNavigationStackState } from 'bpk-component-navigation-stack';
```
sogame: It's quite obvious how it should be imported, but good to make it explicit anyway 😅

packages/bpk-component-navigation-bar/README.md
```diff
@@ -33,7 +32,7 @@ export default () => (
       />
     }
     trailingButton={
-      <BpkButtonLink onClick={() => {}}>Done</BpkButtonLink>
+      <BpkNavigationBarButtonLink onClick={() => {}}>Done</BpkNavigationBarButtonLink>
```
sogame: Maybe with the old branding this wasn't an issue, but given the default colour for `BpkButtonLink` is blue and the background colour of `BpkNavigationBar` is also blue, it's not possible to read the text in this case 😅

PR #1780, opened 2020-01-29
[No JIRA] Remove redundant dependency

k0nserv:
+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10398,7 +10398,7 @@
     },
     "babel-helper-is-nodes-equiv": {
       "version": "0.0.1",
-      "resolved": "https://registry.npmjs.org/babel-helper-is-nodes-equiv/-/babel-helper-is-nodes-equiv-0.0.1.tgz",
+      "resolved": "http://registry.npmjs.org/babel-helper-is-nodes-equiv/-/babel-helper-is-nodes-equiv-0.0.1.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -33092,7 +33023,7 @@
       "dependencies": {
         "got": {
           "version": "6.7.1",
-          "resolved": "https://registry.npmjs.org/got/-/got-6.7.1.tgz",
+          "resolved": "http://registry.npmjs.org/got/-/got-6.7.1.tgz",
```
georgegillams: Could you revert these changes?

k0nserv: They happen when I run `npm install`, which I need to do to remove the dependency

k0nserv: Let me try an alterantive approach

k0nserv: This is just `npm` being `npm` I think.

Here's my env

```
$ npm --version
6.13.4

$ node --version
v12.14.1
```

I ran `npm  uninstall -D static-site-generator-webpack-plugin`

georgegillams: Yeah there's no way to stop NPM being dumb, but we can still be selective about what we check in.
It's not a massive issue, but it makes our git history messier if we always let NPM change stuff back and forth

k0nserv: I'm weary of manually editing `package-lock.json` though

PR #1785, opened 2020-02-03
[BPK-3208] Add flow typing to RN tokens

tiagoengel:
This makes both iOS and Android tokens file the same, the tokens that are not valid for one platform are exported with `undefined` (which is a valid value for all stylesheet props) so flow is happy when people import a platform specific token in a "general" file (no ios or android extension).

It also removes `secondaryColor` from RN tokens which was wrong.

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-tokens/formatters/bpk.react.native.es6.js.js
+++ b/packages/bpk-tokens/formatters/bpk.react.native.es6.js.js
@@ -16,6 +16,9 @@
  * limitations under the License.
  */
 
+import fs from 'fs';
+import path from 'path';
+
 import _ from 'lodash';
 
 import sortTokens from './sort-tokens';
@@ -23,10 +26,24 @@ import adjustTypography from './adjust-typography';
 import { blockComment } from './license-header';
 import valueTemplate from './react-native-value-template';
 
+const iosRawTokens = _.memoize(() =>
+  JSON.parse(
+    fs.readFileSync(path.join(__dirname, '../tokens/base.raw.ios.json')),
+  ),
+);
+
+const androidRawTokens = _.memoize(() =>
+  JSON.parse(
+    fs.readFileSync(path.join(__dirname, '../tokens/base.raw.android.json')),
+  ),
+);
+
 const SEMANTIC_TOKEN_REGEX = /(.*)_(LIGHT|DARK)_(.*)/;
 
 const tokenTemplate = ({ name, value, type }) =>
-  `export const ${_.camelCase(name)} = ${valueTemplate(value, type)};`;
+  `export const ${_.camelCase(name)} = ${
+    value ? valueTemplate(value, type) : value
+  };`;
 
 export const categoryTemplate = (
   categoryName,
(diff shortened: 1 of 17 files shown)


packages/bpk-tokens/formatters/bpk.react.native.es6.js.js
```diff
@@ -16,17 +16,34 @@
...
 import _ from 'lodash';
 
 import sortTokens from './sort-tokens';
 import adjustTypography from './adjust-typography';
 import { blockComment } from './license-header';
 import valueTemplate from './react-native-value-template';
 
+const iosRawTokens = _.memoize(() =>
+  JSON.parse(
+    fs.readFileSync(path.join(__dirname, '../tokens/base.raw.ios.json')),
```
k0nserv: This seems like a bad idea as we are side stepping Theo per their docs we should use the arguments passed to the format function not read the generated raw tokens from the file system. For example if Theo refactors their code to emit these raw files later in their pipepline this would break because we are relying on them being available here. What stops us from just using the values passed to the formatter function?

k0nserv:
All in all it looks really good, just have the one concern about relying on the raw token output

packages/bpk-tokens/formatters/bpk.react.native.es6.js.js
tiagoengel: The raw file is generated by us, I've changed the gulpfile script to generate `raw` files first and made the `platform` tokens dependant on that task.

I tried a lot of other options to find a way to send the `other` platform tokens to this function but I couldn't find any. 

Also the `raw` tokens are generated by us in the gulpfile so I don't think an update in theo will affect this, we guarantee those are generated first in the gulpfile.

k0nserv: Okay, to flip it on its head what's missing from the API provided by Theo here?

tiagoengel: So the way we use it is by calling theo with `src/ios/base.json` or `src/android/base.json` and theo does the magic of importing/merging all files into one JS object and calling the formatter we have registered.

Because android and ios have basically the same tokens I couldn't find a way of making theo parse the two `src/ios/base.json` and `src/android/base.json` and provide them as separate objects to the function. 

What happens in that case is that the files get merged and you end up with all tokens in one js object.

k0nserv: Okay, can we leave a comment to that effect in the code to remember why we did it this way? Also is this something we should raise an issue for Theo for?

tiagoengel: Added the comment.

I'll try to think if this can fit any general use case to add in Theo. My initial feeling is that this is too specific to our usage to be in the lib, but maybe I'm just looking at it the wrong way.

PR #1786, opened 2020-02-04
BpkText readme update

sogame:
* Readme for `BpkText` is missing the new `textStyle` values.
* Changed a bit the code that generates `className` in `BpkText` component, making use of `cssModules` capabilities.

### DIFF
--- a/packages/bpk-component-text/README.md
+++ b/packages/bpk-component-text/README.md
@@ -60,7 +60,7 @@ export default () => (
 | Property  | PropType                                 | Required | Default Value |
 | --------- | ---------------------------------------- | -------- | ------------- |
 | children  | -                                        | true     | -             |
-| textStyle | 'xs', 'sm', 'base', 'lg', 'xl', 'xxl'    | false    | 'base'        |
+| textStyle | 'xs', 'sm', 'base', 'lg', 'xl', 'xxl', 'xxxl', 'xxxxl', 'xxxxxl'    | false    | 'base'        |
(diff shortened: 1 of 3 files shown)


packages/bpk-component-text/src/BpkText.js
```diff
@@ -70,17 +65,15 @@ const BpkText = (props: Props) => {
...
+  const classNames = getClassName(
+    'bpk-text',
+    `bpk-text--${textStyle}`,
+    bold && 'bpk-text--bold',
+    className,
+  );
 
   return (
-    <TagName className={classNames.join(' ')} {...rest}>
+    <TagName className={classNames} {...rest}>
```
sogame: I don't think this requires a changelog entry, but happy to add if if you think it would be better :-)

packages/bpk-component-text/README.md
```diff
@@ -60,7 +60,7 @@ export default () => (
 | Property  | PropType                                 | Required | Default Value |
 | --------- | ---------------------------------------- | -------- | ------------- |
 | children  | -                                        | true     | -             |
-| textStyle | 'xs', 'sm', 'base', 'lg', 'xl', 'xxl'    | false    | 'base'        |
+| textStyle | 'xs', 'sm', 'base', 'lg', 'xl', 'xxl', 'xxxl', 'xxxxl', 'xxxxxl'    | false    | 'base'        |
```
olliecurtis: Good catch!

PR #1795, opened 2020-02-20
ATBT-2578 - Add schema markup for breadcrumbs

mrkalndz:
Adds schema meta data as an optional property for breadcrumbs

More info about structured data can be found here https://developers.google.com/search/docs/guides/intro-structured-data
Breadcrumb schema specifically is here
https://developers.google.com/search/docs/data-types/breadcrumb

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests

To discuss: Do we prefer the implementation here (including dangerouslySetInnerHtml) or adding an additional dependency like: https://github.com/google/react-schemaorg

### DIFF
--- a/packages/bpk-component-breadcrumb/README.md
+++ b/packages/bpk-component-breadcrumb/README.md
@@ -37,6 +37,8 @@ export default class App extends Component {
 | ------------------ | -------- | -------- | ------------- |
 | children           | node     | true     | -             |
 | label              | string   | true     | -             |
+| schemaMetaData     | array    | false    | null          |
+
 
 
 ### BpkBreadcrumbItem
--- a/packages/bpk-component-breadcrumb/README.md
+++ b/packages/bpk-component-breadcrumb/README.md
@@ -30,19 +30,21 @@ export default class App extends Component {
 
 ```
 
-## Schema Markup
+## Structured Data
 
-JSON-LD schema mark up can be added to improve the SEO of the component through the `schemaMetaData` property
+[JSON-LD](https://json-ld.org/) schema mark up can be used to
+add [structured data](https://developers.google.com/search/docs/guides/intro-structured-data)
+to the component to improve the SEO of the component through the `schemaMetaData` property.
 
 ## Props
 
 ### BpkBreadcrumb
 
-| Property           | PropType | Required | Default Value |
-| ------------------ | -------- | -------- | ------------- |
-| children           | node     | true     | -             |
-| label              | string   | true     | -             |
-| schemaMetaData     | array    | false    | null          |
+| Property           | PropType                             | Required | Default Value |
+| ------------------ | ------------------------------------ | -------- | ------------- |
+| children           | node                                 | true     | -             |
+| label              | string                               | true     | -             |
+| schemaMetaData     | array({url: string, label: string})  | false    | null          |
 
 
 
--- a/packages/bpk-component-breadcrumb/README.md
+++ b/packages/bpk-component-breadcrumb/README.md
@@ -29,6 +29,11 @@ export default class App extends Component {
 }
 
 ```
+
+## Schema Markup
+
+JSON-LD schema mark up can be added to improve the SEO of the component through the `schemaMetaData` property
+
 ## Props
 
 ### BpkBreadcrumb
--- a/packages/bpk-component-breadcrumb/src/BpkBreadcrumb.js
+++ b/packages/bpk-component-breadcrumb/src/BpkBreadcrumb.js
@@ -37,6 +37,11 @@ export type Props = {
   label: string,
 };
 
+/*
+  The google structured data reference for the stringified output of
+  the following function is here:
+  https://developers.google.com/search/docs/data-types/breadcrumb
+*/
 const buildMetaData = (schemaMetaData: SchemaMetaDataItem[]): string => {
   const itemListElement = schemaMetaData.map((schemaMetaDataItem, index) => ({
     '@type': 'ListItem',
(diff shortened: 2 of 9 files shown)


packages/bpk-component-modal/package.json
```diff
@@ -16,7 +16,7 @@
     "bpk-component-close-button": "^2.0.35",
     "bpk-component-icon": "^5.0.35",
     "bpk-component-link": "^2.0.34",
-    "bpk-component-navigation-bar": "^2.1.31",
+    "bpk-component-navigation-bar": "^2.1.32",
```
olliecurtis: You won't need to update this as this will be done during the release process 😄

packages/bpk-component-navigation-stack/package.json
```diff
@@ -22,7 +22,7 @@
   "devDependencies": {
     "bpk-component-button": "^3.2.13",
     "bpk-component-icon": "^5.0.35",
-    "bpk-component-navigation-bar": "^2.1.31",
+    "bpk-component-navigation-bar": "^2.1.32",
```
olliecurtis: You won't need to update this as this will be done during the release process 😄

packages/bpk-component-modal/package.json
mrkalndz: Reverted!

packages/bpk-component-navigation-stack/package.json
mrkalndz: Reverted!

mrkalndz:
@olliecurtis Thanks for the quick feedback. I've reverted the package changes, however now CI i failing due to the check dependencies issue. I take it this is ok?

packages/bpk-component-modal/package.json
mrkalndz: And re-reverted

packages/bpk-component-navigation-stack/package.json
mrkalndz: And re-reverted

packages/bpk-component-breadcrumb/README.md
```diff
@@ -29,6 +29,11 @@ export default class App extends Component {
 }
 
 ```
+
+## Schema Markup
+
+JSON-LD schema mark up can be added to improve the SEO of the component through the `schemaMetaData` property
```
shaundon:
```suggestion
JSON-LD schema mark up can be added to improve the SEO of the component through the `schemaMetaData` property.
```
Also, should we make `JSON-LD` a link to a relevant resource for people to know what it is?

packages/bpk-component-breadcrumb/src/BpkBreadcrumb.js
```diff
@@ -26,24 +26,66 @@ import STYLES from './BpkBreadcrumb.scss';
...
+  label: string,
+};
+
 export type Props = {
   children: Node,
+  schemaMetaData: ?(SchemaMetaDataItem[]),
   label: string,
 };
 
+const buildMetaData = (schemaMetaData: SchemaMetaDataItem[]): string => {
```
shaundon: I wonder if it's worth adding a link to the schema in a comment above this function, as I can imagine somebody unfamiliar with it coming along in the future and not knowing why the data needs to be formatted this way.

packages/bpk-component-breadcrumb/README.md
```diff
@@ -37,6 +42,8 @@ export default class App extends Component {
 | ------------------ | -------- | -------- | ------------- |
 | children           | node     | true     | -             |
 | label              | string   | true     | -             |
+| schemaMetaData     | array    | false    | null          |
```
shaundon: Ideally this should align to what's used in the actual prop type, so how about:

`arrayOf({url: string, label: string})`

packages/bpk-component-breadcrumb/src/BpkBreadcrumb.js
mrkalndz: I've added a link to the google structured data reference above

packages/bpk-component-breadcrumb/README.md
mrkalndz: Added the full stop and a link

packages/bpk-component-breadcrumb/README.md
mrkalndz: Agreed 👍 . I've changed

packages/bpk-component-breadcrumb/src/BpkBreadcrumb.js
```diff
@@ -26,24 +26,71 @@ import STYLES from './BpkBreadcrumb.scss';
...
   return (
-    <nav aria-label={label} {...rest}>
-      <ol className={getClassName('bpk-breadcrumb')}>{children}</ol>
-    </nav>
+    <React.Fragment>
+      {Array.isArray(schemaMetaData) && (
+        <script
+          type="application/ld+json"
+          // eslint-disable-next-line react/no-danger
+          dangerouslySetInnerHTML={{ __html: buildMetaData(schemaMetaData) }}
```
olliecurtis: @mrkalndz  one question here: How often does the schema metadata change? Can we avoid re calculating on every render? Just from an efficiency perspective.

If it doesn't change then maybe we can store this so we don't have to re compute every time

olliecurtis:
@shaundon - This was my fault in getting confused and the components were out of date too

packages/bpk-component-breadcrumb/src/BpkBreadcrumb.js
olliecurtis: @mrkalndz one question here: How often does the schema metadata change? Can we avoid re calculating on every render? Just from an efficiency perspective.

If it doesn't change then maybe we can store this so we don't have to re compute every time

shaundon:
@olliecurtis no problemo :)

packages/bpk-component-breadcrumb/src/BpkBreadcrumb.js
mrkalndz: Yeah, these should be computed once per page load as we're telling google about our page structure and that doesn't change at any stage in the lifecycle in the current implementation, so can be computed in the constructor

If at any stage someone needs dynamic breadcrumbs with dynamic markup then they can solve that issue as and when they need to

PR #1801, opened 2020-02-27
[No JIRA] Update some dependencies

shaundon:


packages/bpk-stylesheets/base.css
```diff
@@ -240,8 +240,7 @@ figure {
  * 2. Show the overflow in Edge and IE.
  */
 hr {
-  -webkit-box-sizing: content-box;
-          box-sizing: content-box;
+  box-sizing: content-box;
```
shaundon: Pretty sure this is fine. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing), the vendor-prefixed version of `box-sizing` hasn't been needed since Safari 5.1, which was [released in 2011](https://en.wikipedia.org/wiki/Safari_version_history#Safari_5).

scripts/webpack/postCssPlugins.js
```diff
@@ -22,7 +22,12 @@ const postCssFlexbugsFixes = require('postcss-flexbugs-fixes');
 module.exports = () => [
   postCssFlexbugsFixes,
   autoprefixer({
-    browsers: ['>1%', 'last 4 versions', 'Firefox ESR', 'not ie < 9'],
+    overrideBrowserslist: [
```
shaundon: To remove this warning when building.

Screen Shot 2020-02-27 at 16 51 10

packages/bpk-stylesheets/base.css
olliecurtis: Will this cause browser support issues by this changing?

shaundon: I feel like my comment addresses that already 😅

shaundon: Wait sorry, I see you marked this as resolved. Never mind!

olliecurtis: Yeah I think I clicked submit as you did so our comments went in at the same time

PR #1804, opened 2020-03-02
Bpk 3619 grid button fixes

olliecurtis:
This PR makes spacing changes to the button component as part of the new 8px grid rollout project.

Before
Screenshot 2020-03-02 at 16 02 58

After
Screenshot 2020-03-02 at 16 03 09


packages/bpk-tokens/tokens/base.raw.json
```diff
@@ -893,16 +893,16 @@
     },
     "BUTTON_ICON_BORDER_RADIUS": {
       "category": "buttons",
-      "value": "1.125rem",
+      "value": "50%",
       "type": "size",
-      "originalValue": "{!BORDER_RADIUS_PILL}",
+      "originalValue": "50%",
```
tiagoengel: Just one question about this, shouldn't we be updating and using the same token as before (BORDER_RADIUS_PILL) instead of `50%` directly? 

What happens if people are using `BORDER_RADIUS_PILL` directly in this case?

olliecurtis: As the `BORDER_RADIUS_PILL` token is used by other components it could have unintended sideaffects to those components that need it vs the button. If the other dependent components also need this set to 50% (covered in other tickets) then we can update the token at the route to keep it consistent

tiagoengel: Right, looks go to me then. We should try to revisit this in the future though

olliecurtis: Have made a note to revisit as the project goes along!

PR #1844, opened 2020-03-19
Add "sticky" prop to "BpkNavigationBar"

sogame:
Add a `sticky` prop to `BpkNavigationBar` component to make it easy to keep it at the top when scrolling up.

Before scrolling:
image

After scrolling:
image


packages/bpk-component-navigation-bar/src/BpkNavigationBar.scss
```diff
@@ -63,4 +63,10 @@
       left: $bpk-spacing-sm;
     }
   }
+
+  &__sticky {
+    position: sticky;
+    top: 0;
+    z-index: $bpk-zindex-tooltip - 1; // Allow tooltips/modals/... to be displayed above the navigation bar
```
sogame: a bit hacky, I guess a new token should be created, right?

PR #1859, opened 2020-04-02
Rename "componentWillReceiveProps" to "UNSAFE_componentWillReceiveProps"

sogame:
React 16.3 introduced `UNSAFE_componentWillReceiveProps` as an alias for `componentWillReceiveProps`, as this will be removed in React 17 (see https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html)

Since React 16.9, there’s a warning message when using `componentWillReceiveProps` (see https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#new-deprecations).

Backpack seems to be using React 16.4.2. So there should be no issue replacing `componentWillReceiveProps` with `UNSAFE_componentWillReceiveProps` as the alias was introduced in React 16.3.

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/UNRELEASED.yaml
+++ b/UNRELEASED.yaml
@@ -1,5 +1,15 @@
 # UNRELEASED:
 
+ADDED:
+  - bpk-animate-height:
+  - bpk-component-banner-alert:
+  - bpk-component-calendar:
+  - bpk-component-datatable:
+  - bpk-component-drawer:
+  - bpk-component-image:
+  - bpk-react-utils:
+    - Rename `componentWillReceiveProps` to `UNSAFE_componentWillReceiveProps`.
+
 # How to write a good changelog entry:
 #
 #   1. Add 'BREAKING', 'ADDED' OR 'FIXED' depending on if the change will be major, minor or patch according to [semver.org](semver.org).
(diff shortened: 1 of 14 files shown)


.eslintrc
```diff
@@ -65,6 +65,14 @@
...
         ]
       }
+    ],
+    "camelcase": [
+      "error",
+      {
+        "properties": "never",
+        "ignoreDestructuring": false,
+        "allow": ["^UNSAFE_"]
+      }
```
sogame, line 75: Maybe this could be added to `eslint-config-skyscanner` also?

sogame: I've raised this for `eslint-config-skyscanner`, just in case: https://github.com/Skyscanner/eslint-config-skyscanner/pull/41

.eslintrc
```diff
@@ -65,6 +65,14 @@
           "render"
         ]
       }
+    ],
+    "camelcase": [
```
olliecurtis, line 69: @sogame - will change this to not add it here but through `eslint-config-skyscanner` as done here:  https://github.com/Skyscanner/eslint-config-skyscanner/pull/41

sogame: Done! Updated `eslint-config-skyscanner` instead :-)

sogame:
@olliecurtis I've fixed the linter issues that appeared after updating `eslint-config-skyscanner`

PR #1860, opened 2020-04-02
Add "valid" prop to BpkDatepicker

sogame:
`BpkDatepicker` doesn't have `valid` prop, and this makes it not work correctly with `BpkFieldset`: the label becomes red and error message is displayed, but the input border doesn't change to red nor the exclamation icon is displayed

image

With this changes, the `valid` prop is send to the input field, so it behaves as expected:

image

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests


sogame:
A lot of changes in storrybook for a tiny code change 😅

sogame:
Fixed conflicts in `UNRELEASED.yaml`

PR #1862, opened 2020-04-03
[BPK-3801] Add dark tooltip type

shaundon:
Screen Shot 2020-04-03 at 17 10 23

I'll do docs once this is merged.


packages/bpk-component-tooltip/src/BpkTooltip.scss
```diff
@@ -19,6 +19,7 @@
 @import '~bpk-mixins';
 
 $arrow-size: $bpk-spacing-xs;
+$dark-background-color: $bpk-color-sky-gray;
```
olliecurtis, line 22: Do we need to create an extra variable here? Would it make sense to just use the color directly?

shaundon: I figured it was DRYer this way, if you're using the same value for five different things it should probably be a variable

PR #1868, opened 2020-04-09
[BPK-2118] Add Flow types to some of `bpk-react-utils`

shaundon:
I tried adding Flow to `Portal.js` but it's an absolute BEAST and a lot of what it's doing is strictly against Flow, even though we know it's ok because it works. So many of the errors are interlinked where fixing one opens up ten more in other areas too.

To fix it, we'd either need to totally refactor it or use a lot of undesirable types like `any` to make the errors go away.

In this case I've left it alone and added Flow to the rest of the module, so there's at least _some_ value.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,4 @@
 [ignore]
-<PROJECT_ROOT>/backpack-react-native/.*
 .*/node_modules/jsonlint/.*
 .*/node_modules/config-chain/test/broken.json
 
--- a/packages/bpk-component-code/src/BpkCodeBlock.js
+++ b/packages/bpk-component-code/src/BpkCodeBlock.js
@@ -34,7 +34,7 @@ const BpkCodeBlock = (props: Props) => {
   const { children, alternate, className, ...rest } = props;
   const preClassNames = getClassName(
     'bpk-code__pre',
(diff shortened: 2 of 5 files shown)


.flowconfig
```diff
@@ -1,5 +1,4 @@
 [ignore]
-<PROJECT_ROOT>/backpack-react-native/.*
```
shaundon, line 2: No longer needed as we have the separate repo.

packages/bpk-component-code/src/BpkCodeBlock.js
```diff
@@ -32,21 +32,18 @@ type Props = {
...
-    preClassNames.push(className);
-  }
+  const codeClassNames = getClassName('bpk-code', 'bpk-code--block');
 
   return (
     // $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
-    <pre className={preClassNames.join(' ')} {...rest}>
-      <code className={codeClassNames.join(' ')}>{children}</code>
+    <pre className={preClassNames} {...rest}>
+      <code className={codeClassNames}>{children}</code>
```
shaundon, line 46: I updated this to fix some Flow errors that came up when I added Flow to `bpk-react-utils`.

PR #1870, opened 2020-04-17
[BPK-3765] Fix star rating RTL animation

georgegillams:
+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/UNRELEASED.yaml
+++ b/UNRELEASED.yaml
@@ -1,7 +1,11 @@
 # UNRELEASED:
 FIXED:
+
   - bpk-component-rating:
     - Fixed a couple of minor layout issues in RTL.
+
+  - bpk-component-star-rating:
+    - Fixed a bug that caused a strange animation when selecting a star rating in RTL mode.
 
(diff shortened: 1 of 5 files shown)


UNRELEASED.yaml
```diff
@@ -1,4 +1,7 @@
 # UNRELEASED:
+FIXED:
+  - bpk-component-star-rating:
+    - The RTL interactive star rating had got a bit crazy - must be all this self-isolation. Think we've got it to settle down now.
```
shaundon, line 4:
```suggestion
    - Fixed a bug where the animation when selecting a star rating in RTL mode didn't appear correctly.
```

```
if (rightTime && rightPlace) {
  doAFunnyChangelogEntry()
} else {
  readTheRoomAndKeepItProfesh()
}
```

I know I sound like a killjoy but bear in mind these changelog entries are read by non-native English speakers from across the world, so the best thing we can do for our users is to keep them factual.

PR #1876, opened 2020-04-30
[BPK-3907] Support leading icon in loading button

georgegillams:
image

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests


shaundon:
I know you haven't changed them, but it makes it look like you've changed the background colour of disabled buttons because of your browser extension (I assume). Might not be the best idea as people could misinterpret.

UNRELEASED.yaml
```diff
@@ -1,4 +1,8 @@
 # UNRELEASED:
+ADDED:
+  - bpk-component-loading-button:
+    - Added a `iconPosition` prop which allows consumers to control whether the icon is show on the leading or trailing side of the button.
```
shaundon, line 4:
```suggestion
    - Added a `iconPosition` prop which allows consumers to control whether the icon is shown on the leading or trailing side of the button.
```

PR #1879, opened 2020-05-01
[BPK-3447] Fix bug in setOpacity

shaundon:
Those damn falsy checks and zero!

### DIFF
--- a/packages/bpk-tokens/src/setOpacity.js
+++ b/packages/bpk-tokens/src/setOpacity.js
@@ -19,8 +19,8 @@
 import Color from 'color';
 
 const setOpacity = (colorToken, opacity) => {
-  // Number properties compat is low, so we'll stick to the globals
-  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#Browser_compatibility
+  // IE11 doesn't support `Number.isNaN` so we must use the global.
+  // When IE11 support drops we can migrate.
   // eslint-disable-next-line no-restricted-globals
   if (isNaN(parseFloat(opacity))) {
     throw new Error(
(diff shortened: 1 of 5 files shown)


packages/bpk-tokens/src/setOpacity.js
```diff
@@ -19,7 +19,10 @@
 import Color from 'color';
 
 const setOpacity = (colorToken, opacity) => {
-  if (!parseFloat(opacity)) {
+  // Number properties compat is low, so we'll stick to the globals
```
georgegillams, line 22: I don't really understand what point this is trying to make? Is compatibility with `isNaN` good enough? Or isn't it?

shaundon: `0` is a valid opacity value, but it currently causes an error, because `0 == false`. 

Using `isNaN` is more thorough because it works with `0`.

georgegillams: I can see that, but I still don't understand the comment.

shaundon: Ah, sorry I misunderstood your question.

So when I added `isNaN`, eslint complained. I searched for `isNaN` in our codebase to see what we do in other situations, and every other usage of it has this comment, so it seems like a convention, and I copied it over to here too.

georgegillams: Maybe the comment would be more useful if it said `isNaN is currently unsupported in IE, but that's fine because... `

shaundon: But then I have to update it everywhere FINE I'll do it 😝

shaundon: Done, try now 😇

PR #1882, opened 2020-05-06
[BPK-3915]: Added new Indeterminate checkbox and updated existing checkboxes

olliecurtis:
This PR adds new indeterminate state option to the checkbox and updates existing checkbox designs.

**Indeterminate**
Screenshot 2020-05-05 at 18 48 50

**Checked**
Screenshot 2020-05-05 at 18 48 56

**Disabled**
Screenshot 2020-05-05 at 18 49 13

**Disabled checked**
Screenshot 2020-05-05 at 18 49 19

# Verified Browsers
- [x] IE 11
- [x] Edge
- [x] Safari
- [x] Safari (iOS)
- [x] Firefox
- [x] Chrome
- [x] Chrome (Android)


packages/bpk-component-checkbox/src/BpkCheckbox.js
```diff
@@ -70,12 +69,10 @@ const BpkCheckbox = props => {
         disabled={disabled}
         aria-label={label}
         aria-invalid={isInvalid}
+        // eslint-disable-next-line no-return-assign, no-param-reassign
+        ref={el => el && (el.indeterminate = indeterminate)}
         {...rest}
       />
-      <Tick
```
olliecurtis, line 75: Removed this as it seemed to have no effect on the field when we are using the svgs this was having no effect.

Verified in all browsers IE 11, Edge, Chrome, Firefox and Safari

packages/bpk-component-checkbox/src/BpkCheckbox.scss
```diff
@@ -70,30 +65,6 @@
     }
   }
 
-  &__icon {
```
olliecurtis, line 73: Removed as we no longer use it in the JS

UNRELEASED.yaml
```diff
@@ -11,6 +11,11 @@ FIXED:
   - bpk-component-autosuggest:
     - Upgraded `react-autosuggest` to `10.0.2`.
 
+BREAKING/ADDED:
```
olliecurtis, line 14: Wasn't sure if this was Breaking or just Added due to me removing code. Wdyt?

packages/bpk-component-checkbox/README.md
```diff
@@ -26,15 +26,20 @@ export default () => (
...
+| required      | bool     | false    | false         |
+| disabled      | bool     | false    | false         |
+| white         | bool     | false    | false         |
+| smallLabel    | bool     | false    | false         |
+| valid         | bool     | false    | null          |
+| indeterminate | bool     | false    | false         |
+
+## `indeterminate` prop
+
+This property is used to set the state of the checkbox to `indeterminate` which displays with a minus sign. This is used when there is checkbox group and the parent displays this state when not all children checkboxes are selected.
```
georgegillams, line 42: What happens if the checkbox is `checked` and `indeterminate`?
We should clearly document which takes precidence.

In fact, we should make state an enum so that it's `empty`, `checked` or `indeterminate` but that would be heavily breaking

packages/bpk-component-checkbox/README.md
```diff
@@ -26,15 +26,20 @@ export default () => (
...
+| Property      | PropType | Required | Default Value |
+| ------------- | -------- | -------- | ------------- |
+| name          | string   | true     | -             |
+| label         | node     | true     | -             |
+| required      | bool     | false    | false         |
+| disabled      | bool     | false    | false         |
+| white         | bool     | false    | false         |
+| smallLabel    | bool     | false    | false         |
+| valid         | bool     | false    | null          |
+| indeterminate | bool     | false    | false         |
```
georgegillams, line 38: alphabetical

UNRELEASED.yaml
georgegillams: The removed code is all in non-public files, right? So non-breaking

georgegillams: Unless we consider the visual changes to be breaking?

packages/bpk-component-checkbox/README.md
olliecurtis: They weren't before so wasn't sure - but can update it all

packages/bpk-component-checkbox/README.md
olliecurtis: So it uses psudeo classes so can't make state an enum as its just properties which applies the styles. 

I will check to see what happens with precedence regarding this web standard and how it works, as it will be out of our control

olliecurtis: Resolved by adding more information based on how the standard and functionality works

packages/bpk-component-checkbox/README.md
olliecurtis: Done

UNRELEASED.yaml
olliecurtis: Made it just added

packages/bpk-component-checkbox/src/BpkCheckbox-test.js
```diff
@@ -91,6 +91,15 @@ describe('BpkCheckbox', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('should render correctly with indeterminate attribute', () => {
```
georgegillams, line 94: Can we add a test for checked + indeterminate

packages/bpk-component-checkbox/README.md
```diff
@@ -26,15 +26,20 @@ export default () => (
...
+| disabled      | bool     | false    | false         |
+| indeterminate | bool     | false    | false         |
+| required      | bool     | false    | false         |
+| smallLabel    | bool     | false    | false         |
+| valid         | bool     | false    | null          |
+| white         | bool     | false    | false         |
+
+## `indeterminate` prop
+
+This property is used to set the state of the checkbox to `indeterminate` which displays with a minus sign. If a box is both `indeterminate` and `checked`, then `indeterminate` takes higher precedence until the box is clicked to check it. This is used when there is checkbox group and the parent displays this state when not all children checkboxes are selected.
```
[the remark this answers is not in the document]
tiagoengel: My suggestion is something along the lines of: "the indeterminate flag is only a visual clue, it does not affect the checked state of the checkbox"

packages/bpk-component-checkbox/src/BpkCheckbox.js
```diff
@@ -70,12 +69,10 @@ const BpkCheckbox = props => {
         disabled={disabled}
         aria-label={label}
         aria-invalid={isInvalid}
+        // eslint-disable-next-line no-return-assign, no-param-reassign
+        ref={el => el && (el.indeterminate = indeterminate)}
```
tiagoengel, line 73: Using ref like this doesn't seem right. AFAIK refs are only called during "mount" and not update, so this will fail after the first render.

You should probably set the ref and use it in `componentDidUpdate` to set the indeterminate flag.

olliecurtis: Yeah there was not much online about setting this in react but from the react repo there was an issue opened to add support and both this and the approach mentioned above were acceptable until React adds support.

If the preferred way would be to do `componentDidUpdate` happy to address using that

olliecurtis: I believe either approach will come with side-effects until official support is added in react - according to the React engineers at FB

tiagoengel: Can you link to this issue? The problem here is that if this component is rendered with `indeterminate=false` and then the prop is updated after that, this will not work. So there is actually a bug, using the normal way of setting refs and updating this on `didUpdate` will ensure it works during "mount" and "update"

olliecurtis: `https://github.com/facebook/react/issues/1798`

tiagoengel: you can do something like this 
```
componentDidUpdate(props) {
  this.checkboxRef.indeterminate = props.indeterminate;
}
ref={el => {
   this.checkboxRef = el;
   this.checkboxRef.indeterminate = indeterminate;
}```

olliecurtis: Will we need `componentDidMount` as well?

tiagoengel: The `ref` will cover mount. But I tested this and it seems to work fine during update as well. 
We can keep it like it is, just remove the `eslint-disable` lines and follow the suggested style.
e.g:
```js
ref={e => {
  if (e) {
      e.indeterminate = indeterminate;
   }
}}

olliecurtis: Addressed

PR #1884, opened 2020-05-06
[BPK-3954] Switch component

shaundon:
Kapture 2020-05-06 at 17 31 40

* Checked it in Safari, Chrome, IE11, Edge and even Android 5.1 for the lols, all good.
* Checked it with voiceover, works fine. Uses the same DOM structure as our checkbox component.
* I've deliberately not added things like a 'disabled' prop because right now it's just not needed.
* Thanks to @georgegillams for all the accessibility help when building this!


packages/bpk-component-switch/README.md
```diff
@@ -0,0 +1,37 @@
+# bpk-component-switch
+
+> Backpack example component.
```
shaundon, line 3:
```suggestion
> Backpack switch component.
```

packages/bpk-component-switch/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "bpk-component-switch",
+  "private": true,
```
shaundon, line 3:
```suggestion
```

packages/bpk-component-switch/src/BpkSwitch.scss
```diff
@@ -0,0 +1,86 @@
...
+
+$on-color: $bpk-color-sky-blue;
+$off-color: $bpk-color-sky-gray-tint-06;
+$handle-color: $bpk-color-white;
+
+$height: $bpk-spacing-lg;
+$width: $bpk-spacing-base * 2;
+$border-width: $bpk-one-pixel-rem * 2;
+
+$ease-out-back: cubic-bezier(0.34, 1.56, 0.64, 1);
```
shaundon, line 29:
```suggestion
```

.storybook/config.js
```diff
@@ -111,6 +111,7 @@ configure(() => {
   require('./../packages/bpk-component-slider/stories');
   require('./../packages/bpk-component-spinner/stories');
   require('./../packages/bpk-component-star-rating/stories');
+require('./../packages/bpk-component-switch/stories');
```
olliecurtis, line 114: This needs to be indented

packages/bpk-component-switch/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+## Props
+
+| Property  | PropType | Required | Default Value |
+| --------- | -------- | -------- | ------------- |
+| label     | Node     | true     | -             |
+| className | string   | false    | null          |
+
+This component uses a hidden checkbox, so it supports all the same properties as it (for example `checked`).
```
olliecurtis, line 32: Might be worth linking through to the Checkbox props here for reference/when on the docs site?

packages/bpk-component-switch/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```sh
+npm install bpk-component-switch --save-dev
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkSwitch from 'bpk-component-code';
```
georgegillams, line 15:
```suggestion
import BpkSwitch from 'bpk-component-switch';
```

packages/bpk-component-switch/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "bpk-component-switch",
+  "version": "0.0.51",
```
georgegillams, line 3: Doesn't matter as it will be released 1.0.0, but this is a very random value

packages/bpk-component-switch/src/BpkSwitch.js
```diff
@@ -0,0 +1,60 @@
...
+export type Props = {
+  label: Node,
+  className: ?string,
+};
+const BpkSwitch = (props: Props) => {
+  const { className, label, ...rest } = props;
+  const classNames = getClassName('bpk-switch', className);
+
+  return (
+    <label className={classNames}>
```
georgegillams, line 37: can we just inline this instead of having the `const` above?

packages/bpk-component-switch/src/BpkSwitch.js
```diff
@@ -0,0 +1,60 @@
...
+      <input
+        type="checkbox"
+        className={getClassName('bpk-switch__checkbox')}
```
georgegillams, line 40: Shouldn't this have `aria-label={label}`?

packages/bpk-component-switch/src/BpkSwitch.scss
```diff
@@ -0,0 +1,85 @@
...
+$height: $bpk-spacing-lg;
+$width: $bpk-spacing-base * 2;
+$border-width: $bpk-one-pixel-rem * 2;
+
+
+.bpk-switch {
+  position: relative;
+  display: inline-block;
+
+  // Checkbox is invisible so the switch element can be the visual element.
```
georgegillams, line 34: Nice comment

shaundon:
Thanks for the reviews, all good comments, will address in the morning!

packages/bpk-component-switch/README.md
shaundon: It's not a `BpkCheckbox`, just a regular checkbox. But I'll make that more explicit and link to MDN.

packages/bpk-component-switch/package.json
shaundon: Yeah haha, so this is because Lerna bumps the version of the boilerplate component every time there's a release. It's never been worth fixing just because it's not actually causing a problem.

packages/bpk-component-switch/src/BpkSwitch.js
shaundon: Sure - it was just copypasta'd from the boilerplate process.

packages/bpk-component-switch/src/BpkSwitch.js
shaundon: Yes! Great catch 😅

packages/bpk-component-switch/src/BpkSwitch.scss
shaundon: Why thank you 🙇

shaundon:
All feedback addressed!

PR #1887, opened 2020-05-11
[BPK-2589]: Resolve npm audit issues

olliecurtis:
This PR resolves nom audit issues bringing the total down to 5 vulnerabilities which aren't fixable due to the dependencies that contain do not have newer version that can be upgraded to with the fixes.

In this PR you will also notice all the svgs have been modified - it seems that the gulp tasks have minified/linted the svgs as part of the gulp upgrades (although nothing is documented regarding it).

Testing the icons post this does not seem to have interfered or broken them as they are still displayed correctly so bear that in mind the changes just seem to have been minifying values together for improvements


package.json
```diff
@@ -110,29 +110,29 @@
     "eslint_d": "^8.0.0",
     "extract-text-webpack-plugin": "^3.0.2",
     "flow-bin": "^0.122.0",
-    "flow-typed": "^2.5.1",
+    "flow-typed": "^3.1.0",
```
tiagoengel, line 113: Just to be sure, are we ok with this change? From the changelog it seems the only braking change is that they've dropped support for node < 10.

olliecurtis: @tiagoengel - Yeah I checked using `npm audit` and there was vulnerabilities in pre 3 and when upgrading they went away. Looks like the changelog didn't cover deep bumps

olliecurtis:
@tiagoengel very interesting spot with the Icon fonts going missing as they are appearing locally so no sure why it says they have been deleted

package.json
tiagoengel: My main concern is regarding to us breaking projects using node < 10, if that is not a concern than it's fine by me.

olliecurtis: This will be fine as all projects internally are on node 12 as the LTS, including Backpack so should be good here as this is a devDependency to Backpack only anyway so wouldn't have effect on anyone externally

shaundon:
Should we close this if the ticket has been abandoned?

olliecurtis:
Closing as this was abandoned

PR #1891, opened 2020-05-25
[BPK-3982]: Added event banner alert type

olliecurtis:
Screenshot 2020-05-25 at 17 30 57


shaundon:
Looks good!

 CI is failing because the .css file has changed. I can see you've committed that in this PR, maybe there's a change missing from it?

olliecurtis:
Thanks for raising @shaundon - I have tried running the build locally and nothing is changing 🤔 so will do some digging to find out what is happening

PR #1893, opened 2020-06-08
[BPK-4011]: Added back and forward native icons

olliecurtis:
Added these icons as they were stored separately in the React Native repo, the native android ones are in our svgs so makes sense for the iOS ones too


olliecurtis:
Closing due to ongoing discussion

PR #1895, opened 2020-06-16
[BPK-4041] New icons

georgegillams:
This PR causes the font to be generated with 2 versions of each glyph: `glyph--name` and `glyph--name-sm`. Going forward, the dedicated `sm` variant should be used whenever rendering an icon at the smaller size. 

Once we have the new icons for both large and small:
1. Delete all the svgs in `./packages/bpk-svgs/src/icons/`
2. Add the new large svgs to `./packages/bpk-svgs/src/icons/lg`
3. Add the new small svgs to `./packages/bpk-svgs/src/icons/sm`
4. rebuild

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests


CONTRIBUTING.md
```diff
@@ -115,7 +114,7 @@ Run `npm start` to start the storybook server, then go to [http://localhost:9001
 
 If you want to add icons, they should first be approved by [Chris](https://skyscanner.slack.com/messages/D6BU4R0V8) as he "owns" Skyscanner icons.
```
olliecurtis, line 115: I know you haven't changed this line, but maybe we should remove this in the PR or change it to say message Backpack?

packages/bpk-svgs/README.md
```diff
@@ -29,7 +31,8 @@ This package exposes Backpack SVGs in various formats:
   bpk-svgs/dist/scss/_spinners.scss
   ```
 
-- An icon font
+- Icon fonts
   ```
-  bpk-svgs/dist/font/BpkIcon.ttf
+  bpk-svgs/dist/font/sm/BpkIcon.ttf
```
tiagoengel, line 36: I think this will break the app build process, If I'm not mistaken there is a phase where they copy the font files the RN assets folder

georgegillams: Yeah that's exactly why it's a breaking change. When we upgrade the dependency we'll need to change the location it uses

tiagoengel:
Just playing devil's advocate, but wouldn't it be better if we keep one font file and generate new icons with a suffix, e.g. `hotels_sm`. This will avoid the breaking change and we can handle all internally in the BpkIcon component.

georgegillams:
> Just playing devil's advocate, but wouldn't it be better if we keep one font file and generate new icons with a suffix, e.g. `hotels_sm`. This will avoid the breaking change and we can handle all internally in the BpkIcon component.
> 
> I think it also reduces the complexity of having to deal with loading the correct font file for the icon size. We not going to gain any from separating the files as we are still going to be shipping both files in the app.

That's a pretty neat idea 👍
Are you thinking the large icons would remain un-suffixed so it won't be a breaking change?

CONTRIBUTING.md
georgegillams: Yeah lemme do that as a follow-up

tiagoengel:
> > Just playing devil's advocate, but wouldn't it be better if we keep one font file and generate new icons with a suffix, e.g. `hotels_sm`. This will avoid the breaking change and we can handle all internally in the BpkIcon component.
> > I think it also reduces the complexity of having to deal with loading the correct font file for the icon size. We not going to gain any from separating the files as we are still going to be shipping both files in the app.
> 
> That's a pretty neat idea 👍
> Are you thinking the large icons would remain un-suffixed so it won't be a breaking change?

Yeah, we can keep the current name as large icons, and just add the new ones with the sm suffix.

packages/bpk-svgs/tasks/codepoints.json
```diff
@@ -2,8 +2,6 @@
   "accessibility": 59905,
   "accessibility-sm": 60138,
   "account": 59907,
-  "account female": 60140,
-  "account female-sm": 60139,
   "account name": 60142,
```
olliecurtis, line 5: As with the above removal of `account female` can you do the same here?

PR #1903, opened 2020-06-29
[No Jira] Reinstate icons.js file, needed for docs site icon downloads

shaundon:
When the last PR was merged, it broke the docs site. Turns out, that blank `icons.js` file in `bpk-svgs` actually has a purpose! It's used in the docs site as something to latch onto for downloading a zip file of all the icons. So, I reinstated it.

I tested this locally on a copy of the docs site and it works fine.

### DIFF
--- a/packages/bpk-svgs/src/icons/icons-test.js
+++ b/packages/bpk-svgs/src/icons/icons-test.js
@@ -1,0 +1,14 @@
+import fs from 'fs';
+import path from 'path';
+
+// In order for the docs site to serve a zip-file containing all icon SVGs, the `zip-it-loader` loader
+// must be provided with an empty file within the icons directory.
+// If icons.js is not empty, then it will be served instead of the SVGs
+describe('icons.js', () => {
+  it('icons.js file should be empty', () => {
+    const iconsFile = path.join(__dirname, 'icons.js');
+    const iconsJsFileContents = fs.readFileSync(iconsFile);
+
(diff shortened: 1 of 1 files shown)


georgegillams:
btw danger will fail due to the file not having a license header. We can ignore that check and merge anyway

shaundon:
> btw danger will fail due to the file not having a license header. We can ignore that check and merge anyway

Good point!

packages/bpk-svgs/src/icons/icons.test.js
```diff
@@ -0,0 +1,14 @@
+import fs from 'fs';
+
+// In order for the docs site to serve a zip-file containing all icon SVGs, the `zip-it-loader` loader
+// must be provided with an empty file within the icons directory.
+// If icons.js is not empty, then it will be served instead of the SVGs
+describe('icons.js', () => {
+  it('icons.js file should be empty', () => {
+    const iconsJsFileContents = fs.readFileSync(
+      'packages/bpk-svgs/src/icons/icons.js',
```
georgegillams, line 9: shouldn't this path just be `./icons.js` if the test file is in the same directory? 🤔

shaundon: Yeah it should. I just blindly copied your code because I'm stupid 🤦 Will update.

georgegillams:
Changed a couple of things:
 - Fixed the file path
 - Renamed the file to end `-test.js` otherwise jest won't actually run the test

PR #1904, opened 2020-06-29
[BPK-4061]: Icon support to accordion

olliecurtis:
Screenshot 2020-06-29 at 19 50 05

with RTL

Screenshot 2020-06-30 at 11 56 49


packages/bpk-component-accordion/src/BpkAccordionItem.scss
```diff
@@ -74,6 +74,10 @@
     display: inline-block;
   }
 
+  &__leading-icon {
+    margin-right: $bpk-spacing-xs;
```
shaundon, line 78: This may need to switch to `margin-left` when RTL is enabled. We have a mixin that can do it automatically: https://backpack.github.io/sassdoc/#mixin-bpk-margin-trailing

olliecurtis: Ah yes! Good spot!

packages/bpk-component-accordion/README.md
```diff
@@ -77,6 +98,7 @@ export default () => (
 | id        | string   | true     | -             |
 | title     | string   | true     | -             |
 | expanded  | bool     | false    | false         |
+| icon      | node     | false    | null          |
```
georgegillams, line 101: I wonder if this should be called something else, because you could use it to display anything (eg a badge).
In the past I think we've used "auxilliary view" 🤔

olliecurtis: Yeah - this was because we didn't want to open it up so much to consumers for abuse as it means people could put a button in there wish is undesirable

shaundon: I agree with Ollie, let's use the API name to implicitly limit it to icons.

packages/bpk-component-accordion/src/BpkAccordionItem.scss
```diff
@@ -74,6 +74,15 @@
...
+
+    @include bpk-rtl {
+      margin-right: 0;
+      margin-left: $bpk-spacing-xs;
+    }
+  }
+
```
shaundon, line 85: Why not just use the mixin?

```suggestion
  &__leading-icon {
    @include bpk-margin-trailing($bpk-spacing-xs, true);
  }

```

olliecurtis: Oh I didn't realise there was this mixin over the other one I had used

shaundon: No worries! We added it about a year ago because it was a pattern we were doing a lot

olliecurtis: Solved!

PR #1913, opened 2020-07-06
[BPK-4077] Add type prop to BpkHorizontalNav

shaundon:
Screen Shot 2020-07-06 at 15 51 01


packages/bpk-component-horizontal-nav/README.md
```diff
@@ -70,6 +70,7 @@ export default class App extends Component {
 | leadingScrollIndicatorClassName   | string   | false    | null          |
 | showUnderline                     | bool     | false    | true          |
 | trailingScrollIndicatorClassName  | string   | false    | null          |
+| type                              | oneOf('default', 'light')| false | 'default' |
```
georgegillams, line 73:
```suggestion
| type                              | oneOf('default', 'light') | false | 'default' |
```

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -145,29 +153,40 @@ class BpkHorizontalNav extends Component<Props> {
...
+      type,
       ...rest
     } = this.props;
 
     const classNames = getClassName(
       'bpk-horizontal-nav',
-      showUnderline && 'bpk-horizontal-nav--show-underline',
+      showUnderline &&
+        type === HORIZONTAL_NAV_TYPES.default &&
+        'bpk-horizontal-nav--show-default-underline',
```
georgegillams, line 164: It might be neater to do something like 
```
const underlineClassNames = {
  default: 'bpk-horizontal-nav--show-default-underline',
  light: 'bpk-horizontal-nav--show-light-underline',
}
...
showUnderline && underlineClassNames[type]
```

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -145,29 +153,40 @@ class BpkHorizontalNav extends Component<Props> {
...
+        if (type === HORIZONTAL_NAV_TYPES.light) {
+          childProps.type = HORIZONTAL_NAV_TYPES.light;
         }
 
-        return React.cloneElement(child, {
-          ref: ref => {
-            this.selectedItemRef = ref;
-          },
-        });
+        return React.cloneElement(child, childProps);
```
georgegillams, line 189: Will this blow-up if one of the children is null? I think there was a reason that we were checking that `child` was truthy

packages/bpk-component-horizontal-nav/stories.scss
```diff
@@ -38,4 +38,10 @@ $custom-color: $bpk-color-sagano;
     height: 10rem;
     border-right: 0.1rem lightgray solid;
   }
+
+  &__light-appearance-wrapper {
```
georgegillams, line 42: This is fine for the story, but when you do the docs, please use the `dark` example property instead of using a custom style
```
  {
    id: 'light',
    title: 'Light',
    blurb: [
      <Paragraph>The light badge is a gray color as shown below.</Paragraph>,
    ],
    dark: true,
    examples: [
      <BpkBadge type={BADGE_TYPES.light} className={badgeClassName}>
        Apples
      </BpkBadge>,
      ...
```

shaundon: Will do

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: Ah yeah let me check that

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: Good idea, I thought of an even shorter way:

```
      'bpk-horizontal-nav',
      showUnderline && `bpk-horizontal-nav--show-${type}-underline`,
      className,
    );
```

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.js
shaundon: I _think_ it should be fine. This is a `children.map` loop so it shouldn't iterate over null values. I'll pop a check in anyway just for safety.

PR #1915, opened 2020-07-07
[BPK-4068]: New radio style

olliecurtis:
Splitting BPK-4068 into two PRs: Radios and Checkbox.

This PR utilises the new icon style for Radio buttons using pure html and css. This allows for the component to still fully support theming and remove reliance on SVGs that can change size and style when new versions introduced so takes the headache of maintaining out.

The updated has been reviewed and ok'd by design.

Screenshot 2020-07-07 at 20 58 52

RTL:

Screenshot 2020-07-07 at 20 59 02


packages/bpk-component-radio/src/BpkRadio.js
```diff
@@ -41,18 +40,15 @@ const BpkRadio = props => {
...
-  if (className) {
-    classNames.push(className);
-  }
+  const classNames = [
+    getClassName(
+      'bpk-radio',
+      white && 'bpk-radio--white',
+      disabled && 'bpk-radio--disabled',
+      isInvalid && 'bpk-radio--invalid',
+      className && className,
```
shaundon, line 49: You can just use `className` here, `className && className` is redundant, because if it's `null`, it evaluates to `null && null`.

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -666,7 +670,8 @@
 
 @mixin bpk-radio--invalid {
   input[type='radio'] {
-    background: url(map-get($bpk-elements, radio-invalid)) no-repeat;
+    border: ($bpk-one-pixel-rem * 3) solid $bpk-color-panjin;
```
shaundon, line 673: Is it possible to use the [border size tokens](https://backpack.github.io/tokens/borders) for this? Just to simplify things.

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -679,14 +684,14 @@
 
 @mixin bpk-radio__input {
   position: absolute;
-  top: ($bpk-spacing-base - $bpk-spacing-md) / 2;
+  top: 0;
   left: 0;
-  width: $bpk-spacing-md;
-  height: $bpk-spacing-md;
+  width: $bpk-spacing-base - ($bpk-one-pixel-rem * 4);
+  height: $bpk-spacing-base - ($bpk-one-pixel-rem * 4);
```
shaundon, line 690: As this is repeated for `width` and `height`, perhaps we should declare a variable for this (not a token, just a variable local to this mixin)

packages/bpk-mixins/src/mixins/_forms.scss
shaundon: It might be that we don't wanna use border size because it's not _exactly_ a border, it's more intrinsic to the component's style. Either way, `($bpk-one-pixel-rem * 3) solid <color>` is repeated in this file a couple of times, so we might wanna declare a mixin or variable for this so it's DRYer

shaundon:
This is outside of your PR but just mentioning so it's on your radar — when we update the docs for this, can we add an example of the `white` prop? I just went on the docs site to see what it does and we don't have an example of it.

packages/bpk-component-radio/src/BpkRadio.js
olliecurtis: Ah ok! Copied it from another place in the code and didn't realise!

packages/bpk-mixins/src/mixins/_forms.scss
olliecurtis: Sounds good

packages/bpk-mixins/src/mixins/_forms.scss
olliecurtis: Sure thing, forgot we actually have a 3px border token!

shaundon:
This solution doesn't work in IE11 btw, it falls back to what I assume is the default IE radio styling.

I think this is probably an acceptable tradeoff, given how many travellers use IE (around 5%). And the radios are still functional, nothing is broken (theming and invalid states don't work though, of course), so it's progressive enhancement. What do you think?

Screen Shot 2020-07-08 at 08 45 11

UNRELEASED.yaml
```diff
@@ -1,5 +1,9 @@
 # UNRELEASED:
 
+ADDED:
```
georgegillams, line 3: This is just changing the visual appearance of an existing thing, right?
Maybe it should be `fixed` instead of `added`?

packages/bpk-component-radio/src/BpkRadio.js
```diff
@@ -41,18 +40,15 @@ const BpkRadio = props => {
...
-  if (isInvalid) {
-    classNames.push(getClassName('bpk-radio--invalid'));
-  }
-  if (className) {
-    classNames.push(className);
-  }
+  const classNames = [
+    getClassName(
+      'bpk-radio',
+      white && 'bpk-radio--white',
```
georgegillams, line 46: You could also simplify this a bit with `bpk-radio--${style}`. Probably unnecessary though 🤷

UNRELEASED.yaml
olliecurtis: Good point, I guess if we are removing the old assets from svgs it could be breaking right of someone is directly using them, outside of their recommended usage?

shaundon: I think it's breaking for `bpk-svgs`, patch for `bpk-mixins` and `bpk-component-radio`.

UNRELEASED.yaml
```diff
@@ -1,5 +1,14 @@
 # UNRELEASED:
 
+BREAKING:
+  - bpk-svgs:
+    - Removed radio svgs elements as these are no longer used.
```
shaundon, line 5: If I read this as a consumer of Backpack I wouldn't really know what to make of this information — can you provide some more details on what exactly has been removed and what it means? There was never a React element for these so I think they were only usable from scss.

Something along the lines of "this SVGs that were being used in the radio component have been removed because the radio component no longer uses them. Therefore this change is only breaking if you were using the following Sass variables (or however they were used). Not that exact thing but you get the idea 😅

UNRELEASED.yaml
```diff
@@ -1,5 +1,14 @@
...
+
+ADDED:
+  - bpk-mixins:
+  - bpk-component-radio:
+    - Added new radio styles.
```
shaundon, line 10: We can make this punchier 😄 

```suggestion
    - `BpkRadio` has a refreshed, modern appearance.
```

olliecurtis: Used that one!

UNRELEASED.yaml
olliecurtis: What do you think of the change? Used mostly your sentence as it made sense

PR #1917, opened 2020-07-09
No jira/Add flow annotation to nudger

federico-costa:
I added the annotation `/* @flow strict */` to the index of BpkNudger in order to be able to use the types and avoid adding libdefs.

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/packages/bpk-component-nudger/src/themeAttributes.js
+++ b/packages/bpk-component-nudger/src/themeAttributes.js
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+/* @flow strict */
+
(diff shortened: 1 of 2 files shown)


packages/bpk-component-nudger/src/themeAttributes.js
```diff
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+/* @flow strict */
```
olliecurtis, line 19: We don't normally flow type these themeAttributes file, is there a reason for doing so?

olliecurtis: Actually ignore me!

PR #1918, opened 2020-07-09
Shaun's edits

shaundon:
Will comment on specific parts inline.


packages/bpk-component-checkbox/src/BpkCheckbox.scss
```diff
@@ -103,36 +103,31 @@
...
   &__checkmark {
-    &::before {    
+
+    &::before, &::after {
       position: absolute;
+      border-radius: $bpk-border-size-lg;
+      transform: rotate(45deg);
+      content: '';
+      background-color: $bpk-color-white;
+    }
```
shaundon, line 113: Using combination of before and after to abstract things common to both pseudoclasses.

packages/bpk-component-checkbox/src/BpkCheckbox.scss
```diff
@@ -103,36 +103,31 @@
...
       height: 11px;
       width: 3px;
-      background-color: $bpk-color-white;
-      content: '';
       left: 7px;
       top: 2px;
-      border-radius: $bpk-border-size-lg;
-      transform: rotate(45deg);
-      -ms-transform: rotate(45deg); /* IE 9 */
-      -webkit-transform: rotate(45deg); /* Chrome, Safari, Opera */
```
shaundon, line 131: Vendor-prefixed transforms not needed for the browsers we support.

packages/bpk-component-checkbox/stories.js
```diff
@@ -104,12 +57,7 @@ class StatefulCheckbox extends Component<
```
shaundon, line 57: Simplified the stateful example to use minimal lines of code and not worry about indeterminate.

I could make it even shorter with `useState`, but our React version doesn't support that.

PR #1919, opened 2020-07-09
Bpk 4068 new checkbox design

olliecurtis:
This PR is the second PR for the work with the new radios and checkboxes

Screenshot 2020-07-09 at 15 56 29

RTL
Screenshot 2020-07-09 at 15 57 34


olliecurtis:
Just need to delete the SVGs

PR #1923, opened 2020-07-13
[no jira]: Changing css publish process to change name instead of version

olliecurtis:
Upgrading the process to rather than change the versioning number and strategy, we publish css packages under new names.

### DIFF
--- a/scripts/publish-process/publish-css-tagged-packages.js
+++ b/scripts/publish-process/publish-css-tagged-packages.js
@@ -29,7 +29,7 @@
 
 const {
   getCurrentPackageMeta,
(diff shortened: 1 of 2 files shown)


olliecurtis:
> LGTM - how have you tested it?
> I guess you could just pull the new scripts into master and try running it from there?

I ran it locally and then commenting out the bit where it publishes to check the package files get tested. I guess the final test would be to put it in master and then publish for reals

olliecurtis:
Issue with travis reporting back, CI has passed here: https://travis-ci.org/github/Skyscanner/backpack/builds/707661242

Merging

PR #1924, opened 2020-07-14
[BPK-4085]: BpkText API improvements

olliecurtis:
Updated the API to support passing a `weight` prop to set the font style bold or black on the text.

Also deprecated the `bold` prop in favour of the new `weight` prop - this prop will throw a warning in the console in order to get users to migrate.

Updated textStyles to support an enum instead of just strings to help consumers understand the options available


packages/bpk-component-text/stories.js
```diff
@@ -90,4 +90,22 @@ storiesOf('bpk-component-text', module)
...
+        The man jumped over the shark tank. That was very{' '}
+        <BpkText weight={WEIGHT_STYLES.bold}>bold</BpkText> indeed.
+      </BpkText>
+      <BpkText tagName="p">
+        The man jumped over the shark tank. That was a{' '}
+        <BpkText weight={WEIGHT_STYLES.black} textStyle="xl">
+          heavy
+        </BpkText>{' '}
+        shark.
+      </BpkText>
```
shaundon, line 109: 😂

PR #1925, opened 2020-07-15
[no jira]: Fixing css release process to update backpack imports

olliecurtis:
This PR fixes the issue where when we release the CSS versions of the packages, the imports in the js files for Backpack components were still refering to the old package names which are not available when you install

e.g. `bpk-component-button` was being imported when it should be `bpk-component-button-css` as that is the installed version.

## Testing
To test the PR I commented out the part where it actually does the publish to npm and just ran the publish:css command manually and checked the git changes that were made to verify that the imports were correct

### DIFF
--- a/scripts/publish-process/transform-bpk-imports.js
+++ b/scripts/publish-process/transform-bpk-imports.js
@@ -1,0 +1,68 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2020 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


scripts/publish-process/transform-bpk-imports.js
```diff
@@ -0,0 +1,68 @@
...
+const updateImports = (file, dirs) =>
+  new Promise((resolve, reject) => {
+    fs.readFile(file, 'utf8', (err, data) => {
+      if (err) {
+        reject(err);
+      }
+      let result = data;
+
+      dirs.forEach(pkg => {
+        const split = result.split(`from '${pkg}`);
```
olliecurtis, line 33: We split like this to ensure that we pickup only the imports, as just spliting on the packages updated other things like story names and other references that are not intended

scripts/publish-process/transform-bpk-imports.js
```diff
@@ -0,0 +1,68 @@
...
+        if (err2) return reject(err2);
+        return resolve();
+      });
+    });
+  });
+
+// eslint-disable-next-line no-console
+console.log('Crunching Backpack import paths...');
+
+const dirs = execSync('ls packages | grep -v "react-version-test.js"')
```
olliecurtis, line 50: So that we don't need to manually maintain a list of available packages in a JSON somewhere and its always up to date, this small 'hack' lists all the package names from the `packages` folder based on their dir name in order to help identify package names that we need to replace in the js files.

shaundon: That's pretty slick

PR #1926, opened 2020-07-16
[no jira]: Update bold prop usage

olliecurtis:
PR to update components that were still using deprecated `bold` prop for BpkText

Also minor update to the `deprecated` messaging as it was pointed out by @sogame danger would not see and report the warnings so add 'Warning' so that danger can parse and report back to PR 😄  see here https://github.com/sogame/danger-plugin-toolbox/blob/master/src/rules/common/fileWarnings.js#L21

### DIFF
--- a/packages/bpk-react-utils/src/deprecated.js
+++ b/packages/bpk-react-utils/src/deprecated.js
@@ -26,7 +26,7 @@ const deprecated = (propType: PropType, alternativeSuggestion: string) => (
   ...rest: [any]
 ) => {
   if (props[propName] != null) {
-    const message = `"${propName}" property of "${componentName}" has been deprecated. ${alternativeSuggestion}`;
+    const message = `Warning: "${propName}" property of "${componentName}" has been deprecated. ${alternativeSuggestion}`;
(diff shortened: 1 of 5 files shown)


packages/bpk-react-utils/src/deprecated.js
```diff
@@ -26,7 +26,7 @@ const deprecated = (propType: PropType, alternativeSuggestion: string) => (
   ...rest: [any]
 ) => {
   if (props[propName] != null) {
-    const message = `"${propName}" property of "${componentName}" has been deprecated. ${alternativeSuggestion}`;
+    const message = `Warning: "${propName}" property of "${componentName}" has been deprecated. ${alternativeSuggestion}`;
```
olliecurtis, line 29: So that Danger can recognise and report this

sogame: 🙇

olliecurtis:
https://github.com/Skyscanner/backpack/pull/1926#issuecomment-659303888

So turns out when jest runs the tests these warnings are still thrown regardless of usage or not 🤔 need to double check why that happens

PR #1937, opened 2020-07-29
onClose default prop update for BpkDialog and BpkModal

dominicfraser:
# Change
Updates `onClose` to default to `() => null` rather than `null`. Removes a custom prop type in an instance it is no longer required.

# Reasoning
Our Portal has a [default prop of `() => null`](https://github.com/Skyscanner/backpack/blob/master/packages/bpk-react-utils/src/Portal.js#L293).

If a non-function `onClose` is passed in then an error of `this.props.onClose is not a function` is seen.

A Portal can be created from either `BpkDialog` or `BpkDialog`. Here their `onClose` prop is passed to the `Portal`.

Both can optionally have `onClose` provided. If it is not then the default prop will be used. 

The default prop for both is current `null`.

This is fine at the top level, but this will then be passed to the `Portal`. The default prop of the wrapping component overrides the default of the `Portal`. This results in the error as above.

If we default to `() => null` then this should not happen.

This is not needed in `BpkModalDialog` as props given to this component never end up at the `Portal`.

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests


packages/bpk-component-modal/src/BpkModal.js
```diff
@@ -123,7 +123,7 @@ export const defaultProps = {
   closeOnEscPressed: true,
 };
 
-BpkModal.propTypes = { ...propTypes };
-BpkModal.defaultProps = { ...defaultProps };
+BpkModal.propTypes = { ...propTypes, onClose: PropTypes.func };
```
olliecurtis, line 126: Rather than adding here you can probably move this to as looking at this it would be defining the onClose twice:

https://github.com/Skyscanner/backpack/pull/1937/files#diff-9b7c4907d7e12003dbda71f207de17f5R106

packages/bpk-component-modal/src/BpkModal.js
```diff
@@ -123,7 +123,7 @@ export const defaultProps = {
...
+BpkModal.defaultProps = { ...defaultProps, onClose: () => null };
```
olliecurtis, line 127: Rather than adding here you can probably move this to as looking at this it would be defining the onClose twice:

https://github.com/Skyscanner/backpack/pull/1937/files#diff-9b7c4907d7e12003dbda71f207de17f5R115

packages/bpk-component-modal/src/BpkModal.js
dominicfraser: I initially had - but chose to instead overwrite it here.

The spread propTypes and defaultProps come from BpkModalDialog, where they are defined as:
```
onClose: onClosePropType,
```
and
```
onClose: null,
```
https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-modal/src/BpkModalDialog.js#L140

This is correct afaik for `BpkModalDialog` - it correctly _can_ take null or a function, i.e the Flow type is:
```
onClose: ?(event: SyntheticEvent<>) => void,
```

Because of this I don't want to change anything for `BpkModalDialog`, but as its props are all otherwise exactly the same continue to import and spread them here as before, and override the one that is now different for `BpkModal`.

Does that make sense? Separately, is that correct? 😂

PR #1938, opened 2020-07-29
[BPK-4073] Re-add native icons

georgegillams:
Following a discussion with @k0nserv and Dima, we decided that the native icons that existed in `lg` size before can now be replaced with standard icons from our new set, as they are very similar now.

For ease of adoption, I have simply aliased them for now, so that the changes can be adopted without any more things breaking

+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests


UNRELEASED.yaml
```diff
@@ -1,5 +1,10 @@
 # UNRELEASED:
 
+ADDED:
```
olliecurtis, line 3: Will be Fixed right as this brings back accidently deleted icons?

PR #1939, opened 2020-07-31
[BPK-4116] Rewrote the chip component completely

shaundon:
New chips!

## Done
* Implementation
* Tests
* Unreleased
* Readme
* Fix Flow issues
* Write the migration guide
* Add hover and active states

## Advice for reviewers
* It might be best to think of this as a brand-new component as opposed to an update to an existing one, in terms of looking at the diff.
* The chips in the storybook example are all stateful, so you can click them to your heart's content.

### Accessibility

Have I done a11y correctly?
* I made a required `accessibilityLabel` prop and I'm setting it as the `title` attribute on the button. Should this be optional and instead default to `children`, or is it better to be required?
* I'm using the [`checkbox` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role) which seemed to be the right way to go about it.
* Dismissible chips override `role="checkbox` to `role="button` because they aren't selectable.


.spelling
```diff
@@ -31,6 +31,7 @@ BpkCarouselItem
 
 BpkChip
```
olliecurtis, line 32: We can probably remove this now that `BpkChip` no longer exists?

Unless spelling complains due to the migration guide

shaundon: Yeah probably best to just leave it

PR #1940, opened 2020-07-31
WIP - Excluding bpk-mixins from css process

olliecurtis:
This fix excludes bpk-mixins from being changed to CSS as they are focused on SASS they shouldn't be published as a css package as they don't contain any .css files

### DIFF
--- a/scripts/publish-process/transform-bpk-imports.js
+++ b/scripts/publish-process/transform-bpk-imports.js
@@ -47,7 +47,9 @@ const updateImports = (file, dirs) =>
 // eslint-disable-next-line no-console
 console.log('Crunching Backpack import paths...');
 
-const dirs = execSync('ls packages | grep -v "react-version-test.js"')
+const dirs = execSync(
+  'ls packages | grep -v "react-version-test.js" | grep -v "bpk-mixins"',
+)
   .toString()
   .split('\n')
   .filter(s => s !== '');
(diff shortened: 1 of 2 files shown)


scripts/publish-process/transform-bpk-imports.js
```diff
@@ -47,7 +47,9 @@ const updateImports = (file, dirs) =>
 // eslint-disable-next-line no-console
 console.log('Crunching Backpack import paths...');
 
-const dirs = execSync('ls packages | grep -v "react-version-test.js"')
+const dirs = execSync(
+  'ls packages | grep -v "react-version-test.js" | grep -v "bpk-mixins"',
```
georgegillams, line 51: I wonder if we should have a file called `ignoredPackages` to act as a source of truth for packages to skip. Probably not worth it if it's only one though...

olliecurtis: I guess we could keep it in mind should we need to make it bigger and do more than one?

shaundon:
@olliecurtis what's going on with this? If it needs some work and it's not ready for a review yet, can you mark it as WIP?

shaundon:
Sounds good, thanks!

olliecurtis:
Closing this as these will be addressed when we change delivery of Backpack web

PR #1942, opened 2020-08-03
[BPK-4104]: Increase flow

olliecurtis:
+ [x] `UNRELEASED.yaml`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/UNRELEASED.yaml
+++ b/UNRELEASED.yaml
@@ -1,5 +1,15 @@
 # UNRELEASED:
 
+ADDED:
+  - bpk-component-autosuggest:
+  - bpk-component-card:
+  - bpk-component-checkbox:
+  - bpk-component-heading:
+  - bpk-component-link:
+  - bpk-component-loading-button:
+  - bpk-component-spinner:
+    - Added Flow.
+
 # How to write a good changelog entry:
 #
 #   1. Add 'BREAKING', 'ADDED' OR 'FIXED' depending on if the change will be major, minor or patch according to [semver.org](semver.org).
--- a/packages/bpk-component-autosuggest/src/BpkAutosuggest.js
+++ b/packages/bpk-component-autosuggest/src/BpkAutosuggest.js
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+/* @flow strict */
+
 import React from 'react';
 import BpkInput from 'bpk-component-input';
 import Autosuggest from 'react-autosuggest';
@@ -49,10 +51,16 @@ Autosuggest.defaultProps.theme = {
   sectionTitle: getClassName('bpk-autosuggest__section-title'),
 };
 
-Autosuggest.defaultProps.renderInputComponent = inputProps => {
+type Props = {
+  ref: (?HTMLInputElement) => mixed,
+  inputRef: (?HTMLInputElement) => mixed,
+  autoComplete: string,
+};
+Autosuggest.defaultProps.renderInputComponent = (inputProps: Props) => {
   const { ref, inputRef, autoComplete = 'off', ...rest } = inputProps;
 
   return (
+    // $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
     <BpkInput
       inputRef={element => {
         ref(element);
--- a/packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
+++ b/packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
@@ -16,15 +16,26 @@
  * limitations under the License.
  */
 
+/* @flow strict */
+
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { type Node } from 'react';
 import { cssModules } from 'bpk-react-utils';
 
 import STYLES from './BpkAutosuggest.scss';
 
 const getClassName = cssModules(STYLES);
 
-const BpkSuggestion = props => {
+type Props = {
+  value: Node,
+  indent: boolean,
+  className: ?string,
+  icon: Function,
+  subHeading: ?Node,
+  tertiaryLabel: ?string,
+};
+
+const BpkSuggestion = (props: Props) => {
   const classNames = [getClassName('bpk-autosuggest__suggestion')];
   const {
     indent,
@@ -45,6 +56,7 @@ const BpkSuggestion = props => {
   }
 
   return (
+    // $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
     <section className={classNames.join(' ')} {...rest}>
       {icon ? (
         <Icon className={getClassName('bpk-autosuggest__suggestion-icon')} />
--- a/packages/bpk-component-card/src/BpkCard.js
+++ b/packages/bpk-component-card/src/BpkCard.js
@@ -16,18 +16,28 @@
  * limitations under the License.
  */
 
+/* @flow strict */
+
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { type Node } from 'react';
 import { cssModules } from 'bpk-react-utils';
 
 import STYLES from './BpkCard.scss';
 
 const getClassName = cssModules(STYLES);
 
-const BpkCard = props => {
-  const classNames = [getClassName('bpk-card')];
+type Props = {
+  children: Node,
+  className: ?string,
+  href: ?string,
+  padded: boolean,
+  blank: boolean,
+};
+
+const BpkCard = (props: Props) => {
   const { children, className, href, padded, blank, ...rest } = props;
 
+  const classNames = [getClassName('bpk-card')];
   if (padded) {
     classNames.push(getClassName('bpk-card--padded'));
   }
@@ -45,6 +55,7 @@ const BpkCard = props => {
     }
 
     return (
+      // $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
       <a href={href} className={classNameFinal} {...blankProps} {...rest}>
         {children}
       </a>
@@ -52,6 +63,7 @@ const BpkCard = props => {
   }
 
   return (
+    // $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
     <div role="button" className={classNameFinal} {...rest}>
       {children}
     </div>
--- a/packages/bpk-component-fieldset/src/BpkFieldset.js
+++ b/packages/bpk-component-fieldset/src/BpkFieldset.js
@@ -60,6 +60,7 @@ const BpkFieldset = (props: Props) => {
     return null;
   }
 
+  // $FlowFixMe
   let childId: string = children.props.id;
   if (
     // $FlowFixMe Flow is being dumb here and doesn't let us do this check, even though it's perfectly safe!
--- a/packages/bpk-component-heading/src/BpkHeading.js
+++ b/packages/bpk-component-heading/src/BpkHeading.js
@@ -16,28 +16,40 @@
  * limitations under the License.
  */
 
+/* @flow strict */
+
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { type Node } from 'react';
 import { cssModules } from 'bpk-react-utils';
 
 import STYLES from './BpkHeading.scss';
 
 const getClassName = cssModules(STYLES);
 
-const BpkHeading = props => {
-  const classNames = [getClassName(`bpk-heading-${props.level}`)];
+type Props = {
+  children: Node,
+  level: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6',
+  className: ?string,
+  id: ?string,
+  bottomMargin: boolean,
+};
 
-  if (props.className) {
-    classNames.push(props.className);
+const BpkHeading = (props: Props) => {
+  const { children, level, className, id, bottomMargin } = props;
+  const classNames = [getClassName(`bpk-heading-${level}`)];
+  const Level = level;
+
+  if (className) {
+    classNames.push(className);
   }
-  if (!props.bottomMargin) {
+  if (!bottomMargin) {
     classNames.push(getClassName('bpk-heading--no-bottom-margin'));
   }
 
   return (
-    <props.level className={classNames.join(' ')} id={props.id}>
-      {props.children}
-    </props.level>
+    <Level className={classNames.join(' ')} id={id}>
+      {children}
+    </Level>
   );
 };
 
--- a/packages/bpk-component-loading-button/src/BpkLoadingButton.js
+++ b/packages/bpk-component-loading-button/src/BpkLoadingButton.js
@@ -16,8 +16,10 @@
  * limitations under the License.
  */
 
+/* @flow strict */
+
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { type Node, type Element } from 'react';
 import BpkButton from 'bpk-component-button';
 import { BpkSpinner, BpkLargeSpinner } from 'bpk-component-spinner';
 import {
@@ -45,17 +47,19 @@ const getPropsIcon = props => {
   return icon;
 };
 
-const getSpinner = large =>
+const getSpinner = (large: boolean) =>
   large ? <BpkLargeSpinner alignToButton /> : <BpkSpinner alignToButton />;
 
-const getEnabledIcon = large => {
+const getEnabledIcon = (large: boolean) => {
   const AlignedIcon = large
     ? withLargeButtonAlignment(withRtlSupport(ArrowIconLg))
     : withButtonAlignment(withRtlSupport(ArrowIconSm));
   return <AlignedIcon />;
 };
 
-const getDefaultIcon = props => {
+type IconProps = { loading: boolean, large: boolean };
+
+const getDefaultIcon = (props: IconProps) => {
   const { loading, large } = props;
 
   if (loading) {
@@ -64,7 +68,23 @@ const getDefaultIcon = props => {
   return getEnabledIcon(large);
 };
 
-const BpkLoadingButton = props => {
+type LoadingProps = {
+  children: Node,
+  className: ?string,
+  disabled: boolean,
+  secondary: boolean,
+  destructive: boolean,
+  large: boolean,
+  link: boolean,
+  loading: boolean,
+  iconOnly: boolean,
+  icon: ?Element<any>,
+  iconPosition: string,
+  iconDisabled: ?Element<any>,
+  iconLoading: ?Element<any>,
+};
+
+const BpkLoadingButton = (props: LoadingProps) => {
   const {
     children,
     disabled,
@@ -88,6 +108,7 @@ const BpkLoadingButton = props => {
       : [children, spacer, buttonIcon];
 
   return (
+    // $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
     <BpkButton iconOnly={iconOnly} disabled={showBtnDisabled} {...rest}>
       {child0}
       {child1}
@@ -119,6 +140,7 @@ BpkLoadingButton.defaultProps = {
   disabled: false,
   secondary: false,
   destructive: false,
+  large: false,
   link: false,
   loading: false,
   iconOnly: false,
--- a/packages/bpk-component-loading-button/stories.js
+++ b/packages/bpk-component-loading-button/stories.js
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+/* @flow strict */
+
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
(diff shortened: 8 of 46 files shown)


packages/bpk-component-card/src/BpkCard.js
```diff
@@ -24,10 +26,18 @@ import STYLES from './BpkCard.scss';
 
 const getClassName = cssModules(STYLES);
 
-const BpkCard = props => {
-  const classNames = [getClassName('bpk-card')];
+type Props = {
+  children: any,
```
olliecurtis, line 30: Had to set this to `any` which is not ideal and needs to be looked at more as setting this to the correct type `Node` as done in other components throws the following error:

```
Cannot create BpkCard element because JSX text [1] is incompatible with Node type [2] in property children.
[incompatible-type]
```

This is the same for other `children` props that are set to any since adding flow

packages/bpk-component-fieldset/src/BpkFieldset.js
```diff
@@ -60,6 +60,7 @@ const BpkFieldset = (props: Props) => {
     return null;
   }
 
+  // $FlowFixMe
```
olliecurtis, line 63: Like the comments below flow was throwing this against `BpkCheckbox` which is not even referenced here

packages/bpk-component-heading/src/BpkHeading.js
```diff
@@ -24,19 +26,28 @@ import STYLES from './BpkHeading.scss';
...
   }
-  if (!props.bottomMargin) {
+  if (!bottomMargin) {
     classNames.push(getClassName('bpk-heading--no-bottom-margin'));
   }
 
   return (
-    <props.level className={classNames.join(' ')} id={props.id}>
-      {props.children}
+    <props.level className={classNames.join(' ')} id={id}>
```
olliecurtis, line 49: Kept this as `props.level` as changing it to `level` an error in the console is thrown that `<level>` is not a valid html component and therefore doesn't create the component correctly

packages/bpk-component-loading-button/src/BpkLoadingButton.js
```diff
@@ -119,6 +140,7 @@ BpkLoadingButton.defaultProps = {
   disabled: false,
   secondary: false,
   destructive: false,
+  large: false,
```
olliecurtis, line 143: Added this prop instead of spreading with `...rest` as its used here: 

https://github.com/Skyscanner/backpack/pull/1942/files#diff-ac38633aeb0e164740ff850d58f35e9dR53

and flow complains that its not a valid prop as its not available in the main props of the component when its used for those functions

packages/bpk-component-loading-button/stories.js
```diff
@@ -293,78 +295,102 @@ storiesOf('bpk-component-loading-button', module)
   ))
   .add('Icon-only button', () => (
     <div>
-      <BpkLoadingButton iconOnly onClick={action('iconOnly button clicked')} />
+      <BpkLoadingButton iconOnly onClick={action('iconOnly button clicked')}>
+        <span className="visually-hidden">Search</span>
```
olliecurtis, line 299: Added these as `BpkLoadingButton` requires `children` prop in every instance so excluding it causes flow to report its invalid

packages/bpk-component-modal/src/BpkModalDialog.js
```diff
@@ -40,7 +40,7 @@ export type Props = {
   fullScreen: boolean,
   padded: boolean,
   dialogRef: (ref: ?HTMLElement) => void,
-  onClose: ?(event: SyntheticEvent<>) => void,
+  onClose: (event: SyntheticEvent<>) => void,
```
olliecurtis, line 43: Changed this due to the following flow error with mixing between components

Screenshot 2020-08-04 at 14 50 36

Changed the default to be a function to ensure that null is not passed and caused `BpkButtonLink` not to be used incorrectly

packages/bpk-component-spinner/src/BpkExtraLargeSpinner.js
```diff
@@ -26,18 +28,26 @@ import SPINNER_TYPES from './spinnerTypes';
...
+};
+
+const BpkExtraLargeSpinner = (props: Props) => {
   const { type, className, ...rest } = props;
   const classNames = [
-    'bpk-spinner',
-    'bpk-spinner--extra-large',
-    `bpk-spinner--${type}`,
-  ].map(getClassName);
+    getClassName(
```
olliecurtis, line 39: Flow didn't like using `.map(getClassName)` as the `getClassName` function returns mixed types depending on what is passed to it, so flow cannot validate it correctly.

However this is better way we get classNames across our components so cleaning it up to align with others

packages/bpk-component-spinner/src/BpkLargeSpinner.js
```diff
@@ -26,14 +28,18 @@ import SPINNER_TYPES from './spinnerTypes';
...
+
+const BpkLargeSpinner = (props: Props) => {
   const { type, className, alignToButton, ...rest } = props;
 
   const classNames = [
-    'bpk-spinner',
-    'bpk-spinner--large',
-    `bpk-spinner--${type}`,
-  ].map(getClassName);
+    getClassName('bpk-spinner', 'bpk-spinner--large', `bpk-spinner--${type}`),
```
olliecurtis, line 41: Flow didn't like using `.map(getClassName)` as the `getClassName` function returns mixed types depending on what is passed to it, so flow cannot validate it correctly.

However this is better way we get classNames across our components so cleaning it up to align with others

packages/bpk-component-spinner/src/BpkSpinner.js
```diff
@@ -26,18 +28,24 @@ import SPINNER_TYPES from './spinnerTypes';
...
+  type: string,
+  className: ?string,
+  alignToButton: boolean,
+};
+
+const BpkSpinner = (props: Props) => {
   const { type, className, alignToButton, ...rest } = props;
 
-  const classNames = ['bpk-spinner', `bpk-spinner--${type}`].map(getClassName);
+  const classNames = [getClassName('bpk-spinner', `bpk-spinner--${type}`)];
```
olliecurtis, line 40: Flow didn't like using `.map(getClassName)` as the `getClassName` function returns mixed types depending on what is passed to it, so flow cannot validate it correctly.

However this is better way we get classNames across our components so cleaning it up to align with others

packages/bpk-component-card/src/BpkCard.js
olliecurtis: Turns out I was not importing `type Node` from React so was using some built in `Node` type so fixed this

packages/bpk-component-autosuggest/src/BpkAutosuggest.js
```diff
@@ -49,10 +51,16 @@ Autosuggest.defaultProps.theme = {
   sectionTitle: getClassName('bpk-autosuggest__section-title'),
 };
 
-Autosuggest.defaultProps.renderInputComponent = inputProps => {
+type Props = {
```
olliecurtis, line 54: This one I had to take a rough guess based on other uses of ref (and the component which these props are used) in Backpack as I couldn't see the props in `Autosuggest` lib or in `BpkInput`

packages/bpk-component-autosuggest/src/BpkAutosuggest.js
```diff
@@ -49,10 +51,16 @@ Autosuggest.defaultProps.theme = {
...
+  ref: (?HTMLInputElement) => mixed | ?Function,
+  inputRef: (?HTMLInputElement) => mixed | ?Function,
```
shaundon, line 56: Out of interest, do we need the `?Function` part or will just `mixed` work? I figured `mixed` would include `Function`

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
```diff
@@ -16,15 +16,26 @@
...
 import STYLES from './BpkAutosuggest.scss';
 
 const getClassName = cssModules(STYLES);
 
-const BpkSuggestion = props => {
+type Props = {
+  value: Node,
+  indent: boolean,
+  className: ?string,
+  icon: Function,
```
shaundon, line 33: I'm wondering if this should be a `Node` if it's a React element that can be passed in. `Function` is technically correct, as React components are technically functions, but it feels like it could be too broad.

packages/bpk-component-heading/src/BpkHeading.js
```diff
@@ -16,27 +16,38 @@
...
 
 import STYLES from './BpkHeading.scss';
 
 const getClassName = cssModules(STYLES);
 
-const BpkHeading = props => {
-  const classNames = [getClassName(`bpk-heading-${props.level}`)];
+type Props = {
+  children: Node,
+  level: string,
```
shaundon, line 31: Can we make this more strict to reflect the prop type?

i.e. `'h1' | 'h2' ... etc`

packages/bpk-component-heading/src/BpkHeading.js
shaundon: I think this is fine, but for future reference there's a trick/hack you can use. If you make a copy of the variable that's capitalised, it shouldn't complain:

```
const Level = level
```

packages/bpk-component-loading-button/src/BpkLoadingButton.js
```diff
@@ -64,7 +68,23 @@ const getDefaultIcon = props => {
...
+  children: Node,
+  className: ?string,
+  disabled: boolean,
+  secondary: boolean,
+  destructive: boolean,
+  large: boolean,
+  link: boolean,
+  loading: boolean,
+  iconOnly: boolean,
+  icon: ?any,
```
shaundon, line 81: We try not to use `any`, especially in code we ship to users. How come it's not something more specific here? (Same question with `iconDisabled` and `iconLoading` props too)

packages/bpk-component-spinner/src/BpkExtraLargeSpinner.js
```diff
@@ -26,18 +28,26 @@ import SPINNER_TYPES from './spinnerTypes';
 
 const getClassName = cssModules(STYLES);
 
-const BpkExtraLargeSpinner = props => {
+type Props = {
+  type: string,
```
shaundon, line 32: This is too broad, we should use `$Keys<typeof SPINNER_TYPES>`. As a rule of thumb, Flow types should match what we put in propTypes

packages/bpk-component-spinner/src/BpkExtraLargeSpinner.js
shaundon: This could be simplified a bit more. When we call `getClassName` in this way, it returns a string with the separate classes already separated by spaces. So it feels weird to then wrap the result of that in an array, then do another `.join(' ')` later.

This code will have the same effect in a simpler way:

```jsx
  'bpk-spinner',
  'bpk-spinner--extra-large',
  `bpk-spinner--${type}`,
  className
);
```

You can then remove the `if classname` and the `.join(' ')` parts too.

packages/bpk-component-spinner/src/BpkLargeSpinner.js
```diff
@@ -26,14 +28,18 @@ import SPINNER_TYPES from './spinnerTypes';
 
 const getClassName = cssModules(STYLES);
 
-const BpkLargeSpinner = props => {
+type Props = {
+  type: string,
```
shaundon, line 32: Same comment as earlier about using `$Keys<typeof SPINNER_TYPES>`.

packages/bpk-component-spinner/src/BpkLargeSpinner.js
shaundon: Same comment as before about simplifying this.

packages/bpk-component-spinner/src/BpkSpinner.js
```diff
@@ -26,18 +28,24 @@ import SPINNER_TYPES from './spinnerTypes';
 
 const getClassName = cssModules(STYLES);
 
-const BpkSpinner = props => {
+type Props = {
+  type: string,
```
shaundon, line 32: Same comment as earlier about using `$Keys<typeof SPINNER_TYPES>`.

packages/bpk-component-spinner/src/BpkSpinner.js
shaundon: Same comment as before about simplifying this.

packages/bpk-component-loading-button/src/BpkLoadingButton.js
olliecurtis: Ah good spot I just did this temp whilst doing work but forgot about it

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
olliecurtis: Will have a look at this, took this from another example in Backpack codebase

packages/bpk-component-autosuggest/src/BpkAutosuggest.js
olliecurtis: I think because in the useage below we check for a function and I thought that was right over a HtmlInputElement

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
olliecurtis: I chose this as the propTypes at the bottom says its type is `func` so matched the flow type to it, should it still be changed?

packages/bpk-component-heading/src/BpkHeading.js
olliecurtis: Did the above suggestion and seems to be ok

packages/bpk-component-spinner/src/BpkExtraLargeSpinner.js
olliecurtis: Did this suggestion and seems to work fine

packages/bpk-component-spinner/src/BpkLargeSpinner.js
olliecurtis: Did this suggestion and seems to work fine

packages/bpk-component-spinner/src/BpkSpinner.js
olliecurtis: Did this suggestion and seems to work fine

packages/bpk-component-autosuggest/src/BpkAutosuggest.js
olliecurtis: Changed and seems to be fine

packages/bpk-component-heading/src/BpkHeading.js
olliecurtis: Done

packages/bpk-component-loading-button/src/BpkLoadingButton.js
olliecurtis: Done the correct changes

packages/bpk-component-spinner/src/BpkExtraLargeSpinner.js
olliecurtis: Did this suggestion and seems to work fine

packages/bpk-component-spinner/src/BpkLargeSpinner.js
olliecurtis: Did this suggestion and seems to work fine

packages/bpk-component-spinner/src/BpkSpinner.js
olliecurtis: Did this suggestion and seems to work fine

packages/bpk-component-spinner/src/BpkExtraLargeSpinner.js
```diff
@@ -26,19 +28,22 @@ import SPINNER_TYPES from './spinnerTypes';
...
+  const classNames = getClassName(
     'bpk-spinner',
     'bpk-spinner--extra-large',
     `bpk-spinner--${type}`,
-  ].map(getClassName);
-
-  if (className) {
-    classNames.push(className);
-  }
+    className && className,
```
shaundon, line 42:
```suggestion
    className,
```

You don't need to do the `className && className` here, just `className` works

I'm getting deja vu as I'm sure I left this exact comment on another PR of yours about a month ago 😛

packages/bpk-component-spinner/src/BpkLargeSpinner.js
```diff
@@ -26,23 +28,25 @@ import SPINNER_TYPES from './spinnerTypes';
...
-  ].map(getClassName);
-
-  if (alignToButton) {
-    classNames.push(getClassName('bpk-spinner--align-to-large-button'));
-  }
-  if (className) {
-    classNames.push(className);
-  }
+    alignToButton && 'bpk-spinner--align-to-large-button',
+    className && className,
```
shaundon, line 45:
```suggestion
    className,
```

packages/bpk-component-spinner/src/BpkSpinner.js
```diff
@@ -26,19 +28,24 @@ import SPINNER_TYPES from './spinnerTypes';
...
-    classNames.push(getClassName('bpk-spinner--align-to-button'));
-  }
-  if (className) {
-    classNames.push(className);
-  }
+  const classNames = getClassName(
+    'bpk-spinner',
+    `bpk-spinner--${type}`,
+    alignToButton && 'bpk-spinner--align-to-button',
+    className && className,
```
shaundon, line 44:
```suggestion
    className,
```

UNRELEASED.yaml
```diff
@@ -1,5 +1,15 @@
...
 
+ADDED:
+  - bpk-component-autosuggest:
+  - bpk-component-card:
+  - bpk-component-checkbox:
+  - bpk-component-heading:
+  - bpk-component-link:
+  - bpk-component-loading-button:
+  - bpk-component-spinner:
+    - Added flow typing to components
```
shaundon, line 11:
```suggestion
    - Added Flow typing to components.
```

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
shaundon: Ah ok. In this case I would still use a function but make it more specific, like `() => mixed`

packages/bpk-component-spinner/src/BpkExtraLargeSpinner.js
olliecurtis: Yeah my ide was throwing errors when I did this yesterday so had to put that in for now

shaundon: Are you sure? We use this in many different components so it would be unusual for it to only error on these instances and not all of them, if it's really broken

olliecurtis: As I said was just my ide yesterday was planning on fixing it this morning after a pc restart

shaundon:
Have tested locally and all good, so I'm happy to approve once the changes mentioned above are addressed

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
olliecurtis: I probably try using `Element` instead of `() => mixed` as that is too generic whereas we want to be more specific here to restrict to the correct expectation which is what we have done in other places that do this

UNRELEASED.yaml
olliecurtis: Done

packages/bpk-component-spinner/src/BpkSpinner.js
olliecurtis: Done

packages/bpk-component-spinner/src/BpkLargeSpinner.js
olliecurtis: Done

packages/bpk-component-spinner/src/BpkExtraLargeSpinner.js
olliecurtis: Done

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
olliecurtis: So had a play around with different types here and Function is the best we can use here given the prop is declared as a func in `propTypes` as setting to `() => mixed` or `Element` throws errors that `null` is not a valid function see attached
Screenshot 2020-08-05 at 10 37 25

Screenshot 2020-08-05 at 10 36 18

PR #1944, opened 2020-08-06
[BPK-4012] Upgrade to latest eslint-config-skyscanner and fix issues

shaundon:
Initially I just upgraded `eslint-config-skyscanner`, and got this error when running the linter:

```
Error: /Users/shaundonnelly/projects/backpack/node_modules/eslint-config-airbnb-base/rules/style.js:
	Configuration for rule "padded-blocks" is invalid:
	Value [{"blocks":"never","classes":"never","switches":"never"},{"allowSingleLineBlocks":true}] should NOT have more than 1 items.

Referenced from: /Users/shaundonnelly/projects/backpack/node_modules/eslint-config-airbnb-base/index.js
Referenced from: /Users/shaundonnelly/projects/backpack/node_modules/eslint-config-airbnb/index.js
Referenced from: /Users/shaundonnelly/projects/backpack/node_modules/eslint-config-skyscanner/index.js
Referenced from: /Users/shaundonnelly/projects/backpack/.eslintrc
```

I couldn't find reference to this anywhere in our config, and I know that this config works fine in a vanilla project, so I upgraded `eslint_d` to see if that would help. It did, but it found some new errors so I've fixed all of those in this PR too.


.eslintrc
```diff
@@ -24,6 +24,7 @@
     }
   },
   "rules": {
+    "backpack/use-components": "off",
```
georgegillams, line 27: Presumably this is off because we can't use a backpack component inside a backpack component?

packages/bpk-component-radio/src/BpkRadio.js
```diff
@@ -52,6 +52,9 @@ const BpkRadio = props => {
 
   return (
     <label className={classNames.join(' ')}>
+      {/* Deciding to support this because this prop does often work with voiceover
```
georgegillams, line 55:
```suggestion
      {/* Deciding to support this because `aria-invalid` does often work with voiceover
```

packages/bpk-component-table/src/BpkTableHead-test.js
```diff
@@ -26,7 +26,7 @@ describe('BpkTableHead', () => {
     const tree = renderer
       .create(
         <BpkTableHead>
-          <th />
+          <th>Skyscanner</th>
```
georgegillams, line 29: Was this deliberate?

.eslintrc
shaundon: Yeah, there are legit places where we need to use HTML elements, for example we use `<code />` inside `<BpkCode />`. There were over 100 errors like this. @olliecurtis and I discussed it and decided this brute force solution was better because we already know to use our own components, and it saves the tedium of adding over 100 opt outs.

packages/bpk-component-table/src/BpkTableHead-test.js
shaundon: Yeah, it was to solve this error:

```
error  A control must be associated with a text label  jsx-a11y/control-has-associated-label
```

.eslintrc
georgegillams: Yeah I guess if we haven't used backpack there's probably a very good reason for it 😉

packages/bpk-component-table/src/BpkTableHead-test.js
georgegillams: Seems strange - what control is it referring to?

shaundon: The `th` element — it makes sense, because `<th />` evaluates to `<th></th>` which is invalid, it needs to have content. So it's just enforcing that. You'll see a similar update in this PR on a `<Button />`.

georgegillams: I'm an idiot. I thought you were removing the text, but you're adding it which totally makes sense 👍

shaundon: Haha no worries

shaundon:
Merging anyway because Flow won't work on CI.

PR #1945, opened 2020-08-06
[BPK-4015]: Increase flow coverage

olliecurtis:
Added flow to the following:

  - bpk-component-accordion
  - bpk-component-breakpoint
  - bpk-component-paragraph
  - bpk-component-radio
  - bpk-component-star-rating
  - bpk-component-table
  - bpk-component-textarea

### DIFF
--- a/packages/bpk-component-accordion/src/withSingleItemAccordionState.js
+++ b/packages/bpk-component-accordion/src/withSingleItemAccordionState.js
@@ -16,8 +16,17 @@
  * limitations under the License.
  */
 
+/* @flow strict */
+
 import PropTypes from 'prop-types';
-import React, { Component, Children, cloneElement } from 'react';
+import React, {
+  Component,
+  Children,
+  cloneElement,
+  type Node,
+  type Element,
+  type ComponentType,
+} from 'react';
 import { wrapDisplayName } from 'bpk-react-utils';
 
 const getInitiallyExpanded = children => {
@@ -29,9 +38,19 @@ const getInitiallyExpanded = children => {
   return (result || {}).key || null;
 };
 
-const withSingleItemAccordionState = ComposedComponent => {
-  class WithSingleItemAccordionState extends Component {
-    constructor(props) {
+type Props = {
+  children: Node,
+};
+
+type State = {
+  expanded: ?string | ?number,
+};
+
+const withSingleItemAccordionState = (
+  ComposedComponent: ComponentType<any>,
+) => {
+  class WithSingleItemAccordionState extends Component<Props, State> {
+    constructor(props: Props) {
       super(props);
 
       this.state = {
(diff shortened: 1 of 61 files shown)


packages/bpk-component-accordion/src/withSingleItemAccordionState.js
```diff
@@ -29,21 +38,31 @@ const getInitiallyExpanded = children => {
...
+    constructor(props: Props) {
       super(props);
 
       this.state = {
         expanded: getInitiallyExpanded(this.props.children),
       };
     }
 
-    openAccordionItem = key => {
+    openAccordionItem = (key: ?string | ?number) => {
```
olliecurtis, line 61: This gave mixed errors for both string and number - so added it to be either or as per the React error being thrown here.

packages/bpk-component-accordion/src/withSingleItemAccordionState.js
```diff
@@ -29,21 +38,31 @@ const getInitiallyExpanded = children => {
...
 
-const withSingleItemAccordionState = ComposedComponent => {
-  class WithSingleItemAccordionState extends Component {
-    constructor(props) {
+type Props = {
+  children: Node,
+};
+
+type State = {
+  expanded: ?string | ?number,
```
olliecurtis, line 46: This gave mixed errors for both string and number - so added it to be either or as per the React error being thrown here.

packages/bpk-component-breakpoint/src/BpkBreakpoint.js
```diff
@@ -29,11 +31,18 @@ const BREAKPOINTS = {
   ABOVE_TABLET: TOKENS.breakpointQueryAboveTablet,
 };
 
-const BpkBreakpoint = props => (
-  <MediaQuery query={props.query}>{props.children}</MediaQuery>
-);
+type Props = {
+  children: Node | (() => mixed),
```
shaundon, line 35: This file doesn't import `type Node` from React, so this is probably referring to the wrong Node.

packages/bpk-component-breakpoint/src/BpkBreakpoint.js
```diff
@@ -29,11 +31,18 @@ const BREAKPOINTS = {
...
+  query: string | $Keys<typeof BREAKPOINTS>,
```
shaundon, line 36: This should take a function like it does in `propTypes`, I'll fix it.

shaundon: Ah I see what you were doing here now, and I messed it up 😅

shaundon: So one important distinction here is that I had to change it to `string | $Values<typeof BREAKPOINTS>,` (values instead of keys).

If we allow `$Keys` that evaluates to `'MOBILE' | 'TABLET' | 'TABLET_ONLY' | ..etc`, but what users will actually pass in is the values that the keys refer to: `'32.25rem' | '71.25rem' | ..etc`. `$Values` achieves that.

Fun fact, when we first adopted Flow the `$Values` type didn't exist and stuff like this was a huge headache!

packages/bpk-component-breakpoint/src/BpkBreakpoint.js
olliecurtis: Hmm it's strange that it somehow still accepted it though!

shaundon: Yeah, seemed weird but I figured I should fix it anyway!

Also: ➡️ 🌴 ❗

PR #1960, opened 2020-08-28
[BPK-4160]: Upgrade Flow to 0.132.0

olliecurtis:
Upgraded Flow to v 0.132.0

Changed syntax of `FlowFixMe` to include the error code that Flow will report on as part of the new requirement of 0.132.0

Updated some `FlowFixMe` to use `FlowIgnore` for things that we can't/won't fix and `FlowIssue` where new issues have been found that are problems with Flow

### DIFF
--- a/decisions/flowfixme.md
+++ b/decisions/flowfixme.md
@@ -5,7 +5,7 @@ When we tried to upgrade to Flow 0.112, all of our components that used `...rest
 We decided to suppress these comments using the following comment:
 
 ```
-// $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
+// $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
 ```
 
 If you need to do this in future, add the same comment. Using this format will allow us to easily find and remove the suppression comments once a better solution is found.
--- a/packages/bpk-component-label/src/BpkLabel.js
+++ b/packages/bpk-component-label/src/BpkLabel.js
@@ -26,6 +26,7 @@ import STYLES from './BpkLabel.scss';
 const getClassName = cssModules(STYLES);
 
 export type Props = {
+  // $FlowIssue[value-as-type] - Node is a type import from prop-types and is mishandled by Flow
   children: Node,
   className: ?string,
   disabled: boolean,
@@ -61,7 +62,7 @@ const BpkLabel = (props: Props) => {
   }
 
   return (
-    // $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
+    // $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
     <label className={classNames.join(' ')} {...rest}>
       {children}
       {!disabled && required && (
--- a/packages/bpk-component-label/src/BpkLabel.js
+++ b/packages/bpk-component-label/src/BpkLabel.js
@@ -17,8 +17,8 @@
  */
 /* @flow strict */
 
-import PropTypes, { type Node } from 'prop-types';
-import React from 'react';
+import PropTypes from 'prop-types';
+import React, { type Node } from 'react';
 import { cssModules } from 'bpk-react-utils';
 
 import STYLES from './BpkLabel.scss';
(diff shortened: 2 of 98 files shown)


decisions/flowfixme.md
```diff
@@ -5,7 +5,7 @@ When we tried to upgrade to Flow 0.112, all of our components that used `...rest
 We decided to suppress these comments using the following comment:
 
 ```
-// $FlowFixMe - inexact rest. See 'decisions/flowfixme.md'.
+// $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
```
olliecurtis, line 8: As flow now expects an error code to be provided updating example comment for easy use.

packages/bpk-component-label/src/BpkLabel.js
```diff
@@ -26,6 +26,7 @@ import STYLES from './BpkLabel.scss';
 const getClassName = cssModules(STYLES);
 
 export type Props = {
+  // $FlowIssue[value-as-type] - Node is a type import from prop-types and is mishandled by Flow
```
olliecurtis, line 29: An example where a new issue has been introduced from the new version.

We could potentially solve this by changing the import to

```js
import { Node } from 'prop-types'
```

and then do 

```js
children: typeof Node
```

but not sure if that is what we should do vs this being an issue with Flow?

packages/bpk-react-utils/src/deprecated.js
```diff
@@ -19,6 +19,7 @@
 /* @flow strict */
 import { type PropType } from 'prop-types';
 
+// $FlowIssue[value-as-type] - PropType is imported as a type so is incorrectly reporting the PropType is not a valid type
```
olliecurtis, line 22: Same as https://github.com/Skyscanner/backpack/pull/1960/files#r479434855

packages/bpk-component-select/src/BpkSelect.js
```diff
@@ -35,6 +35,7 @@ export type Props = {
   dockedFirst: boolean,
   dockedLast: boolean,
   dockedMiddle: boolean,
+  // $FlowIssue[value-as-type] - Node is a type import from prop-types and is mishandled by Flow
```
olliecurtis, line 38: Same as https://github.com/Skyscanner/backpack/pull/1960/files#r479434855

packages/bpk-component-label/src/BpkLabel.js
shaundon: Just wanna check, as this might be a legit bug we're covering up — usually we do `import { type Node } from 'react'`, i.e. not `'prop-types'`. I've never seen it imported from prop-types so I think it's a mistake, although I might be wrong.

If you change it to import from `react` does this error go away?

decisions/flowfixme.md
shaundon: This is so much better, I'm happy they added this

olliecurtis: Yeah so cleaner

packages/bpk-component-label/src/BpkLabel.js
olliecurtis: Will have a look!

olliecurtis: This might be from an old implementation before React had support so if that solves it and we can remove this line that will be perfect!

shaundon: 🤞

shaundon (approved):
All looks good! There's the outstanding comment about importing `Node` from `prop-types` but once you take a look and see if it can be removed or not, all good from my end.

I didn't fun this locally, I'm relying on CI

shaundon:
Oh CI failed.. I could have sworn it was passing when I first opened this

packages/bpk-component-label/src/BpkLabel.js
olliecurtis: Yep looks like this was fine to change to 😄

shaundon (approved):

packages/bpk-react-utils/src/deprecated.js
olliecurtis: For this one PropType does not exist in `react` so leaving as is

shaundon (approved):

PR #1966, opened 2020-09-02
[BPK-4151]: Added new BpkPriceMarker to map component

olliecurtis:
Maps


packages/bpk-component-map/src/BpkPriceMarker.js
```diff
@@ -0,0 +1,131 @@
...
+  TEXT_STYLES,
+} from 'bpk-component-text/src/BpkText';
+
+import { LatLongPropType, type LatLong } from './common-types';
+import BpkBasicMapMarker from './BpkBasicMapMarker';
+import STYLES from './BpkPriceMarker.scss';
+
+const getClassName = cssModules(STYLES);
+
+export const MARKER_STATUSES = {
```
shaundon, line 36: Can we name this `PRICE_MARKER_STATUSES`? It'll make it clearer to consumers that it applies to the price marker component, otherwise they might think it's something to be used with `BpkMapMarker`.

Sidenote: You've already named it `PRICE_MARKER_STATUSES` a couple of times in the readme, probably copypasta'd from the RN readme?

packages/bpk-component-map/src/BpkPriceMarker.js
```diff
@@ -0,0 +1,131 @@
...
+  viewed: 'viewed',
+};
+
+type Props = {
+  label: string,
+  position: LatLong,
+  disabled: boolean,
+  className: ?string,
+  arrowClassName: ?string,
+  onClick: ?(event: SyntheticEvent<>) => mixed,
```
shaundon, line 48: Just checking, should `onClick` be optional? I made it required in RN because I figured it's like a type of button and should always have an action. Happy to discuss though, I could be wrong

packages/bpk-component-map/src/BpkPriceMarker.scss
```diff
@@ -0,0 +1,143 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins';
+
+.bpk-map-marker {
```
shaundon, line 21: Can we name this `bpk-price-marker` instead?

packages/bpk-component-map/stories.js
```diff
@@ -216,4 +316,5 @@ storiesOf('bpk-component-map', module)
         icon={<AlignedBusIconLg />}
       />
     </StoryMap>
-  ));
+  ))
+  .add('Stateful price markers', () => <StatefulBpkPriceMarker />);
```
shaundon, line 320: Nitpick, but we can probably just call this 'price markers' as it's the only example we have of them (i.e. no reason to call them out as being stateful).

```suggestion
  .add('Price markers', () => <StatefulBpkPriceMarker />);
```

shaundon (changes requested):
Couple of small naming things but otherwise looks great. Will pull down to try locally now.

packages/bpk-component-map/src/BpkPriceMarker.js
olliecurtis: Sure thing sounds good!

packages/bpk-component-map/src/BpkPriceMarker.js
olliecurtis: Let me check as I took this from the `BpkMapMarker.js` so might be from historical reasons. I guess it might have been that not always an action was needed and was used to just display but not sure the reasoning from when it was built

olliecurtis: Actually you might know as you built the component 😆

packages/bpk-component-map/src/BpkPriceMarker.scss
olliecurtis: Yep sounds good!

packages/bpk-component-map/stories.js
olliecurtis: 👍

packages/bpk-component-map/src/BpkPriceMarker.js
shaundon: spider-man pointing at self meme

Yeah let's just leave it optional then, as I guess it'll be weird if it's optional in `BpkMapMarker` but required in `BpkPriceMarker`.

shaundon (changes requested):
I've run locally and all good, I just one further request — can we use Storybook's `action` function to display a message when the examples are clicked? 

Here's an example: https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-button/stories.js#L73

shaundon:
Also, we should make sure it's been tested in mobile Safari and IE as it's a new component

olliecurtis:
@shaundon all comments have been addressed and tested in the browsers mentioned so this should be ready to review again 😄

shaundon (approved):
`README.md` still mentions the old `MARKER_STATUSES` object in a couple of places, might be worth running a find and replace on it. Otherwise all good once that's fixed

PR #1976, opened 2020-09-11
WIP - [no-jira]:Post publish dependency upgrades

olliecurtis:
Exploring possible post publish upgrade strategy for upgrading dependencies in the docs site.


olliecurtis:
Parking for now as we want to explore potentially using GH Actions

PR #1977, opened 2020-09-11
[No Jira] Made some tweaks to Danger to make it less noisy

shaundon:
As discussed on Slack, some of our well meaning Danger notices were being ignored, which reduces the impact of the important ones.

## Changed behaviours

* The browser support checklist, which we never filled in, is now a notice reminding you to check it in multiple browsers, especially IE11.
* Dependabot PRs don't get the "you're amazing thanks xoxo" messages we send to external contributors.

## Removed behaviours

* "It looks like you are introducing a new component. Ensure the component style is extensible via `className`."
* "`package-lock.json` was updated. Ensure that this was intentional."

## Added behaviours

* I added a notice to reviewers that only appears for Dependabot PRs. I'm hoping that this will help give some quick instructions on how to keep the number down in future. I'm not entirely sure if my logic check (`const isPrFromDependabot = author === 'dependabot';`) will work but I figured the best way to test it is to push it and see what happens.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -85,7 +85,7 @@ if (componentChangedOrCreated) {
   message(`
   ## Browser support
 
-  If this is a significant change, make sure you've tested it in multiple browsers,
+  If this is a visual change, make sure you've tested it in multiple browsers,
   particularly IE11.
   `);
 }
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -41,7 +41,6 @@ const BACKPACK_SQUAD_MEMBERS = meta.maintainers.map(
 );
 const author = danger.github.pr.user.login;
 const isPrExternal = !BACKPACK_SQUAD_MEMBERS.includes(author);
-const isPrFromDependabot = author === 'dependabot';
 
 const createdFiles = danger.git.created_files;
 const modifiedFiles = danger.git.modified_files;
@@ -62,22 +61,7 @@ const thanksGifs = [
   'https://media.giphy.com/media/1lk1IcVgqPLkA/giphy.gif', // Cap salute
 ];
 
-if (isPrFromDependabot) {
-  markdown(`
-  # Notes for reviewers about Dependabot PRs
-
-  If CI passes, it's usually ok to merge. We should be confident in CI's ability to catch issues.
-  If a PR passes CI and *does* break something later, we should make tickets to improve CI to cover whatever it missed.
-
-  ## Closing PRs
-
-  If a PR fails CI or is quite old, it's ok to close it. Ideally our dependencies would always be up to date, but
-  it's ok to miss a few dependency versions. As long as we don't get too behind, it's fine.
-
-  When we close a Dependabot PR, it'll make another when a new version is released, so we always get the chance
-  to look at it again.
-  `);
-} else if (isPrExternal) {
+if (isPrExternal) {
   markdown(`
   # Hi ${author}!
 
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -41,11 +41,11 @@ const BACKPACK_SQUAD_MEMBERS = meta.maintainers.map(
 );
 const author = danger.github.pr.user.login;
 const isPrExternal = !BACKPACK_SQUAD_MEMBERS.includes(author);
+const isPrFromDependabot = author === 'dependabot';
 
 const createdFiles = danger.git.created_files;
 const modifiedFiles = danger.git.modified_files;
 const fileChanges = [...modifiedFiles, ...createdFiles];
-const declaredTrivial = danger.github.pr.title.includes('#trivial');
 const markdownChanges = fileChanges.filter(path => path.endsWith('md'));
 
 const thanksGifs = [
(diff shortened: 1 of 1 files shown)


dangerfile.js
```diff
@@ -114,7 +111,7 @@ const unreleasedModified = includes(modifiedFiles, 'UNRELEASED.yaml');
 const packagesModified = fileChanges.some(
   filePath => filePath.startsWith('packages/') && !filePath.endsWith('.md'),
 );
-if (packagesModified && !unreleasedModified && !declaredTrivial) {
+if (packagesModified && !unreleasedModified) {
```
georgegillams, line 114: So this is even noisier now?

shaundon: Have you ever seen a PR where somebody actually added `#trivial` to the title to suppress an error? What's the point of having it if it's not used?

georgegillams: If it was a label, I'd probably use it.

but yeah I guess it's so rare that we'd probably only remember the option after we see the warning anyway

georgegillams (approved):

dangerfile.js
```diff
@@ -62,8 +62,22 @@ const thanksGifs = [
   'https://media.giphy.com/media/1lk1IcVgqPLkA/giphy.gif', // Cap salute
 ];
 
-// Be nice to our neighbours.
-if (isPrExternal) {
+if (isPrFromDependabot) {
+  markdown(`
+  # Notes for reviewers about Dependabot PRs
+
+  If CI passes, it's usually ok to merge. We should be confident in CI's ability to catch issues.
```
olliecurtis, line 69: I would argue this as sometimes Dependabot PRs just update the dependency in the `package-lock.json` and not update it in the `package.json` if a direct dependency in the project so maybe we need to word this for us to check if the dependency its upgrading is a direct dependency and if so update the package.json

dangerfile.js
shaundon: Yeahhh - to clarify, I think it's a good idea but it's clearly not used as is, so I figured remove it and then we can explore a better solution.

dangerfile.js
```diff
@@ -98,14 +101,8 @@ if (componentChangedOrCreated) {
...
 
-  Please complete this list of browsers that you've checked this works in.
-
-  - [ ] IE11
-  - [ ] Edge
-  - [ ] Safari (iOS)
-  - [ ] Chrome (Android)
-  - [ ] Chrome (Desktop)
-  - [ ] Firefox (Desktop)
+  If this is a significant change, make sure you've tested it in multiple browsers,
```
olliecurtis, line 104: Maybe instead of this being from Danger we should keep the checklist and add it to the PR template instead to make it explicit as even small changes can have significant impact especially on IE - so we should make sure all changes are tested until IE 11 support is dropped

olliecurtis (changes requested):

dangerfile.js
shaundon: If I understand correctly (it's hard to read when you type one huge run-on sentence 😉), you're saying that the Dependabot PRs that only upgrade the lockfile need additional manual work by us before they can be merged?

If so, it sounds like Dependabot is failing to make usable PRs, which is a separate problem. We can't have it making PRs that aren't mergeable, I'd sooner just switch it off entirely.

olliecurtis: Yes since we changed the settings it will now sometimes just do lock file upgrades for dependencies that are direct in the project.

So we either further down the line need to update manually or in the same PR or when a major gets released.

olliecurtis: ⬆️ hopefully those are better shorter sentences 😉

dangerfile.js
olliecurtis: Where is even `declaredTrivial` declared? Call me crazy but I can't seem to find it in the original file 😆 so must have done nothing anyway?

dangerfile.js
shaundon: Yeah that feels kind of wrong to me. I've merged loads of PRs that Dependabot made that only changed the lockfile, because I assumed they were ok to merge. 

Dependabot's PRs needs to be treated exactly like ones we'd get from a human. If somebody made a PR that only did part of the job, we would ask them to stop doing that.

olliecurtis: Yeah that is something and why we need to treat these with care and see if there are other tools available to do this maybe?

georgegillams: On my repos, it bumps the `package.json` for direct dependencies 🤔 
https://github.com/georgegillams/browser-scripts/pull/49/files

olliecurtis: Looking at that repo @georgegillams it looks like you are still running the preview version over the official GH version (no .github/dependabot.yml file) and the default settings on that will do the `package.json` bump but we recently changed it for some reason will need to dig out why which has now caused it.

Or internally they have made a change to Dependabot

dangerfile.js
georgegillams: I think we should adopt the same attitude we use elsewhere - which is that our tests should be sufficient to catch anything, so if CI passes everything should be fine

dangerfile.js
georgegillams: 👍

shaundon: Ok cool, well let's talk about it on Monday and get to the bottom of it. 

I feel like everything that I wrote in the string above is how it **should** be, and if that's not the case we should do something about it so that we **can** be confident in the PRs that Dependabot makes

dangerfile.js
olliecurtis: I don't feel that is a good approach here as our CI doesn't do complex testing over different browsers - if it did then yes if CI is fine then we can merge but visually it could be broken on older browsers or non mainstream ones

shaundon: I think you're both right! 

Ideally our CI would do this but for now it's not feasible, so we have this message as a manual reminder. It would be cool to get to this as an end state though!

dangerfile.js
shaundon: I'm gonna revert this change so as not to block this PR, and we can discuss it separately on Slack

dangerfile.js
shaundon: With regards to having it in the PR template or Danger, I'd like to keep it in Danger because when I make PRs the first thing I do is remove the template, because I feel it's more for external contributors. 

I like it when Danger makes the comment because I get a notification to remind me, and it's something important, which is really the essence of what Danger is for.

Let's keep this but also add it to the PR template, there's no downside to having it in both places.

olliecurtis: Sounds good! Will we look at updating the wording for this as any visual change could have an affect across browsers not just significant

shaundon: Ah yeah I'll change that too

shaundon: Done — changed 'significant' to 'visual'

olliecurtis (approved):

PR #1978, opened 2020-09-14
[BPK-3438]: Support bold in accordion

olliecurtis:
The following PR allows the support of a `weight` prop that can be passed to `BpkAccordionItem` to style the title of the accordion.

- [x] `UNRELEASED.yaml`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:

Screenshot:

Screenshot 2020-09-14 at 12 45 58

### DIFF
--- a/packages/bpk-component-accordion/README.md
+++ b/packages/bpk-component-accordion/README.md
@@ -97,7 +97,7 @@ const AlignedStopsIcon = withAlignment(StopsIcon, lineHeightBase, iconSizeSm);
 | children  | node     | true     | -             |
 | id        | string   | true     | -             |
 | title     | string   | true     | -             |
-| weight    | See [`BpkText`](https://backpack.github.io/components/text?platform=web#readme) for valid options     | false    | WEIGHT_STYLES.regular         |
+| weight    | See prop details     | false    | WEIGHT_STYLES.regular         |
 | expanded  | bool     | false    | false         |
 | icon      | node     | false    | null          |
 | onClick   | func     | false    | () => null    |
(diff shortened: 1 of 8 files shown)


shaundon:
CI is failing because the prop is `black`, not `heavy` (I said `heavy` in my Slack message but I guessed wrong 🙈 )

shaundon (commented):
It looks good, one thing I'm not sure about though — should we re-export `WEIGHT_STYLES` from `bpk-component-accordion`, so that consumers don't need to import it from `bpk-component-text`?

I'm not certain because I like the simplicity of having them go to the source for it, but then it feels a bit odd that they need to use two different packages to pass this prop in. What do you think?

olliecurtis:
@shaundon - That is a good and interesting point! I guess re-exporting is probably a good idea, especially if a consumer doesn't pass `BpkText` as a child through to the component so won't necessarily have that package installed.

I guess one other potential thing we could do is similar to the `tagName` and `textStyle` we just make consumers pass in the string that they want but use the styles in `BpkText` docs as a reference for supported names?

shaundon:
I guess re-exporting doesn't really have any downsides, so maybe we should go with that?

packages/bpk-component-accordion/README.md
```diff
@@ -115,6 +116,12 @@ const AlignedStopsIcon = withAlignment(StopsIcon, lineHeightBase, iconSizeSm);
 | ------------------------------ | -------- | -------- | ------------- |
 | ~~expanded~~ initiallyExpanded | bool     | false    | false         |
 
+## Prop Details
+
+#### weight
+
+This prop takes `WEIGHT_STYLES` that has been re-exported from `BpkText`. For valid options please refer to the [BpkText](https://backpack.github.io/components/text?platform=web#readme) docs.
+
```
shaundon, line 124: Nice 🤩

shaundon (approved):
L G T M
G
T
M

PR #1979, opened 2020-09-14
[BPK-4150]: Removed italic placeholder style

olliecurtis:
This PR removed the italic placeholder `font-style` and the `italic` bpk token. Turns out we were only using this in the `input` and `textarea` component so was a easy removal.

This shouldn't be a breaking change for our consumers as this token should be only used internally for the component, however as the token is being removed its a breaking change.

- [x] `UNRELEASED.yaml`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:

Screenshot 2020-09-14 at 14 23 51
Screenshot 2020-09-14 at 14 24 12


shaundon (approved):
Looks great!

PR #1981, opened 2020-09-24
[ZEP-688] Allow consumers of BpkDatePicker to control isOpen from the outside

TudorRotarus25:
At the moment the isOpen state from the BpkDatePicker can be only initialised as opened or closed through a prop but the component doesn't re-render when the prop changes. To fix that I added a check in componentDidUpdate to update the state when the prop changes.

- [x] `UNRELEASED.yaml`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:

### DIFF
--- a/packages/bpk-component-datepicker/src/BpkDatepicker-test.js
+++ b/packages/bpk-component-datepicker/src/BpkDatepicker-test.js
@@ -233,7 +233,7 @@ describe('BpkDatepicker', () => {
   });
 
   it('should update state when a date is selected', () => {
-    const onCloseHook = jest.fn();
+    const onOpenChangeHook = jest.fn();
 
     const datepicker = mount(
       <BpkDatepicker
@@ -251,21 +251,22 @@ describe('BpkDatepicker', () => {
         minDate={new Date(2010, 1, 15)}
         maxDate={new Date(2010, 2, 15)}
         date={new Date(2010, 1, 15)}
-        onClose={onCloseHook}
+        onOpenChange={onOpenChangeHook}
       />,
     );
 
     datepicker.find('BpkInput').simulate('click');
     expect(datepicker.state('isOpen')).toEqual(true);
+    expect(onOpenChangeHook).toHaveBeenCalledWith(true);
 
     const date = new Date(2010, 1, 15);
     datepicker.instance().handleDateSelect(date);
     expect(datepicker.state('isOpen')).toEqual(false);
-    expect(onCloseHook).toHaveBeenCalledTimes(1);
+    expect(onOpenChangeHook).toHaveBeenCalledWith(false);
   });
 
   it('should close when `onClose` is called', () => {
-    const onCloseHook = jest.fn();
+    const onOpenChangeHook = jest.fn();
 
     const datepicker = mount(
       <BpkDatepicker
@@ -283,15 +284,16 @@ describe('BpkDatepicker', () => {
         minDate={new Date(2010, 1, 15)}
         maxDate={new Date(2010, 2, 15)}
         date={new Date(2010, 1, 15)}
-        onClose={onCloseHook}
+        onOpenChange={onOpenChangeHook}
       />,
     );
 
     datepicker.find('BpkInput').simulate('click');
     expect(datepicker.state('isOpen')).toEqual(true);
+    expect(onOpenChangeHook).toHaveBeenCalledWith(true);
 
     datepicker.instance().onClose();
     expect(datepicker.state('isOpen')).toEqual(false);
-    expect(onCloseHook).toHaveBeenCalledTimes(1);
+    expect(onOpenChangeHook).toHaveBeenCalledWith(false);
   });
 });
--- a/packages/bpk-component-datepicker/src/BpkDatepicker-test.js
+++ b/packages/bpk-component-datepicker/src/BpkDatepicker-test.js
@@ -233,6 +233,8 @@ describe('BpkDatepicker', () => {
   });
 
   it('should update state when a date is selected', () => {
+    const onCloseHook = jest.fn();
+
     const datepicker = mount(
       <BpkDatepicker
         id="myDatepicker"
@@ -249,6 +251,7 @@ describe('BpkDatepicker', () => {
         minDate={new Date(2010, 1, 15)}
         maxDate={new Date(2010, 2, 15)}
         date={new Date(2010, 1, 15)}
+        onClose={onCloseHook}
       />,
     );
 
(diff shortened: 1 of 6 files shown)


packages/bpk-component-datepicker/src/BpkDatepicker-test.js
```diff
@@ -204,6 +204,34 @@ describe('BpkDatepicker', () => {
     expect(datepicker.state('isOpen')).toEqual(true);
   });
 
+  it('should open when the isOpen prop is changed from the outside', () => {
```
jasperck, line 207: Maybe a test for we should not change isOpen when `prevProps.isOpen === isOpen` and `props.isOpen !== state.isOpen`?

jasperck:
Would it it better to move https://github.com/Skyscanner/backpack/blob/b0823df7eebea5ea6048870fc38ef2e835d97b0b/packages/bpk-component-datepicker/src/BpkDatepicker.js#L107 to L100, e.g.
```
  isOpen,
  ...rest
} = this.props
```

mrtu (approved):

jasperck (approved):
others LGTM!

packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -40,6 +40,17 @@ class BpkDatepicker extends Component {
     };
   }
 
+  componentDidUpdate(prevProps) {
```
mbfisher, line 43: I think `getDerivedStateFromProps` would be a better alternative - see [this React blog](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html). It would also mean you don't have to disable the `react/no-did-update-set-state` rule.

nicholas-l: That would mean bumping the peerDep for react though to 16.4 (currently [16.0](https://github.com/Skyscanner/backpack/blob/fbe1d4ddb7643025ba366a5ad03e924629e4117c/packages/bpk-component-datepicker/package.json#L13)).

rimadu (approved):

nicholas-l (approved):

packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -40,6 +40,17 @@ class BpkDatepicker extends Component {
...
+    const { isOpen } = this.props;
+
+    if (prevProps.isOpen !== isOpen) {
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({
+        isOpen,
+      });
```
georgegillams, line 50: Would it make sense for these to call the existing `onOpen` and `onClose` methods? That way if those methods are responsible for anything else (such as calling `props.onClose`) then we don't need to duplicate the logic?

packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -59,6 +73,9 @@ class BpkDatepicker extends Component {
     if (this.props.onDateSelect) {
       this.props.onDateSelect(dateObj);
     }
+    if (this.props.onClose) {
+      this.props.onClose();
+    }
```
georgegillams, line 78: Likewise I think we could call `onClose` here instead of setting state and calling `props.onClose`

packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -86,15 +103,16 @@ class BpkDatepicker extends Component {
       initiallyFocusedDate,
       renderTarget,
       valid,
+      // onDateSelect, onClose and isOpen are decomposed here so they don't end
+      // up in the rest object
+      onDateSelect,
+      onClose,
```
georgegillams, line 109: We shouldn't declare variables unless we plan to use them. I think the `delete` syntax was more appropriate.

packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -70,6 +73,9 @@ class BpkDatepicker extends Component {
     if (this.props.onDateSelect) {
       this.props.onDateSelect(dateObj);
     }
+    if (this.props.onClose) {
+      this.props.onClose();
+    }
```
jasperck, line 78: Why do we need to call this in handleDateSelect, wondering when date selected, open status should rely on state not props change? Looking at stories, it seems like passing a new props to close date picker, but date picker is also  setting isOpen to false at L70-72.

georgegillams (changes requested):
Thanks for the PR 🎉

Got a couple of comments

packages/bpk-component-datepicker/src/BpkDatepicker.js
TudorRotarus25: The onClose prop is not really for closing the date picker, it's a callback function that gets called when the date selector pop-up gets closed. This allows consumers to fully control the open/close state from the outside. I also added an use example there: https://github.com/Skyscanner/backpack/pull/1981/files#diff-9cb602e1e5145c854adc323bec617145R212

Does that answer the question?

TudorRotarus25: I see what you mean, the state and prop can get de-synchronised with only the onClose hook. I changed it to a onOpenChange hook instead so we can keep them in sync

packages/bpk-component-datepicker/src/BpkDatepicker.js
TudorRotarus25: That makes sense, much cleaner this way. Fixed now

packages/bpk-component-datepicker/src/BpkDatepicker.js
TudorRotarus25: Cool, I like to avoid delete statements but I don't have a strong preference. I reverted to the way it was before

packages/bpk-component-datepicker/src/BpkDatepicker-test.js
```diff
@@ -221,18 +251,23 @@ describe('BpkDatepicker', () => {
         minDate={new Date(2010, 1, 15)}
         maxDate={new Date(2010, 2, 15)}
         date={new Date(2010, 1, 15)}
+        onOpenChange={onOpenChangeHook}
```
georgegillams, line 254: Sorry - another minor thing. Could we rename this `onOpenChangeMock`, otherwise someone might be confused into thinking it has something to do with React Hooks?

TudorRotarus25: Yeah, good point, sorry about that. Fixed now

georgegillams (approved):
Beautiful 🎉 

Happy for me to merge and release?

TudorRotarus25:
> Beautiful 🎉
> 
> Happy for me to merge and release?

Yes please! Thank you!

PR #1988, opened 2020-10-02
[BPK-4125] New Icon Dialog

shaundon:


shaundon (commented):
I can't approve this because I'm the original author, but it all looks good. CI is failing because the CSS has changed — just needs `npm run build` then committing.

olliecurtis (approved):

PR #1992, opened 2020-10-06
[BPK-4218]: Flare dialog component

olliecurtis:
Added new flare dialog variant.

As part of this PR I introduced a new `flareComponent` prop to the `BpkModal` - this prop is internal so that we can apply the flare in the correct position when the dialog is created.

As we wish this to only be used by the `BpkDialog` component and not by end users it has not been documented for this purpose.

Screenshot 2020-10-06 at 17 52 25
Screenshot 2020-10-06 at 17 52 32


packages/bpk-component-dialog/src/BpkDialog.scss
```diff
@@ -29,6 +29,12 @@
     }
   }
 
+  &--with-flare {
+    height: $bpk-spacing-base * 10;
+    border-radius: inherit;
```
olliecurtis, line 34: This is so that the corners will round to the same radius as the modal container itself

packages/bpk-component-modal/src/BpkModalDialog.js
```diff
@@ -36,6 +36,7 @@ export type Props = {
   wide: boolean,
   isIphone: boolean,
   showHeader: boolean,
+  flareComponent: ?Element<any>,
```
olliecurtis, line 39: New undocumented prop as its only to be used by the Dialog

georgegillams:
Should the `X` be at the top?

Screenshot 2020-10-06 at 17 57 53

olliecurtis:
@georgegillams - designs don't show this - the way this is meant to be used is like this with buttons to dismiss. Will change the example 😄 
Screenshot 2020-10-06 at 17 59 11

olliecurtis:
Example updated

Screenshot 2020-10-06 at 18 09 03

packages/bpk-component-dialog/src/BpkDialog.scss
```diff
@@ -29,6 +29,12 @@
     }
   }
 
+  &--with-flare {
+    height: $bpk-spacing-base * 10;
```
georgegillams, line 33: This can be overridden by consumers, right?

olliecurtis: Yup using `flareClassName` it will allow consumers to override all of the styles that are set here

olliecurtis: For reference for reviewers https://github.com/Skyscanner/backpack/blob/28c9ee095a27cf2fb58920dcefaa9d34ead5679f/packages/bpk-component-dialog/src/BpkDialog.js#L67

packages/bpk-component-dialog/src/BpkDialog.scss
```diff
@@ -29,6 +29,12 @@
     }
   }
 
+  &--with-flare {
```
shaundon, line 32: The BEM naming here isn't correct — this isn't a modifier on an existing element, the flare is its own element. It should be `&__flare` (so that when used in code it would be `bpk-dialog__flare`)

packages/bpk-component-dialog/src/BpkDialog.scss
shaundon: Let's add a quick comment to that effect in the code, then it's documented for the future too

packages/bpk-component-modal/src/BpkModalDialog.js
shaundon: Should we give this a more generic name? From `BpkModalDialog`'s point of view, this is 'any arbitrary component to display at the top', not necessarily a flare. What about `headerComponent` or something like that?

shaundon (changes requested):

packages/bpk-component-dialog/src/BpkDialog.scss
olliecurtis: Sounds good I followed the with-icon as this is a dialog with flare as apposed to flare dialog as it's an addition to the dialog 🤷‍♀️

shaundon:
One thing I'm not sure about — I assume the design intention is for people to put illustrations and things inside the flare, and that the blue background seen in Figma is more of a placeholder. Currently, adding a custom background image requires some custom CSS, here's an example I just did locally:

```
// stories.scss
.bpk-dialog-custom-background {
  background: url('http://placekitten.com/200/300');
}


// stories.js
  .add('With flare', () => (
    <DialogContainer
      flare
      dismissible={false}
      flareClassName={getClassName('bpk-dialog-custom-background')}
    >
      <Paragraph>
        This is a default dialog. You can put anything you want in here.
      </Paragraph>
    </DialogContainer>
  ));
```

Screen Shot 2020-10-07 at 08 42 12

I'm wondering if there's a better API for this, where the image can be passed in using JS. For example we might add an optional `flareBackgroundImage` component that would be a path to the image, then we could set up the CSS for the consumer in the background.

Wondering what @georgegillams thinks as the original author of the flare component too.

packages/bpk-component-dialog/src/BpkDialog.scss
shaundon: That's different — `bpk-dialog--with-icon` is a modifier on the `BpkModal` component, but the icon itself is an element which is why it has `bpk-dialog__icon` as its class name.

olliecurtis:
> One thing I'm not sure about — I assume the design intention is for people to put illustrations and things inside the flare, and that the blue background seen in Figma is more of a placeholder. Currently, adding a custom background image requires some custom CSS, here's an example I just did locally:
> 
> ```
> // stories.scss
> .bpk-dialog-custom-background {
>   background: url('http://placekitten.com/200/300');
> }
> 
> 
> // stories.js
>   .add('With flare', () => (
>     <DialogContainer
>       flare
>       dismissible={false}
>       flareClassName={getClassName('bpk-dialog-custom-background')}
>     >
>       <Paragraph>
>         This is a default dialog. You can put anything you want in here.
>       </Paragraph>
>     </DialogContainer>
>   ));
> ```
> 
> Screen Shot 2020-10-07 at 08 42 12
> 
> I'm wondering if there's a better API for this, where the image can be passed in using JS. For example we might add an optional `flareBackgroundImage` component that would be a path to the image, then we could set up the CSS for the consumer in the background.
> 
> Wondering what @georgegillams thinks as the original author of the flare component too.
> 

@shaundon Let me check how the flare component is used and intended for use, maybe we open this up a bit more and instead of us pre building a flare it's on the consumer to provide it that way they can use it however they want?

shaundon:
>  maybe we open this up a bit more and instead of us pre building a flare it's on the consumer to provide it that way they can use it however they want?

Yeah that's another good option. I guess it's a tradeoff, as the current API is quite nice because consumers don't need to do too much. I like the idea of it being locked to a flare right now to stop people putting literally anything in there, so maybe there's a good middle ground we can inhabit

olliecurtis:
@shaundon looking at your comment about adding a background image by providing CSS - that makes sense that the consumer would do that as that is how our current flare implementation works already by setting it through `className` so I think this existing API is fine as its on the consumer already to do.

They would already need to provide it to change the colour or the height anyway if they OOTB version doesn't suit so I think this is acceptable given the flare API itself

https://backpack.github.io/components/flare?platform=web

olliecurtis:
@shaundon - on the comment 

> Should we give this a more generic name? From BpkModalDialog's point of view, this is 'any arbitrary component to display at the top', not necessarily a flare. What about headerComponent or something like that?

I would prefer to keep this as flareComponent as this is just added for the sole purpose of being a) internal b) for use solely by the dialog component as a flare so it aligns properly without CSS hacking and c) as the dialog uses `BpkModalDialog` under the hood so had to be put here.

If we open it up there is a risk of abuse of such prop if found by consumers as they can put anything in there and also we don't want to allow for additional components to the header as it does not align with our guidelines or is a request from consumers, I would like to keep it strict so people know what it is for if they stumble across it. What I am thinking is we try and make it a strict Flare component type instead of a generic `any` so we give an expectation its a flare that is required here.

shaundon:
I wish GitHub had threaded replies, it's so hard having multiple conversations in one comment stream!

Your comment about the background image — ok, all sounds good, let's leave it as is.

I kind of don't agree about the flare component name though. I'm not advocating for opening it up to consumers, I just think the name should be more generic. It doesn't make sense to call it `flareComponent` if you can technically pass anything in. All of the other reasons you cited still apply regardless of what the name is (for example if people want to abuse this prop, they will). @georgegillams @k0nserv what are your thoughts?

georgegillams:
What if we just allow a `flareChildren` prop?
That's essentially what we do on iOS

k0nserv:
Yeah I think `flareChildren` is the way to go? We just need to document a good way to do it with an image(using `BpkImage` ideally)

olliecurtis:
May need to reach out @k0nserv or @georgegillams just to understand what you mean by this but sounds like the way to go!

shaundon:
Let's check that it even **can** work with a `BpkImage` first, because it might only work correctly with a CSS `background-image`, because of the way the image needs to blend into the flare at the bottom

shaundon:
@georgegillams @k0nserv what about your thoughts on the flare prop name in `BpkModalDialog`? This comment: https://github.com/Skyscanner/backpack/pull/1992#discussion_r500792125

olliecurtis:
>Let's check that it even can work with a BpkImage first, because it might only work correctly with a CSS background-image, because of the way the image needs to blend into the flare at the bottom

I don't think the flare even accepts this and the way to apply an image is through CSS looking at the source code for the flare

olliecurtis:
@k0nserv our current flare doesn't take a BpkImage to add an image to the flare and CSS is the way we apply it. See the storybook example for the flare with image. This is how the component was built, I believe due to the point @shaundon suggested with the image blending into the pointer.

https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-flare/stories.js#L84
https://github.com/Skyscanner/backpack/blob/master/packages/bpk-component-flare/stories.scss#L40

packages/bpk-component-dialog/src/BpkDialog.js
```diff
@@ -74,6 +80,9 @@ const BpkDialog = (props: Props) => {
       fullScreenOnMobile={false}
       isIphone={false}
       contentClassName={headerIcon ? contentClassNames : null}
+      flareComponent={
```
olliecurtis, line 83: Forgot to update this 🤦 just doing it now

shaundon (approved):
L G T M

PR #1993, opened 2020-10-08
Use "currentColor" as "fill" in Icon documentation

sogame:
Using `currentColor` for `fill` in icons is very useful, as it allows not duplicating the same color (and makes components more reusable).
`currentColor` is not know by many people, so I though it would be good to use it in the example (adding a link to some documentation).

What do you think about this? Might not be the best way to promote its usage 😅

### DIFF
--- a/packages/bpk-component-icon/README.md
+++ b/packages/bpk-component-icon/README.md
@@ -30,7 +30,7 @@ export default () => (
 @import '~bpk-mixins';
 
 .abc-icon__flight {
-  fill: $bpk-color-white;
+  fill: currentColor; // see https://css-tricks.com/currentcolor/
 }
 
 .abc-icon__a11y {


packages/bpk-component-icon/README.md
```diff
@@ -30,7 +30,7 @@ export default () => (
 @import '~bpk-mixins';
 
 .abc-icon__flight {
-  fill: $bpk-color-white;
+  fill: currentColor; // https://www.w3schools.com/colors/colors_currentcolor.asp
```
k0nserv, line 33: Can we link to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentColor) instead of W3Schools? Or mabye [CSS Tricks](https://css-tricks.com/currentcolor/)

sogame: Makes sense, added both :-P

olliecurtis: I think just having one might be the best here as it looks to be a bit messy having both and overkill - I am in favour of CSS tricks. Can we just revert to one?

sogame: You're right, nicer with just one 😅 
Changed to CSS tricks.

olliecurtis (approved):

PR #1996, opened 2020-10-12
[No Jira] Fix warning Danger keeps reporting

shaundon:
Lately, Danger keeps reporting build log warnings in every PR — like this: https://github.com/Skyscanner/backpack/pull/1995#issuecomment-707092530.

I looked into it and it's a warning from `handlebars`. We don't use this, but it's a dependency of `jest`. Apparently they messed something up in a release and it was causing errors when `jest --coverage` is run.

I followed https://github.com/facebook/jest/issues/9396#issuecomment-573328488 to fix it, by making the coverage reports just print in text (which is how we were using it before, so nothing should change).

If it's worked, Danger shouldn't report any build log warnings on this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,6 +175,7 @@
       "^.+\\.scss$": "<rootDir>/scripts/stubs/styleStub.js",
       "^.+\\.svg$": "<rootDir>/scripts/stubs/fileStub.js"
     },
+    "coverageReporters": ["text"],
     "coverageThreshold": {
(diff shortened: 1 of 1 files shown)


shaundon:
I'm going to YOLO this through and see if it breaks CI on master. It's a one liner in our package.json so I can't see how it's breaking anything! If it breaks master I'll jump on it and fix.

PR #1998, opened 2020-10-12
[BPK-4240]: Add new iOS and Android Radius tokens

olliecurtis:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:


packages/bpk-tokens/src/android/base/aliases.json
```diff
@@ -41,7 +41,9 @@
     "BORDER_SIZE_SM": 1,
     "BORDER_SIZE_LG": 2,
     "BORDER_SIZE_XL": 3,
+    "BORDER_RADIUS_XS": 4,
     "BORDER_RADIUS_SM": 4,
```
olliecurtis, line 45: This value has been kept for now until the work in the platform to use is in progress so that if any urgent work comes up we aren't forced to adopt the breaking change of moving this to `6`

packages/bpk-tokens/src/ios/base/aliases.json
```diff
@@ -40,7 +40,9 @@
     "BORDER_SIZE_SM": 1,
     "BORDER_SIZE_LG": 2,
     "BORDER_SIZE_XL": 3,
+    "BORDER_RADIUS_XS": 4,
     "BORDER_RADIUS_SM": 4,
```
olliecurtis, line 44: This value has been kept for now until the work in the platform to use is in progress so that if any urgent work comes up we aren't forced to adopt the breaking change of moving this to `6`

packages/bpk-tokens/src/android/base/aliases.json
```diff
@@ -41,7 +41,9 @@
...
+    "BORDER_RADIUS_MD": 12,
```
georgegillams, line 46: So this unblocks the card/panel change, but not the [button change](https://gojira.skyscanner.net/browse/BPK-4232). Is that correct?

olliecurtis: Yes that is correct

olliecurtis: Yes that is correct as we are not immediately working on the button change we don't want to make the breaking change until we are ready to work on that and adopt it - as if any other urgent fixes are needed before that work starts it would mean we are forced to adopt the breaking change

georgegillams (approved):
Beautiful!

k0nserv (approved):

PR #2000, opened 2020-10-13
[skip ci] Format package.json correctly

shaundon:
Just committing this so that I can do a release without stuff changing!


georgegillams (changes requested):
We should just use `npm ci`.
From the `npm ci` docs:
```
It will never write to package.json or any of the package-locks: installs are essentially frozen.
```

shaundon:
@georgegillams I agree, but it should be formatted this way anyway so this is still valid

georgegillams:
> @georgegillams I agree, but it should be formatted this way anyway so this is still valid

Should it? I don't have anything against formatting it like this so feel free to merge 👍

PR #2003, opened 2020-10-16
[No Jira] Remove unused 'create-react-class' dep from slider component

shaundon:
Closes #2002 

We got a Snyk PR to upgrade this dep, but I looked in the code and it's not used anywhere. I ran the tests and storybook after removing it just to be safe too.

### DIFF
--- a/packages/bpk-component-slider/package-lock.json
+++ b/packages/bpk-component-slider/package-lock.json
@@ -1,15 +1,15 @@
 {
 	"name": "bpk-component-slider",
-	"version": "2.0.13",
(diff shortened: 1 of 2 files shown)


georgegillams:
Is it a rogue dependency of any other packages?

georgegillams (approved):

shaundon:
Not that I'm aware of, I searched the entire codebase for it and it only appeared in the package.json and lockfile of this one. Bit of a weird dep, it's for codebases that don't support ES6..

PR #2006, opened 2020-10-22
[BPK-4247] Reinstate iOS tokens for RN

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:


packages/bpk-tokens/src/ios/base/borders.json
```diff
@@ -4,6 +4,18 @@
     "type": "size"
   },
   "props": {
+    "BORDER_SIZE_SM": {
```
georgegillams, line 7: I previously removed these as they're no longer used by native iOS, but I didn't realise that removing them would remove them from RN where we need to still use them.

packages/bpk-tokens/src/ios/base/borders.json
```diff
@@ -35,6 +47,26 @@
     "CORNER_RADIUS_PILL": {
       "value": "{!BORDER_RADIUS_PILL}",
       "category": "radii"
+    },
+    "BORDER_RADIUS_XS": {
```
olliecurtis, line 51: Do we need these given that they are named corner radius above? It seems we are just creating duplicates of the same tokens 🤔

packages/bpk-tokens/src/react-native-output-test.js
```diff
@@ -24,14 +24,6 @@ import androidTokens from '../tokens/base.raw.android.json';
 import iosTokens from '../tokens/base.raw.ios.json';
 
 const ANDROID_ONLY_PROPS = [
-  'BORDER_RADIUS_LG',
```
olliecurtis, line 27: These are still needed on Android, why have they gone missing?

packages/bpk-tokens/tokens/base.raw.ios.json
```diff
@@ -393,6 +393,62 @@
       "originalValue": "{!WHITE}",
       "name": "BACKGROUND_TERTIARY_LIGHT_COLOR"
     },
+    "BORDER_RADIUS_LG": {
```
olliecurtis, line 396: Same as my comment before around do we need them given they have corner radius

packages/bpk-tokens/tokens/ios/base.react.native.es6.js
```diff
@@ -433,14 +449,6 @@ lineHeightXs,
 lineHeightXxl,
 };
 // Android only tokens
-export const borderRadiusLg = undefined;
```
olliecurtis, line 436: Same comment as in the test file, we probably need to keep these as they are Android only

olliecurtis (changes requested):

packages/bpk-tokens/src/react-native-output-test.js
georgegillams: Because they're not `ANDROID_ONLY` anymore.
This is a list of all the values that are only on Android. As these are available on both again, they're not in the list.

packages/bpk-tokens/tokens/base.raw.ios.json
georgegillams: Yes, otherwise every RN usage has to look like this:
`borderRadius: Platform.OS === 'iOS' ? cornerRadiusSm : borderRadiusSm`

And in RN `borderRadius` is a more 1st-class name. Corner radius is very Native iOS

packages/bpk-tokens/tokens/ios/base.react.native.es6.js
georgegillams: But they're not anymore. I'm reinstating them for iOS

packages/bpk-tokens/src/ios/base/borders.json
georgegillams: See https://github.com/Skyscanner/backpack/pull/2006#discussion_r509957948

packages/bpk-tokens/tokens/base.raw.ios.json
olliecurtis: Sure thing! This is interesting as myself and Shaun had this convo in slack and somehow even though we use `borderRadius: borderRadiusSm` in code and have not done your suggestion it always worked even though we didn't define the `cornerRadiusSm` 🤯

olliecurtis (approved):

packages/bpk-tokens/tokens/base.raw.ios.json
georgegillams: Yeah because `bpk-tokens` in the react-native codebase is 2 major versions behind 🙈

PR #2008, opened 2020-10-26
[BPK-4249] Only apply hover state is onRowClick is set

shaundon:
Kapture 2020-10-26 at 11 44 14


packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -139,6 +139,9 @@ class BpkDataTable extends Component {
     if (this.state.rowSelected === index) {
       classNames.push(getClassName('bpk-data-table__row--selected'));
     }
+    if (this.props.onRowClick !== undefined) {
```
georgegillams, line 142: In case they provide `null` for some reason

```suggestion
    if (this.props.onRowClick) {
```

olliecurtis (commented):
LGTM - Need to run it locally!

packages/bpk-component-datatable/src/BpkDataTable.js
shaundon: I did it this way to maintain parity with the logic in `onRowClicked`:

```js
    if (this.props.onRowClick !== undefined) {
      this.props.onRowClick(this.state.sortedList[index]);
    }
```

We could change it there too but I worry it'll have unforeseen consequences.

georgegillams: If that's already a limitation there then this is fine.
If they pass in null (or any other non-function except `undefined`) then this will blow up though

shaundon: I wonder why we did it this way in the first place 🤔

olliecurtis (approved):
LGTM - Tested locally and all good!

PR #2010, opened 2020-10-26
[NO JIRA]: Migrate to gh actions

olliecurtis:
Migrating from Travis CI to GH Actions.

One last outstanding item is to create the tokens to allow docs to be deployed once built.


georgegillams (changes requested):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+    name: Trigger docs Travis CI build
+    if: github.ref == 'refs/heads/master'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Trigger docs Travis CI
+        env:
+          TRAVIS_API_KEY: ${{ secrets.TRAVIS_API_KEY }}
+        run: |
+          curl -f -s -X POST -H "Content-Type: application/json" -H "Accept: application/json" -H "Travis-API-Version: 3" -H "Authorization: token $TRAVIS_API_KEY" -d '{"request": {"branch":"master"}}' https://api.travis-ci.org/repo/Skyscanner%2Fbackpack-docs/requests
```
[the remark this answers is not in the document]
olliecurtis: Ok awesome will keep it in for now until that or is merged, just incase this goes in first so still need it 🙂

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+      run: npm ci
+    
+    - name: Build Backpack
+      run: npm run build
+
+    - name: Check package-lock
+      run: ./scripts/check-pristine-state package-lock.json
+    
+    - name: Test
+      run: (set -o pipefail; npm test |& tee test.log)
```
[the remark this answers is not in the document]
olliecurtis: I can change it, this was just taken directly out of the travis step so left it so as not to case issues or loss of info 🙂

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    
+    - name: nvm install
+      shell: bash -l {0}
```
shaundon, line 17: What does this line do?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
+name: Backpack CI
+
+on:
+  push:
```
k0nserv, line 4: What about tags? Should we run on those too?(in which case the checks with `refs/head/master` will need to be tweaked as the ref will be `refs/tags/x.y.z`

olliecurtis: Sure thing! Wasn't sure about this one as when we publish tags we publish multiple ones (one for each package) so not sure how that will play with the deploy based on tags as we don't want to kick off 30 builds at once 🤔

k0nserv: That is a good point 🤔 Probably wouldn't be great if that happened

.github/workflows/ci.yml
olliecurtis: That is a good question I need to ask @k0nserv as I took this from backpack-ios GH Actions

shaundon: Ah ok, yeah I'm sure it's there for a good reason, it just wasn't immediately obvious (in which case maybe we should pop a comment above it)

k0nserv: See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

.github/workflows/ci.yml
olliecurtis: Yeah we need to think how to do this responsibly - I took the root of just doing pushes based on master as we don't tend to make changes directly on master without them going through a PR so means if we are merging they should be good to deploy the docs straight away?

.github/workflows/ci.yml
shaundon: It says `bash` is the default though, so what does this line actually do that we don't already get built in?

olliecurtis: We add the `-l` which from official bash docs says

```
-l        Make bash act as if it had been invoked as a login shell (see
                 INVOCATION below).
```

shaundon: Why do we need to set this?

olliecurtis: Referring again to @k0nserv as he set this up in iOS - guessing there was some issues with setting it up? I was planning to chat to him about some of these things to understand more what they meant

olliecurtis: So tried removing it and it seems to fail, I think using the logged in parameter allows it to assume a state of a logged in shell so has access for tooling?

Screenshot 2020-10-27 at 11 23 36

shaundon: Ah fair enough! Thanks for checking it

shaundon (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+    
+    - name: Deploy storybook
+      uses: peaceiris/actions-gh-pages@v3
+      if: github.ref == 'refs/heads/master'
+      with:
+        personal_token: ${{ secrets.DEPLOY_TOKEN }}
+        publish_dir: dist-storybook/
+        keep_files: true
+        external_repository: backpack/storybook
+        publish_branch: master
```
georgegillams, line 62: Can we have a new line at the end of the file?

georgegillams (approved):

.github/workflows/ci.yml
olliecurtis: Merged before I saw this, will add something to master now :)

PR #2011, opened 2020-10-27
[no jira]: Add danger step to CI

olliecurtis:
We had previously disabled Danger as part of CI but now we have a way to configure and run this as part of CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,11 @@ jobs:
         run: |
           nvm use
           npm test
+      
+      - name: Danger
+        run: npm run danger
+        env:
+          GITHUB_TOKEN: ${{ secrets.DANGER_GITHUB_API_TOKEN }}
 
       - name: Build Storybook
         run: |


.github/workflows/ci.yml
```diff
@@ -33,6 +33,12 @@ jobs:
     - name: Test
       run: npm test
     
+    - name: Danger
+      run: npm run danger
```
sogame, line 37: I'm doing the same in `danger-plugin-toolbox` (https://github.com/sogame/danger-plugin-toolbox/blob/master/.github/workflows/pull_request.yml#L39-L42) and I found a problem: when running the build on master after merging it was failing because Danger can't find the PR data.
It seems like it's not able to detect it's not pr (as it does when running in Travis, for example). I think this should get solved with https://github.com/danger/danger-js/issues/1066

I ended up creating two workflows, one for pr and one for push: https://github.com/sogame/danger-plugin-toolbox/pull/403
The only difference is that the push one doesn't run the danger step.

I'm new to GH Actions, so I might have something wrong in my repo, thought 😅

olliecurtis: @sogame coming back to this and reading the issue more, if its only on the danger step that you need to worry about you won't need to duplicate the logic and steps, all you need to do is have your danger step only run if its a PR which you can simply do by running an if statement

`if: github.ref != 'refs/heads/main'`

Similar to how we do with running our PR builds/deployment of storybook in the repo, see here https://github.com/Skyscanner/backpack/blob/main/.github/workflows/ci.yml#L88 - **Note:** in this example we only run the step if the branch is not on main and is not an external PR outside the repo due to the availablity of secrets in CI

sogame: Oh, nice! Thanks @olliecurtis! I didn't know about this (I'm quite new to github actions 😅 )
I'll apply to my repo, thank a lot!!

PR #2014, opened 2020-11-02
[BPK-3895] Increase link button hit area

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:


UNRELEASED.md
```diff
@@ -1,6 +1,9 @@
 # Unreleased
 
 > Place your changes below this line.
+**Fixed:**
```
georgegillams, line 4: Maybe this is visually breaking?

shaundon (commented):
I'm really confused by this PR — `BpkButton.scss` has been removed so it *should* be broken, right? But I'm running it locally and it seems fine. What's going on?

georgegillams:
> I'm really confused by this PR — `BpkButton.scss` has been removed so it _should_ be broken, right? But I'm running it locally and it seems fine. What's going on?

It's not used. When we refactored the button we must have forgotten to remove it

shaundon:
This will be breaking as you said, and it's going to be difficult to assess the impact of it with consumers. I'm wondering if instead of changing *all* link buttons, we should have an optional `padded` prop to set this, to make it non breaking?

It's not that I'm afraid of making breaking changes, I'm just thinking about how it would hamper adoption of the latest version

georgegillams:
> This will be breaking as you said, and it's going to be difficult to assess the impact of it with consumers. I'm wondering if instead of changing _all_ link buttons, we should have an optional `padded` prop to set this, to make it non breaking?
> 
> It's not that I'm afraid of making breaking changes, I'm just thinking about how it would hamper adoption of the latest version

Yeah I agree - I'll make that change.
Do you think it's best if we still change the mixin (not many people should be using that directly) but make the component non breaking?
Or should we offer a new mixin for padded button instead?

shaundon:
If it's easy to make a new mixin, I'd say do that, just on the off chance anybody is using it directly

packages/bpk-component-button/src/BpkButtonLink.scss
```diff
@@ -18,6 +18,12 @@
 
 @import '~bpk-mixins/index';
 
-.bpk-button--link {
-  @include bpk-button--link;
+.bpk-button {
+  &--link {
+    @include bpk-button--link;
+
+    &--padded {
```
shaundon, line 25: This violates BEM, we can't have a modifier on top of a modifier. If you change it to `&-padded` it's ok, because then the end result will be `--link-padded`. 

I know it's silly because it's a single character difference, but gotta be consistent with how we apply the rules!

.nvmrc
```diff
@@ -1 +1 @@
-lts/erbium
+12.13.0
```
shaundon, line 1: pourquoi?

shaundon (changes requested):

shaundon:
Have run locally and all good, just my small comment about BEM left

packages/bpk-component-button/src/BpkButtonLink.scss
georgegillams: Would it make more sense if it was `bpk-button bpk-button--link bpk-button--padded` as `padded` and `link` are both modifiers?

.nvmrc
georgegillams: Oh - github actions was failing because it was using the wrong version. Will revert if it's not needed 👍

packages/bpk-component-button/src/BpkButtonLink.scss
shaundon: Yeah that sounds perfect!

shaundon (approved):

shaundon (approved):

PR #2015, opened 2020-11-03
NO-JIRA: BpkNudger internal props usage adjustment

dominicfraser:
# Description 

In local development I see errors like:

```
Warning: Received `false` for a non-boolean attribute `outline`.

If you want to write it to the DOM, pass a string instead: outline="false" or outline={value.toString()}.

If you used to conditionally omit it with outline={condition && value}, pass outline={condition ? value : undefined} instead.
    in button (at BpkButtonBase.js:103)
    in BpkButton (at BpkButtonSecondary.js:41)
    in BpkButtonSecondary (at BpkConfigurableNudger.js:99)
    in div (at BpkConfigurableNudger.js:98)
    in BpkConfigurableNudger (at BpkNudger.js:48)
```

Poking around I think we are passing props unnecessarily.

Just above the changed lines we do:

```
const ButtonComponent =
    buttonType === 'secondary' ? BpkButtonSecondary : BpkButtonOutline;
```

This means at the point we pass in `outline` or `secondary` they are redundant; the type of button is already specific and no longer accepts these props.

# Checklist

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,11 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+
+ - bpk-component-nudger:
+   - Fixed use of internal props that generated console warnings due to their unneeded usage
+
 ## How to write a good changelog entry
 
 1. Add 'Breaking', 'Added' or 'Fixed' in bold depending on if the change will be major, minor or patch according to [semver](semver.org).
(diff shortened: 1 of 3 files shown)


olliecurtis (commented):
LGTM! Just going to run it locally to verify the changes!

dominicfraser:
❤️

olliecurtis (approved):
Tested locally and works as expected! Thanks for the contribution 🎉

PR #2017, opened 2020-11-10
[NO-JIRA] Neaten up CI workflow

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,11 +18,9 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: nvm install
-        shell: bash -l {0}
         run: nvm install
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -18,11 +18,9 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: nvm install
-        shell: bash -l {0}
         run: nvm install
 
       - name: nvm use
```
olliecurtis, line 23: You could probably go further and get rid of this step completely as we call nvm use before the install and build commands anyway so this is redundant

olliecurtis (approved):

PR #2020, opened 2020-11-17
[BPK-4286] Prevent navigation bar titles overlapping buttons

shaundon:
Previously we were using absolute positioning for the buttons, meaning the title was always centered but the buttons could overlap when the title got long.

Now we're using flexbox, and to work around the centering problem I've added a workaround involving cloning the accessory view when only one is provided. This does have a downside that when both accessory views are provided, if they're different widths, the title is off centre. Tbh this does suck but it feels like the least-worst solution I can think of, and as @georgegillams  pointed out, if it affects any consumers we can fall back to them setting a max width on their accessory view.

I did try using CSS grid but that doesn't help here either.

There is one other thing we could do but it doesn't seem great, which is to set a fixed size for the accessory views at all times, meaning the title would always be centered, but longer accessory views would wrap which doesn't seem great.

Anyway here's some screenshots

_Note how the fourth one down has accessory views of different widths so the title is slightly off centre. Also note how the second one from the bottom has a big leading padding — that's the invisible clone of the 'Done' button._
Screen Shot 2020-11-17 at 09 52 25

_This looks fine_
Screen Shot 2020-11-17 at 09 53 05

_This looks ok too, although note the large leading padding due to the cloned X button._
Screen Shot 2020-11-17 at 10 02 26

_This one isn't great tbh because 'Back to results' is so much longer than the X button._
Screen Shot 2020-11-17 at 09 55 45

_If we reduce it to just say 'Back' it's not as bad though._
Screen Shot 2020-11-17 at 09 57 32

### DIFF
--- a/packages/bpk-component-navigation-bar/src/BpkNavigationBar.scss
+++ b/packages/bpk-component-navigation-bar/src/BpkNavigationBar.scss
@@ -24,7 +24,8 @@
   min-height: $bpk-spacing-xxl + $bpk-spacing-xs;
   padding: $bpk-spacing-sm;
   flex-direction: row;
-  justify-content: center;
+  justify-content: space-between;
+  align-items: center;
 
   @include bpk-themeable-property(
     background-color,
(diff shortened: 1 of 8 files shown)


packages/bpk-component-navigation-bar/src/BpkNavigationBar.scss
```diff
@@ -41,32 +42,25 @@
...
-    right: $bpk-spacing-sm;
-
-    @include bpk-rtl {
-      right: auto;
-      left: $bpk-spacing-sm;
-    }
-  }
-
-  &__sticky {
+  &--sticky {
```
shaundon, line 49: This was incorrectly applying BEM before, as `sticky` is a modifier.

packages/bpk-component-navigation-bar/src/BpkNavigationBar.scss
```diff
@@ -41,32 +42,25 @@
...
+    &-leading {
+      @include bpk-margin-trailing($bpk-spacing-sm);
+    }
+    &-trailing {
+      @include bpk-margin-leading($bpk-spacing-sm);
+    }
+    &--invisible {
+      visibility: hidden;
+    }
+  }
```
shaundon, line 65: Refactored the `item` classes a little to make the CSS more readable.

PR #2021, opened 2020-11-17
Replace layout issue on BpkModal with a different one

shaundon:
Round 2 of my attempt to fix the overlapping modal!

This time I'm fixing it with custom CSS inside BpkModal and not attempting to change BpkNavigationBar. Technically we should refactor BpkNavigationBar to allow both leading and centered titles, but in the interest of time I think this is probably acceptable for the moment. We were already doing some custom CSS in BpkModal so I'm not _really_ introducing a hack, I'm just changing one we already had... that's what I'm telling myself anyway. And now you, reader.

_Regular modal_
Screen Shot 2020-11-17 at 16 10 40

_Long title_
Screen Shot 2020-11-17 at 16 10 48

_Slightly-broken one with an accessory view (those earlier #justflexboxthings problems still apply)_
Screen Shot 2020-11-17 at 16 10 56


UNRELEASED.md
```diff
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Fixed:**
```
georgegillams, line 5: Is this breaking? Consumers who were using `float` in CSS before will need to change it, right? 
```suggestion
**Breaking:**
```

shaundon: I don't think so — the button was absolutely positioned before, so `float` would have had no effect. Here's proof (this was recorded on `master`)

Kapture 2020-11-17 at 17 27 04

georgegillams (approved):

PR #2023, opened 2020-11-18
BPK-4032: Fix Modal a11y issue

olliecurtis:
A11y fix whereby aria-labelled by was incorrect when applied on the modals/dialogs

Before:
Screenshot 2020-11-18 at 13 59 56

After:
Screenshot 2020-11-18 at 13 59 14

### DIFF
--- a/packages/bpk-component-modal/src/BpkModalDialog.js
+++ b/packages/bpk-component-modal/src/BpkModalDialog.js
@@ -77,7 +77,6 @@ const BpkModalDialog = (props: Props) => {
   }
 
   const headingId = `bpk-modal-heading-${props.id}`;
-  const navigationId = `bpk-modal-navigation-${props.id}`;
 
   const accessoryViewFinal = props.accessoryView ? (
     <span className={getClassName('bpk-modal__accessory-view')}>
@@ -95,7 +94,7 @@ const BpkModalDialog = (props: Props) => {
         id={props.id}
         tabIndex="-1"
         role="dialog"
-        aria-labelledby={headingId}
+        aria-labelledby={props.showHeader ? headingId : null}
         className={classNames.join(' ')}
         ref={props.dialogRef}
       >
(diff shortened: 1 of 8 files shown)


packages/bpk-component-modal/src/BpkModalDialog.js
```diff
@@ -95,15 +94,15 @@ const BpkModalDialog = (props: Props) => {
...
+        aria-labelledby={props.showHeader ? headingId : null}
         className={classNames.join(' ')}
         ref={props.dialogRef}
       >
         {props.INTERNAL__outerComponent}
         {props.showHeader && (
           <header className={getClassName('bpk-modal__header')}>
             <BpkNavigationBar
-              id={navigationId}
+              id={headingId}
```
olliecurtis, line 105: We pass `headingId` instead of `navigationId` as the `id` prop is purely used for setting `aria-labelledby` in the Navigation component so it ensures the navbar can also target the correct title component

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -52,7 +52,10 @@ const BpkNavigationBar = (props: Props) => {
     ...rest
   } = props;
 
-  const titleId = `${id}-bpk-navigation-bar-title`;
+  // If the title is a component that sets its own id we want the aria-labelledby on the nav to match this so it can find the element
+  // Otherwise if its just a string we set the id on the title component.
+  const titleId =
+    typeof title === 'string' ? `${id}-bpk-navigation-bar-title` : id;
```
georgegillams, line 58: I'm not entirely clear on the reason for this type check?

olliecurtis: So if the the title passed is not a string and a component 

E.g. H2 or BpkText then we would already be passing an id which then when passed from something like the modal which has an `aria-labelledby` if we didn't use the id that was passed it would have the extra string appended to the id which would then mismatch with the `aria-labelledby` in the parent component so throws an a11y error.

Maybe that isn't too clear in the comment above, happy to have a call to discuss/clarify

georgegillams (approved):

PR #2035, opened 2020-11-24
[BPK-4299] Aria-live example tweaks

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:


shaundon (changes requested):
Looks good, but while you have this PR open could you also update the readme to say it supports all the other relevant ARIA props? I was gonna add it on the docs site page but it feels like it should go next to the props table.

shaundon (approved):
*chefs kiss*

PR #2037, opened 2020-11-26
[No Jira] Fix Flow in bpk-component-image

shaundon:


olliecurtis (approved):

UNRELEASED.md
```diff
@@ -2,6 +2,11 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+- bpk-component-image:
+  - Fixed an issue where `BpkBackgroundImageProps` weren't being exported correctly.
+  - Fixed an issue where Flow was not enabled for `bpk-component-image`.
```
olliecurtis, line 8: Would this still be 'Fixed' as we are now adding flow and functionailty to the component so may impact consumers here?

shaundon: Ah yeah good catch, I'll make it a minor

shaundon: Done

PR #2042, opened 2020-12-02
[BPK-4308] Fix issue with checkbox checks not always appearing

shaundon:
Previously, the checkmark's appearance was based on React seeing the `checked` prop, but the blue background was based on the CSS `:checked` selector. This meant that when using `element.checked = true` in JS, the background would become blue but because the checkmark didn't get added.

This PR changes the logic for the checkmark to also use the `:checked` selector, ensuring that it works as should.

I've attached a before and after gif to prove it works. A snapshot test can't capture this because it's about rendering, not markup.

**Before**
before

**After**
after


UNRELEASED.md
```diff
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+- bpk-component-checkbox:
+  - Fixed an issue where checked checkboxes didn't always display the check.
```
shaundon, line 7: Try and say this out loud ten times

packages/bpk-component-checkbox/src/BpkCheckbox.js
```diff
@@ -72,7 +72,6 @@ const BpkCheckbox = (props: Props) => {
   const inputClasses: string = getClassName(
     'bpk-checkbox__input',
     white && 'bpk-checkbox__input-white',
-    checked && !indeterminate && 'bpk-checkbox__input-checkmark',
```
shaundon, line 75: No longer needed because it's handled in the CSS.

packages/bpk-component-checkbox/src/BpkCheckbox.scss
```diff
@@ -44,38 +44,6 @@
...
-      }
-
-      &:disabled {
-        &::before,
-        &::after {
-          background-color: $bpk-color-sky-gray-tint-04;
-        }
-      }
-    }
-
```
shaundon, line 78: Moved to `_forms.scss`

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -611,6 +650,10 @@
     }
   }
 
+  &:checked:not(:indeterminate) {
+    @include bpk-checkbox__checkmark;
+  }
+
```
shaundon, line 656: `:not(:indeterminate)` needed so that indeterminate still takes precedence over checked.

UNRELEASED.md
olliecurtis: Yep I failed at this 😆

olliecurtis (approved):
LGTM!

PR #2046, opened 2020-12-07
DINGO 3759: Upgrade some deps to use latest RTG

igunsaya:
The apparent breaking changes for `react-transition-group` upgrade is
- `3.x.x` - use new style react context - requires react v16.6 or above
- `4.x.x` - in environments where esm is supported importing from commonjs requires explicitly adding the .default after require() when resolving to the esm build

More info here https://github.com/reactjs/react-transition-group/blob/master/CHANGELOG.md

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests

React 16.4 compatibility:

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 > Place your changes below this line.
-**Breaking**
+**Fixed**
 
  - bpk-component-banner-alert
    - Upgrade `react-transition-group` dependency to `4.4.1`
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,6 +1,16 @@
 # Unreleased
 
 > Place your changes below this line.
+**Breaking**
+
+ - bpk-component-banner-alert
+   - Upgrade `react-transition-group` dependency to `4.4.1`
+ - bpk-component-drawer
+   - Upgrade `react-transition-group` dependency to `4.4.1`
+ - bpk-component-image
+   - Upgrade `react-transition-group` dependency to `4.4.1`
+ - bpk-react-utils
+   - Upgrade `react-transition-group` dependency to `4.4.1`
 
 ## How to write a good changelog entry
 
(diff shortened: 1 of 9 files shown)


UNRELEASED.md
```diff
@@ -1,6 +1,16 @@
 # Unreleased
 
 > Place your changes below this line.
+**Breaking**
```
olliecurtis, line 4: Unless this has breaking changes to our consumers and how the components work, we can probably make this a patch. As breaking for us means it requires action on the consumer or visually breaking 😄

olliecurtis (changes requested):
LGTM - one minor comment and just going to run it locally!

Thanks 🎉

UNRELEASED.md
igunsaya: Ahh got you. I thought breaking would apply because they were major dep upgrade :-)

olliecurtis:
@igunsaya - Just a quick question do you think we would see an issue with upgrading this and consumers using this version if the react context has changed in v3 and now requires `16.6` of react, when the version we use in prod being `16.4.2`? 🤔

igunsaya:
Good question. I'm checking the dependencies of the `react-transition-group` (in lock file) and don't see any hard `react` dependency. Digging a little further, it looks like they use newer versions of react as devDeps. I assume this means that they themselves use it. Not sure why they would mark it as a major version. Let me dig a little deeper. Atm, I doesn't look like an issue

olliecurtis:
So tried it locally changing the react version down and it seems that it breaks running storybook and rendering the components as it seems this latest versions are using hooks which is not compatible with older versions 😞

Screenshot 2020-12-07 at 16 44 06

igunsaya:
Closing this due to fact that the react peerDependencies of react transition group v3 is react v16.6.0. However, the version most of the company uses is react v16.4.2. context/hooks wont work

PR #2047, opened 2020-12-08
Upgrading storybook

olliecurtis:
This PR upgrades storybook to the latest version. 

You'll notice here that we are using the alpha release here this is due to a babel issue that exists within storybook, which babel are working to solve but SB have already fixed (with a workaround) in the next `6.2` release, so in order not to block us, add any hacks and after e2e testing it is safe for us to proceed wiht the alpha version and once released will move to the stable 6.2

You'll notice that the only storybook add-on not upgraded is the `addon-a11y` - this is due to an issue that when storybook is running the addon constantly is stuck in an `Initalising` state so never properly loads. 

I think this maybe due to an issue with our current storybook config setup which we need to migrate to the new style but will be done in a separate PR as requires a lot more changes.


package.json
```diff
@@ -108,20 +108,20 @@
...
   "devDependencies": {
-    "@babel/core": "^7.8.4",
-    "@babel/plugin-proposal-class-properties": "^7.8.3",
-    "@babel/plugin-proposal-object-rest-spread": "^7.8.3",
-    "@babel/plugin-transform-runtime": "^7.8.3",
-    "@babel/preset-env": "^7.4.4",
-    "@babel/preset-flow": "^7.8.3",
-    "@babel/preset-react": "^7.8.3",
-    "@babel/register": "^7.4.4",
+    "@babel/core": "^7.12.9",
```
olliecurtis, line 111: Babel was upgraded in this PR in order to be compatible with 6.1+

shaundon (approved):
Have run locally, all good

georgegillams (approved):

PR #2049, opened 2020-12-09
[BPK-3976] Add skip link component

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:

## This is what I'm thinking for docs page

Skip links mostly benefit sighted keyboard/switch users by providing them with a way to skip over lengthy menus or filters.
Skip links should be absolute-positioned to ensure that when they become visible they do not shift other content. It’s ok for a skip link to overlap other content on the page, as long as you are still able to tab forwards/backwards to that content.

The skip link is very noticeable, and doesn’t disappear immediately. These are deliberate design decisions to ensure that a user tabbing quickly through the page won’t miss the skip link.

https://webaim.org/techniques/skipnav/


packages/bpk-component-skip-link/src/BpkSkipLink-test.js
```diff
@@ -0,0 +1,42 @@
...
+/* @flow strict */
+
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import BpkSkipLink from './BpkSkipLink';
+
+describe('BpkSkipLink', () => {
+  it('should render correctly', () => {
+    const tree = renderer.create(<BpkSkipLink />).toJSON();
```
olliecurtis, line 27: The docs say href and label are required as props yet we are not passing them here, is that expected?

packages/bpk-component-skip-link/src/BpkSkipLink.scss
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+@import '~bpk-mixins/index';
+
+.bpk-skip-link {
+  display: inline-block;
+  padding: $bpk-spacing-sm;
+  transition: none;
+  border-radius: $bpk-button-border-radius;
+  background-color: $bpk-color-panjin;
```
olliecurtis, line 26: Haven't seen the design spec but do we possibly need this to support theming? Just from a partner branding perspective so that it wouldn't clash?

olliecurtis (changes requested):
LGTM just a couple of queries - going to try it locally now

olliecurtis:
Will we also be releasing it after this is merged? As we will probably need a UNRELEASED entry too right?

olliecurtis (commented):
Run locally and all seems to be working and fine as expected - one feedback has been addressed happy to approve it!

olliecurtis (approved):
LGTM!

PR #2052, opened 2020-12-15
[BPK-4333] Fix popover input focus

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,9 +1,10 @@
 # Unreleased
 
 > Place your changes below this line.
-**Fixed:**
- - bpk-component-skip-link:
-   - Component now works with no client-side JS.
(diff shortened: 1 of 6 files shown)


georgegillams:
After: https://backpack.github.io/storybook-prs/2052/iframe.html?id=bpk-component-popover--alongside-input&viewMode=story

 - Using escape or the close button should cause focus to return to the button
 - Tabbing should not be possible when the popover is open (note that you can still tab outside the iframe though, so view outside the storybook UI)
 - Clicking on the input or theme selector should close the popover, but otherwise work as expected

shaundon (approved):

PR #2058, opened 2020-12-18
[BPK-4341]: Update banner alert styles

olliecurtis:
This PR updates the banner alert to the new design.

However this currently needs design review in regards to the expandable implementation to decide on the hover and active states colours to ensure they are accessible and work for the colours


packages/bpk-component-banner-alert/stories.js
```diff
@@ -128,4 +203,50 @@ storiesOf('bpk-component-banner-alert', module)
       type={ALERT_TYPES.SUCCESS}
       icon={CurrencyIcon}
     />
+  ))
+  .add('All expandable', () => (
```
olliecurtis, line 207: Added this purely for debugging and testing out the styles will be removed when implementation of state colours is complete

packages/bpk-tokens/src/web/base/notifications.json
```diff
@@ -6,7 +6,7 @@
     "category": "notifications"
   },
   "props": {
-    "BANNER_ALERT_BACKGROUND_COLOR": {
+    "BANNER_ALERT_COLOR": {
```
olliecurtis, line 9: As we now use solid colours for the background we don't need the token to be background but instead for text

packages/bpk-component-banner-alert/stories.js
```diff
@@ -47,10 +47,40 @@ const BpkBannerAlertExpandableState = withBannerAlertState(
 
 storiesOf('bpk-component-banner-alert', module)
   .add('Neutral', () => (
-    <BpkBannerAlert message={message} type={ALERT_TYPES.NEUTRAL} />
+    <div>
```
olliecurtis, line 50: Added group examples to each story to show what they look like in each variant and during testing of code updates.

Will resolve the styling later once we have full colour sign-off

olliecurtis (commented):
Tests are expected to fail here as the colours are all still in WIP so don't want to edit them given they will change later once styles for expandable states confirmed

PR #2059, opened 2021-01-03
[NO-JIRA] Update license headers 2021

georgegillams:

- [ ] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests

React 16.4 compatibility:


olliecurtis:
Just a tip in future to prevent us from having to do this every year we can just update the year to be -present many OSS projects follow this and makes it easier so we don't have to remember in the future 🙂

PR #2060, opened 2021-01-08
[NO-JIRA] Fix column header sorting icons in datatable

jorgedc93:
The arrow icons being used in the datatable for sorting were inverted

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [x] Tests

React 16.4 compatibility:

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Fixed:**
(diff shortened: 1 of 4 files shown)


georgegillams (approved):
LGTM 👍
I don't think the README needs to change, and docs will be updated automatically by a version bump after this is released. So I'm happy to merge and release once CI passes

PR #2061, opened 2021-01-12
[BPK-4343]: Add semantic colors for web

olliecurtis:
This PR adds all the semantic colours for web to mirror that we have for our mobile platforms.

As part of this PR the svg generation script has been changed so as to ignore creating svgs for colours that are used for text and line colours which are not applicable for the icons.

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:


packages/bpk-svgs/gulpfile.babel.js
```diff
@@ -54,6 +54,15 @@ const largeIconPxSize = remToPx(largeIconSize);
 
 const colors = _(tokens.props)
   .filter({ category: 'colors', type: 'color' })
+  .filter(color => {
+    // We don't want to generate SVGs for colors that begine with Background or Line as these are not meant for icons.
+    // We want it for raw colors only.
+    if (color.name.startsWith('BACKGROUND') || color.name.startsWith('LINE')) {
+      return;
```
georgegillams, line 61: I guess undefined is falsy? Should we be more explicit and return false?

packages/bpk-tokens/src/web/base/colors.json
```diff
@@ -5,6 +5,78 @@
     "category": "colors"
   },
   "props": {
+    "LINE_LIGHT_COLOR": {
+      "value": "{!SKY_GRAY_TINT_05}"
+    },
+    "LINE_DARK_COLOR": {
+      "value": "{!BLACK_TINT_04}"
+    },
```
georgegillams, line 13: Should we expose dark mode colours? We don't support DM on web so in theory no one should be using them 🤔

olliecurtis: Isn't that though how our semantic colours are created? It takes both a light and dark colour to generate it or have I got that wrong?

https://backpack.github.io/guidelines/colour/#dark-mode-dynamic-colours

packages/bpk-svgs/gulpfile.babel.js
olliecurtis: Yep that sounds good! Means I can also remove the eslint disable rule just below it too

packages/bpk-tokens/src/web/base/colors.json
georgegillams: Semantic just means that the alias has a specific meaning. eg `backgroundColor`, `dangerColor`, `successColor`, `calendarLowPriceColor`.
The semantic colours for app are mostly dynamic (light and dark pairs) because the apps support light and dark mode, but on web where we only support light the semantic colours should probably just have one variant

packages/bpk-tokens/src/web/base/colors.json
```diff
@@ -110,6 +131,27 @@
...
+      "value": "{!BLACK_TINT_03}"
+    },
+    "COLOR_BLACK_TINT_04": {
+      "value": "{!BLACK_TINT_04}"
+    },
+    "COLOR_BLACK_TINT_05": {
+      "value": "{!BLACK_TINT_05}"
+    },
+    "COLOR_BLACK_TINT_06": {
+      "value": "{!BLACK_TINT_06}"
```
georgegillams, line 154: These should never be used in LM AFAIK?

olliecurtis: Yup you are right shall remove them!

georgegillams (approved):
Noice!

PR #2062, opened 2021-01-12
[NOJIRA] Proxy custom sort function to react-virtualized Table

jorgedc93:
Adds props `sort`, `sortBy` and `sortDirection` that when present leave all the sorting to the underlying react-virtualized `Table`

- [ ] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests

React 16.4 compatibility:


jorgedc93:
I was not sure if updating the `README` was needed since it already says:
`Supports all properties defined in [Table](https://github.com/bvaughn/react-virtualized/blob/master/docs/Table.md) (from react-virtualized)`

And this is adding props that are in that file

smrpr (approved):

olliecurtis:
Hey @jorgedc93 - As we are now exposing these props to consumers and adding them explicitly I would say yes we should update the README to reflect this 😄

jorgedc93:
Hey @olliecurtis, I just updated the README with the new props and an example.

packages/bpk-component-datatable/src/BpkDataTable-test.js
```diff
@@ -222,6 +223,99 @@ describe('BpkDataTable', () => {
...
+      {
+        name: 'John',
+        description: 'Some other guy',
+        seat: { office: 'Barcelona', desk: 15 },
+      },
+    ];
+    let sortByValue = 'seat';
+    let sortDirectionValue = 'DESC';
+    const sortFunction = ({ sortBy, sortDirection }) => {
+      if (sortBy === 'seat') {
```
olliecurtis, line 247: We could probably make this sortFunction a little short as you are never testing pressing the sort in any other way so this else would never get tested.

Unless we want to add another part of the test which is to click another column in the rendered output to really test this.

packages/bpk-component-datatable/src/BpkDataTable-test.js
```diff
@@ -222,6 +223,99 @@ describe('BpkDataTable', () => {
...
+        complexRows = _sortBy(complexRows, [
+          row => row.seat.office,
+          row => row.seat.desk,
+        ]);
+      } else {
+        complexRows = _sortBy(complexRows, sortBy);
+      }
+      if (sortDirection === 'DESC') {
```
olliecurtis, line 255: Similar comment as above here

packages/bpk-component-datatable/README.md
```diff
@@ -55,6 +131,9 @@ in addition to the following:
 | headerHeight           | number                      | false    | 60                   |
 | rowHeight              | number                      | false    | 60                   |
 | defaultColumnSortIndex | number                      | false    | 0                    |
+| sort                   | func                        | false    | null                 |
+| sortBy                 | string                      | false    | null                 |
+| sortDirection          | oneOf('ASC', 'DESC')        | false    | null                 |
```
olliecurtis, line 136: Its mentioned sortDirection its type is `oneOf('ASC', 'DESC')` but the prop type defined in the implementation just accepts a string. 

So probably if we are restricting it to be one of these two values we need to make the prop explcit in the implementation

so in the `propTypes` declaration in the main file set it to 

`sortDirection: PropTypes.oneOf('ASC', 'DESC')`

Left the suggestion below

k0nserv (commented):
I feel like the complexity of this component has grown quite a bit, especially with added conditionals everywhere.

Can I suggest that we add a tiny bit of abstraction to make this cleaner? I'm thinking something like a `Sorter` abstraction with two implementations, one that is the previous naive behaviour and another with the custom sort function. This way we'd only need to keep the sorter in the state and we can use polymorphism to get rid of most conditions

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -212,6 +236,9 @@ BpkDataTable.propTypes = {
   headerHeight: PropTypes.number,
   className: PropTypes.string,
   defaultColumnSortIndex: PropTypes.number,
+  sort: PropTypes.func,
+  sortBy: PropTypes.string,
+  sortDirection: PropTypes.string,
```
olliecurtis, line 241:
```suggestion
  sortDirection: PropTypes.oneOf('ASC', 'DESC'),
```

olliecurtis (changes requested):
Few review comments. We should have a story for this to test and show it in action in `stories.js` for our storybook

packages/bpk-component-datatable/src/BpkDataTable.js
olliecurtis: ⬆️ this is because in the README we are choosing to be explict about the valid values

jorgedc93:
> I feel like the complexity of this component has grown quite a bit, especially with added conditionals everywhere.
> 
> Can I suggest that we add a tiny bit of abstraction to make this cleaner? I'm thinking something like a `Sorter` abstraction with two implementations, one that is the previous naive behaviour and another with the custom sort function. This way we'd only need to keep the sorter in the state and we can use polymorphism to get rid of most conditions

Extracted all the sorting related code to a couple Sorters, let me know what you think @k0nserv

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -177,14 +116,15 @@ class BpkDataTable extends Component {
...
-        rowGetter={({ index }) => sortedList[index]}
+        rowCount={this.state.sorter.rowCount}
+        rowGetter={({ index }) => this.state.sorter.getRow(index)}
         headerClassName={headerClassNames.join(' ')}
         rowClassName={row => this.rowClassName(rowClassName, row)}
         onRowClick={this.onRowClicked}
         onHeaderClick={this.onHeaderClick}
-        sortBy={sortBy}
-        sortDirection={sortDirection}
+        sortBy={this.state.sorter.sortProps.sortBy}
```
k0nserv, line 125: This part could be something like

```javascript
{...this.state.sorter.sortProps()}
```

and we could move the repsonsibility of providing `sort`, `sortDirection`, and `sortBy` to the sorter

packages/bpk-component-datatable/src/sorter.js
```diff
@@ -0,0 +1,118 @@
...
+      SortDirection.ASC;
+    this.sortProps = {
+      sortBy,
+      sortDirection,
+    };
+    this.list = props.rows;
+    this.sort();
+  }
+
+  getRow = index => this.sortedList[index];
```
k0nserv, line 41: This is subtley different from

```javascript
getRow(index) {
    return this.sortedList[index];
}
```

With the current way of writing this each instance of the class will delcare and allocate it's own copy of the function. With the second way of doing it each instance of the class will share the function via the class's [prototype chain](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain). 

There are a few cases where the style of declaration you've used is desirable, for example when the function needs to be bound to `this`, but in this instance we should prefer the second alternative for all these methods.

k0nserv (commented):
Thanks @jorgedc93, it's greatly improved with the addition of the sorter abstraction. Just a few tiny comments

packages/bpk-component-datatable/src/sorter.js
jorgedc93: Thanks, good to know about this. I just pushed a new commit with these changes

packages/bpk-component-datatable/src/sorter.js
```diff
@@ -0,0 +1,122 @@
...
+    this.list = props.rows;
+    this.rowCount = props.rows.length;
+  }
+
+  getRow(index) {
+    return this.list[index];
+  }
+
+  /* eslint-disable-next-line class-methods-use-this */
+  sort() {}
```
k0nserv, line 95: Is this part of the public interface of sorters? If not we can just omit it in this class imo

jorgedc93: Yeah, you're right, I think with the last refactor it's only called internally, removed it for this one.

k0nserv (approved):
LGTM

k0nserv:
Merging despite CI failing because it's only the last part with the docs preview that is failing

k0nserv:
Thanks for the contribution @jorgedc93

PR #2064, opened 2021-01-14
[No JIRA] Flow type bpk-component-datatable

k0nserv:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

React 16.4 compatibility:

### DIFF
--- a/packages/bpk-component-datatable/src/BpkDataTable.js
+++ b/packages/bpk-component-datatable/src/BpkDataTable.js
@@ -15,6 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* @flow strict */
 
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
@@ -25,13 +26,31 @@ import _omit from 'lodash/omit';
 import STYLES from './BpkDataTable.scss';
 import BpkDataTableColumn from './BpkDataTableColumn';
 import hasChildrenOfType from './hasChildrenOfType';
-import makeSorter from './sorter';
+import makeSorter, { type Sorter } from './sorter';
+import type { Props } from './common-types';
 
 const getClassName = cssModules(STYLES);
 const omittedTableProps = ['rowGetter', 'rowCount', 'onHeaderClick'];
 
-class BpkDataTable extends Component {
-  constructor(props) {
+type State<Row> = {
+  sorter: Sorter<Row>,
+  rowSelected: ?number,
+};
+
+class BpkDataTable<Row> extends Component<Props<Row>, State<Row>> {
+  static defaultProps = {
+    ...Table.defaultProps,
+    width: null,
+    headerHeight: 60,
+    rowHeight: 60,
+    gridStyle: { direction: undefined }, // This is required for rows to automatically respect rtl
+    defaultColumnSortIndex: 0,
+    sort: null,
+    sortBy: null,
+    sortDirection: null,
+  };
+
+  constructor(props: Props<Row>) {
     super();
 
     this.state = {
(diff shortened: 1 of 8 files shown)


olliecurtis (approved):
LGTM!

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -25,13 +26,31 @@ import _omit from 'lodash/omit';
...
 const omittedTableProps = ['rowGetter', 'rowCount', 'onHeaderClick'];
 
-class BpkDataTable extends Component {
-  constructor(props) {
+type State<Row> = {
+  sorter: Sorter<Row>,
+  rowSelected: ?number,
+};
+
+class BpkDataTable<Row> extends Component<Props<Row>, State<Row>> {
```
olliecurtis, line 40: Should we have a `type Props` object similar to `type State` where we are checking the props passed that are explicitly called out in the props table https://github.com/Skyscanner/backpack/tree/master/packages/bpk-component-datatable or is it sufficiently covered by the propTypes declaration at the bottom of this file?

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -57,24 +76,34 @@ class BpkDataTable extends Component {
...
+      return;
+    }
+
     if (column.props.disableSort === true) {
       return;
     }
 
     // See: https://reactjs.org/docs/events.html#event-pooling
-    const eventTarget = event.target;
+    const eventTarget = event.currentTarget;
```
k0nserv, line 99: This is an actual code change, it seems like the correct option however

packages/bpk-component-datatable/src/bpkHeaderRenderer.js
```diff
@@ -35,23 +37,35 @@ const downIconClassName = getClassName(
   'bpk-data-table-column__sort-icon--down',
 );
 
-export const getSortIconDirection = element => {
+export const getSortIconDirection = (element: Element): ?SortDirectionType => {
   if (
     hasClassName(element, upIconClassName) ||
-    hasClassName(element.parentNode, upIconClassName)
+    hasClassName(element.parentElement, upIconClassName)
```
olliecurtis, line 43: Is this change due to the type that we are assigning `element`?

packages/bpk-component-datatable/src/BpkDataTable.js
olliecurtis: What is the thinking behind it being the correct option here?

jorgedc93 (approved):

packages/bpk-component-datatable/src/bpkHeaderRenderer.js
k0nserv: Yes, The type of `parentNode` is `Node` whereas the type of `parentElement` is `Element` we need it to be element because that's where we get `getAttributes` from which is used by `hasClassName`. `parentNode` and `parentElement` is the exact same instance here so this is fine

packages/bpk-component-datatable/src/BpkDataTable.js
k0nserv: > Note: To get the element instance, like HTMLButtonElement in the example above, it is a common mistake to use event.target instead of event.currentTarget. The reason why you want to use event.currentTarget is that event.target may be the wrong element due to event propagation.

From the Flow docs, https://flow.org/en/docs/react/events/

k0nserv: `event.target` isn't typed correctly so it makes things difficult

k0nserv: I suppose we could verify that `event.target` is an `Element` rather than `Document` or `Window` which are the two other options, but this change seems like the more correct option anyuway

olliecurtis (approved):

PR #2065, opened 2021-01-14
[BPK-4358]: Autosuggest fix for compatibility

olliecurtis:
**Changes** 
- Downgrade react-autosuggest to v9 
- Downgrade react version to be compatible with change and align with company standard
- Downgraded storybook as the newer version will not work with old react version

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests


.github/pull_request_template.md
```diff
@@ -9,7 +9,3 @@ Remember to include the following changes:
 - [ ] `README.md`
 - [ ] Tests
 - [ ] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)
-
-React 16.4 compatibility:
```
olliecurtis, line 13: Removed this as we are no longer using a version of React that supports hooks

packages/bpk-component-autosuggest/package.json
```diff
@@ -18,7 +18,7 @@
     "bpk-mixins": "^20.1.17",
     "bpk-react-utils": "^3.0.11",
     "prop-types": "^15.6.2",
-    "react-autosuggest": "^10.0.2"
+    "react-autosuggest": "^9.4.3"
```
georgegillams, line 21: What's to prevent this being upgraded again in the future?
Could we write a test that verifies the version. The test can also serve as documentation with a comment to say why we're testing it 👍

olliecurtis: That is a good shout will add that now!

georgegillams: You can use `import packageJson from './package.json';` and then just read the value, and assert that it's `"^9.4.3"`

packages/bpk-component-autosuggest/check-autosuggest-version-test.js
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow strict */
+
+import packageFile from './package.json';
+
+describe('Check version of react-autosuggest', () => {
+  it('it should not be v10 or higher due to compatibility issues with older react versions', () => {
```
georgegillams, line 24:
```suggestion
  it('it should not be v10 or higher due to compatibility issues with react version < 16.3', () => {
```

scripts/npm/check-react-versions.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+});
+
+if (reactVersionErrors.length > 0) {
+  console.log('\nThere were React version mismatches');
+  console.log(reactVersionErrors.join('\n'));
+  process.exit(1);
+}
+
+console.log('\nAll react versions match 👍');
```
georgegillams, line 44: https://www.youtube.com/watch?v=ki6v0-fscJI

georgegillams (approved):

PR #2067, opened 2021-01-15
NEON-167 DataTable rowClassName can be a function

jorgedc93:
While working on a feature to dynamically change the background of a row depending on a value, I noticed that `BpkDataTable` did not allow to pass a function for `rowClassName` while the original react-virtualized `Table` [does](https://github.com/bvaughn/react-virtualized/blob/master/docs/Table.md).

This PR extends the component to allow passing a function as in `Table`

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,11 @@
 
 > Place your changes below this line.
 
+**Added:**
+
+- bpk-component-datatable:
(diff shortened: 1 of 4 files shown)


olliecurtis (approved):
LGTM! Just going to run it locally

PR #2072, opened 2021-01-28
[BPK-4371] Move accordion component to new stories format

shaundon:
We should form a convention of how we do this. My proposal is to have `examples.js` where the examples, agnostic to any particular technology, can live, then in `stories.js` we simply import and use them. 

This way we're not bound to importing `stories.js` on the docs site, as the docs site shouldn't have to have any knowledge of Storybook. Similarly, in future we might wish to import `examples.js` into our Jest tests.


packages/bpk-component-accordion/stories.js
```diff
@@ -18,304 +18,31 @@
...
+  )
+  .add('Multiple items open', MultipleItemsOpenExample)
+  .add(
+    'Multiple items open (second & third items initially expanded)',
+    MultipleItemsOpenInitiallyExpandedExample,
+  )
+  .add('Custom', CustomExample)
+  .add('Custom title textStyle', CustomTitleTextStyleExample)
+  .add('With icons', WithIconsExample)
+  .add('With bold titles', WithBoldTitlesExample);
```
georgegillams, line 48: I wonder if we could use a helper that does something similar to CSF where we give it a list of components and it automatically creates the stories from them.
In other words we give it `SingleItemOnly` and it infers the story name `Single item only`.

Maybe yagni?

shaundon: Yeah I was wondering about something like this too actually. The advantage would be that we could build something to automatically create the stories based on whatever is in `examples.js` so we wouldn't even need to have all these `stories.js` files.

It's a nice idea but it'd be a lot more work so I reckon _for now_ we call it out of scope and we might come back to it later?

shaundon:
@georgegillams can I have ✅ if you're good with it?

georgegillams (approved):

olliecurtis (approved):

PR #2082, opened 2021-02-11
[BPK-4379] Add bpk-storybook-utils

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-storybook-utils/src/BpkStorybookUtils.js
+++ b/packages/bpk-storybook-utils/src/BpkStorybookUtils.js
@@ -15,15 +15,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 /* @flow strict */
 
-import React from 'react';
-import { storiesOf } from '@storybook/react';
-import { action } from 'bpk-storybook-utils';
+/* eslint-disable global-require, no-console, import/no-mutable-exports */
+let action = (...args: Array<any>) => {
+  return () => console.info(args);
+};
 
-import BpkCloseButton from './index';
+try {
+  const storybookAction = require('@storybook/addon-actions').action;
+  if (storybookAction) {
+    action = storybookAction;
+  }
+} catch (e) {
+  console.info(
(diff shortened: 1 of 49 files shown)


packages/bpk-storybook-utils/src/BpkStorybookUtils.js
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* @flow strict */
+
+/* eslint-disable global-require, no-console, import/no-mutable-exports */
+let action = (...args: Array<any>) => {
+  return () => console.log(args);
+};
```
shaundon, line 23: Not sure if this would work, but maybe worth a try for simplification purposes?

```suggestion
let action = console.log
```

packages/bpk-storybook-utils/src/BpkStorybookUtils.js
```diff
@@ -0,0 +1,37 @@
...
+
+try {
+  const storybookAction = require('@storybook/addon-actions').action;
+  if (storybookAction) {
+    action = storybookAction;
+  }
+} catch (e) {
+  console.log(
```
shaundon, line 31: How about using `console.info` here instead? It'll make it easier to spot this message and it feels better semantically. 

Screen Shot 2021-02-12 at 11 05 26


```suggestion
  console.info(
```

packages/bpk-storybook-utils/src/BpkStorybookUtils.js
georgegillams: No because the way it's used is `onClick={action('...')}
If we use `let action = console.log` then the logging will happen when the component is rendered, not when `onClick` happens

shaundon (approved):

PR #2089, opened 2021-02-17
[NEON-160] Fix datatable sortDirection

alexandraharet:
This PR fixes a bug introduced when Flow type was added to the component [here](https://github.com/Skyscanner/backpack/pull/2064/files#diff-a7c17bd431e0a4bdaab2b571688246b564385f1d8de2fa152319d8d68339eb0aR99).

`event.currentTarget` will contain the element that actually has the event listener, but in this case we need to detect clicks on inner elements (and leverage event propagation), to keep the `sortDirection` functionality assigned to the Up / Down arrows in the columnHeaders - see Before / After screen recording below.

The PR fixes this bug, and adds new unit tests that check the correct behaviour of click events on these icons, to prevent future accidental bugs of similar nature. 

**Before:**
https://user-images.githubusercontent.com/18306755/108337699-af6c1180-71d5-11eb-9497-2fd339df750f.mp4

**After:**
https://user-images.githubusercontent.com/18306755/108337835-d88ca200-71d5-11eb-8b29-6259e047be24.mp4

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [x] Tests


packages/bpk-component-datatable/src/bpkHeaderRenderer.js
```diff
@@ -37,16 +37,24 @@ const downIconClassName = getClassName(
   'bpk-data-table-column__sort-icon--down',
 );
 
-export const getSortIconDirection = (element: Element): ?SortDirectionType => {
+export const getSortIconDirection = (
+  eventTarget: EventTarget,
+): ?SortDirectionType => {
+  /* Flow does not work well with EventTarget type, but we need to use it in this case, because we need
```
k0nserv, line 43: This is false AFAIK, Flow is correct and this code has a bug. 

[`EventTarget`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) does not actually have a `parentElement` property which is why Flow is complaining.  In particular, all of `Window`, `Document`, and `Element` are examples of types that implement the `EventTarget` interface. However, `parentElement` is a property on `Node` which `Document` and `Element` are subclasses of but `Window` *is not*.

This code used `Element` precisely because it solves this problem as all `Element`s have a `parentElement` property

packages/bpk-component-datatable/src/utils.js
```diff
@@ -18,12 +18,20 @@
...
     return false;
   }
 
-  const nodeClassName = node.getAttribute('class');
+  /* Flow does not work well with EventTarget type, but we need to use it in this case, because we need
+   * event propagation; the eventHadler is placed on the outer Header element, but we want to know when a
+   * click has happened on the inner 'arrowUp' or 'arrowDown' icons, in order to apply the correct sortDirection.
+   */
+  // $FlowFixMe[prop-missing] - see above
+  const nodeClassName = eventTarget.getAttribute('class');
```
k0nserv, line 34: This is wrong for the same reason as above, Flow is correct

k0nserv (changes requested):
I don't understand why we needed to change the use of `Element` to `EventTarget` at all? `Element` is a super set of `EventTarget` after all.

packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -95,8 +95,8 @@ class BpkDataTable<Row> extends Component<Props<Row>, State<Row>> {
       return;
     }
 
-    // See: https://reactjs.org/docs/events.html#event-pooling
-    const eventTarget = event.currentTarget;
+    // See: https://reactjs.org/docs/legacy-event-pooling.html
+    const eventTarget = event.target;
```
k0nserv, line 99: It looks like I introduced this bug FWIW. See https://github.com/Skyscanner/backpack/pull/2064/files#r557349473. I think what we should do is my suggestion from there i.e. checking the type to ensure it's `Element`(which it will always be)

k0nserv:
```diff
diff --git a/packages/bpk-component-datatable/src/BpkDataTable.js b/packages/bpk-component-datatable/src/BpkDataTable.js
index 116ff92fe..8a4195565 100644
--- a/packages/bpk-component-datatable/src/BpkDataTable.js
+++ b/packages/bpk-component-datatable/src/BpkDataTable.js
@@ -96,11 +96,13 @@ class BpkDataTable<Row> extends Component<Props<Row>, State<Row>> {
     }
 
     // See: https://reactjs.org/docs/events.html#event-pooling
-    const eventTarget = event.currentTarget;
+    const eventTarget = event.target;
 
-    this.setState(prevState => ({
-      sorter: prevState.sorter.onHeaderClick(sortBy, eventTarget, column),
-    }));
+    if (eventTarget instanceof HTMLElement) {
+      this.setState(prevState => ({
+        sorter: prevState.sorter.onHeaderClick(sortBy, eventTarget, column),
+      }));
+    }
   };
 
   rowClassName = (consumerClassName: ?string, { index }: { index: number }) => {
```

This is the change that I think is correct and also pleases Flow

UNRELEASED.md
```diff
@@ -28,3 +29,5 @@ See [`CHANGELOG.md`](CHANGELOG.md) for real-world examples of good changelog ent
 
 - bpk-component-horcrux:
   - Fixed issue where `BpkHorcrux` would occasionally possess the living.
+- bpk-component-datatable:
```
k0nserv, line 32: This should be above under # Unreleased

packages/bpk-component-datatable/src/bpkHeaderRenderer.js
alexandraharet: You're right, done!

UNRELEASED.md
alexandraharet: Done

k0nserv (approved):

PR #2090, opened 2021-02-18
Excluding storybook from deploying if on a fork

olliecurtis:
In order to deploy the storybook to our PR site it requires the use of secret tokens that are not available to forks, so the purpose of this PR is to prevent this step running on forks and causing builds to fail at this step.

How to verify change:

- [x] Test on this branch to make sure CI still works
- [x] Create a fork and test CI to make sure it works as expected - see #2092

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,13 +72,13 @@ jobs:
           mkdir build
           mv dist-storybook $PR_NUMBER
           cp -R $PR_NUMBER build/
-        if: github.ref != 'refs/heads/main'
+        if: github.ref != 'refs/heads/main' && github.repository == github.event.pull_request.head.repo.full_name
         env:
           PR_NUMBER: ${{ github.event.pull_request.number }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -72,13 +72,13 @@ jobs:
           mkdir build
           mv dist-storybook $PR_NUMBER
           cp -R $PR_NUMBER build/
-        if: github.ref != 'refs/heads/main'
+        if: github.ref != 'refs/heads/main' || github.repository == github.event.pull_request.head.repo.full_name
```
georgegillams, line 75: Surely we should check the build works even if we won't deploy it?

olliecurtis: This is already done in a previous step as if you look at this step its preparing to be deployed not building and just copies files from one directory to another

georgegillams: Should this logic be `branch is not main AND is PR from repo`?

olliecurtis: Ah yes good shout!

georgegillams (approved):

PR #2095, opened 2021-02-22
[BPK-4415] Disabled chips are no longer pressable

shaundon:


olliecurtis (commented):
LGTM! Once PR is deployed will check and approve!

packages/bpk-mixins/src/mixins/_chips.scss
```diff
@@ -35,7 +35,7 @@
       background-color: $bpk-color-sky-gray-tint-07;
     }
 
-    &:active {
+    &:active:not(:disabled) {
```
georgegillams, line 38: Another way to achieve this would be to explicitly set the disabled background colour as disabled style takes precedence over active style.

I wonder if the cursor should be `not-allowed` for disabled ones?

shaundon: > I wonder if the cursor should be not-allowed for disabled ones?

I don't think we should do that – we let the browser decide in every other case so it'd be an unusual thing for us to do.

i don't think

olliecurtis (approved):

packages/bpk-mixins/src/mixins/_chips.scss
georgegillams: We literally do that for button

PR #2105, opened 2021-03-02
[BPK-4420] Strip fill rule from icons

shaundon:
* I've updated the dangerfile to check and provide advice, but I don't think it's actually running with our current CI process. Fixing it felt outside the scope of this PR but we should definitely do it soon.
* As part of this, most of our icons have changed to not include `fill-rule`, which is nice as it'll save some bytes. I loaded them in Storybook alongside the ones we have in prod and I'm pretty sure none of them have changed visually.


decisions/fill-rule-in-svgs.md
```diff
@@ -0,0 +1,11 @@
+# We can't use SVGs that include `fill-rule`
+
+Whenever our icons contain the `fill-rule` property, they don't render correctly on iOS. This is because the process we use to convert SVGs into an icon font seems to strip out `fill-rule`.
```
georgegillams, line 3: This is _**technically**_ wrong as they could use `fill-rule` unnecessarily. This might be more technically correct
```suggestion
Whenever our icons rely on `fill-rule="evenodd"`, they don't render correctly on iOS. This is because the process we use to convert SVGs into an icon the `fill-rule` is not applied.
```

georgegillams (changes requested):

georgegillams (commented):
calendar, wallet and cloakroom have changed. Wanna pair on fixing them?

georgegillams (approved):

PR #2112, opened 2021-03-10
[DINGO-4320]: Calendar focus on null date

dominicfraser:
# Description

Currently the flow is:
-  with no date selected the focus outline is on the `minDate` (See `determineFocusedDate` function) 
- a date is selected on the calendar, that date has a focus outline
- then the selected date is set to `null`, the focus outline will remain on the previously selected date

The desired flow is for the final point to change to:
- then the selected date is set to `null`, the focus outline will return to the `minDate`

At a code level what is happening is:
- The new `selectedDate` prop is given to the `BpkCalendarContainer`
- It runs `UNSAFE_componentWillReceiveProps`
- That runs `focusedDateHasChanged`

As the date could be passed in as `Date.now()` but we only care about the Day, not the time, then `isSameDay` is used to compare two Date type values.

Previously the check was:
```
rawNextSelectedDate && !isSameDay(rawNextSelectedDate, rawSelectedDate)
```

We can see here that in the case of `rawNextSelectedDate` being null then `false would be returned`.

The possible cases are:

| prevDate / nextDate | Is handled? |
| -- | --| 
| null / null     | This PR adds  |
| Date / null   | This PR adds, previously would be false|
| null / Date   | This PR adds |
| Date / Date | Already handled |

# Checklist

- [x] `UNRELEASED.md`
- ~[ ] `README.md`~
- [x] Tests

### DIFF
--- a/packages/bpk-component-calendar/src/BpkCalendarContainer.js
+++ b/packages/bpk-component-calendar/src/BpkCalendarContainer.js
@@ -41,7 +41,14 @@ const focusedDateHasChanged = (currentProps, nextProps) => {
   const rawNextSelectedDate = nextProps.selectedDate || nextProps.date;
   const rawSelectedDate = currentProps.selectedDate || currentProps.date;
 
-  if (!rawNextSelectedDate) {
+  if (!rawSelectedDate && !rawNextSelectedDate) {
+    return false;
+  }
+
+  if (
+    (rawSelectedDate && !rawNextSelectedDate) ||
+    (!rawSelectedDate && rawNextSelectedDate)
+  ) {
     return true;
   }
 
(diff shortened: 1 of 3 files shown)


olliecurtis (approved):
LGTM

packages/bpk-component-calendar/src/BpkCalendarContainer.js
```diff
@@ -41,9 +41,18 @@ const focusedDateHasChanged = (currentProps, nextProps) => {
...
-  return (
-    rawNextSelectedDate && !isSameDay(rawNextSelectedDate, rawSelectedDate)
-  );
+  if (!rawSelectedDate && !rawNextSelectedDate) {
+    return false;
+  }
+
+  if (
+    (rawSelectedDate && !rawNextSelectedDate) ||
+    (!rawSelectedDate && rawNextSelectedDate)
```
georgegillams, line 50: This could be simplified, but maybe you were deliberately trying to make it explicit?
```suggestion
    !rawSelectedDate || !rawNextSelectedDate
```

dominicfraser: Going for deliberately explicit to:
- make it easy for future readers to scan and know the intent is 'prev is null and next is not OR prev is null/undefined and next is not'
- make it as easy as possible for reviewers to spot mistakes in my intent, as this is the first time I have looked at this component so may have missed context myself

PR #2115, opened 2021-03-11
[BPK-3607] Make tooltips focusable and add required ariaLabel prop

shaundon:
In addition to this, I'm going to add a note to the docs site page to mention that tooltips aren't a great pattern as they don't work well on touchscreens.

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-tooltip/README.md
+++ b/packages/bpk-component-tooltip/README.md
@@ -49,6 +49,8 @@ const App = () => (
 
 Tooltips are invisible to assistive technologies such as screen readers. To improve accessibility, `ariaLabel` is required to describe the content of the tooltip to assistive technologies.
 
+The label will be used on the `target` element, so any existing `aria-label` attached to `target` will be overridden.
+
 #### `popperModifiers`
 
 Please refer to the [documentation](https://github.com/FezVrasta/popper.js/blob/v1.12.9/docs/_includes/popper-documentation.md#modifiers) for the underlying positioning library "Popper.js". You can achieve various behaviours such as allowing the tooltip to overflow the viewport etc.
--- a/packages/bpk-component-tooltip/README.md
+++ b/packages/bpk-component-tooltip/README.md
@@ -17,6 +17,7 @@ import BpkTooltip from 'bpk-component-tooltip';
 
 const App = () => (
   <BpkTooltip
+    ariaLabel="London Heathrow"
     id="my-tooltip"
     target={<BpkText textStyle="lg">LHR</BpkText>}
   >
@@ -29,6 +30,7 @@ const App = () => (
 
 | Property              | PropType                                       | Required | Default Value       |
 | --------------------- | ---------------------------------------------- | -------- | ------------------- |
+| ariaLabel             | string                                         | true     | -                   |
 | id                    | string                                         | true     | -                   |
 | children              | node                                           | true     | -                   |
 | target                | node                                           | true     | -                   |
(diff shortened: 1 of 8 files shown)


packages/bpk-component-tooltip/src/BpkTooltip.js
```diff
@@ -18,6 +18,12 @@
 
 /* @flow strict */
 
+/*
+This is the component for the tooltip that is displayed to users.
+
+The actual component that developers create (i.e. the default export from this package) is BpkTooltipPortal.
+*/
+
```
shaundon, line 26: I was confused by our weird naming – the component consumers interface with is actually `BpkTooltipPortal`, not `BpkTooltip`, so I added this as a hint for future people.

Another option would be to rename them:

`BpkTooltip` ➡️ `BpkTooltipInner`
`BpkTooltipPortal` ➡️ `BpkTooltip`

But I wanted to see what people think in the PR first. I don't think renaming them would be breaking as it's an internal implementation detail and would only break usage for people who were using the non-public interface.

packages/bpk-component-tooltip/src/BpkTooltipPortal.js
```diff
@@ -42,7 +42,8 @@ const hasTouchSupport = () =>
 
 export type Props = {
   ...$Exact<TooltipProps>,
-  target: Node,
+  ariaLabel: string,
+  target: Element<any>,
```
shaundon, line 46: Had to change this as it caused a Flow error when I cloned it.

packages/bpk-component-tooltip/src/BpkTooltipPortal.js
```diff
@@ -98,6 +100,8 @@ class BpkTooltipPortal extends Component<Props, State> {
     if (this.targetRef) {
       const ref = this.targetRef;
 
+      ref.addEventListener('focusin', this.openTooltip);
+      ref.addEventListener('focusout', this.closeTooltip);
```
shaundon, line 104: Ensure that tooltips open/close when the element is focused.

packages/bpk-component-tooltip/src/BpkTooltipPortal.js
```diff
@@ -169,13 +176,18 @@ class BpkTooltipPortal extends Component<Props, State> {
     const classNames = [getClassName('bpk-tooltip-portal')];
     const renderPortal = !hasTouchSupport() || !hideOnTouchDevices;
 
+    const targetWithAccessibilityProps = React.cloneElement(target, {
+      tabIndex: '0',
+      'aria-label': ariaLabel,
```
olliecurtis, line 181: What happens if the target a consumer supplies already has an aria-label set? Do we worry that it would then get overriden by this one?

e.g.

```
  <BpkTooltip
    ariaLabel="London Heathrow"
    id="my-tooltip"
    target={<BpkText ariaLabel="London Heathrow Airport" textStyle="lg">LHR</BpkText>}
  >
  >
    London Heathrow
  </BpkTooltip>
```

Maybe we need to make sure its clear in the docs around how the aria-label would be used so that if someone does the above scenario they know they should be passing it through the prop rather than just supply it straight to the target?

olliecurtis: I know its rare but best to think about it and cover our bases just in case?

olliecurtis (commented):
LGTM! Will check it once deployed

packages/bpk-component-tooltip/src/BpkTooltipPortal.js
shaundon: Yeah it would override it, that's a good point. I'll add a note to the docs now

georgegillams: Maybe in the changeling we should say "if you're already providing an ariaLabel to the target, ensure you move this to the tooltip"?

packages/bpk-component-tooltip/README.md
```diff
@@ -43,6 +45,12 @@ const App = () => (
 
 ### Prop Details
 
-#### popperModifiers
+#### `ariaLabel`
+
+Tooltips are invisible to assistive technologies such as screen readers. To improve accessibility, `ariaLabel` is required to describe the content of the tooltip to assistive technologies.
+
+The label will be used on the `target` element, so any existing `aria-label` attached to `target` will be overridden.
```
georgegillams, line 52: 👍

olliecurtis: 👍

olliecurtis (approved):

PR #2121, opened 2021-03-17
DINGO-4358: Android screenreader datepicker bug

dominicfraser:
# Description 

The Datepicker component renders an input that when clicked opens either a popover (on desktop) or a full screen modal (mobile) containing a calendar.

On mobile the `readOnly` prop is applied to the input as a way to prevent the native keyboard opening when the input receives focus. The keyboard would be both interruptive and useless when the inputs value is controlled by the calendar component.

On iOS this works without problem, but on Android it has been found that the modal (containing the calendar) cannot be opened when a screen reader (TalkBack) is in use. The input receives accessibility focus, reads out that it is 'read only' but does not trigger either click or touch events. TalkBack seems to be trying to be clever and stopping these events from triggering due to the read only nature of the input.

Quick experimentation shows that `onFocus` does trigger, so is available for use. It triggers not on initial focus but after 'double tap to activate' has been used. This can be seen to first be assistive technology focus that then triggers the full browser focus.

The input in the Datepicker is constructed using `withOpenEvents(BpkInput)`. If we look inside `withOpenEvents` we can see that `onFocus` is already set up for desktop, but is not used on mobile. By allowing this to also work on mobile we see Android now works as expected, and iOS appears unaffected (which is good).

We can also add the `aria-readonly=false` attribute to prevent the screen reader confusing the user by announcing that the input is read only.

The only [reference to this externally](https://stackoverflow.com/questions/57719547/onclick-event-for-input-field-doesnt-work-while-accessibility-screen-reader-is) that I can find indicate that in older (v8) versions of Android this was not the behaviour - so when built this component likely did not have this problem but with newer versions of Android it has been introduced as a side effect.

# Checklist

- [x] `UNRELEASED.md`
- ~[ ] `README.md`~
- [x] Tests

### DIFF
--- a/packages/bpk-component-datepicker/src/BpkDatepicker.js
+++ b/packages/bpk-component-datepicker/src/BpkDatepicker.js
@@ -124,7 +124,6 @@ class BpkDatepicker extends Component {
         onChange={() => null}
         onOpen={this.onOpen}
         isOpen={this.state.isOpen}
-        readOnly
         valid={valid}
         {...inputProps}
       />
--- a/packages/bpk-component-input/src/withOpenEvents.js
+++ b/packages/bpk-component-input/src/withOpenEvents.js
@@ -115,16 +115,20 @@ const withOpenEvents = InputComponent => {
       };
 
       if (hasTouchSupport) {
-        // Prevents the mobile keyboard from opening (iOS / Android)
+        // Prevents the mobile keyboard from opening (iOS / Android), while not announcing it as 'read only' to a screen reader
         eventHandlers.readOnly = 'readOnly';
+        eventHandlers['aria-readonly'] = false;
+
         eventHandlers.onTouchEnd = withEventHandler(
           this.handleTouchEnd,
           onTouchEnd,
         );
-      } else {
-        eventHandlers.onFocus = withEventHandler(this.handleFocus, onFocus);
-        eventHandlers.onBlur = withEventHandler(this.handleBlur, onBlur);
       }
+
+      // Needed on desktop to allow the intended behaviour of opening on focus
+      // Needed on mobile as some Android devices do not trigger onClick or onTouch when TalkBack is active but do trigger onFocus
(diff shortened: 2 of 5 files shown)


packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -124,7 +124,6 @@ class BpkDatepicker extends Component {
         onChange={() => null}
         onOpen={this.onOpen}
         isOpen={this.state.isOpen}
-        readOnly
```
dominicfraser, line 127: This is optional to include - but I think adding this is redundant here as we already apply it within `withOpenEvents`. If we do want to add it here too I'd probably say we also want to pair it with `aria-readonly=false` - but I think we can just remove it?

packages/bpk-component-input/src/withOpenEvents.js
```diff
@@ -115,17 +115,21 @@ const withOpenEvents = InputComponent => {
...
-      } else {
-        eventHandlers.onFocus = withEventHandler(this.handleFocus, onFocus);
-        eventHandlers.onBlur = withEventHandler(this.handleBlur, onBlur);
       }
 
+      // Needed on desktop to allow the intended behaviour of opening on focus
+      // Needed on mobile as some Android devices do not trigger onClick or onTouch when TalkBack is active but do trigger onFocus
+      eventHandlers.onFocus = withEventHandler(this.handleFocus, onFocus);
+      eventHandlers.onBlur = withEventHandler(this.handleBlur, onBlur);
+
```
dominicfraser, line 132: Left more inline comments as someone new to the file likely wouldn't be able to tell the context of this being necessary on mobile even if we reversed the existing desktop behaviour at a future point. WDYT?

olliecurtis (commented):
LGTM! As discussed yesterday we will first release as a beta to ensure there is no strange side effects of this and if all is 🟢 then we can merge and release a stable

olliecurtis (approved):

PR #2124, opened 2021-03-22
DINGO-4393: Calendar screen reader tweaks

dominicfraser:
# Description 

On Android Chrome screen readers (TalkBack) currently reads out the column headings (e.g. 'Mon', 'Tues') and disabled dates in the Calendar, used in the BpkDatepicker.

It looks like the title is already intended not to be read, but on Android specifically TalkBack takes the `title` attribute and uses it as the accessible label.

For the dates they are given a negative tab index, but on mobile devices all elements are read through so also need `aria-hidden` applied.

Screenshot 2021-03-22 at 22 17 12

# Changes
- 'blocked' dates are now not read by screen readers
- column headings (e.g. 'Mon', 'Tues') are now consistently not read by screen readers

- [x] `UNRELEASED.md`
- ~[ ] `README.md`~
- [x] Tests


packages/bpk-component-calendar/src/BpkCalendarDate.js
```diff
@@ -139,7 +139,8 @@ class BpkCalendarDate extends PureComponent {
         type="button"
         style={style}
         className={classNames.join(' ')}
-        aria-label={date.getDate()}
+        aria-hidden={isBlocked}
+        aria-label={isBlocked ? '' : date.getDate()}
```
dominicfraser, line 143: Date change 1 - added here as date can be used as its own component.

packages/bpk-component-calendar/src/BpkCalendarGridHeader.js
```diff
@@ -59,8 +59,8 @@ const WeekDay = props => {
   }
 
   return (
-    <Element className={classNames.join(' ')} title={weekDay.name}>
-      <span aria-hidden="true">{weekDay[weekDayKey]}</span>
+    <Element className={classNames.join(' ')} title={weekDay.name} aria-hidden>
+      <span>{weekDay[weekDayKey]}</span>
```
dominicfraser, line 63: Heading change - move `aria-hidden` up by one element. The `title` appears to be being read when using TalkBack otherwise.

packages/bpk-component-calendar/src/Week.js
```diff
@@ -205,39 +205,41 @@ class Week extends Component {
...
-              {...dateProps}
-            />
-          </DateContainer>
-        ))}
+            >
+              <DateComponent
+                date={date}
+                modifiers={dateModifiers}
+                aria-hidden={isBlocked}
+                aria-label={isBlocked ? '' : formatDateFull(date)}
```
dominicfraser, line 228: Date change 2 - apply here as the Week can have any custom date component passed in.

Make `isBlocked` a const and use in `aria-hidden` and `aria-label`.

UNRELEASED.md
```diff
@@ -6,6 +6,11 @@
 - bpk-svgs:
   - New `center-location` icon.
 
+**Fixed:**
+
+- bpk-component-calendar:
+  - Calendar Grid column headings and out-of-range dates now have aria-hidden applied.
```
georgegillams, line 12:
```suggestion
  - Calendar Grid column headings and out-of-range dates are now hidden to assistive technology.
```

packages/bpk-component-calendar/src/BpkCalendarDate-test.js
```diff
@@ -40,6 +40,15 @@ describe('BpkCalendarDate', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('should render a disabled and blocked button', () => {
+    const tree = renderer
+      .create(
+        <BpkCalendarDate date={new Date(2010, 1, 15)} disabled isBlocked />,
```
georgegillams, line 46: What's the difference between `disabled` and `isBlocked`? Can we use one prop to do both?

packages/bpk-component-calendar/src/BpkCalendarGridHeader.js
georgegillams: Would it make sense to just apply `aria-hidden="true"` to the entire header instead of each date?

packages/bpk-component-calendar/src/BpkCalendarDate-test.js
dominicfraser: It's a little complex due to the fact that we allow components to be used both in isolation and as wrapped components that are provided specific props.

BpkCalendarDate takes both a `buttonProps` implicit `rest` that can be spread onto the button, and a separate specific `isBlocked` prop.

`disabled` is used in some tests, but isn't explicitly part of the API. Rather, `isBlocked` is used to set `disabled` - `disabled={isBlocked}`. The test shows how `disabled` can also be passed in as a singular prop.

However, `Week.js` provides `isBlocked` to its `DateComponent`, with `DateComponent` being passed in as a prop. In all cases I've seen this `DateComponent` is `BpkCalendarDate`, but it could technically be anything. From the existing code it looks like `isBlocked` is intended to be the declarative prop passed around.

So, to answer the question - we _almost_ do use just one - if only `isBlocked` is provided it also sets `disabled`. The API allows for unsetting this using `isBlocked disabled={false}` if someone really wanted to. The test is a little redundant having both set here - `isBlocked` will set `disabled=true` and then `disabled` will set `disabled=true`. It does represent something possible with the API, but unlikely. 

I could make the test only pass `isBlocked`, or could pass both to highlight it is possible - do you have a preference? Possibly the former.

packages/bpk-component-calendar/src/BpkCalendarDate.js
k0nserv: Is this change needed if we have `aria-hidden`?

packages/bpk-component-calendar/src/Week.js
```diff
@@ -205,39 +205,41 @@ class Week extends Component {
...
-              isKeyboardFocusable={isKeyboardFocusable}
-              isFocused={isSameDay(date, focusedDate)}
-              isSelected={isSameDay(date, selectedDate)}
-              isBlocked={
-                minDate && maxDate
-                  ? !isWithinRange(date, minDate, maxDate)
-                  : false
+        {this.props.dates.map(date => {
+          const isBlocked =
+            minDate && maxDate ? !isWithinRange(date, minDate, maxDate) : false;
```
k0nserv, line 210: Can't this be `minDate && maxDate && !isWithinRange(date, minDate, maxDate)` which avoids the ternary?

packages/bpk-component-calendar/src/Week.js
```diff
@@ -205,39 +205,41 @@ class Week extends Component {
...
+
+          return (
+            <DateContainer
```
georgegillams, line 213: By default, VoiceOver navigates through each cell of a table. As these are now essentially empty, I think we need to apply `aria-hidden` to each `DateContainer` too, to avoid this:
Screenshot 2021-03-23 at 08 33 14 (2)

dominicfraser:
> Is this change needed if we have aria-hidden
Technically no - aria-hidden takes precedence. I added it since as it is hidden the aria-label is redundant, so we can avoid the 'work' of setting the label.

packages/bpk-component-calendar/src/Week.js
dominicfraser: Yes - I just lifted and shifted how it had been declared by the previous writer when it was declared inline in the prop - can update if preferred?

packages/bpk-component-calendar/src/BpkCalendarDate-test.js
georgegillams: I'd probably prefer just `isBlocked` as the `disabled` prop is redundant

packages/bpk-component-calendar/src/Week.js
dominicfraser: Hmm, this is with `control + option + arrow`? 

When using TalkBack it completely skips these boxes.

dominicfraser: Let me have a play adjusting for desktop 👀

packages/bpk-component-calendar/src/BpkCalendarDate.js
dominicfraser: Technically no - aria-hidden takes precedence. I added it since as it is hidden the aria-label is redundant, so we can avoid the 'work' of setting the label.

packages/bpk-component-calendar/src/BpkCalendarGridHeader.js
dominicfraser: It possible - was aiming for as minimal chance of unspotted side effects as possible by moving just one level - but I have no objection if we wanted to move it higher.

packages/bpk-component-calendar/src/Week.js
dominicfraser: Pushed - in my testing this now works. Added to `DateContainer`, so could remove from `DateComponent` I think?

k0nserv:
> Technically no - aria-hidden takes precedence. I added it since as it is hidden the aria-label is redundant, so we can avoid the 'work' of setting the label.

Is this really worth the extra code compexity?

packages/bpk-component-calendar/src/BpkCalendarGridHeader.js
dominicfraser: Moved up

packages/bpk-component-calendar/src/Week.js
dominicfraser: Removed from `DateComponent`

packages/bpk-component-calendar/src/BpkCalendarDate.js
dominicfraser: > Is this really worth the extra code complexity?

No strong opinion, so have removed.

UNRELEASED.md
```diff
@@ -2,6 +2,15 @@
 
 > Place your changes below this line.
 
+**Added:**
+- bpk-svgs:
+  - New `center-location` icon.
```
georgegillams, line 7: I think this probably got added in a rebase?

dominicfraser: 🤦

dominicfraser: Removed 👍

georgegillams (approved):

packages/bpk-component-calendar/src/Week.js
k0nserv: Ah, I didn't realise. Let's keep it as is to reduce the number of changes

k0nserv (approved):

PR #2125, opened 2021-03-25
Add script post publish to update bpk deps in the docs site

olliecurtis:
So that we don't need to worry about the docs site becoming out of date with Web components or updates (e.g. icons being released), created a simple script that will run post publishing our packages so that we don't forget 😄 

The script is **not** made to be super intense and do loads of bells and whilstes but just simply take the effort out of changing projects, updating and creating a PR through the website.

The script will: 

- clone the docs site locally, 
- run an update for bpk packages, 
- install the new packages 
- use the gh cli to push and create an automatic PR to the docs site
- Cleanup after itself (litterally just an `rm -rf`)

To show it working a sample PR is available here: https://github.com/olliecurtis/docs-testground/pull/4


scripts/publish-process/update-docs-deps.sh
```diff
@@ -0,0 +1,44 @@
...
+
+BRANCH_NAME="autobump-upgrade-bpk-dependencies"
+
+echo "--- Cloning Backpack docs ---"
+git clone git@github.com:skyscanner/backpack-docs.git
+
+cd backpack-docs
+
+echo "--- Checkout upgrade branch ---"
+git checkout -b $BRANCH_NAME
```
georgegillams, line 27: This will fail if the branch already exists remotely. Maybe it should be this + timestamp?

scripts/publish-process/update-docs-deps.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "--- Updating bpk dependencies ---"
+npx npm-check-updates "/^bpk\\-(.*)$/" -u && npm i
+
+echo "--- Commiting to GitHub ---"
+git add .
+git commit -m "Automated upgrade of bpk dependencies"
+git push --set-upstream origin $BRANCH_NAME
+
+echo "--- Create PR on GitHub ---"
+gh pr create --title "Automated bump for Backpack dependencies" --body "This is an automated PR generated as part of the publish process in Backpack"
```
georgegillams, line 38: This assumes the publisher is signed in to GH for command line. Probably an acceptable assumption.

Do we need to set the title? By default doesn't it use the commit message?

scripts/publish-process/update-docs-deps.sh
olliecurtis: Only the case if you doon't delete the branch after merging the PRs ;)

scripts/publish-process/update-docs-deps.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "--- Cloning Backpack docs ---"
+git clone git@github.com:skyscanner/backpack-docs.git
+
+cd backpack-docs
+
+echo "--- Checkout upgrade branch ---"
+git checkout -b $BRANCH_NAME
+
+echo "--- Updating bpk dependencies ---"
+npx npm-check-updates "/^bpk\\-(.*)$/" -u && npm i
```
georgegillams, line 30: I feel like maybe the backpack-docs repo should have it's own script for bumping which this script could just use? But maybe that's over engineering it?

scripts/publish-process/update-docs-deps.sh
olliecurtis: Yeah I guess we can exclude the title - I don't think personally it matters what we do for this and if we need to get picky about setting it or not

scripts/publish-process/update-docs-deps.sh
olliecurtis: Yeah - I don't think we need to over engineer here, this is just a simple script to help out, as the docs site would never auto run a script.

I just translated my workflow into a script here for convience and having steps across both repos adds to the maintenace burden which we want to avoid

scripts/publish-process/update-docs-deps.sh
georgegillams: Which could easily happen, especially as we sometimes release a lot

olliecurtis: Its also worth noting we do have this setting turned on in the docs repo meaning that the branch should never exist on remote once the PR has been merged 🤷 

Screenshot 2021-03-25 at 13 38 27

scripts/publish-process/update-docs-deps.sh
olliecurtis: I do also however see we have the npx part as a npm command we can invoke so could just use that instead

https://github.com/Skyscanner/backpack-docs/blob/main/package.json#L45

scripts/publish-process/update-docs-deps.sh
georgegillams: Yeah but suppose I release something small, I might not look at the docs at all, and then someone releases something later and suddenly the PR they actually want isn't made because the branch already existed

scripts/publish-process/update-docs-deps.sh
georgegillams: Yeah. I guess maybe that should do the `npm i` as well as it hasn't technically upgraded them until that's done

scripts/publish-process/update-docs-deps.sh
olliecurtis: That is true, I just went on the assumption our PRs channel would alert that a PR has been opened and we would be jumping on it asap as normal, but I do see this as a potential clash!

scripts/publish-process/update-docs-deps.sh
olliecurtis: Docs site script has been fixed to include this and now the step has been simplified! Thanks for the suggestion looks cleaner now!

scripts/publish-process/update-docs-deps.sh
olliecurtis: Changed this to use the date and time at the end to avoid the problem @georgegillams has mentioned above

georgegillams (approved):
Excellent

shaundon (changes requested):
I have a couple of small suggestions so I'm just sending this now to block the PR before it automerges 😅 I'll add full comment in a moment

.gitignore
```diff
@@ -16,3 +16,6 @@ sassdoc-data.json
 backpack.iml
 yarn.lock
 /packages/*/changelog.md
+
+# Incase the script does not fully remove this we don't want to push to git
```
shaundon, line 20: Good to have this comment here, but I could foresee a situation in the future where somebody new doesn't know what we mean by 'the script', so I've expanded the comment to add some more context.

```suggestion
# On postpublish, we clone and run some scripts on the backpack-docs repo. This ensures we don't accidentally commit it.
```

package.json
```diff
@@ -49,7 +49,7 @@
     "postinstall": "npm run bootstrap",
     "preinstall": "npx ensure-node-env",
     "prettier": "prettier --config .prettierrc --write \"**/*.js\"",
-    "publish": "npm run check-pristine && npm run check-owners && npm run build && git pull --rebase && flow stop && npm run test && lerna publish && npm run publish:css",
+    "publish": "npm run check-pristine && npm run check-owners && npm run build && git pull --rebase && flow stop && npm run test && lerna publish && npm run publish:css && ./scripts/publish/process/update-docs-deps.sh",
```
shaundon, line 52: npm has a built-in `postpublish` script that runs automatically after `publish` completes, could we use that for this instead? It'll just help the command from getting too long, and make more sense semantically.

```suggestion
    "publish": "npm run check-pristine && npm run check-owners && npm run build && git pull --rebase && flow stop && npm run test && lerna publish && npm run publish:css",
    "postpublish": "./scripts/publish/process/update-docs-deps.sh",
```

scripts/publish-process/update-docs-deps.sh
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+BRANCH_NAME="autobump-bpk-dependencies-`date +%Y-%m-%d.%H%M`"
+
+echo "--- Cloning Backpack docs ---"
+git clone git@github.com:skyscanner/backpack-docs.git
```
shaundon, line 22: `tmp` discussion.

package.json
olliecurtis: Let me take a look at this and see what happens - now I saw it in the npm docs will give it a go!

shaundon (approved):

PR #2127, opened 2021-03-26
[BPK-4448] Proof of concept of adding jest-axe to BpkButton tests

shaundon:
This PR adds a single a11y test to as a proof of concept. 

To make it work, I had to add a dependency on `@testing-library/react`, as jest-axe won't work with `react-test-renderer`. This is because it requires real HTML, not the pseudo-HTML structures generated by our tests.

As a follow up to this PR, I suggest we do two things:

## 1. Migrate our tests

We should migrate **all** our tests to use `@testing-library/react`. We shouldn't have two different test renderers, so we should standardise on this better library. It's more capable, used elsewhere in Skyscanner (like in Banana) and it's the testing library built into new repos made with `create-react-app`, so all signs point to this being the one to use.

This could be mostly automated with a regex so it isn't a lot of overhead. It would mean that our snapshots would change, like so:

**Using `react-test-renderer` (current method)**
```js
  const tree = renderer.create(<BpkButton>My button</BpkButton>).toJSON();
});


// Generated snapshot. 
// Note that `className` and `onClick` are not valid HTML attributes, they are React attributes.
exports[`BpkButton should render correctly 1`] = `
<button
  className="bpk-button"
  disabled={false}
  onClick={null}
  type="button"
>
  My button
</button>
`;
```

**Using `@testing-library/react` (proposed method)**
```js
  const { asFragment } = render(<BpkButton>My button</BpkButton>);
});

// Output from `@testing-library/react`. Note that it's now valid HTML, e.g. 'className' has been replaced with 'class'.
exports[`BpkButton should render correctly 1`] = `
<DocumentFragment>
  <button
    class="bpk-button"
    type="button"
  >
    My button
  </button>
</DocumentFragment>
`;
```

## 2. Add a11y tests to all our test files

This is definitely non trivial (we have almost 200 test files), but it should be _relatively_ straightforward. The code will largely be copy pasted with some minor tweaks each time. I'd expect it to be doable as a five point Jira card.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
       "^.+\\.svg$": "<rootDir>/scripts/stubs/fileStub.js",
       "react-transition-group/CSSTransition": "<rootDir>/scripts/stubs/cssTransitionStub.js"
     },
-    "setupFiles": [
+    "setupFilesAfterEnv": [
       "<rootDir>/scripts/jest/setup.js"
     ],
     "testPathIgnorePatterns": [
@@ -124,6 +124,7 @@
     "@storybook/addon-links": "^5.3.14",
     "@storybook/addon-viewport": "^5.3.14",
     "@storybook/react": "^5.3.14",
+    "@testing-library/react": "^11.2.5",
     "autoprefixer": "^9.4.7",
     "babel-jest": "^24.9.0",
     "babel-loader": "^8.0.0",
@@ -159,6 +160,7 @@
     "gulp-theo": "^2.0.0",
     "husky": "^4.0.2",
     "jest": "^24.8.0",
+    "jest-axe": "^4.1.0",
     "js-yaml": "^3.13.1",
     "lerna": "^3.15.0",
     "lint-staged": "^10.0.6",
--- a/packages/bpk-component-button/src/BpkButton-test.js
+++ b/packages/bpk-component-button/src/BpkButton-test.js
@@ -23,14 +23,11 @@ import { render } from '@testing-library/react';
 
 import BpkButton from './BpkButton';
 
-const defaultTestProps = {
-  onClick: () => {},
-  children: 'My button',
-};
-
 describe('BpkButton accessibility tests', () => {
   it('should not have programmatically detectable accessibility issues', async () => {
-    const { container } = render(<BpkButton {...defaultTestProps} />);
+    const { container } = render(
+      <BpkButton onClick={() => {}}>My button</BpkButton>,
+    );
     const results = await axe(container);
     expect(results).toHaveNoViolations();
   });
--- a/packages/bpk-component-button/src/BpkButton-test.js
+++ b/packages/bpk-component-button/src/BpkButton-test.js
@@ -17,9 +17,24 @@
  */
 
 import React from 'react';
+import { axe } from 'jest-axe';
 import renderer from 'react-test-renderer';
+import { render } from '@testing-library/react';
 
 import BpkButton from './BpkButton';
+
+const defaultTestProps = {
+  onClick: () => {},
+  children: 'My button',
+};
+
+describe('BpkButton accessibility tests', () => {
+  it('should not have programmatically detectable accessibility issues', async () => {
+    const { container } = render(<BpkButton {...defaultTestProps} />);
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+});
 
 describe('BpkButton', () => {
   it('should render correctly', () => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -9302,10 +9302,265 @@
         }
       }
     },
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -97,7 +97,7 @@
       "^.+\\.svg$": "<rootDir>/scripts/stubs/fileStub.js",
       "react-transition-group/CSSTransition": "<rootDir>/scripts/stubs/cssTransitionStub.js"
     },
-    "setupFiles": [
+    "setupFilesAfterEnv": [
```
shaundon, line 100: This was required to work around a problem with integrating `jest-axe` into our setup: https://github.com/testing-library/jest-dom/issues/122#issuecomment-650520461

packages/bpk-component-button/src/BpkButton-test.js
```diff
@@ -17,10 +17,25 @@
...
+};
+
+describe('BpkButton accessibility tests', () => {
+  it('should not have programmatically detectable accessibility issues', async () => {
+    const { container } = render(<BpkButton {...defaultTestProps} />);
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+});
+
```
georgegillams, line 38: Are we able to reuse code in existing snapshot tests or examples instead of having a 3rd place that consumes the components?

olliecurtis: @georgegillams could you expand on this comment? Looking at this test it looks the exact same as how we would make any of the others below. Would be interested to chat to understand this comment more

georgegillams: As things stand, we need to create an example usage of BpkButton for `examples.js`. Then we also need to create it in `BpkButton-test.js` to test how it's rendered, and again to test is for accessibility.

Maybe instead we could define an example, and then reuse that example for both snapshot and a11y testing?

shaundon: Yeah that's definitely an option. We could do this:

```js
  const { asFragment, container } = render(<BpkButton>My button</BpkButton>);
  const results = await axe(container);
  expect(results).toHaveNoViolations();
});
```

The reason I didn't go for this was for two reasons:

1. It feels a little like it breaks single responsibility principle?
2. Banana doesn't do this and I'm using them as the gold standard of automated a11y testing

shaundon: Actually, let's ask Dom for his opinion. I'll loop him in.

shaundon: With regards to sharing `examples.js` in tests, I think that's out of scope for this. If we choose to do that it can be done separately without affecting this work.

packages/bpk-component-button/src/BpkButton-test.js
```diff
@@ -17,10 +17,25 @@
  */
 
 import React from 'react';
+import { axe } from 'jest-axe';
 import renderer from 'react-test-renderer';
+import { render } from '@testing-library/react';
```
dominicfraser, line 22: > To make it work, I had to add a dependency on `@testing-library/react`, as `jest-axe` won't work with react-test-renderer. This is because it requires real HTML, not the pseudo-HTML structures generated by our tests.

While, for the reasons stated in the PR description, I agree about using a) only one test framework and b) the ideal would be `@testing-library/react` we don't _have_ to introduce it in this PR.

It looks like the Backpack codebase has both `react-test-renderer` and `enzyme` already installed.

`enzyme` can be used as so: https://github.com/nickcolley/jest-axe#testing-react-with-enzyme

while it's also possible to do `ReactDOMServer.renderToString`: https://codeburst.io/automated-accessibility-testing-tool-a11y-pa11y-jest-storybook-2ad294bfe71a

If we're committed to the migration to `@testing-library/react` then there is no time like the present to start, but if we're unsure then `jest-axe` is not blocked on it AFAIK.

shaundon: Oh thanks, that's good to know!

packages/bpk-component-button/src/BpkButton-test.js
dominicfraser: I would always have the a11y test as a separate test.

However, it can be within the same describe block (just a new `it()`), and doesn't need to use the `{...getDefaultTestProps()}` pattern if the test file doesn't use it. That pattern is one we use only if there are large amount of props we want to have test defaults shared across tests, but in this case the two props are also easy to declare just within the individual test.

olliecurtis (approved):
After discussion on call about the approach for this happy to approve this approach and then have follow up PRs to changing libs, implementation, etc

Nice one!

https://media3.giphy.com/media/3ohhwi25ISXC7Z4tMs/giphy.gif

PR #2131, opened 2021-03-31
[NO-JIRA] Add aria-hidden note to aria-live README

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests


georgegillams:
Actually, you should not do this as it breaks Voice over

PR #2132, opened 2021-03-31
Improve usability of calendar navigation nudgers and dropdown when using a screenreader

Pyroboris:
Changes:
 - Add bpk-aria-live region to calendar navigation for month.
 - Add parameters to allow for custom previous/next month nudger text.
 - In order to keep backwards compatibility, or for those that wish to keep the current nudgers text the new parameters are optional.

Accessibility discussions and work has raised that the nudgers and the month dropdown can be confusing when used with screenreaders. Below are the three components addressed in this PR, marked 15, 16 and 17.

Screenshot 2021-04-01 at 00 27 38
When reading the Calendar Navigation with a screen reader (Android):   

|Before|After|
| 15. February, button. disabled |15. go to previous month, button, disabled.|
| 16. change month. Collapsed. march. change month. double tap to edit.  |16. Collapsed. march. change month. double tap to edit. |
| 17. April, Button. |17. go to next month, button |

The aria-live region that will update the user when either nudger is used, this means they will be informed of the new month selected.

- [ ] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests


packages/bpk-component-calendar/src/BpkCalendarNav.js
```diff
@@ -81,17 +84,20 @@ const BpkCalendarNav = props => {
         >
           <ArrowLeftIcon className={getClassName('bpk-calendar-nav__icon')} />
           <span className={getClassName('bpk-calendar-nav__text--hidden')}>
-            {formatMonth(addMonths(baseMonth, -1))}
+            {previousMonthLabel || formatMonth(prevMonth)}
```
Pyroboris, line 87: reusing [prevMonth from here](https://github.com/Skyscanner/backpack/pull/2132/files?file-filters%5B%5D=.js#diff-d4348e792a2389203dd012f08910249b4d0fb5c56ab4a319d3450468f65d7d9fR66) seems like it was duplicate code initially.

Using either previousMonthLabel or formatMonth(prevMonth) in order to keep the existing behaviour for those that don't have translations/custom text for their nudgers.

packages/bpk-component-calendar/src/BpkCalendarNav.js
```diff
@@ -81,17 +84,20 @@ const BpkCalendarNav = props => {
...
           </span>
         </button>
       </div>
       <div className={getClassName('bpk-calendar-nav__month')}>
         <label
           htmlFor={`${id}_select`}
           className={getClassName('bpk-calendar-nav__text--hidden')}
+          aria-hidden="true"
```
Pyroboris, line 95: The label is repeated again in the BpkSelect, by using `aria-hidden="true"` it's still used in the select, but is no longer read twice.

packages/bpk-component-calendar/src/BpkCalendarNav.js
```diff
@@ -81,17 +84,20 @@ const BpkCalendarNav = props => {
...
         >
           {changeMonthLabel}
         </label>
+
+        <BpkAriaLive aria-hidden="true">{formatMonth(baseMonth)}</BpkAriaLive>
```
Pyroboris, line 100: `aria-hidden="true"` In this case, we don't want the live region to be read out unless it's changed, otherwise this results in duplication of what is read out.

packages/bpk-component-calendar/src/composeCalendar.js
```diff
@@ -43,10 +43,12 @@ const composeCalendar = (Nav, GridHeader, Grid, CalendarDate) => {
       maxDate,
       minDate,
       month,
+      nextMonthLabel,
```
Pyroboris, line 46: For both of these labels, I attempted to follow the same pattern as `changeMonthLabel` the only difference being that they are optional.

packages/bpk-component-calendar/src/BpkCalendarNav-test.js
```diff
@@ -55,6 +55,38 @@ describe('BpkCalendarNav', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
+  it('should use given previousMonth and nextMonth labels correctly', () => {
```
Pyroboris, line 58: New tests to check that there isn't any regressions in labels, or in the default logic in the future. Other functionality is already covered in the existing tests.

packages/bpk-component-calendar/package.json
```diff
@@ -26,6 +26,7 @@
     "react": "^16.0.0"
   },
   "devDependencies": {
+    "bpk-component-aria-live": "^1.0.38",
```
Pyroboris, line 29: Added this dependency, had a lot of issue defaulting to a specific Artifactory and so didn't push the package-lock.json changes. I'd appreciate some specific attention around this change 🙇

georgegillams: No problem - I can push the change.

georgegillams:
Thanks for the PR.
I think the previous/next month labels should be required. In compose calendar, we only require `changeMonthLabel` if `Nav` is being provided. We can follow that pattern for these.

We can release a breaking change and write a migration guide. If we submit the strings too, then anyone can reuse them, so upgrading won't be hard.

packages/bpk-component-calendar/src/BpkCalendarNav.js
```diff
@@ -130,10 +136,14 @@ const BpkCalendarNav = props => {
...
           <ArrowRightIcon className={getClassName('bpk-calendar-nav__icon')} />
           <span className={getClassName('bpk-calendar-nav__text--hidden')}>
-            {formatMonth(addMonths(baseMonth, 1))}
+            {nextMonthLabel}
           </span>
         </button>
       </div>
+
+      <BpkAriaLive>
+        {`${nowDisplayingLabel} ${formatMonth(baseMonth)}`}
```
Pyroboris, line 145: Though aria-hidden *works* with BpkAriaLive on Android, it doesn't on iOS. In order to remove the aria-hidden and have the month read out all the time, the label text was made more descriptive to prevent confusion.

packages/bpk-component-calendar/src/BpkCalendarNav.js
```diff
@@ -146,6 +156,9 @@ BpkCalendarNav.propTypes = {
   maxDate: PropTypes.instanceOf(Date).isRequired,
   minDate: PropTypes.instanceOf(Date).isRequired,
   month: PropTypes.instanceOf(Date).isRequired,
+  nextMonthLabel: PropTypes.string.isRequired,
+  nowDisplayingLabel: PropTypes.string.isRequired,
+  previousMonthLabel: PropTypes.string.isRequired,
```
Pyroboris, line 161: This is 4 labels total now that are being passed down to the NAV from the datepicker, how do others feel about that?

packages/bpk-component-calendar/src/__snapshots__/BpkCalendarContainer-test.js.snap
```diff
@@ -31,19 +31,27 @@ exports[`BpkCalendarContainer should focus the correct date when \`initiallyFocu
...
       >
         <label
+          aria-hidden="true"
           class="bpk-calendar-nav__text--hidden"
           for="myCalendar__bpk_calendar_nav_select"
         >
           Change month
         </label>
+        <div
+          aria-hidden="true"
```
georgegillams, line 49: Looks like snapshots need to be updated

CHANGELOG.md
```diff
@@ -1,12 +1,18 @@
 # Backpack changelog
 
 [Unreleased changes](./UNRELEASED.md).
+# 2021-04-02
+
+**Added:**
+- bpk-component-calendar:
+  - Calendar Nav updated to use labels for nudgers and month aria-live region added for assistive technology.
```
georgegillams, line 8: Sorry to be a pain, could you move this to `UNRELEASED.md`?

georgegillams (approved):

PR #2133, opened 2021-04-01
[KOA-4458] Add some jest-axe tests and tweak accordion to pass a11y test

shaundon:
I migrated a few components just to validate the approach, and if we're happy with it I'll continue and do the rest.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+- bpk-component-accordion:
+  - Improved accessibility by removing invalid `aria-level="3"` attribute on `BpkAccordionItem`.
+
 ## How to write a good changelog entry
 
 1. Add 'Breaking', 'Added' or 'Fixed' in bold depending on if the change will be major, minor or patch according to [semver](semver.org).
--- a/packages/bpk-component-button/src/BpkButton-test.js
+++ b/packages/bpk-component-button/src/BpkButton-test.js
@@ -23,10 +23,8 @@ import { render } from '@testing-library/react';
 import BpkButton from './BpkButton';
 
 describe('BpkButton accessibility tests', () => {
-  it('should not have programmatically detectable accessibility issues', async () => {
-    const { container } = render(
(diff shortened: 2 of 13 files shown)


olliecurtis (approved):
LGTM!

olliecurtis (approved):

UNRELEASED.md
```diff
@@ -2,6 +2,10 @@
 
 > Place your changes below this line.
 
+**Fixed:**
+- bpk-component-accordion:
+  - Improved accessibility by removing invalid `aria-level="3"` attribute on `BpkAccordionItem`.
```
georgegillams, line 7: This property wasn't actually achieving anything as it was invalid, so I don't think we need to even have an entry 🤷

packages/bpk-component-button/src/BpkButton-test.js
```diff
@@ -23,10 +23,8 @@ import { render } from '@testing-library/react';
 import BpkButton from './BpkButton';
 
 describe('BpkButton accessibility tests', () => {
-  it('should not have programmatically detectable accessibility issues', async () => {
-    const { container } = render(
-      <BpkButton onClick={() => {}}>My button</BpkButton>,
-    );
+  it('should not have programmatically-detectable accessibility issues', async () => {
+    const { container } = render(<BpkButton>My button</BpkButton>);
```
georgegillams, line 27: The component actually renders something completely different if you provide a `href`, so I think we should test that too

UNRELEASED.md
shaundon: I was just thinking just in case anybody is using deep rendering in their snapshots, causing them to break, this will explain why

packages/bpk-component-button/src/BpkButton-test.js
shaundon: Good catch

UNRELEASED.md
olliecurtis: I guess if a consumer is skipping a test or supressing any accessibility tests for using the component this will allow them to re-enable or remove any skipping

PR #2134, opened 2021-04-01
[KOA-4458] Add a11y tests for more components

shaundon:
I've done A-G, a bit more than a third of components.


packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -190,7 +190,7 @@ BpkDataTable.propTypes = {
   defaultColumnSortIndex: PropTypes.number,
   sort: PropTypes.func,
   sortBy: PropTypes.string,
-  sortDirection: PropTypes.oneOf('ASC', 'DESC'),
+  sortDirection: PropTypes.oneOf(['ASC', 'DESC']),
```
shaundon, line 193: This fixes an error that was coming up whenever we ran snapshot tests.

packages/bpk-component-banner-alert/src/BpkBannerAlertInner.js
```diff
@@ -185,7 +185,7 @@ const BpkBannerAlertInner = (props: Props) => {
       {...rest}
     >
       <section className={sectionClassNames.join(' ')}>
-        <header
+        <div
```
shaundon, line 188: `header` isn't valid in this use case, and the test rightly caught it. It should be used as a page or section header (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header)

olliecurtis (approved):

PR #2135, opened 2021-04-02
[KOA-4466] Automatically move unreleased entries to changelog after publishing

shaundon:
I removed the boilerplate from unreleased.md and changelog.md because keeping them made for a more complicated script, which outweighed the (very limited) purpose they served.


olliecurtis (approved):
Love it!

PR #2137, opened 2021-04-02
[DINGO-4409]-Adding labels to connect calendar and datepicker.

Pyroboris:
This change adds parameters that links the datepicker to the calendar. They appear to be missing in #2132.

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [x] Tests


UNRELEASED.md
```diff
@@ -1,2 +1,5 @@
+- bpk-component-calendar:
```
georgegillams, line 1:
```suggestion
- bpk-component-datepicker:
```

georgegillams (approved):
good catch - I'd just change the unreleased entry as suggested as this technically only changes the `datepicker` package

PR #2138, opened 2021-04-05
[KOA-4459, KOA-4460] Added jest-axe tests for all remaining components

shaundon:
There are two or three places where I've skipped a test and left a comment about why. In all cases, I've added a ticket to our backlog to address the underlying issue and reinstate the test.

### DIFF
--- a/packages/bpk-component-slider/src/accessibility-test.js
+++ b/packages/bpk-component-slider/src/accessibility-test.js
@@ -1,0 +1,49 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016-2021 Skyscanner Ltd
+ *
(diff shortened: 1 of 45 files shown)


packages/bpk-component-slider/src/accessibility-test.js
```diff
@@ -0,0 +1,49 @@
...
+import { axe } from 'jest-axe';
+
+import BpkSlider from './BpkSlider';
+
+describe('BpkSlider accessibility tests', () => {
+  /*
+  Currently this test fails because the third-party library
+  we're using doesn't pass down the ARIA props we add.
+  We should investigate and fix this so we can reinstate the test.
+  */
```
georgegillams, line 30: So does this test need to be excluded then?

georgegillams (approved):

PR #2141, opened 2021-04-09
[NO JIRA]: Updating check owners script to ignore non published folders

olliecurtis:
As we don't actually publish `bpk-component-boilerplate` and is used for automated component creation in our project, adding it as a project to be ignored when checking owners

- [ ] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/scripts/npm/check-owners.js
+++ b/scripts/npm/check-owners.js
@@ -119,7 +119,7 @@ const verifyMaintainers = data => {
 console.log(`Maintainers are:\n  ${owners.join('\n  ')}\n`);
 
 readdir('packages/')
-  .then(packages => packages.filter(i => !i.startsWith('.')))
+  .then(packages => packages.filter(i => i !== 'bpk-component-boilerplate'))
   .then(packages => {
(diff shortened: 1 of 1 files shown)


shaundon (approved):

scripts/npm/check-owners.js
```diff
@@ -119,7 +119,7 @@ const verifyMaintainers = data => {
 console.log(`Maintainers are:\n  ${owners.join('\n  ')}\n`);
 
 readdir('packages/')
-  .then(packages => packages.filter(i => !i.startsWith('.')))
+  .then(packages => packages.filter(i => i !== 'bpk-component-boilerplate'))
```
georgegillams, line 122: Shouldn't we read whether the package is private from package.json?

Maybe that's over-engineering. To make this more readable we could use 
```
const privatePackages = ['bpk-component-boilerplate'];
...
.then(packages => packages.filter(i => !privatePackages.contains(i)))
```

PR #2143, opened 2021-04-09
Koa 4467 voiceover bug fix for calendar on iOS

olliecurtis:
This fixes an issue with iOS and iPadOS with Asssitive Technologies when changing the month from the calendar dropdown meant that voiceover would get stuck in the current selected week and unable to navigate to new date rows

To do this we check if the device is an iOS device and return the calendar without grid transitions as this was causing the bug, a follow up PR/ticket will be done to review the transitions. 

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-calendar/src/BpkCalendarGrid.js
+++ b/packages/bpk-component-calendar/src/BpkCalendarGrid.js
@@ -215,6 +215,7 @@ BpkCalendarGrid.defaultProps = {
   cellClassName: null,
 };
 
+// On iOS, having transitions causes accessibility issues, so we disable it (KOA-4467).
 const BpkCalendarGridWithTransition = isDeviceIos()
   ? BpkCalendarGrid
   : addCalendarGridTransition(BpkCalendarGrid);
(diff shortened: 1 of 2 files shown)


packages/bpk-component-calendar/src/BpkCalendarGrid.js
```diff
@@ -215,9 +215,9 @@ BpkCalendarGrid.defaultProps = {
   cellClassName: null,
 };
 
-const BpkCalendarGridWithTransition = addCalendarGridTransition(
-  BpkCalendarGrid,
-);
+const BpkCalendarGridWithTransition = isDeviceIos()
```
shaundon, line 218: We should probably add a comment here to explain why we're doing this, as it's not immediately obvious. 

```suggestion
// On iOS, having transitions causes accessibility issues, so we disable it (KOA-4467).
const BpkCalendarGridWithTransition = isDeviceIos()
```

olliecurtis: Sounds good to me!

shaundon (approved):

PR #2144, opened 2021-04-12
[NO-JIRA] Fix card tabIndex

georgegillams:

- [x] `UNRELEASED.md`
- [x] `README.md`
- [x] Tests


olliecurtis (changes requested):
Does this affect any A11y? In which case is there a test we reenable or add for this?

georgegillams:
> Does this affect any A11y? In which case is there a test we reenable or add for this?

Somehow it wasn't failing before even though the test seems to be enabled 🤔

olliecurtis (approved):

PR #2152, opened 2021-04-23
A11y fixes for nudger axe tests

olliecurtis:
Fixed failing axe tests - turns out the test only needed a `<label>` or `aria-label`.

Also modifed the storybook example for the buttons to be more descriptive and add an `aria-label` to give context of the field value and when `aria-live` updates.

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,4 @@
+**Added:**
+- bpk-component-nudger:
+  - Fixed A11y jest axe test
+  - Added ability to spread `rest` across the input component to add `aria` and other `input` props.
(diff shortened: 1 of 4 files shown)


UNRELEASED.md
```diff
@@ -0,0 +1,4 @@
+**Added:**
+- bpk-component-nudger:
+  - Fixed A11y jest axe test
```
shaundon, line 3:
```suggestion
```

We don't usually mention things that aren't consumer facing, like changes to tests.

shaundon (changes requested):

shaundon:
Tested the deployed Storybook with Voiceover and it feels quite usable to me now, so I'm otherwise happy with the changes.

UNRELEASED.md
olliecurtis: Removed and resolved!

shaundon (approved):

PR #2156, opened 2021-04-27
[KOA-4560]: Add new spacing functions for tokens

olliecurtis:
As part of the new spacing updates and the PoC conducted, adding the new function tokens which will be used by components and consumers to switch between the old and new spacing.

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [x] Tests


packages/bpk-tokens/src/web/base/spacing.json
```diff
@@ -7,24 +7,52 @@
     "category": "spacings"
   },
   "props": {
+    "SPACING_FUNCTION_XS": {
+      "value": "@function bpk-spacing-xs() { @if global-variable-exists('new-spacing-values') and $new-spacing-values { @return .25rem; } @return {!SPACING_XS}; }",
```
georgegillams, line 11: Should this be something like
```suggestion
      "value": "@function bpk-spacing-xs() { @if global-variable-exists('new-spacing-values') and $new-spacing-values { @return {!SPACING_XS_NEW} } @return {!SPACING_XS}; }",
```

georgegillams (approved):

PR #2163, opened 2021-05-03
[KOA-4563] Upgrade `date-fns`

shaundon:
The changes were quite substantial so I had to update a lot to get it all working. I'll add inline comments for context where needed.

### DIFF
--- a/packages/bpk-component-calendar/README.md
+++ b/packages/bpk-component-calendar/README.md
@@ -16,8 +16,8 @@ import BpkCalendar from 'bpk-component-calendar';
 import BpkInput, { INPUT_TYPES } from 'bpk-component-input';
 import format from 'date-fns/format';
 
-const formatDateFull = date => format(date, 'dddd, Do MMMM YYYY');
-const formatMonth = date => format(date, 'MMMM YYYY');
+const formatDateFull = date => format(date, 'EEEE, do MMMM yyyy');
+const formatMonth = date => format(date, 'MMMM yyyy');
 const daysOfWeek = [
   {
     name: 'Sunday',
@@ -313,7 +313,7 @@ A function to format a full, human-readable date, for example: "Monday, 8th Janu
 ```js
 import format from 'date-fns/format';
 
-const formatDateFull = date => format(date, 'dddd, Do MMMM YYYY');
+const formatDateFull = date => format(date, 'EEEE, do MMMM yyyy');
 ```
 
 #### formatMonth
@@ -325,7 +325,7 @@ If you just need to quickly prototype, use the following from [`date-fns`](https
 ```js
 import format from 'date-fns/format';
 
-const formatMonth = date => format(date, 'MMMM YYYY');
+const formatMonth = date => format(date, 'MMMM yyyy');
 ```
 
 #### weekStartsOn
--- a/packages/bpk-component-calendar/examples-components.js
+++ b/packages/bpk-component-calendar/examples-components.js
@@ -21,7 +21,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import BpkButton from 'bpk-component-button';
-import addMonths from 'date-fns/add_months';
+import addMonths from 'date-fns/addMonths';
 import {
   colorWhite,
   colorPanjin,
--- a/packages/bpk-component-calendar/examples.js
+++ b/packages/bpk-component-calendar/examples.js
@@ -349,13 +349,13 @@ const WeekExample = () => {
         <tbody>
           <Week
             {...weekProps}
-            selectionStart={parseDate(`19800611`)}
-            selectionEnd={parseDate(`19800616`)}
+            selectionStart={parseDate(`19800611`, 'yyyyMMdd', new Date())}
+            selectionEnd={parseDate(`19800616`, 'yyyyMMdd', new Date())}
           />
           <Week
             {...weekProps}
-            selectionStart={parseDate(`19800601`)}
-            selectionEnd={parseDate(`19800607`)}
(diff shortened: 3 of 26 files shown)


packages/bpk-component-scrollable-calendar/examples.js
```diff
@@ -96,9 +96,9 @@ const DefaultExample = () => (
   />
 );
 
-const WeekStartsOnSeven = () => (
+const WeekStartsOnSix = () => (
   <ScrollableCal
-    weekStartsOn={7}
+    weekStartsOn={6}
```
shaundon, line 101: `date-fns` now throws an error if you use a number not between 0-6 for this, so I changed it.

packages/bpk-component-calendar/README.md
```diff
@@ -16,8 +16,8 @@ import BpkCalendar from 'bpk-component-calendar';
 import BpkInput, { INPUT_TYPES } from 'bpk-component-input';
 import format from 'date-fns/format';
 
-const formatDateFull = date => format(date, 'dddd, Do MMMM YYYY');
```
shaundon, line 19: Anywhere you see a change in the format like this, it's because 1.x was using a slightly custom format, and in 2.x they updated it to be consistent with ISO formatting for dates. It doesn't change what users see, as evidenced by the snapshots not changing.

packages/bpk-component-calendar/examples-components.js
```diff
@@ -21,7 +21,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import BpkButton from 'bpk-component-button';
-import addMonths from 'date-fns/add_months';
+import addMonths from 'date-fns/addMonths';
```
shaundon, line 24: 2.x changed all imports to use camelCase instead of snake_case. There are a few places you see this throughout this PR.

packages/bpk-component-calendar/examples.js
```diff
@@ -349,13 +349,13 @@ const WeekExample = () => {
         <tbody>
           <Week
             {...weekProps}
-            selectionStart={parseDate(`19800611`)}
-            selectionEnd={parseDate(`19800616`)}
+            selectionStart={parseDate(`19800611`, 'yyyyMMdd', new Date())}
+            selectionEnd={parseDate(`19800616`, 'yyyyMMdd', new Date())}
```
shaundon, line 353: `parseDate` now requires you to specify how it's formatted, hence the addition of `yyyyMMdd`. The `new Date()` part is a backup to return if it couldn't parse it.

packages/bpk-component-calendar/src/BpkCalendarGridTransition.js
```diff
@@ -157,7 +161,7 @@ class BpkCalendarGridTransition extends Component {
           onTransitionEnd={this.onMonthTransitionEnd}
         >
           {this.state.months.map((m, index) =>
-            isWithinRange(m, min, max) ? (
+            min && max && isWithinRange(m, { start: min, end: max }) ? (
```
shaundon, line 164: `isWithinRange` now throws an exception if you don't pass in valid dates, so we use this check to make sure we don't do that.

packages/bpk-component-calendar/src/BpkCalendarGridTransition.js
```diff
@@ -147,7 +147,11 @@ class BpkCalendarGridTransition extends Component {
       isTransitioning && 'bpk-calendar-grid-transition__strip--transitioning',
     );
 
-    const { min, max } = getMonthRange(rest.minDate, rest.maxDate);
+    let min;
+    let max;
+    if (rest.minDate && rest.maxDate) {
+      ({ min, max } = getMonthRange(rest.minDate, rest.maxDate));
```
shaundon, line 153: This check ensures that `getMonthRange` isn't called on `null`, which now returns `Invalid Date`.

packages/bpk-component-calendar/src/BpkCalendarGridTransition.js
```diff
@@ -147,7 +147,11 @@ class BpkCalendarGridTransition extends Component {
       isTransitioning && 'bpk-calendar-grid-transition__strip--transitioning',
     );
 
-    const { min, max } = getMonthRange(rest.minDate, rest.maxDate);
+    let min;
+    let max;
```
georgegillams, line 151: I think this could be changed to 1 line but not a biggie
```suggestion
    let min, max;
```

packages/bpk-component-scrollable-calendar/stories.js
```diff
@@ -20,7 +20,7 @@ import { storiesOf } from '@storybook/react';
 
 import {
   DefaultExample,
-  WeekStartsOnSeven,
+  WeekStartsOnSix,
```
georgegillams, line 23: Is this correct? Used to be `Seven` and now it's `Six`?

georgegillams (commented):
Just a couple of minor comments

packages/bpk-component-scrollable-calendar/stories.js
shaundon: I left a comment elsewhere about this – date-fns now throws an error if you pass in anything other than `0-6` for this, so I changed it.

packages/bpk-component-calendar/src/BpkCalendarGridTransition.js
shaundon: Nice idea

shaundon: lol this broke CI – eslint forbids this. I'll revert

georgegillams (approved):

PR #2164, opened 2021-05-04
Upgrading backpack node sass to add licensing to .css files

olliecurtis:
Upgrading backpack-node-sass package to the latest and utilising the `--prefixComment` command to add licenses to our generated CSS files


shaundon (approved):

scripts/check-pristine-state
```diff
@@ -32,6 +32,14 @@ const init = async () => {
         ', ',
       )}). Changed files: ${changedFiles.join(', ')}`,
     );
+    try {
```
olliecurtis, line 35: This is just to debug why CI is reporting changes when I don't get any locally

PR #2176, opened 2021-05-07
[KOA-4450] Add screenshot testing

shaundon:
This PR adds visual regression tests (VRTs), which run in CI.

### About the tests

They use [Storyshots](https://www.npmjs.com/package/@storybook/addon-storyshots) and [jest-image-snapshot](https://www.npmjs.com/package/jest-image-snapshot), which I based from how Acorn does VRTs. They currently only run in Chromium. We may be able to tweak this later to add more, but this should deliver a good amount of value for now.

To prevent the tests taking lots of time to run, each component has a story called 'Visual tests', and only these run. Also we don't test anything that can't be tested easily, like the modal component (as the story is only valuable if interacted with).

### How to run them

The tests use the **built** Storybook instance, in the `dist-storybook/` folder.

```bash
# if you don't already have a built Storybook
npm run storybook:dist

npm run jest:visual-tests
npm run jest:visual-tests:update
```

### On CI

If the tests fail, the failure diffs are uploaded as an artifact for you to download, and CI will post a message notifying you:

Screen Shot 2021-05-20 at 10 09 15

If they pass, you get a message to make it clear that they only run on Chromium to avoid overconfidence:

Screen Shot 2021-05-20 at 10 37 48


.storybook/storyshots-test.js
```diff
@@ -0,0 +1,51 @@
...
+import { imageSnapshot } from '@storybook/addon-storyshots-puppeteer';
+
+// This small timeout is to allow the component to settle.
+// See: https://www.npmjs.com/package/@storybook/addon-storyshots-puppeteer#specifying-options-to-jest-image-snapshots
+// Its primary function is to delay the screenshot being taken to prevent regressions due to mounting animations & images loading.
+const beforeScreenshot = () =>
+  new Promise(resolve =>
+    setTimeout(() => {
+      resolve();
+    }, 800),
```
georgegillams, line 33: This might be more self-documenting if we make this a variable called `DELAY_MS` or something?

CONTRIBUTING.md
```diff
@@ -155,7 +155,6 @@ We use [CSS Modules](https://github.com/css-modules/css-modules) along with [BEM
 To maintain React 16.4 compatibility, please do not use newer features of React such as Hooks.
 
 When creating (S)CSS files, follow the CSS Module naming convention by using the `.module.(s)css` extension.
-
```
georgegillams, line 158: Think this should still be here to ensure correct processing of headers?

.storybook/storyshots-test.js
shaundon: Yeah excellent idea, will change it.

packages/bpk-component-code/examples.js
```diff
@@ -44,4 +44,13 @@ const BlockAlternate = () => (
...
+const MixedExample = () => (
+  <div>
+    <Inline />
+    <InlineAlternate />
+    <Block />
+    <BlockAlternate />
+  </div>
+);
+
+export { Inline, InlineAlternate, Block, BlockAlternate, MixedExample };
```
georgegillams, line 56: I wonder if we should have a function called `withMixedExample` so we can do `export withMixedExample({e1, e2...})`

Something like
```
withMixedExample = examples => {
  const MixedExample = <div>{Object.keys(examples).map(Example => <Example />)}</div>;
  return {...examples, MixedExample}
}
```

georgegillams (approved):

CONTRIBUTING.md
shaundon: Yeah I'm not sure how that got removed! I'll add it back in

PR #2205, opened 2021-06-23
[No JIRA] Create chip radiogroup example

k0nserv:

- [ ] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-chip/examples.js
+++ b/packages/bpk-component-chip/examples.js
@@ -21,7 +21,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { cssModules } from 'bpk-react-utils';
-import BpkText, { TEXT_STYLES } from 'bpk-component-text';
+import BpkText, { TEXT_STYLES, WEIGHT_STYLES } from 'bpk-component-text';
 import BeachIconSm from 'bpk-component-icon/sm/beach';
 import CarWashIconSm from 'bpk-component-icon/sm/car-wash';
 import PlusIconSm from 'bpk-component-icon/sm/plus';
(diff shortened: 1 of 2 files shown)


packages/bpk-component-chip/examples.js
```diff
@@ -200,6 +268,10 @@ const DismissibleChipsExample = () => (
 
 const SelectableChipsExample = () => <StatefulSelectableChips />;
 
+const RadiogroupChipsExample = () => (
+  <StatefulRadioGroupChipsExample type={CHIP_TYPES.primar} />
```
olliecurtis, line 272:
```suggestion
  <StatefulRadioGroupChipsExample type={CHIP_TYPES.primary} />
```

packages/bpk-component-chip/examples.js
```diff
@@ -200,6 +268,10 @@ const DismissibleChipsExample = () => (
 
 const SelectableChipsExample = () => <StatefulSelectableChips />;
 
+const RadiogroupChipsExample = () => (
```
olliecurtis, line 271:
```suggestion
const RadioGroupChipsExample = () => (
```

packages/bpk-component-chip/examples.js
k0nserv: Thanks :) I think parts of my editor experience is broken since this should've been caught. Haven't done web in a long time

packages/bpk-component-chip/examples.js
k0nserv: Open question if it should be one or the other?

olliecurtis (changes requested):
Left some suggestions for naming improvements and a fix for incorrect prop

k0nserv:
Why is `accessibilityLabel` required on `BpkSelectableChip`? It feels like it shouldn't be, at least in this instance it doesn't add anything

packages/bpk-component-chip/examples.js
```diff
@@ -151,6 +151,74 @@ class StatefulDismissibleChipsExample extends React.Component<
...
+              {chip}
+            </BpkSelectableChip>
+          ))}
+        </div>
+        <AriaLiveDemo
+          className={getClassName('bpk-banner-alert-examples__component')}
+        >
+          {this.state.updates.map(u => (
+            <>
+              {u}
```
shaundon, line 212: Single-letter variable names?

i don't think

Seriously though, can't hurt to make it more explicit and aid self documentation.

packages/bpk-component-chip/examples.js
shaundon: React's own naming convention is with Ollie – `Radiogroup` isn't a valid word, so it should be `RadioGroup` to reflect that.

olliecurtis (approved):

PR #2206, opened 2021-06-23
Modify banner alert roles for accessibility

AHB:
`<BpkBannerAlert>` with `isExpandable` would result in a role of `alert button` which was being flagged as invalid

Instead I've moved the `role=alert` up to the section, rather than the `div`, which is applicable for all banner alerts, and then if it is an expandable alert, add `role=button` on the div as it used to.

This way there are two elements with a single role, rather than one element with two roles on

- [X] `UNRELEASED.md`
- [ ] `README.md`
- [X] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,4 @@
+**Fixed:**
+
+- bpk-component-banner-alert:
+  - Updated role attribute for expandable alerts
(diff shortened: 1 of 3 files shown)


georgegillams (approved):
LGTM 👍

PR #2213, opened 2021-07-08
[KOA-4554]: Badge text style update

olliecurtis:
**Before:**
Screenshot 2021-07-08 at 17 06 02

**After:**
Screenshot 2021-07-08 at 17 03 22

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [x] Tests


shaundon (approved):
Approved, but isn't it kinda weird that the visual tests passed? I guess this was below the threshold that would trigger a fail

PR #2215, opened 2021-07-13
[KOA-4552]: BpkSwitch to use new spacing system

olliecurtis:
**Before:**
Screenshot 2021-07-13 at 15 21 22

**After:**
Screenshot 2021-07-22 at 17 28 01

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [x] Tests


packages/bpk-component-switch/src/BpkSwitch.module.scss
```diff
@@ -18,13 +18,15 @@
...
+
 $primary-on-color: $bpk-color-sky-blue;
 $event-on-color: $bpk-color-abisko;
 $off-color: $bpk-color-sky-gray-tint-06;
 $handle-color: $bpk-color-white;
 
-$height: $bpk-spacing-lg;
-$width: $bpk-spacing-base * 2;
+$height: bpk-spacing-xl();
+$width: bpk-spacing-lg() * 2;
```
georgegillams, line 29: I'm not quite sure about these changes. The switch just doesn't quite look right here 🤔

georgegillams: It it is going to be taller, I think it needs to be wider too to maintain proportions. wdyt?

olliecurtis: I was wondering about this too and it not looking too right. I tried to make it a bit wider but would possibly mean making the track wider too which then made it look a bit weirder 🤔

olliecurtis: Might be similar to our checkbox and radio components we make this one use the sizes it was before with the magic of CSS maths just to make it look balanced otherwise either way it might look a bit off, wdyt? Will also check in with design on this too

marianeum (approved):

PR #2217, opened 2021-07-20
NO-JIRA: Update react-responsive to v6 in bpk-component-breakpoint

dominicfraser:
`bpk-component-breakpoint` uses an older version of `react-responsive` that produces the following warnings when running locally:

Screenshot 2021-07-20 at 11 25 27

This PR updates to a newer version, but still not the latest.

v6 should fix this and requires 16.3.0, while v7+ require 16.8.0 which Backpack currently does not support as a minimum version.

https://github.com/contra/react-responsive/blob/master/CHANGELOG.md#v600

- [x] `UNRELEASED.md`
- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,2 +1,4 @@
+**Fixed**
+
 - bpk-component-breakpoint:
   - Updated `react-responsive` dependency to 6.1.2. This requires a minimum React version of 16.3.0 as it migrates some life cycle methods to their `UNSAFE_` names.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,2 @@
+- bpk-component-breakpoint:
+  - Updated `react-responsive` dependency to 6.1.2. This requires a minimum React version of 16.3.0 as it migrates some life cycle methods to their `UNSAFE_` names.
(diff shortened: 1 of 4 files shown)


UNRELEASED.md
```diff
@@ -0,0 +1,2 @@
+- bpk-component-breakpoint:
```
olliecurtis, line 1: Can you add if this change is 'Breaking', 'Added' or 'Fixed' as per https://github.com/Skyscanner/backpack/blob/main/CHANGELOG_FORMAT.md?

olliecurtis (changes requested):
Looks good to me, just one minor comment, otherwise will test and see if all good!

UNRELEASED.md
dominicfraser: Ahh, this used to be listed in UNRELEASED.md - it was gone and couldn't find it.

Should I update the PR description or UNRELEASED.md itself to point to it?

olliecurtis: Yeah we had moved it for automation purposes when we release, so now this is kept in a separate file!

olliecurtis (approved):

PR #2218, opened 2021-07-21
[KOA-3661]: Slider spacing to 4px grid

olliecurtis:
**Before:**

Slider component before

**After:**

Slider component after

- [ ] `README.md`
- [x] Tests


packages/bpk-component-slider/src/BpkSlider.module.scss
```diff
@@ -18,20 +18,24 @@
...
+$slider-large-control-inset-size: bpk-spacing-sm() * 7;
 
 .bpk-slider {
-  height: $bpk-spacing-lg;
+  height: bpk-spacing-xl();
   -webkit-tap-highlight-color: transparent;
 
   &--large {
-    height: $bpk-spacing-xxl;
+    height: bpk-spacing-xxl();
```
marianeum, line 31: should one of these be width or what effect does defining height twice have?

olliecurtis: Ah this was one value I missed whilst testing sizes out and forgot to clean up! Good spot, it just needs to be one height definition

marianeum (approved):

PR #2220, opened 2021-07-28
[no jira]: Fix babel config for storybook to fix rendering issues on browsers

olliecurtis:

### DIFF
--- a/.storybook/babel.config.js
+++ b/.storybook/babel.config.js
@@ -16,6 +16,13 @@
  * limitations under the License.
  */
 
+/*
+ * The purpose of this file is to provide babel config for configuration and use with storybook.
+ *
+ * For babel config used in building and distributing Backpack components, there is a separate file
+ * located in the root of the project.
+ */
+
 module.exports = {
   presets: [
     [
--- a/.storybook/babel.config.js
+++ b/.storybook/babel.config.js
@@ -16,4 +16,35 @@
  * limitations under the License.
  */
 
-module.exports = require('../babel.config.js');
+module.exports = {
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        targets: {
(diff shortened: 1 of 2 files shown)


babel.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
       '@babel/preset-env',
       {
         targets: {
+          browsers: ['ie >= 11'],
```
olliecurtis, line 25: Setting this allows storybook to render correctly on ie11 to avoid having to directly go to the component URL

.storybook/babel.config.js
```diff
@@ -16,4 +16,35 @@
  * limitations under the License.
  */
 
-module.exports = require('../babel.config.js');
+module.exports = {
```
olliecurtis, line 19: This file now writes the babel config instead of importing it as we need to set the babel targets which we can't do at the root as it causes compliation errors for our outputed css in bpk-stylesheets, so we just modify the babel config to accept the correct targets

georgegillams: As a lot of this file is similar, it might be worth importing the root one as before and then changing anything we need to?
However I can see the counter-argument that says these are effectively config files for 2 separate use cases, so duplicating isn't necessarily a bad idea?

georgegillams: Might be worth adding a comment to the root one though stating that the storybook config is separate/different

olliecurtis: Sure will add the appropriate comments! 

I tried to keep them the same and just doing an import and modifying all that we needed but build/running didn't like that.

So the only option was to have a separate config, but as the point you mentioned they are specific configs for different usecases so should be ok.

georgegillams (approved):
tenor-205780809

PR #2221, opened 2021-07-30
[KOA-3622]: Select spacing updates

olliecurtis:
**Before:**

See it running: https://backpack.github.io/storybook/?path=/story/bpk-component-select--example

Default Size:
Screenshot 2021-07-30 at 09 47 18

Large Size:
Screenshot 2021-07-30 at 09 47 57

Default with image:
Screenshot 2021-07-30 at 09 49 41

Large with image:
Screenshot 2021-07-30 at 09 50 03

**After:**

See it running: https://backpack.github.io/storybook-prs/2221/?path=/story/bpk-component-select--example

Default Size:
Screenshot 2021-07-30 at 09 45 48

Large Size:
Screensho0t 2021-07-30 at 09 45 54

Default with image:
Screenshot 2021-07-30 at 09 50 21

Large with image:
Screenshot 2021-07-30 at 09 50 37

- [N/A] `README.md`
- [x] Tests


olliecurtis:
There will be some screenshots updated in this PR due to the select component being used for theme dropdown

UNRELEASED.md
```diff
@@ -0,0 +1,4 @@
+**Fixed:**
```
olliecurtis, line 1: Am wondering if actually this should be breaking as visually it is bigger so will have an impact on alignment when in situ

georgegillams: Breaking is probably safer. Just try to make it clear that it's visually breaking only, to not put folk off upgrading 👍

packages/bpk-component-select/src/BpkSelect.module.scss
```diff
@@ -42,38 +44,42 @@
...
+        /* stylelint-disable-next-line declaration-no-important */
+        padding-right: bpk-spacing-xxl() !important;
+        /* stylelint-disable-next-line declaration-no-important */
+        padding-left: bpk-spacing-base() !important;
       }
     }
 
     @media screen\0 {
-      padding-left: $bpk-spacing-xl !important; /* stylelint-disable-line declaration-no-important */
+      /* stylelint-disable-next-line declaration-no-important */
```
georgegillams, line 62: I appreciate you didn't add these, but every lint-disable comment should be accompanied by an explanation.
I assume the reason for these is that we need to ensure we override the browser styles?

georgegillams: A single explanation at the top of the file should be sufficient

georgegillams (approved):

UNRELEASED.md
olliecurtis: Sounds good to me!

packages/bpk-component-select/src/BpkSelect.module.scss
olliecurtis: Good shout! Better to improve things now as we are making changes. Yeah I assume that is why, its an IE override thing but will put that in a comment at the top!

PR #2224, opened 2021-08-05
[no-jira]: fixing bpk dependencies check script to support @skyscanner packages

olliecurtis:
Whilst reviewing dependency versions I found that the `check-bpk-dependencies` script didn't work as expected for working with the external foundations packages and therefore not updating foundations dependencies correctly in projects (particularly if we run major releases).

This PR solves this and fixes the package.json files to update to the latest versions of foundations and svgs (not lockfiles as these install in range already).

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "create-component": "node scripts/npm/create-component.js",
     "danger": "danger ci",
     "fix-bpk-dependencies": "node scripts/npm/check-bpk-dependencies.js --fix",
+    "upgrade-foundations": "node scripts/npm/check-bpk-dependencies.js --fix --upgrade-foundations",
     "flow": "flow --max-warnings 0",
     "flow-typed": "flow-typed",
     "format:packagejson": "format-package -w",
--- a/packages/bpk-animate-height/package.json
+++ b/packages/bpk-animate-height/package.json
@@ -20,7 +20,7 @@
     "react": "^16.3.0"
   },
   "devDependencies": {
-    "@skyscanner/bpk-foundations-web": "^1.0.0",
+    "@skyscanner/bpk-foundations-web": "^1.1.0",
     "bpk-component-button": "^5.0.4"
   }
 }
--- a/packages/bpk-component-accordion/package.json
+++ b/packages/bpk-component-accordion/package.json
@@ -25,7 +25,7 @@
     "react": "^16.3.0"
   },
   "devDependencies": {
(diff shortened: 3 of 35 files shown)


georgegillams (changes requested):
I think this is going to do more harm than good. Every time someone makes a PR CI will fail if a new version of an external dependency happens to have been released.

The purpose of this script was to ensure that we don't create the cross-dependencies "invalid" by referencing versions that aren't defined in the mono-repo.

If we want to enforce consistent dependency versions in each package, maybe we should have a config file at the top of the tree with `commonPackageVersions`. This script could then ensure that any package using those dependencies is on the aligned version.

olliecurtis:
@georgegillams thanks for confirmation on this point, I guess we could split the foundations logic to a separate script.

To just make any bumps we make and may need in the future easier? That way we can run one command that would help us go into each package and update as it seems like dependabot isn't seeing the updates and raising PRs for it.

This was my thinking about the purpose of the existing script but I was wrong which caused this confusion

georgegillams:
> @georgegillams thanks for confirmation on this point, I guess we could split the foundations logic to a separate script.
> 
> To just make any bumps we make and may need in the future easier? That way we can run one command that would help us go into each package and update as it seems like dependabot isn't seeing the updates and raising PRs for it.
> 
> This was my thinking about the purpose of the existing script but I was wrong which caused this confusion

Yeah I think that's a good idea. We could still potentially share some logic with this script, but I just think we need the foundations versions to be explicit instead of being pulled from NPM.
What do you think of having a `common deps` file at the top of the repo? I guess we could use it to ensure that react versions etc are also aligned?

olliecurtis:
> Yeah I think that's a good idea. We could still potentially share some logic with this script, but I just think we need the foundations versions to be explicit instead of being pulled from NPM.

Just out of interest why is that and how would we keep it up to date? The way I am seeing the script is just purely from a automated process of doing the find and replace or going into each package individually and running the dependency update.

For me thats what I thought the original script was for so I was just catering for the ease of developer experience. All this pain will go away in the future though when we move to one lib as it will be one dependency defined at the root like we do in backpack-react-native, so might just close this and ignore for now.

> I guess we could use it to ensure that react versions etc are also aligned?

We already have a script for this that I created some time ago to make sure that the version we use in this project is aligned to the company version 😄

georgegillams (approved):

PR #2238, opened 2021-08-12
[KOA-4704]Fix tertiary label wrapping when using hanzi (Chinese characters)

giriawu:
## Context

See Slack thread: https://skyscanner.slack.com/archives/C0JHPDSSU/p1628653629032400

### Why Are We Making This Change?

`BpkAutosuggestSuggestion` tertiary label will force line breaks for Asian characters, which will occupy a large height when the word is long. 

### What Predictions Do We Have About When This Goes Live? 

- FIxed the bug and make sure that hanzi will look correct in tertiary label.

### Change Description

- add `word-break: keep-all;` to `BpkAutosuggest.module.scss` to adjust hanzi but keep other characters normal
- add releted tests

#### Mobile

| Before                                                       | After                                                        |
| ------------------------------------------------------------ | ------------------------------------------------------------ |
| image-20210812145006273 | image-20210812145048669 |

#### Desktop

| Before                                                       | After                                                        |
| ------------------------------------------------------------ | ------------------------------------------------------------ |
| image-20210812144826782 | image-20210812144800106 |

- [ ] `README.md`
- [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,5 @@
+# 2021-08-11
+**Fixed:**
+
+- `bpk-component-autosuggest`
+  - Fix tertiary label wrapping when using `hanzi` (Chinese characters) to make sure it won't occupy a large height.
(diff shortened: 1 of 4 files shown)


georgegillams (commented):
Thanks for contributing this fix!
To get CI to pass would you be able to just run `npm run build` again and check in the CSS changes?
Once that's done I can merge and release 👍

georgegillams (approved):

PR #2242, opened 2021-08-12
[no jira]: Allow CI to run on PRs to PRs

olliecurtis:
This PR allows CI to run on both PRs to main but also on pull requests to open pull requests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
 
 defaults:
   run:
--- a/packages/bpk-component-datepicker/package.json
+++ b/packages/bpk-component-datepicker/package.json
@@ -15,8 +15,8 @@
   "gitHead": "f9f94f86d7cfc75e43bc6ad5eb01b23e62a89cc8",
   "dependencies": {
     "bpk-component-breakpoint": "^3.0.7",
-    "bpk-component-calendar": "^9.0.12",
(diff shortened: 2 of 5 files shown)


georgegillams (approved):

.github/workflows/ci.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
```
georgegillams, line 5: Should this go too?

olliecurtis: No as I think we will still want to run CI on main after we merge? Just to make sure things are ok and if we didn't rebase before clicking merge

georgegillams: Ah right.
If I create the branch structure `main <-- A <-- B`, create a PR for B, and then push additional changes to B, will this mean it won't re-run on push?

olliecurtis: It should do as its still a PR, and therefore satisfies the conditions of the logic for running it.

So the `push` option says when a push is made to main then it can be run (so PRs being merged)

The the `pull_request` option says to run on any PRs, so in the old logic 

```
pull_request:
   branches: [main]
```

It was basically telling CI that it should only run on PRs that are made to main, so even with additional commits and activity on the PR it would still run as the changes are on a PR to main.

Whereas now we are saying whereever there is a PR regardless of branch run the CI, so it will run on `main <-- A` and also now `A <-- B`

Hope that makes sense?

georgegillams (approved):

PR #2248, opened 2021-08-17
[KOA-3657] Update nudger component to use the new spacing grid

anambl:
Before: https://backpack.github.io/storybook/?path=/story/bpk-component-nudger--default
After: https://backpack.github.io/storybook-prs/2248/?path=/story/bpk-component-nudger--default

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-nudger/src/BpkNudger.module.scss
+++ b/packages/bpk-component-nudger/src/BpkNudger.module.scss
@@ -18,6 +18,8 @@
 
 @import '~bpk-mixins/index';
 
+$bpk-spacing-v2: true;
+
 .bpk-nudger {
   &__button {
     position: relative;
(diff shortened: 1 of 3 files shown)


packages/bpk-component-nudger/src/BpkNudger.module.scss
```diff
@@ -39,7 +41,7 @@
     }
 
     &--numeric {
-      $size: $bpk-button-line-height + ($bpk-button-padding-y * 2);
+      $size: bpk-spacing-lg() + (bpk-spacing-md() * 2);
```
olliecurtis, line 44: For the `$bpk-button-line-height`, has this changed to aid with the new grid size, as keeping it in use causes the size to be off?

olliecurtis (changes requested):
Just a small comment from me otherwise looks great!

olliecurtis (approved):
LGTM!

packages/bpk-component-nudger/src/BpkNudger.module.scss
anambl: `$bpk-button-line-height` uses a spacing token under the hood and also it's used for setting the width

PR #2252, opened 2021-08-23
Update pagination component to use the new spacing grid

anambl:
|Before https://backpack.github.io/storybook/?path=/story/bpk-component-pagination--pagination-default|After https://backpack.github.io/storybook-prs/2252/?path=/story/bpk-component-pagination--pagination-default|
|-|-|
|image|image|

- [ ] `README.md`
- [ ] Tests


packages/bpk-component-pagination/src/BpkPaginationPage.module.scss
```diff
@@ -18,11 +18,13 @@
...
+$bpk-spacing-v2: true;
+
 .bpk-pagination-page {
-  min-width: $bpk-spacing-xl;
-  height: $bpk-spacing-xl;
+  min-width: bpk-spacing-xxl();
+  height: bpk-spacing-xxl();
   padding: 0;
-  border-radius: $bpk-border-radius-pill;
+  border-radius: $bpk-border-radius-pill-lg;
```
k0nserv, line 27: Should we just use 50% instead? The intent here is to have a circle and while `border-radius: $bpk-border-radius-pill-lg;` accidentally achieves this it seems more future proof to just use 50%.

anambl: It seems like the tokens are used somewhat more frequently in our code rather than percentages so it seemed like a better choice in terms of consistency. Happy to use 50% if that's safer and we know it will most likely always remain a circle

k0nserv: You're definitely correct, but I feel like 50% better expresses the intent here. Maybe we can see what @olliecurtis and @georgegillams thinks

georgegillams: That's definitely not what we want...

Screenshot 2021-08-23 at 18 25 22

georgegillams: Oh this is for the pagination ones. Yeah as they're circular that works 👍

georgegillams: I agree with @k0nserv that `50%` expresses the intent more accurately.

anambl: Makes sense 👍 Thanks both!

k0nserv (approved):
Code LGTM

packages/bpk-component-pagination/src/BpkPaginationPage.module.scss
```diff
@@ -18,11 +18,13 @@
 
 @import '~bpk-mixins/index';
 
+$bpk-spacing-v2: true;
+
 .bpk-pagination-page {
-  min-width: $bpk-spacing-xl;
-  height: $bpk-spacing-xl;
+  min-width: bpk-spacing-xl() + bpk-spacing-sm();
```
georgegillams, line 24: Is it worth including a comment to explain why we use this combo? (ie to match the height of the button)

georgegillams (approved):

packages/bpk-component-pagination/src/BpkPaginationPage.module.scss
anambl: I think it's worth having this info in Figma perhaps as it's a design decision rather than having it in our code?

olliecurtis: Plus one to @anambl comment - given this is something from design rather than us and to keep in line with our own linting tools (instead of using a raw value), this makes sense.

As a lot of the changes might be just 'because that is the design given' vs. having a specific reason as to why we do this in code. We are just 'codifying the design language' and not necessarily having a reason for making a combo

olliecurtis (approved):
LGTM - Ship it!

PR #2260, opened 2021-09-08
[NO-JIRA] Avoid Google translate numbers in BpkCalendarDate

seraco:
…nslate translating numbers

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-calendar/src/BpkCalendarDate.js
+++ b/packages/bpk-component-calendar/src/BpkCalendarDate.js
@@ -153,7 +153,9 @@ class BpkCalendarDate extends PureComponent {
         ref={this.getButtonRef}
         {...buttonProps}
       >
-        <span aria-hidden="true">{date.getDate()}</span>
+        <span aria-hidden="true" className="notranslate">
+          {date.getDate()}
+        </span>
       </button>
     );
   }
--- a/packages/bpk-component-calendar/src/__snapshots__/BpkCalendarContainer-test.js.snap
+++ b/packages/bpk-component-calendar/src/__snapshots__/BpkCalendarContainer-test.js.snap
@@ -264,6 +264,7 @@ exports[`BpkCalendarContainer should focus the correct date when \`initiallyFocu
                 >
                   <span
                     aria-hidden="true"
+                    class="notranslate"
                   >
                     1
                   </span>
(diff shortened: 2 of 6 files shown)


UNRELEASED.md
```diff
@@ -0,0 +1,36 @@
+# Unreleased
```
olliecurtis, line 1: Can you just keep this file to have, otherwise it breaks our automation and is not how the Unreleased should be setup 😄 

```
**Fixed:**
- `BpkSwitch`:
  - Prevent Google Translate from translating numbers in BpkCalendarDate
 ```

File: UNRELEASED.md [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+
+> Place your changes below this line.
+>
+
+**Fixed:**
+- `BpkSwitch`:
```
olliecurtis, line 7: This should be `bpk-component-calendar`?

olliecurtis (changes requested):
Thanks for the contribution! Just a couple of minor points from me

UNRELEASED.md
olliecurtis: As this also affects the scrollable calendar and the datepicker - might be good to include them in here too

UNRELEASED.md
seraco: Thanks for letting me know. 🙂 Addressed by d22ae113972fcdb3d787b76a256c8d4b26258ca4.

UNRELEASED.md
seraco: Oops, I forgot to change it. Addressed by d22ae113972fcdb3d787b76a256c8d4b26258ca4.

seraco: Also by d0149658957af35617bfe76ac3c2651111733b97.

zgeor:
This seems to be a problem with Google Translate. Navigating to https://translate.google.co.uk/?sl=es&tl=ar&text=11&op=translate shows that the translator makes the decision. Furthermore, if you look at 23 https://translate.google.co.uk/?sl=es&tl=ar&text=23&op=translate you can see that it too appears broken. 

For reference bing.com would make the correct translation in the cases above. 

**Note:** I tried a number of translations from different languages and could not reproduce the issue from above with any pair.

We have submitted feedback to google about the issue. Given this will impact all users using different translation engines, we should not merge this PR.

PR #2261, opened 2021-09-08
[NO JIRA]: Updating propTypes package and fixing incorrect proptypes

olliecurtis:
This PR solves two things:

- Upgrades `propTypes` to latest version to allow use of `elementType` prop ([see more](https://github.com/facebook/prop-types/pull/211))
- Fixed an issue where component props were being set as just `func` but in newer React versions its not strictly the case

- [x] `README.md`
- [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,9 @@
+**Fixed:**
+- Updated `prop-types` to version `15.7.2`
+
+- bpk-component-barchart:
+- bpk-component-calendar:
+- bpk-component-datepicker:
+- bpk-component-infinite-scroll:
+- bpk-theming:
+  - Corrected component props to allow `elementType` for the correct props when components are being used.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,4 +6,4 @@
 - bpk-component-datepicker:
 - bpk-component-infinite-scroll:
 - bpk-theming:
-  - Corrected component props to use `elementType` for the props were a component type is expected.
+  - Corrected component props to use `elementType` for the props where a component type is expected.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,4 +6,4 @@
 - bpk-component-datepicker:
 - bpk-component-infinite-scroll:
 - bpk-theming:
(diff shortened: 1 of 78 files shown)


UNRELEASED.md
```diff
@@ -0,0 +1,9 @@
+**Fixed:**
+- Updated `prop-types` to version `15.7.2`
+
+- bpk-component-barchart:
+- bpk-component-calendar:
+- bpk-component-datepicker:
+- bpk-component-infinite-scroll:
+- bpk-theming:
+  - Corrected component props to allow `elementType` for the correct props when components are being used.
```
k0nserv, line 9:
```suggestion
  - Corrected component props to use `elementType` for the props were a component type is expected.
```

packages/bpk-component-infinite-scroll/src/withInfiniteScroll.js
```diff
@@ -75,8 +75,8 @@ const propTypes = {
   loaderIntersectionTrigger: PropTypes.oneOf(['small', 'half', 'full']),
   onScroll: PropTypes.func,
   onScrollFinished: PropTypes.func,
-  renderLoadingComponent: PropTypes.func,
```
k0nserv, line 78: Both of these are actually treated as functions, not React components so this is not correct. That said I don't think there's a down side to start treating them as components

olliecurtis: Ah yes I see, maybe it does make sense to keep it as that given how its used in comparios to other things

packages/bpk-component-infinite-scroll/README.md
```diff
@@ -170,8 +170,8 @@ Updates the internal array and triggers all listeners.
 | loaderMinDisplay        | oneOf(['small', 'half', 'full']) | false    | 'full'        |
 | onScroll                | func                             | false    | null          |
 | onScrollFinished        | func                             | false    | null          |
-| renderLoadingComponent  | func                             | false    | null          |
```
k0nserv, line 173: This should be changed back to `func` too

olliecurtis: Good catch! Solved!

k0nserv (approved):

PR #2269, opened 2021-10-01
[KOA-4759]: Add new props to support ranges

olliecurtis:

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,5 @@
+**Added:**
+
+- bpk-component-calendar:
+  - Added new `calendarConfiguration` property to Backpack calendar to prepare for range support.
+  - Deprecated `selectedDate` property in favour of `calendarConfiguration`.
--- a/packages/bpk-component-calendar/README.md
+++ b/packages/bpk-component-calendar/README.md
@@ -173,6 +173,7 @@ withCalendarState(composeCalendar(
 | onDateSelect          | func                 | false               | null             |
 | onMonthChange         | func                 | false               | null             |
 | selectedDate          | Date                 | false               | null             |
+| calendarConfiguration | object               | false               | { type: 'single', date: null }  |
 | showWeekendSeparator  | bool                 | false               | true             |
 | navProps              | object               | false               | null             |
 | headerProps           | object               | false               | null             |
@@ -227,6 +228,7 @@ The BpkCalendarGrid component displays a month as a table.
 | formatMonth           | func                 | true     | -                |
 | month                 | Date                 | true     | -                |
 | weekStartsOn          | number               | true     | -                |
+| calendarConfiguration | object               | false    | { type: 'single', date: null }  |
 | focusedDate           | Date                 | false    | null             |
 | isKeyboardFocusable   | bool                 | false    | true             |
 | markOutsideDays       | bool                 | false    | true             |
@@ -248,6 +250,7 @@ The BpkCalendarDate component is used to render the content of a cell
 | --------------------- | -------------------- | -------- | ---------------- |
 | date                  | Date                 | true     | -                |
 | cellType              | oneOf(CELL_TYPES.default, CELL_TYPES.positive, CELL_TYPES.neutral, CELL_TYPES.negative)                     | false    | null             |
+| dateType              | oneOf(DATE_TYPES.single, DATE_TYPES.start, DATE_TYPES.middle, DATE_TYPES.end)                               | false    | DATE_TYPES.single             |
 | isBlocked             | bool                 | false    | false            |
 | isFocused             | bool                 | false    | false            |
 | isKeyboardFocusable   | bool                 | false    | true             |
@@ -268,7 +271,39 @@ This prop determines what the colour date cell is to be displayed.
 - `CELL_TYPES.negative` - sets the calendar cell to `Panjin`
 - `CELL_TYPES.default` - sets the calendar cell to `Sky Gray Tint 02`
 
+#### `dateType` prop
+
+This property determines which selected styles will be applied to the date cell. If using ranges use `start`, `middle` and `end` to apply the correct range styles.
+
+- `DATE_TYPES.single` - sets the calendar cell to standard `Sky Blue` colour
+- `DATE_TYPES.start` - sets the calendar cell to `Sky Blue` along with the start `Sky Blue Tint 02` range background.
+- `DATE_TYPES.middle` - sets the calendar cell to `Sky Blue Tint 02` background.
+- `DATE_TYPES.end` - sets the calendar cell to `Sky Blue` along with the end `Sky Blue Tint 02` range background.
+
 ### Prop details
+
+#### calendarConfiguration
+
+An object to indicate which configuration of the calendar is being used. Choices are `single` date selection or `range` date selection.
+
+##### Single date
+
+```json
+{
+  type: 'single',
+  date: date e.g. new Date()
+}
+```
+
+##### Range
+
+```json
+{
+  type: 'range',
+  startDate: date e.g. new Date(),
+  endDate: date e.g. new Date()
+}
+```
 
 #### daysOfWeek
 
--- a/packages/bpk-component-calendar/src/BpkCalendarDate.js
+++ b/packages/bpk-component-calendar/src/BpkCalendarDate.js
@@ -32,6 +32,13 @@ export const CELL_TYPES = {
   positive: 'positive',
 };
 
+export const DATE_TYPES = {
+  single: 'single',
+  start: 'start',
+  middle: 'middle',
+  end: 'end',
+};
+
 const navigatedByMonthNudger = () =>
   document.activeElement.id &&
   document.activeElement.id.indexOf('month_nudger') !== -1;
@@ -165,6 +172,7 @@ export const propTypes = {
   // Optional
   cellType: PropTypes.oneOf(Object.keys(CELL_TYPES)),
   className: PropTypes.string,
+  dateType: PropTypes.oneOf(Object.keys(DATE_TYPES)),
   isBlocked: PropTypes.bool,
   isFocused: PropTypes.bool,
   isKeyboardFocusable: PropTypes.bool,
@@ -183,6 +191,7 @@ BpkCalendarDate.propTypes = { ...propTypes };
 export const defaultProps = {
   className: null,
   cellType: null,
+  dateType: DATE_TYPES.single,
   isBlocked: false,
   isFocused: false,
   isKeyboardFocusable: true,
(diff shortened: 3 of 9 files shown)


anambl (approved):
Looks good to me!

packages/bpk-component-calendar/README.md
```diff
@@ -268,8 +271,40 @@ This prop determines what the colour date cell is to be displayed.
 - `CELL_TYPES.negative` - sets the calendar cell to `Panjin`
 - `CELL_TYPES.default` - sets the calendar cell to `Sky Gray Tint 02`
 
+#### `dateType` prop
+
+This property determines which selected styles will be applied to the date cell. If using ranges use `start`, `middle` and `end` to apply the correct range styles.
+
+- `DATE_TYPES.single` - sets the calendar cell to standard `Sky Blue` colour
```
k0nserv, line 278: I think rather than explain the resulting style this should explain when this is used

e.g. "When the date is selected individually i.e. not as part of a range"

packages/bpk-component-calendar/README.md
```diff
@@ -268,8 +271,40 @@ This prop determines what the colour date cell is to be displayed.
...
 
+#### calendarConfiguration
+
+An object to indicate which configuration of the calendar is being used. Choices are `single` date selection or `range` date selection.
+
+##### Single date
+
+```json
+{
+  type: 'single',
```
k0nserv, line 293: Should we have a constant for `single` and `range`? `CALENDAR_SELECTION_TYPE.single`?

packages/bpk-component-calendar/src/BpkCalendarDate.js
```diff
@@ -165,6 +172,7 @@ export const propTypes = {
   // Optional
   cellType: PropTypes.oneOf(Object.keys(CELL_TYPES)),
   className: PropTypes.string,
+  dateType: PropTypes.oneOf(Object.keys(DATE_TYPES)),
```
k0nserv, line 175: `selectionType`?

packages/bpk-component-calendar/src/custom-proptypes.js
```diff
@@ -18,6 +18,22 @@
...
 
+const CalendarConfigurationSingle = PropTypes.shape({
+  type: 'single',
+  date: PropTypes.instanceOf(Date),
+});
+
+const CalendarConfigurationRange = PropTypes.shape({
+  type: 'range',
+  startDate: PropTypes.instanceOf(Date),
+  endDate: PropTypes.instanceOf(Date),
```
k0nserv, line 29: Maybe we ought to have a custom prop type that verifies that `endDate > startDate`?

UNRELEASED.md
```diff
@@ -1,4 +1,7 @@
 **Added:**
 
 - bpk-component-calendar:
-  - Added range support to BpkCalendar
\ No newline at end of file
+  - Added range support to BpkCalendar
+    - Added new `calendarConfiguration` property to Backpack calendar to allow for range support.
```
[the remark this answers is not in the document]
k0nserv: `selectionConfiguration` perhaps?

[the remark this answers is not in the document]
olliecurtis: Sounds good to me, seems to align with the iOS one!

packages/bpk-component-calendar/README.md
olliecurtis: Good idea!

packages/bpk-component-calendar/src/custom-proptypes.js
```diff
@@ -18,20 +18,43 @@
 
 import PropTypes from 'prop-types';
 
-const CalendarConfigurationSingle = PropTypes.shape({
-  type: 'single',
+const startDateType = props => {
+  if (props.startDate > props.endDate) {
```
k0nserv, line 22: I think we also need to explicitly check that they are both defined and handle that too? 

Further, does this get passed only the sub-object  passed as `selectionConfiguration` not the whole props object? 

We might consider adding some tests for this too

packages/bpk-component-calendar/src/custom-proptypes.js
```diff
@@ -18,20 +18,43 @@
...
+    return null;
+  }
+  return new Error(
+    `Start date \`${props.startDate}\` cannot be after end date \`${props.endDate}\`.`,
+  );
+};
+
+const endDateType = props => {
```
k0nserv, line 30: Same question as a above

packages/bpk-component-calendar/README.md
```diff
@@ -250,7 +250,7 @@ The BpkCalendarDate component is used to render the content of a cell
 | --------------------- | -------------------- | -------- | ---------------- |
 | date                  | Date                 | true     | -                |
 | cellType              | oneOf(CELL_TYPES.default, CELL_TYPES.positive, CELL_TYPES.neutral, CELL_TYPES.negative)                     | false    | null             |
-| dateType              | oneOf(DATE_TYPES.single, DATE_TYPES.start, DATE_TYPES.middle, DATE_TYPES.end)                               | false    | DATE_TYPES.single             |
+| selectionType         | oneOf(SELECTION_TYPES.single, SELECTION_TYPES.start, SELECTION_TYPES.middle, SELECTION_TYPES.end)      | false    | SELECTION_TYPES.single             |
```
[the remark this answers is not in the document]
olliecurtis: Probably need a none, our designs don't cover off a same day so probably may need to go back to design for that one actually.

What is a row type? I was looking at the link you have provided but there is little comments/docs around what that does so can't be sure if it needs to be considered for Web. (Here is my interpretation its if the range highlighting needs to go fully across a row if its at the end of a month or start of a month?)

packages/bpk-component-calendar/src/custom-proptypes.js
olliecurtis: Yeah I am not sure what you were getting at with this custom type so was my stab at implementing it. 

> Further, does this get passed only the sub-object passed as selectionConfiguration not the whole props object?

Good question I will check but just reused this thinking from other components, as I am not 100% sure how you would pass just the `selectionConfiguration` option through from the props

would it be `  startDate: startDateType(props.selectionConfiguration.startDate),`? As I don't think it would have specific knowledge of the particular props in this file at runtime

olliecurtis: Also I have also realised the problem of 

`props.startDate > props.endDate` should actually be `props.selectionConfiguration.startDate > props.selectionConfiguration.endDate` as these dates are in that object

packages/bpk-component-calendar/src/custom-proptypes.js
```diff
@@ -18,20 +18,43 @@
...
+  if (props.endDate > props.startDate) {
```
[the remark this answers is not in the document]
olliecurtis: You mean also doing a `===` kinda thing for the dates too?

packages/bpk-component-calendar/src/custom-proptypes.js
olliecurtis: If I am honest with you making a custom type like this came as a suprise as this seems more like logic I was going to be implementing for the range itself vs doing it at the earlier prop checking state

As how would this handle if the consumer selected endDate first instead of startDate and the value is null?

k0nserv: Something like this is what I am thinking

```diff
diff --git a/packages/bpk-component-calendar/src/custom-proptypes.js b/packages/bpk-component-calendar/src/custom-proptypes.js
index 56f522a39..7988b25e7 100644
--- a/packages/bpk-component-calendar/src/custom-proptypes.js
+++ b/packages/bpk-component-calendar/src/custom-proptypes.js
@@ -18,13 +18,35 @@
 
 import PropTypes from 'prop-types';
 
-const startDateType = props => {
-  if (props.startDate > props.endDate) {
+import { isBefore, isSameDay } from './date-utils';
+
+const startDateType = key => props => {
+  const { startDate, endDate } = props[key];
+
+  // No range selected
+  if (!startDate && !endDate) {
     return null;
   }
-  return new Error(
-    `Start date \`${props.startDate}\` cannot be after end date \`${props.endDate}\`.`,
-  );
+
+  // End date with not start date is not allowed
+  if (!startDate && endDate) {
+    return new Error(
+      `Cannot specify \`endDate\` without \`startDate\` in ${key}.`,
+    );
+  }
+
+  // Start date without an end date is always valid
+  if (startDate && !endDate) {
+    return null;
+  }
+
+  if (isBefore(endDate, startDate) && !isSameDay(endDate, startDate)) {
+    return new Error(
+      `Start date \`${startDate}\` cannot be after end date \`${endDate}\` in \`props.${key}\`.`,
+    );
+  }
+
+  return null;
 };
 
 const endDateType = props => {
@@ -48,7 +70,7 @@ const SelectionConfigurationSingle = PropTypes.shape({
 
 const SelectionConfigurationRange = PropTypes.shape({
   type: CALENDAR_SELECTION_TYPE.range,
-  startDate: startDateType,
+  startDate: startDateType('selectionConfiguration'),
   endDate: endDateType,
 });
 
diff --git a/packages/bpk-component-calendar/src/date-utils.js b/packages/bpk-component-calendar/src/date-utils.js
index 9c7e04a2e..e567b7fea 100644
--- a/packages/bpk-component-calendar/src/date-utils.js
+++ b/packages/bpk-component-calendar/src/date-utils.js
@@ -195,6 +195,7 @@ export {
   getMonthRange,
   getDay,
   dateToBoundaries,
+  isBefore,
   isWithinRange,
   isSaturday,
   isSunday,

```

packages/bpk-component-calendar/src/custom-proptypes.js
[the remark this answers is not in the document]
k0nserv: Using `isBefore` and `isSameDay` as I did in this [comment](https://github.com/Skyscanner/backpack/pull/2269#discussion_r723191534)

Just using `>` is probably dangerous, in particular for same day 

image

packages/bpk-component-calendar/src/custom-proptypes.js
k0nserv: > As how would this handle if the consumer selected endDate first instead of startDate and the value is null?
This isn't allowed I think. If you have no selection or have already selected a range the next selection should cause the `startDate` to be set.

packages/bpk-component-calendar/README.md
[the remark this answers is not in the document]
k0nserv: It has to do with this behaviour. On iOS we need to know that the selection of the 31st should render across the rest of the screen width and likewise for the 1st we need the same behaviour for the leading portion. This covers `RowTypeEnd` and `RowTypeStart` respectively 

simulator_screenshot_91408349-CF61-4496-8651-884B014A93F0

PR #2271, opened 2021-10-04
[KOA-4546] Update dialog component to use new spacing grid

anambl:

- [ ] `README.md`
- [ ] Tests


UNRELEASED.md
```diff
@@ -1,4 +1,7 @@
-# 2021-10-04
+**Breaking:**
+
+- bpk-component-dialog:
+  - Updating the dialog component to use the new spacing grid. This component will now appear larger so should be checked this does not alter layout.
```
olliecurtis, line 4: Am wondering if this is breaking just because this will always open as an overlay vs aligned to other content so size shouldn't be a factor here 🤔

PR #2282, opened 2021-11-04
[KOA-4761]: Implement range styles

olliecurtis:
This PR implements the styles for supporting ranges and also some tweaks to the calendar styles to match the expectation in Figma.

as part of this change and to best support it `showWeekendSeparator` prop has been deprecated and styles removed as these no longer exist in Figma and caused issues with the range styles of the component.

- [x] `README.md`
- [x] Tests


zgeor:
The custom-composed-calendar example does not render correctly

image

zgeor:
The figma fixes and the deprecation of `showWeekendSeparator` should have been direct contributions to main. It would have reduced the risk of the changes and the size of the PRs.

Furthermore, the weekend separator is currently being used in the product. We should talk to the impacted teams and their designers before making this change. I am conscious that this is used in every calendar I've seen on web.

olliecurtis:
> The figma fixes and the deprecation of `showWeekendSeparator` should have been direct contributions to main. It would have reduced the risk of the changes and the size of the PRs.
> 
> Furthermore, the weekend separator is currently being used in the product. We should talk to the impacted teams and their designers before making this change. I am conscious that this is used in every calendar I've seen on web.

@zgeor sure thing, we will probably need to check this out with them, the reason I opted for the change in this PR was due to issue it raised with using ranges, as the spacing between rows had to be increased it meant there were gaps between the weekend separators so could not longer be a single line down the calendar and would be broken between each row.

So yes we should speak to design but also be conscious we can't achieve their design without removing it because otherwise the calendars will show a dashed non continuous line where the separator is

packages/bpk-component-calendar/src/Week.js
```diff
@@ -80,6 +78,61 @@ function getSelectedDate(date, selectionConfiguration) {
...
 }
 
+/**
+ * Gets the correct selection type for the current date
+ * @param {Date} date the current date of the calendar
+ * @param {Object} selectionConfiguration the current selection configuration
+ * @param {Function} formatDateFull function to format dates
+ * @returns {String} selection type to be passed to the date
+ */
+function getSelectionType(date, selectionConfiguration, formatDateFull) {
```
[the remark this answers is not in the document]
olliecurtis: Nested are painful which I wanted to avoid but given they return different values it maybe a little more difficult to simplify, could be I add some additional comments around it to help?

For the unit test will see if I can do that, may have to make this function an export in order to do so, but from running jest and its coverage, it gets covered as part of other rendering but will see what I can do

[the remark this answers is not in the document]
zgeor: Perhaps consider using if/elseif rather than just if. That way it will be easier to follow through what is part of the same logical block.

.storybook/storyshots-test.js
```diff
@@ -42,7 +42,7 @@ const getMatchOptions = () => ({
 
 initStoryshots({
   suite: 'Visual tests',
-  storyNameRegex: /Visual\stest$/i,
+  storyNameRegex: /Visual\stest\s?([a-z]*)?/i,
```
olliecurtis, line 45: Updating this regex to allow us to create more visual tests for components (think variations like single calendar and range calendar) - with the current regex it would only match on visual test and not for example visual test range.

If we named them the same in the stories the snapshot tests complain that a test with the same name has been defined so this losens this up

packages/bpk-component-calendar/src/Week.js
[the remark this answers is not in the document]
olliecurtis: Sure thing! Looking at the code again, I swear I had done that but for some reason looks like `eslint` is trying to be clever and undo these every time I save, so may need to tweak some settings locally and run the eslint cli to see why its suggesting that and what rule it thinks is right!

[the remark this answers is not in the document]
olliecurtis: So trying to change it to an if/elseif `eslint` looks to be trying to be smart and avoid an issue with its rules - see screenshot below of what is thrown when changing to if/elseif

Screenshot 2021-11-10 at 17 22 13

packages/bpk-component-calendar/src/Week.js
```diff
@@ -21,9 +21,6 @@ import React, { Component } from 'react';
 import { cssModules, deprecated } from 'bpk-react-utils';
 
 import {
-  getDay,
-  getFirstDayOfWeekend,
```
zgeor, line 25: Does this need to be removed from bpk-react-utils, or is it used elsewhere ?

packages/bpk-component-calendar/examples.js
```diff
@@ -77,13 +73,12 @@ const CalendarGridExample = () => (
...
     formatDateFull={formatDateFull}
     DateComponent={BpkCalendarDate}
-    showWeekendSeparator
     preventKeyboardFocus
   />
 );
 
 const CalendarGridAndHeaderExample = () => (
-  <div>
+  <div style={{ display: 'flex', flexDirection: 'column' }}>
```
anambl, line 81: Why do we have different styles for header and grid in terms of width? are they not generally supposed to be used together?
Also, might need a similar fix for the Week example

zgeor (approved):

packages/bpk-component-calendar/src/Week.js
olliecurtis: Ooo good question will take a look!

packages/bpk-component-calendar/examples.js
olliecurtis: This is due to when they are being used independently for custom part of the calendar - so when people are making custom calendars instead of using just `BpkCalendar` so building up individual parts they then need to take these changes into account. So this is an example of when using the calendar in separate parts

olliecurtis: >Also, might need a similar fix for the Week example

Good spot just seen that!

PR #2283, opened 2021-11-05
[KOA-4545] Update datatable component to use new spacing grid

anambl:
|Before https://backpack.github.io/storybook/?path=/story/bpk-component-datatable--autowidth-example|After https://backpack.github.io/storybook-prs/2283/?path=/story/bpk-component-datatable--autowidth-example|
|-|-|
|image|image|

- [ ] `README.md`
- [ ] Tests


packages/bpk-component-datatable/examples.js
```diff
@@ -54,59 +54,80 @@ const onRowClick = row => alert(JSON.stringify(row));
 
 const AutowidthExample = () => (
   <BpkDataTable rows={rows} height={400} onRowClick={onRowClick}>
-    <BpkDataTableColumn label="Name" dataKey="name" width={100} />
+    <BpkDataTableColumn label="Name" dataKey="name" width={100} flexGrow={1} />
```
olliecurtis, line 57: In order to solve this in the examples we are now adding flexGrow to each of the table columns, will consumers need to do this in order to make the component work at its best?

Looking at the [datatable readme](https://github.com/Skyscanner/backpack/tree/main/packages/bpk-component-datatable) it seems our example doesn't share with consumers they should be doing this, so we should update the readme here, or if its needed in order to make it work and space as expected we should add it as a prop people need to pass through similar to [`width prop`](https://github.com/Skyscanner/backpack/tree/main/packages/bpk-component-datatable#props)

What happens if the `flexGrow` for the description column is not 4?

olliecurtis (changes requested):

packages/bpk-component-datatable/examples.js
anambl: > In order to solve this in the examples we are now adding flexGrow to each of the table columns, will consumers need to do this in order to make the component work at its best?
The flexGrow property ensures the width of the column is flexible and can grow. If there is at least one column which should be flexible, then yes consumers would need to set it for each column.

> Looking at the datatable readme it seems our example doesn't share with consumers they should be doing this, so we should update the readme here, 
Yes, that's a very good point and it also looks like one of the links in there is broken so I will update this

> or if its needed in order to make it work and space as expected we should add it as a prop people need to pass through similar to width prop
Not sure I understand what you mean here. Do you mean make it required?

> What happens if the flexGrow for the description column is not 4?
That's ok, I set it to 4 just so that it is larger than the rest, but 1 could work too in which case all columns will have the same width.

PR #2286, opened 2021-11-10
[no jira]: Updating changed snapshots and tweaking threshold for snapshot differences

olliecurtis:
It seems we have outdated snapshots as the threshold for spotting differences is currently 10% in order to trigger failures. Comparing the differences it looks like the select component is changed here and the padding from the left has increased, possibly from a spacing change we recently made

Currently set it to 0% difference as I imagine this would capture all on both CI and local - but looking at the comments in the code it seems to be set to allow some deviation between ci and local environments, so we could tweak this but it may affect local and put us in a similar position?

### DIFF
--- a/.storybook/storyshots-test.js
+++ b/.storybook/storyshots-test.js
@@ -36,13 +36,13 @@ const beforeScreenshot = () =>
 // Allow a small amount of deviation to account for
 // CI running on a different OS than local development.
 const getMatchOptions = () => ({
-  failureThreshold: 0.1,
+  failureThreshold: 0.05,
   failureThresholdType: 'percent',
 });
 
 initStoryshots({
   suite: 'Visual tests',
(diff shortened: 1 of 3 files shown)


.storybook/storyshots-test.js
```diff
@@ -36,13 +36,13 @@ const beforeScreenshot = () =>
 // Allow a small amount of deviation to account for
 // CI running on a different OS than local development.
 const getMatchOptions = () => ({
-  failureThreshold: 0.1,
+  failureThreshold: 0,
```
olliecurtis, line 39: Changing this so we can actually capture deviences in visual changes - not sure on its affect on CI

olliecurtis:
Looks like font differences is causing the CI on this PR to fail

olliecurtis:
Closing as I messed up Git

PR #2287, opened 2021-11-10
[no jira]: Tweaking threshold for snapshot differences and config

olliecurtis:
It seems we have outdated snapshots as the threshold for spotting differences is currently 10% in order to trigger failures. 

Currently set it to 5% difference as I imagine this would capture all on both CI and local - but looking at the comments in the code it seems to be set to allow some deviation between ci and local environments, so we could tweak this but it may affect local and put us in a similar position?

### DIFF
--- a/.storybook/storyshots-test.js
+++ b/.storybook/storyshots-test.js
@@ -36,13 +36,13 @@ const beforeScreenshot = () =>
 // Allow a small amount of deviation to account for
 // CI running on a different OS than local development.
 const getMatchOptions = () => ({
-  failureThreshold: 0.1,
+  failureThreshold: 0.05,
   failureThresholdType: 'percent',
 });
 
 initStoryshots({
   suite: 'Visual tests',
-  storyNameRegex: /Visual\stest$/i,
+  storyNameRegex: /Visual\stest\s?([a-z]*)?/i,
   test: imageSnapshot({
     storybookUrl: `file://${path.resolve(__dirname, '../dist-storybook')}`,
     getMatchOptions,


.storybook/storyshots-test.js
```diff
@@ -36,13 +36,13 @@ const beforeScreenshot = () =>
...
 const getMatchOptions = () => ({
-  failureThreshold: 0.1,
+  failureThreshold: 0.05,
   failureThresholdType: 'percent',
 });
 
 initStoryshots({
   suite: 'Visual tests',
-  storyNameRegex: /Visual\stest$/i,
+  storyNameRegex: /Visual\stest\s?([a-z]*)?/i,
```
olliecurtis, line 45: Updating this regex to allow us to create more visual tests for components (think variations like single calendar and range calendar) - with the current regex it would only match on visual test and not for example visual test range.

If we named them the same in the stories the snapshot tests complain that a test with the same name has been defined so this losens this up

.storybook/storyshots-test.js
```diff
@@ -36,13 +36,13 @@ const beforeScreenshot = () =>
 // Allow a small amount of deviation to account for
 // CI running on a different OS than local development.
 const getMatchOptions = () => ({
-  failureThreshold: 0.1,
+  failureThreshold: 0.05,
```
olliecurtis, line 39: Changing this so we can actually capture deviences in visual changes - not sure on its affect on CI

olliecurtis:
Closing for now as we may need to think of other things for this!

PR #2289, opened 2021-11-11
[KOA-4761]: Cleaning up changes as part of range implementation

olliecurtis:
This pull request ties up some final parts of implementin the range logic and any bugs that have popped as part of this.

Adds some missed tests for the logic implementation
- Visual snapshots
- Additional unit tests for ranges
- Additional A11y tests for ranges

Fixed a range bug with BpkDatepicker

- [ ] `README.md`
- [x] Tests


packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -99,17 +99,6 @@ class BpkScrollableCalendarGridList extends React.Component {
         rowRenderer={this.rowRenderer}
         rowCount={this.state.months.length}
         overscanRowCount={0}
-        scrollToIndex={
```
olliecurtis, line 102: Removing this code as it was reintroduced during the reverting of the library being used for the scrollable calendar, and an issue that was raised via slack regarding the calendar jumping around when a date is selected, as its trying to scroll to indexes.

This will probably be raised as a separate PR or can be done in this cleanup?

zgeor (approved):

PR #2290, opened 2021-11-11
[KOA-4808]: Radio spacing alignment to Figma

olliecurtis:
PR to update the spacing between input and label for the BpkRadio component similar to what we did in #2219 .

In this PR I also noticed when in RTL the sizes were incorrect and misaligned which we missed in #2219 , 

Also when doing this I changed the definition of the spacing to make it clearer of the value we are applying instead of just taking the smallest size and doing a multiple to define it

| Before | After |
| -- | -- |
| Screenshot 2021-11-11 at 16 06 59 | Screenshot 2021-11-11 at 16 06 46 |

- [ ] `README.md`
- [x] Tests


packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -486,13 +486,13 @@ $bpk-spacing-v2: true;
...
   display: inline-block;
-  padding-left: bpk-spacing-sm() * 7;
+  padding-left: bpk-spacing-lg() + bpk-spacing-sm();
 
   @include bpk-text;
   @include bpk-text--base;
 
   @include bpk-rtl {
-    padding-right: $bpk-spacing-base;
+    padding-right: bpk-spacing-lg() + bpk-spacing-sm();
```
olliecurtis, line 495: Missed this in the original PR! As you see the before is still using the old spacing style!

packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -486,13 +486,13 @@ $bpk-spacing-v2: true;
 @mixin bpk-checkbox {
   position: relative;
   display: inline-block;
-  padding-left: bpk-spacing-sm() * 7;
+  padding-left: bpk-spacing-lg() + bpk-spacing-sm();
```
olliecurtis, line 489: Changed the tokens used to make it clearer for the spacing - the value is the exact same as before just using a different combinations of tokens to achieve the same

anambl (approved):

PR #2293, opened 2021-11-18
[KOA-4770]: Multiple inputs for calendars

olliecurtis:
Adding support for multiple inputs for calendars
 - Added new `inputComponent` property for consumers to provide their own element of multiple fields - this was easier than hacking around the component as it would mean some of the a11y features would no longer be flexible for the inputs and ids, without adding even more properties to the component
 - Created a new story to show usage and for use on the docs site

- [x] `README.md`
- [x] Tests


zgeor:
I've been looking at the datepicker example on storybook. And the following occurs

I click on the start date field,
then select a date,
then click away from the calendar so it goes away and end date remains unselected,
then click again on the start date and select an earlier date,
then click away so the calendar closes,
I observe that the calendar cannot be opened again.

packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -313,6 +325,7 @@ BpkDatepicker.propTypes = {
 
 BpkDatepicker.defaultProps = {
   calendarComponent: BpkCalendar,
+  InputComponent: null,
```
anambl, line 328: should this not start with a lowercase 'i' as it is a prop?

olliecurtis: Good spot this would have been from autocomplete in my editor

packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -132,8 +132,20 @@ class BpkDatepicker extends Component {
       selectionConfiguration,
       minDate,
       maxDate,
+      onClose,
     } = this.props;
-    this.onClose();
+
+    // When the calendar is a single date we always want to close on when a date is selected
+    // or if there is a start date but not yet an end date we don't want the calendar to close until the range is selected.
```
zgeor, line 139: This comment looks incorrect. The if statement suggests that we want the the selection to be range and have an end date.

olliecurtis: Ah yes you are correct since changing the function from original implementation, I missed updating the comment

anambl (approved):

PR #2294, opened 2021-11-22
Export CALENDAR_SELECTION_TYPE from bpk-component-datepicker

sogame:
When using `bpk-component-datepicker`, it feels weird doing two different imports, specially taking into account `bpk-component-calendar` might not even be used in the project, as it's just an internal dependency that users don't need to know about:

```
import BpkDatepicker from 'bpk-component-datepicker';
import { CALENDAR_SELECTION_TYPE } from 'bpk-component-calendar';
```

I think it would be better to export `CALENDAR_SELECTION_TYPE` in `bpk-component-datepicker`, so only a single export is needed (and only in the package users have installed):

```
import BpkDatepicker, { CALENDAR_SELECTION_TYPE } from 'bpk-component-datepicker';
```

What do you think about this?

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,5 @@
+**Added:**
+
+- bpk-component-datepicker
+- bpk-component-scrollable-calendar
+  - Export `CALENDAR_SELECTION_TYPE` so it can be imported from `bpk-component-datepicker` and `bpk-component-scrollable-calendar` instead of `bpk-component-calendar`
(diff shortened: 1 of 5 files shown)


olliecurtis:
Good spot @sogame, thanks! We can probably do the same for the `bpk-scrollable-calendar` as well!

sogame:
> Good spot @sogame, thanks! We can probably do the same for the `bpk-scrollable-calendar` as well!

Added, @olliecurtis :-)
I've also added the import in `readme.md`, as it was missing there 😬

olliecurtis (approved):

PR #2308, opened 2021-12-07
[NO JIRA]: Fix incorrect custom prop type in calendar

olliecurtis:
Fixed an issue whereby the custom `SelectionConfiguration` prop type was failing due to incorrect typing of `type` variable in the object. This change correctly types the field to use a specific value as per the `prop-types` spec.

Which would throw a console error in the browser of `Warning: Failed prop type: checker is not a function`

- [x] Tests


sogame (approved):
LGTM

PR #2311, opened 2021-12-13
[KOA-4856]: Form validation spacing fix

olliecurtis:
In this PR we have moved the `margin-top` which was used to put the correct spacing when the validation message was used in the fieldset component to the form validation message so when the validation message is used as standalone component it still applies the correct spacing as when used in the fieldset component

- [x] Tests


packages/bpk-mixins/src/mixins/_forms.scss
```diff
@@ -342,7 +342,7 @@ $bpk-spacing-v2: true;
     bpk-spacing-md() center;
   background-size: $bpk-input-height / 2;
   color: $bpk-select-color;
-  line-height: $bpk-line-height-base - (2 * $bpk-select-border-width);
+  line-height: ($bpk-line-height-base - (2 * $bpk-select-border-width));
```
olliecurtis, line 345: The linter through an error when I made the further down change with the following so to fix it surrounded it with `()`

Screenshot 2021-12-13 at 13 17 08

anambl (approved):

PR #2312, opened 2021-12-13
[KOA-4858] Fix popover arrow size

anambl:
|Before|After|
|-|-|
|image|image|

- [ ] `README.md`
- [ ] Tests


packages/bpk-component-autosuggest/src/BpkAutosuggest.module.scss
```diff
@@ -20,6 +20,8 @@
 
 $bpk-spacing-v2: true;
 
+$arrow-size: bpk-spacing-sm() * 3;
```
anambl, line 23: Value stays the same, only added this as I didn't know the border is the actual arrow and maybe this could make it easier for anyone looking at this in the future

olliecurtis (approved):

PR #2315, opened 2021-12-24
[LEOP-269]Remove sassFunctions and node-sass (As peerDependency) from bpk-mixins

SophiaChen2021:
bpk-mixins has been used in many areas at Skyscanner, but it provides not only styling but also a sassFunctions for encoding SVG content to Base64 content, this case makes it be installed as a (peer)dependency in other packages, it is time to delete the sass-functions.js and instead written it in where needs it, and also the make it only provide styling.


olliecurtis:
This change is dependent on first the work in BRS to go through before it can be removed: https://github.com/Skyscanner/backpack-react-scripts/pull/105

PR #2316, opened 2022-01-04
Updating licence to 2022

olliecurtis:
Updating the licence to 2022 as its a new year! Also cleaned up CONTRIBUTING.md to link to licence file for contents instead of duplicating the same code and requiring two places to update each year

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,30 +17,7 @@ By contributing your code, you agree to license your contribution under the term
 
 All files are released with the Apache 2.0 licence.
 
-If you are adding a new file it should have a header comment containing licence information:
-
-<details>
-<summary>Show/hide licence header</summary>
-
-```
-Backpack - Skyscanner's Design System
-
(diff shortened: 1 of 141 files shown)


CONTRIBUTING.md
```diff
@@ -17,30 +17,7 @@ By contributing your code, you agree to license your contribution under the term
...
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-```
-
-</details>
+If you are adding a new file it should have a header comment containing licence information from the following file: [license](./license.txt).
```
olliecurtis, line 20: Changed this in favour to linking to the license file as its the exact same but just pasted here, so ensures we only need to update one place and can't miss it here

zgeor (approved):

frugoman (approved):

PR #2318, opened 2022-01-06
[KOA-4727]: Adding dependabot config to open automatic PRs

olliecurtis:
This config was missing therefore no automatic PRs to fix issues were being raised.

The dependabot config is now setup also to be a lot less noisy similar to the backpack-docs config where it targets direct dependencies so should reduce some of the noise down and fix the critical things!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
           path: .storybook/__image_snapshots__/__diff_output__
 
       - name: Post a comment about the visual tests (if they pass)
-        uses: unsplash/comment-on-pr@master
+        uses: unsplash/comment-on-pr@v1.3.0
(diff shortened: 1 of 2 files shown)


marianeum (approved):

.github/workflows/ci.yml
```diff
@@ -62,15 +62,15 @@ jobs:
           path: .storybook/__image_snapshots__/__diff_output__
 
       - name: Post a comment about the visual tests (if they pass)
-        uses: unsplash/comment-on-pr@master
+        uses: unsplash/comment-on-pr@v1.3.0
```
olliecurtis, line 65: Setting the version explicity which is best practise but also it looks that in the latest version there was a rewrite of this action into a ruby version which is now causing failures when run on CI.

It might be in the future we look at changing this to not use a third part action for comments and follow Android in using the native GH way of doing this

PR #2335, opened 2022-01-12
bugfix: [KOA-4876] - Datepicker modal fix

olliecurtis:
This PR allows the `fixedWidth` prop to become available to the datepicker so that when in the datepicker opens in a popover (when device is mobile) the calendar will fit to the correct viewport. By default the datepicker has a fixed width but always on mobile it will not.

It also changes the calendar on mobile to not use the calendar transitions as this requires a fixed width in order to calculate the transition and alignment which cannot be always determined by the device being viewed on.

| Before | After |
| -- | -- |
| Screenshot 2022-01-12 at 13 13 06 | Screenshot 2022-01-12 at 13 13 20 |

- [x] `README.md`
- [x] Tests


packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -331,11 +346,12 @@ BpkDatepicker.propTypes = {
 };
 
 BpkDatepicker.defaultProps = {
-  calendarComponent: BpkCalendar,
+  calendarComponent: DefaultCalender,
```
anambl, line 349: small typo here 😄

packages/bpk-component-datepicker/src/BpkDatepicker.js
```diff
@@ -331,11 +346,12 @@ BpkDatepicker.propTypes = {
...
   inputComponent: null,
   date: null,
   dateModifiers: BpkCalendar.defaultProps.dateModifiers,
```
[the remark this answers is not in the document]
olliecurtis: Good catch!

packages/bpk-component-datepicker/src/BpkDatepicker.js
olliecurtis: Good catch!

anambl (approved):
LGTM! 🎉

PR #2344, opened 2022-01-19
[NO JIRA]: Updating licence headers

olliecurtis:
The final update to never have to update it again.

- [ ] `README.md`
- [ ] Tests


packages/bpk-component-mobile-scroll-container/examples.js
```diff
@@ -27,24 +27,22 @@ import BpkMobileScrollContainer from './index';
...
-            index % 2 === 0 && 'bpk-scroll-block__item--alternate',
-          )}
-        >
-          {index}
-        </div>
-      ))}
-    </div>
-  );
-};
+const ScrollExampleBlock = () => (
```
zgeor: Is this change expected?

olliecurtis: Looks like eslint formatting has passed through this which is what has caused that when comparing to the original changes

zgeor (approved):

PR #2359, opened 2022-02-01
[KOA-4812]: Adding new text styles

olliecurtis:

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,15 @@
+**Added:**
+  - bpk-mixins:
+    - Added new Typography mixins.
+      - `bpk-caption`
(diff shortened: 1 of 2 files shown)


anambl:
Looks good! The only thing is that we need to add the letter spacing too for all hero styles

olliecurtis:
> Looks good! The only thing is that we need to add the letter spacing too for all hero styles

Ah you are right 🤦‍♂️ in my head I thought I saw the hero styles didn't have letter spacing in the styles doc but now I have checked and they do! Clearly got carried away with copy and pasting that I just missed it!

anambl (approved):
LGTM 🎉

PR #2361, opened 2022-02-03
[KOA-4879]: Deprecate heading component

olliecurtis:
This PR adds missing deprecation work when this component was previously deprecated back in 2017!

- Adds deprecation note to BpkHeading component README.
- Sets the package to `private` so that the component is ignored during publish.

The code will then probably be removed when we ship Backpack as a single package


anambl (approved):
LGTM!

PR #2364, opened 2022-02-07
[KOA-4937]: Remove margins from heading mixins

olliecurtis:

- [x] Tests


packages/bpk-mixins/src/mixins/_typography.scss
```diff
@@ -718,6 +700,7 @@ $bpk-spacing-v2: true;
 ///     @include bpk-heading-1();
 ///     @include bpk-heading--no-bottom-margin();
 ///   }
+/// @deprecated Headings no longer set margins.
```
olliecurtis, line 703: This mixin is only used by bpk-component-heading and given that is deprecated we can deprecate this

anambl (approved):
LGTM!

PR #2381, opened 2022-02-18
Migrate bpk-mixins and bpk-stylesheets to foundations

olliecurtis:
As part of shipping Backpack as a single package we need to migrate bpk-mixins and bpk-stylesheets to foundations so they can be shipped independently from components as these are foundational packages and can be used without components.

Within this PR we also move the stories for bpk-mixins `.storybook/additional-stories/bpk-mixins` to here as it means we can utilise bpk-react-utils for use in the stories without making it a dependency of mixins just for a story, similarly it allows us to test bpk-mixins working as expected as a library by having it external to the lib.


package.json
```diff
@@ -54,7 +53,7 @@
     "publish:css": "node ./scripts/publish-process/transform-js-scss-css-imports.js && node ./scripts/publish-process/transform-bpk-imports.js && node ./scripts/publish-process/publish-css-tagged-packages.js && git reset --hard HEAD",
     "postpublish": "./scripts/publish-process/update-docs-deps.sh && node ./scripts/publish-process/update-changelog.js",
     "release": "npm run publish",
-    "sassdoc": "sassdoc {packages/bpk-mixins/src/**/*,node_modules/@skyscanner/bpk-svgs/dist/*,node_modules/@skyscanner/bpk-foundations-web/tokens/base.default}.scss -d dist-sassdoc -v --strict",
+    "sassdoc": "sassdoc {node_modules/bpk-mixins/src/**/*,node_modules/@skyscanner/bpk-svgs/dist/*,node_modules/@skyscanner/bpk-foundations-web/tokens/base.default}.scss -d dist-sassdoc -v --strict",
```
zgeor, line 56: Does this need to be here? This looks like a responsibility of the project that hosts those files.

README.md
```diff
@@ -85,9 +85,9 @@ To contribute please see [contributing.md](CONTRIBUTING.md).
 | [`bpk-theming`](/packages/bpk-theming) | [![npm version](https://badge.fury.io/js/bpk-theming.svg)](http://badge.fury.io/js/bpk-theming) |
 | [`bpk-component-ticket`](/packages/bpk-component-ticket) | [![npm version](https://badge.fury.io/js/bpk-component-ticket.svg)](http://badge.fury.io/js/bpk-component-ticket) |
 | [`bpk-component-tooltip`](/packages/bpk-component-tooltip) | [![npm version](https://badge.fury.io/js/bpk-component-tooltip.svg)](http://badge.fury.io/js/bpk-component-tooltip) |
-| [`bpk-mixins`](/packages/bpk-mixins) | [![npm version](https://badge.fury.io/js/bpk-mixins.svg)](http://badge.fury.io/js/bpk-mixins) |
+| [`bpk-mixins`](https://github.com/Skyscanner/backpack-foundations/tree/main/packages/bpk-mixins) | [![npm version](https://badge.fury.io/js/bpk-mixins.svg)](http://badge.fury.io/js/bpk-mixins) |
```
zgeor, line 88: These should be removed from this repo as the responsibility has moved to foundations.

zgeor: Or alternatively there should be an indicator that these 2 are external to this project

package.json
olliecurtis: Agree, good shout!

zgeor (approved):

PR #2387, opened 2022-03-01
[KOA-4962]: Removing deprecated packages

olliecurtis:
This PR removes deprecated packages as part of shipping Backpack as a single package.

As these components are deprecated and we want to avoid people using them we are removing the following components. This will also make the size of the package we ship lighter as we won't be shipping packages people are not required to use.

- bpk-component-heading
- bpk-component-paragraph
- bpk-tether

- [x] `README.md`
- [x] Tests


anambl (approved):
LGTM!

PR #2388, opened 2022-03-01
[KOA-4881]: Update Backpack components to new text api

olliecurtis:
Updates Backpack components (mostly examples) to use the updated `BpkText` API and the new styles provided by Backpack

- [ ] `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-accordion/src/BpkAccordionItem.js
+++ b/packages/bpk-component-accordion/src/BpkAccordionItem.js
@@ -144,7 +144,7 @@ BpkAccordionItem.defaultProps = {
   icon: null,
   onClick: () => null,
   tagName: 'span',
-  textStyle: TEXT_STYLES.base,
+  textStyle: TEXT_STYLES.bodyDefault,
 };
 
 export default BpkAccordionItem;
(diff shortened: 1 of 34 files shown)


packages/bpk-component-accordion/src/BpkAccordionItem.js
```diff
@@ -144,7 +144,7 @@ BpkAccordionItem.defaultProps = {
   icon: null,
   onClick: () => null,
   tagName: 'span',
-  textStyle: TEXT_STYLES.base,
+  textStyle: TEXT_STYLES.bodyDefault,
```
anambl, line 147: I'm thinking whether we should remove the weight from accordion completely as part of this ticket? Otherwise we are still using the text component with weight 🤔 what do you think?

olliecurtis: I think that might be something we should do separately, unless we know the users of the `weight` prop have already upgraded to remove that from their code?

anambl: Looking at the code now, `weight` is still used in one project

anambl (approved):
LGTM!

PR #2389, opened 2022-03-02
[NO JIRA]: Removing unused  property from accordion

olliecurtis:
As part of the new typography changes we deprecated the `weight` property from the Accordion component. Now we have confirmed this usage has been removed in the product we can remove this property as we tie up changes.

- [x] `README.md`
- [x] Tests


anambl (approved):
LGTM!

PR #2409, opened 2022-04-07
[KOA-5002]: Fix AT where a user cannot select a date

olliecurtis:
Fixes the issue where users interacting with the calendar with Assistive Technologies cannot select a date. This came from recent changes with the range support added to the calendar and due to a change in the styles to use a CSS table.

To resolve this we have moved away from using tables and implementing the grids with `div` elements instead to allow full access to the calendar, selecting a date and also keeping the padding between rows

- [ ] `README.md`
- [x] Tests


anambl (approved):
LGTM!

PR #2411, opened 2022-04-08
Add support for Mobile/Tablet/Desktop images for BpkGraphicPromo

robaw:
This change allows different images to be loaded depending on the viewport width.

# Screenshots
## Desktop
Screenshot 2022-04-12 at 12 39 41
## Tablet
Screenshot 2022-04-12 at 12 39 53
## Mobile
Screenshot 2022-04-12 at 12 40 06

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/examples/bpk-component-graphic-promotion/examples.js
+++ b/examples/bpk-component-graphic-promotion/examples.js
@@ -18,16 +18,13 @@
 
 import React from 'react';
 
+import STYLES from './examples.module.scss';
+
 import BpkGraphicPromo, {
   TEXT_ALIGN,
 } from '../../packages/bpk-component-graphic-promotion';
-
-const backgroundImageDesktop =
-  'https://content.skyscnr.com/m/31ebf33b07194794/original/Hiker-looking-out-over-mountain.jpg?crop=2096px:800px&quality=90';
-const backgroundImageTablet =
-  'https://content.skyscnr.com/m/31ebf33b07194794/original/Hiker-looking-out-over-mountain.jpg?crop=820px:1180px&quality=90';
-const backgroundImageMobile =
-  'https://content.skyscnr.com/m/31ebf33b07194794/original/Hiker-looking-out-over-mountain.jpg?crop=390:844px&quality=90';
+import { cssModules } from '../../packages/bpk-react-utils';
+const getClassName = cssModules(STYLES);
 const sponsor = {
   label: 'Sponsored',
   logo: 'https://js.skyscnr.com/sttc/bpk-content/skyland-a76916b4.png',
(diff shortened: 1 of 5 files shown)


examples/bpk-component-graphic-promotion/examples.js
```diff
@@ -22,8 +22,12 @@ import BpkGraphicPromo, {
...
 } from '../../packages/bpk-component-graphic-promotion';
 
-const image =
-  'https://content.skyscnr.com/m/31ebf33b07194794/original/Hiker-looking-out-over-mountain.jpg?crop=1048px:400px&quality=90';
+const backgroundImageDesktop =
+  'https://content.skyscnr.com/m/31ebf33b07194794/original/Hiker-looking-out-over-mountain.jpg?crop=2096px:800px&quality=90';
+const backgroundImageTablet =
+  'https://content.skyscnr.com/m/31ebf33b07194794/original/Hiker-looking-out-over-mountain.jpg?crop=820px:1180px&quality=90';
+const backgroundImageMobile =
+  'https://content.skyscnr.com/m/31ebf33b07194794/original/Hiker-looking-out-over-mountain.jpg?crop=390:844px&quality=90';
```
olliecurtis, line 30: This is something I recently spoke to design about and using the correct ratios for the images, in the end we found `1048px:400px` is the one we should use which gives an aspect ratio that works across the different breakpoints - the guidance for this should be available in the Backpack figma libraries

olliecurtis (changes requested):

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -43,7 +43,9 @@ export type Props = {
   tagline: ?string,
   headline: string,
   subheading: ?string,
-  image: string,
+  backgroundImageMobile: string,
+  backgroundImageTablet: string,
+  backgroundImageDesktop: string,
```
angel3cu, line 48: I am nitpicking here, but just because we have too many properties in the top-level, can we nest these images under an `image` property, as suggested [here](https://confluence.skyscannertools.net/display/SRVT/Front-End+Work+for+Graphic+Promos)? Feel free to use `desktop`, `tablet` and `mobile`, if these are the agreed variations instead of `landscape` and `portrait`.

Thanks! 😄

angel3cu (commented):
Nice job! I am so happy this solution works, because I haven't fully tested it before. Your implementation is great.

I left a small comment on nesting the properties. I think it looks a bit neater that way, because we have too many properties on that component already. However, I am happy to backtrack if the majority disagrees. 🙂

examples/bpk-component-graphic-promotion/examples.js
robaw: Resolving this since we've taken the approach of using 3 different images

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
robaw: Resolving since we're no longer passing through these props

olliecurtis (changes requested):
Would you be able to update the UNRELEASED file?

robaw:
Closing this PR in favour of https://github.com/Skyscanner/backpack/pull/2415

PR #2415, opened 2022-04-12
Add support for Mobile/Tablet/Desktop images for BpkGraphicPromo

robaw:
This change allows different images to be loaded depending on the viewport width.

# Screenshots
## Desktop
Screenshot 2022-04-12 at 12 39 41
## Tablet
Screenshot 2022-04-12 at 12 39 53
## Mobile
Screenshot 2022-04-12 at 12 40 06

# TODO
- Update tests

- [ ] `README.md`
- [ ] Tests


LouiseReid:
@angel3cu we noticed when screen resizing you get a flash of the blue background prior to the new image rendering. We could maybe handle this with pre rendering images in the consumer but is there a reason its set to blue?

For banana I think we would want it to be white but maybe to suit all cases maybe a light grey so its subtle?

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -188,6 +189,7 @@ BpkGraphicPromo.propTypes = {
   invertVertically: PropTypes.bool,
   textAlign: PropTypes.oneOf(Object.values(TEXT_ALIGN)).isRequired,
   textColor: PropTypes.oneOf(Object.values(TEXT_COLORS)),
+  style: PropTypes.string,
```
LouiseReid, line 192: I think this type should be `style: { [key: PropTypes.string]: PropTypes.string };`

angel3cu:
> @angel3cu we noticed when screen resizing you get a flash of the blue background prior to the new image rendering. We could maybe handle this with pre rendering images in the consumer but is there a reason its set to blue?
> 
> For banana I think we would want it to be white but maybe to suit all cases maybe a light grey so its subtle?

Hi! Yes, this is a bit unfortunate and is happening because of the time it takes the browser to load the image. The default background is a Skyscanner blue, but we could make it configurable it it is a big issue.

LouiseReid:
> > @angel3cu we noticed when screen resizing you get a flash of the blue background prior to the new image rendering. We could maybe handle this with pre rendering images in the consumer but is there a reason its set to blue?
> > For banana I think we would want it to be white but maybe to suit all cases maybe a light grey so its subtle?
> 
> Hi! Yes, this is a bit unfortunate and is happening because of the time it takes the browser to load the image. The default background is a Skyscanner blue, but we could make it configurable if it is a big issue.

Is there a reason its blue though or can we change this?

robaw:
> We just picked the default Skyscanner colour from the theme at the time. No particular reason.
Will add backgroundColor as a prop to allow the consumer to decide what colour of background they want (we'll be using white so we don't see a flash of blue before the image is swapped in), as if we remove it entirely then without adding a `background-image` in the consumer's CSS then we will end up rendering white text on a white background.

angel3cu (commented):
@olliecurtis Apologies for the late question. Was the plan to pass a `style` or a `className`? Thanks!

olliecurtis:
> @olliecurtis Apologies for the late question. Was the plan to pass a `style` or a `className`? Thanks!

Hey @angel3cu yes this would be via `className` I know this one is currently applied to the container so if we want it for the image then adding a new property `imageClassName`

This is how we make styles extensible to our current components and recommended for React and other design systems 🙂

olliecurtis:
> > We just picked the default Skyscanner colour from the theme at the time. No particular reason.
> 
> Will add backgroundColor as a prop to allow the consumer to decide what colour of background they want (we'll be using white so we don't see a flash of blue before the image is swapped in), as if we remove it entirely then without adding a `background-image` in the consumer's CSS then we will end up rendering white text on a white background.

Hey @robaw I would say this will be applicable via `className` rather than a property and we can confirm the default colour with design.

I imagine in the future when we have it we can use a loading skeleton or a spinnerto do a loading state whilst the image is loading

UNRELEASED.md
```diff
@@ -0,0 +1,7 @@
+**Breaking**
+- `bpk-component-graphic-promotion` - Now accepts a style prop and no longer accepts an image prop, leaving it up to the consumer to populate the background-image through styling
+
+**Fixed:**
```
olliecurtis, line 4: Can you remove this from this file as this change has been released upstream!

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -109,11 +103,10 @@ const BpkGraphicPromo = (props: Props) => {
 
   const getTextClasses = (baseClass: string) =>
     getClassName(baseClass, `${baseClass}--${textAlign}`);
-
   return (
     <BpkCard
       className={cardClasses}
-      style={{ backgroundImage: `url(${image})`, color: textColor }}
+      style={style}
```
olliecurtis, line 109: The recommended way would be via CSS, this is already possible with `className` as this property is exposed and when passed in gets applied to the styles of the `BpkCard`

So doing it with CSS classes and styles is how this is done

olliecurtis (changes requested):

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
robaw: As discussed, I don’t think we can achieve having multiple background images without using style, as it’s the style attribute that we’re able to inject the correct backgroundImages to be used at runtime, otherwise we’d just have static CSS classes with the same images. Marking as resolved.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
angel3cu:
```typescript
style: PropTypes.shape({
  <something>
})
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -53,7 +48,7 @@ export type Props = {
   onClick: () => void,
   invertVertically: boolean,
   textAlign: TEXT_ALIGN,
-  textColor: TEXT_COLORS,
```
angel3cu, line 56: Have we agreed that we don't need this property anymore?

angel3cu (commented):
Looks good to me.

angel3cu:
We may need to update some snapshots because the pipeline is failing. 🙂

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
olliecurtis: Yes as this will be provided via `className` and the component should provide a default style OOTB 😄

examples/bpk-component-graphic-promotion/stories.js
```diff
@@ -30,11 +31,12 @@ import {
 } from './examples';
 
 storiesOf('bpk-component-graphic-promotion', module)
-  .add('Default styling', NoStyleExample)
-  .add('Default', DefaultExample)
+  .add('Default no background image', NoStyleExample)
```
olliecurtis, line 34: Do we ever expect the GP to be used without a Background image or is it always expected?

examples/bpk-component-graphic-promotion/stories.js
```diff
@@ -30,11 +31,12 @@ import {
...
+  .add('Default with background image', DefaultExample)
   .add('Center Aligned', CenterAlignedExample)
   .add('Right Aligned', RightAlignedExample)
   .add('Inverted Portrait Mode', InvertedPortraitExample)
   .add('Minimalistic', MinimalisticExample)
   .add('Non-Sponsored', NonSponsoredExample)
-  .add('Visual test', MixedExample);
+  .add('Mixed', MixedExample)
```
olliecurtis, line 41: Given we have changed up the visual test and the other stories exist individually already we won't need to have this story anymore so can be removed 😄

olliecurtis (changes requested):

UNRELEASED.md
robaw: Removed this

examples/bpk-component-graphic-promotion/stories.js
robaw: Well out of the box it won't have a background image, I don't know if that would ever be valuable but I think we should be aware what the out-of-the-box component looks like.

olliecurtis (approved):

olliecurtis (commented):
Just had a thought do we need to update the tests to reflect these changes?

olliecurtis (approved):
LGTM!

Wondering if we need an additional test though to be testing the new style prop that has been introduced as I don't see that here 🤔

PR #2423, opened 2022-04-28
[BD-6033] migrate bpk-component-map from react-google-maps to @react-google-maps/api

giriawu:
# Context

The [react-google-maps](https://github.com/tomchentw/react-google-maps) used in bpk-component-map is no longer supported, and the old version of bpk-component-map caused some issues during the upgrade to React v17

The upgraded map based on  [@react-google-maps/api ](https://github.com/JustFly1984/react-google-maps-api) will provide better support for others to upgrade react v17.

## What Predictions Do We Have About When This Goes Live?

- The map component will not depend on the old `react-google-maps` library which is no longer supported, but will depend on the new `@react-google-maps/api` library which is continuously updated and has better support for React v17
- Most of the usage and functions of the map will be maintained as they were, and there will be no visual impact
- Upgraded the `WithGoogleMapsScript` component to support loading multiple maps at the same time without downloading multiple scripts
- Improve the style extensibility of the map component

## ⚠️ Attention

The `@react-google-maps/api` library and the new features require a backpack upgrade to **React v16.8+** or higher, otherwise errors will occur.

## Change Description

- Bumped React main version to 16.8 to support map feature
- Migrate `react-google-maps` to `@react-google-maps/api` and optimize map loading performance
- Improve map customization scalability
- removed some unused props
- Added example of multi maps to storybook for better presentation
- Updated readme file and some snapshots

## Related Doc

- `@react-google-maps/api` doc: https://react-google-maps-api-docs.netlify.app/#googlemap

- [x] `README.md`
- [ ] Tests
- [ ] Docs(have edited examples in `backpack/examples/bpk-component-map/examples` and README.md, checked that no modification to the backpack-docs project is needed. Will update bpk-component-map version after this PR merged.)


packages/bpk-component-map/README.md
```diff
@@ -129,31 +128,34 @@ export default () => (
...
+| onZoom                | func                                                                                            | false    | null                             |
+| onTilesLoaded         | func                                                                                            | false    | null                             |
+| panEnabled            | bool                                                                                            | false    | true                             |
+| showControls          | bool                                                                                            | false    | true                             |
+| zoom                  | number                                                                                          | false    | 15                               |
+| className             | string                                                                                          | false    | null                             |
+| mapOptionStyles       | [array](https://developers.google.com/maps/documentation/javascript/reference/map#MapTypeStyle) | false    | null                             |
```
craigchen1990, line 144: Please keep the property list in order, same as below

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -145,30 +180,32 @@ BpkMap.propTypes = {
...
   showControls: PropTypes.bool,
+  onTilesLoaded: PropTypes.func,
   zoom: PropTypes.number,
+  mapOptionStyles: MapTypeStyle,
 };
 
 BpkMap.defaultProps = {
   bounds: null,
-  center: null,
+  center: undefined,
```
craigchen1990, line 197: why the default value need to change to undefined?

giriawu: I followed the default value from the [documentation](https://react-google-maps-api-docs.netlify.app/#googlemap) and it suggest that it should be `undefined` when no value is set.

packages/bpk-component-map/README.md
giriawu: Thanks for your review! Have updated.

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -114,27 +146,30 @@ const BpkMap = withGoogleMap((props: Props) => {
         streetViewControl: false,
         fullscreenControl: false,
         rotateControl: false,
+        clickableIcons: false,
```
jasonhuang-sky, line 149: So, it's to make the icons unclickable, which means the infobox won't pop up when a icon is clicked
https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions.clickableIcons
image

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -114,27 +146,30 @@ const BpkMap = withGoogleMap((props: Props) => {
...
+        styles: mapOptionStyles,
```
jasonhuang-sky, line 150: It can support us to hide the hotels icons on the maps after adding some custom markers on the map to avoid the conflicts
image

packages/bpk-component-map/README.md
```diff
@@ -129,31 +128,34 @@ export default () => (
...
+| Property              | PropType                                                                                        | Required | Default Value                    |
+| --------------------- | ----------------------------------------------------------------------------------------------- | -------- | -------------------------------- |
+| bounds                | shape({north: number, east: number, south: number, west: number})                               | false    | null                             |
+| center                | shape({latitude: number, longitude: number})                                                    | false    | null                             |
+| className             | string                                                                                          | false    | null                             |
+| greedyGestureHandling | bool                                                                                            | false    | false                            |
+| mapOptionStyles       | [array](https://developers.google.com/maps/documentation/javascript/reference/map#MapTypeStyle) | false    | null                             |
```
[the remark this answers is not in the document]
giriawu: Thanks for your suggestion! Have updated.

package.json
```diff
@@ -183,8 +183,8 @@
     "punycode": "^2.1.1",
     "puppeteer": "^9.1.1",
     "raf": "^3.4.1",
-    "react": "16.4.2",
-    "react-dom": "16.4.2",
+    "react": "16.8.3",
+    "react-dom": "16.8.3",
```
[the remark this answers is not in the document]
giriawu: I made such an attempt, but it will cause the map component to fail to run normally and report an error because the main version of react in the project v16.3 does not support hooks😞. So I update it.

jasonhuang-sky (approved):

packages/bpk-component-map/src/BpkMap.js
```diff
@@ -82,29 +106,37 @@ const BpkMap = withGoogleMap((props: Props) => {
     gestureHandling = 'greedy';
   }
 
+  const ref = useRef(null);
+  const mapContainerClassName = getClassName('bpk-map', className);
```
giriawu, line 110: This property will style the map's div container, defaulting to a height of 100% to render it fully.
You can set the container style directly here through className, or you can wrap another layer of div in the outer layer to specify the style like in the example.js.

packages/bpk-component-map/src/withGoogleMapsScript.js
```diff
@@ -19,23 +19,63 @@
...
+    version: PropTypes.string,
+    preventGoogleFontsLoading: PropTypes.bool,
   };
 
   WithGoogleMapsScript.defaultProps = {
     loadingElement: <DefaultLoadingElement />,
+    preventGoogleFontsLoading: true,
+    libraries: ['geometry', 'drawing', 'places'],
+    // https://github.com/JustFly1984/react-google-maps-api/issues/2963
+    version: '3.46',
```
olliecurtis, line 78: Should we be allowing people to customise which version to use or would it be worth us just setting a version in the component to use?

packages/bpk-component-radio/src/__snapshots__/BpkRadio-test.js.snap
```diff
@@ -108,6 +108,7 @@ exports[`BpkRadio should render correctly with checked attribute 1`] = `
     <input
       aria-invalid="false"
       aria-label="Direct"
+      checked=""
```
olliecurtis, line 111: Might be worth checking why this has changed in this PR? Possibly revert it as its not related and expected with this change

packages/bpk-component-switch/src/__snapshots__/BpkSwitch-test.js.snap
```diff
@@ -31,6 +31,7 @@ exports[`BpkSwitch should render correctly when checked 1`] = `
   >
     <input
       aria-label="Switch"
+      checked=""
```
olliecurtis, line 34: Same here

olliecurtis (changes requested):
Could you add an entry to the UNRELEASED file with a `Breaking` change?

packages/bpk-component-map/src/withGoogleMapsScript.js
jasonhuang-sky: Make sense! The issue in the link is only on `3.48.6` and it may be fixed in the next version, but after setting this default version, if we don't pass this param, we will always stay at this old version or need to update periodically instead of getting the latest version through the official default value `weekly`

https://react-google-maps-api-docs.netlify.app/#usejsapiloader
https://developers.google.com/maps/documentation/javascript/versions#release-channels-and-version-numbers

packages/bpk-component-checkbox/src/__snapshots__/BpkCheckbox-test.js.snap
```diff
@@ -78,6 +78,7 @@ exports[`BpkCheckbox should render correctly with checked attribute 1`] = `
     <input
       aria-invalid="false"
       aria-label="Prefer directs"
+      checked=""
```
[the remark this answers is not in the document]
giriawu: Yes, I noticed it at the first time, and had reverted, but I found that when I tested it, it would report an error, it told me that these four files must be updated. I think the only explanation is most likely that I modified the React version.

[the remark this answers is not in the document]
giriawu: I checkout another PR for testing, I found that after upgrading to React v16.8+, in a structure similar to 
`<label><input/></label>`, the input will be automatically added with the checked attribute. 
This seems to make sense, since the components that contain these structures are indeed all related to checked🤔.

packages/bpk-component-map/src/withGoogleMapsScript.js
olliecurtis: I wonder if by not fixing the version are we at risk of having the component break if there is an issue in newer versions so we should instead be monitoring the versions and make sure it is ok before just accepting in the default weekly version?

As originally you had fixed the value because of a bug so by setting a default to weekly we could at some point have broken maps because of a new version having a bug

jasonhuang-sky: Thanks Ollie! 
Considering as a base component, we might be able to leave this choice to the component consumer to think about:

- Pass a `version` parameter when using this component to ensure that there are no bugs
- Do not pass `version` to automatically get the new version, in this case Google API will use `weekly` as the default version, but there may be a risk of bugs in the new version

So I don't think the default `version` here is needed. 
Consumers could choose the `version` they want, otherwise they won't be able to choose the second one. 
In our project we will choose the first one to void the risk. What do you think @olliecurtis ?

olliecurtis: I guess with passing a version parameter is there a risk that could have people using a version that the library is not compatible with? 🤔 I would say having a default fixed value means consumers could change the version but whilst working out what features, etc they have/need/require they could just setup the minimum amount of change then can control it as it happens

jasonhuang-sky: I think it might be, if we don't pass the fixed version value to this component consumer.
So to void the risk, we could not expose the version as a parameter, and if we need any new features of the new version, we could just update the version in this component

olliecurtis (approved):

PR #2429, opened 2022-05-09
[KOA-5029]: Upgrade web to Node 14

olliecurtis:
Upgrades Backpack Web to Node 14 as an interim step to move to a supported Node version.

Follow up work will be completed to upgrade libraries and implementations to fully support Node 16 as currently upgrading to Node 16 causes test failures and has some bugs in npm v8 which is a large task.

So we will break this down in order to support it properly and not make a large PRs that might be unstable or less easy to manage

### DIFF
--- a/.naverc
+++ b/.naverc
@@ -1,1 +1,1 @@
-lts/carbon
+lts/fermium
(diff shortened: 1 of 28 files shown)


anambl (approved):
LGTM!

PR #2434, opened 2022-05-16
[no jira]: Removing unused dependencies

olliecurtis:
This PR removes some unused dips since various migrations.

Benefits of this:
 - Less deps to maintain
 - Increased security as any vulnerable libs removed
 - Good practice to cleanup


anambl (approved):
LGTM! 🎉

PR #2442, opened 2022-05-24
Migrate tests to React Testing Library

anambl:
- Bump `jest` and `babel-jest` to v26
- Temporarily disable screenshot testing due to `Storybook` v5 being incompatible with `jest` v26 
- Bump `@testing-library/react` to v12
- Migrate the tests of the following component to RTL:
  - `bpk-animate-height`
  - `bpk-component-banner-alert`
  - `bpk-component-breakpoint`
  - `bpk-component-image`
  - `bpk-component-input`
  -  `bpk-component-nudger`
- Fix accessibility tests in the following components due to `aria-label` not working without a role:
  - `bpk-component-barchart` - add `graphics-symbol` role when bars are not clickable
  - `bpk-component-rating` & `bpk-component-star-rating` - add `img` role to rating components
 - Temporarily disable accessibility tests in `bpk-component-graphic-promotion` due to tests failing when using nested interactive elements


.github/workflows/ci.yml
```diff
@@ -47,44 +47,7 @@ jobs:
...
-        if: failure() && steps.visualTests.outcome == 'failure' && github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]'
-        with:
-          script: |
-            github.rest.issues.createComment({
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: "Visual regression tests failed 😢. You can download the failure diffs from the 'Artifacts' section of the failed CI run. To update the tests, run `npm run jest:visual-tests:update` locally."
-            })
-
```
olliecurtis, line 87: Can we just comment this instead of removing so its easier to reactive in the future?

olliecurtis (changes requested):
Small change!

olliecurtis (approved):

packages/bpk-component-rating/src/BpkRating.js
```diff
@@ -102,7 +102,12 @@ const BpkRating = (props: Props) => {
 
   return (
     // $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
-    <div className={classNames.join(' ')} aria-label={ariaLabel} {...rest}>
+    <div
+      className={classNames.join(' ')}
+      aria-label={ariaLabel}
+      role="img"
```
olliecurtis, line 108: This might not always be an img which would be incorrect if its text as would be a mis representation

This might be worth having it instead: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/article_role

or even: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/complementary_role

olliecurtis (changes requested):

olliecurtis (approved):
LGTM!

PR #2445, opened 2022-05-30
[BD-6043] Support rating value with string type

Supremeyh:
##  Context
Subject: Support rating value with string type
Ticket:  https://gojira.skyscanner.net/browse/BD-6043
UX:  [figma](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=7986%3A7553)
Confluence:  [Backpack Rating component](https://confluence.skyscannertools.net/display/BD/Rating+component)
Visual link: https://backpack.github.io/storybook-prs/2430/?path=/story/bpk-component-rating--visual-test

## Why Are We Making This Change?
- There is case in some language, the rating value would be string type, such as `4,5` stands for `4.5`. In terms of supporting this case, we make some adjustment for rating component

image

## Change Description
 - Support rating value with string type
 - Remove `toFixed` function for rating value, therefore when pass value with `4`, it won't fixed to `4.0`
 - Deprecate `size` value `lg` and replace with `large` to show vertical style.
 - Update `README.md` and add one more example Hide scale rating 
 - Update `UNRELEASED.md`

### Screenshots

| Before                        | After                         |
|  |  |

## Checklist

- [x] `README.md`
- [x] Tests


packages/bpk-component-rating/src/common-types.js
```diff
@@ -18,7 +18,7 @@
 
 /* @flow */
 
-export const RATING_SIZES = { base: 'base', large: 'lg' };
+export const RATING_SIZES = { base: 'base', large: 'large' };
```
olliecurtis, line 21: What's the reason behind this? This would make this change breaking and might cause more issues for the consumer

olliecurtis (changes requested):
LGTM! One comment

packages/bpk-component-rating/src/common-types.js
Supremeyh: Hi Ollie, thanks for your review. It due to there is conflict between [propTypes](https://github.com/Skyscanner/backpack/blob/ced06446d5dd6be464b17bee92876848d26bfb34/packages/bpk-component-rating/src/BpkRating.js#L157) and  [typescript Props](https://github.com/Skyscanner/backpack/blob/ced06446d5dd6be464b17bee92876848d26bfb34/packages/bpk-component-rating/src/BpkRating.js#L51). It would throw waring as bellow.
image

UNRELEASED.md
```diff
@@ -0,0 +1,4 @@
+**Fixed:**
+
+- bpk-component-rating:
+   - Support rating `value` when type is `string`.
```
jasonhuang-sky, line 4:
```suggestion
   - Support rating `value` when type is a localized `string` number.
   - Fixed `RATING_SIZES.large` not aligning with baseline
   - Update the value of `RATING_SIZES.large` from `lg` to `large` to avoid propTypes check warning 
```

packages/bpk-component-rating/src/common-types.js
Supremeyh: Since it would change the value what consumer pass, it change to `Breaking` in `UNRELEASED.md`

UNRELEASED.md
Supremeyh: Thanks your suggestions, i've update the `UNRELEASED.md`

packages/bpk-component-rating/src/BpkRating-test.js
```diff
@@ -212,4 +212,16 @@ describe('BpkRating', () => {
...
     expect(asFragment()).toMatchSnapshot();
   });
+
+  it('should correctly when value is string type', () => {
+    const { asFragment } = render(
+      <BpkRating
+        ariaLabel="4,6 Wonderful Wise choice"
+        title="Wonderful"
+        subtitle="Wise choice"
+        value="4,6"
```
Supremeyh, line 222: In some language, the decimal point might be `,` rather than `.` .Some online example in production. See [link](https://www.skyscanner.net/hotels/search?entity_id=27544008&checkin=2022-06-04&checkout=2022-06-05&adults=1&rooms=1&impression_id=b2702e51-e8ce-4a6b-880e-d9efa81dc5f8).
image

jasonhuang-sky: We are using `formatNumber` from `saddlebag-localisation` to localize numbers to localized number strings, so usually we have to pass a string as the value instead of passing a number

olliecurtis (approved):

PR #2449, opened 2022-05-30
[KOA-5132] Run audit fix and resolve minimist alert

anambl:
- Run `npm audit fix` to fix vulnerabilities:
  - Bump `cli-progress` (minor)
  - Bump `prompt` (minor)
 
- Remove `backpack-node-sass` package & step to publish css packages


package.json
```diff
@@ -131,10 +131,10 @@
     "babel-jest": "^24.9.0",
     "babel-loader": "^8.0.0",
     "babel-plugin-require-context-hook": "^1.0.0",
-    "backpack-node-sass": "^0.7.1",
+    "backpack-node-sass": "^0.9.0",
```
olliecurtis, line 134: You may need to install a newer version of `node-sass` as this newer version installs minimum v7 which at the same time means we need to bump `sass-loader`

anambl: will remove it altogether as we're not actually building the css files as part of our release process anymore and no one is using the css packages anyways so we can remove it!

olliecurtis (approved):

PR #2452, opened 2022-06-01
[KOA-5098]: Popper JS to v2

olliecurtis:
Updates PopperJS used by the tooltip and portal components to the latest upstream version, from the Skyscanner fork.

As we don't know what is happening in our Skyscanner fork and v1 causes Node 16 and React 17 compatibility issues, we have decided to move to the latest version v2.

In this PR I have added a test to skip, this being due to some strange issue with `focusStore` and `keyboardStoreFocus` not being called during the test, but they are when rendered as a component. So as not to block this work going through and our progression to Node 16, I have disabled it temporarily with a follow up task created to dix

- [ ] `README.md`
- [x] Tests


examples/bpk-component-tooltip/examples.js
```diff
@@ -127,9 +127,12 @@ const PopperModifiersExample = () => (
       ariaLabel="Berlin Brandenburg Airport"
       id="my-tooltip"
       target={<Heading>BER</Heading>}
-      popperModifiers={{
-        flip: { enabled: false },
-      }}
+      popperModifiers={[
```
anambl, line 130: is this a change consumers would have to make?
We might need to also update the README.md of the tooltip component to point to the right documentation link for this prop

olliecurtis: Ah yes you are right good spot on this! Hopefully it will not be massively used property so won't have a big impact but still an breaking change!

anambl (approved):
LGTM!

PR #2454, opened 2022-06-02
[BD-6116] Add a new component: bpk-component-price

colorwin:
Ticket: https://gojira.skyscanner.net/browse/BD-6116
Figma: https://www.figma.com/file/v6aKHTyuA6yXGZPnOv4kYq/Inventory-Card-Components?node-id=3627%3A41603

The price component is widely used in the hotel scene, there's a high priority to refresh across the using scene and contribute to Backpack. 

## Snapshot

image

RTL:
image

- [x] `README.md`
- [x] Tests


packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+const BpkPrice = (props: Props) => {
+  const { className, description, subtitle, title, ...rest } = props;
+
+  return (
+    <BpkBreakpoint query={BREAKPOINTS.MOBILE}>
+      {(isMobile) =>
+        isMobile ? (
+          <div
```
olliecurtis, line 44: A lot of this code is duplicated across when it's mobile or desktop. Could we check to see if we can simplify this instead

For example using isMobile when setting specific class names or !isMobile for when wanting to set classes inside the getClassNames function

olliecurtis (changes requested):

packages/bpk-component-price/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```sh
+npm install bpk-component-price --save-dev
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkPrice from 'bpk-component-code';
```
Supremeyh, line 15: It should be `bpk-component-price`

examples/bpk-component-price/examples.js
```diff
@@ -0,0 +1,37 @@
...
+import BpkPrice from '../../packages/bpk-component-price';
+
+const DefaultExample = () => <BpkPrice title="£1,830" />;
+
+const WithDescriptionExample = () => (
+  <BpkPrice title="£1,830" description="/ night" />
+);
+
+const WithSubtitleDescriptionExample = () => (
+  <BpkPrice subtitle="£209" title="£1,830" description="/ night" />
```
Supremeyh, line 30: For `description` i guess it would better only pass `night` and `/` in `BpkPrice` component.

colorwin: Thank you, Ezreal. I think one field is enough and more flexible.

UNRELEASED.md
```diff
@@ -0,0 +1,7 @@
+**Added:**
+
+- bpk-component-price
+  - Price is a new component.
+  - `title` is a required attribute to show price.
+  - `subtitle` is a optional attributes to show original price before the discount.
+  - `description` is a optional attribute which could be used to show the price is for how many nights.
```
jasonhuang-sky, line 7: Since this is a new one, I think we can just add something to describe what the component is for, find the example [here](https://github.com/Skyscanner/backpack/blob/main/CHANGELOG.md)

examples/bpk-component-price/examples.js
jasonhuang-sky: I agree with @Supremeyh if we always need to pass `/`, why not stick it in this component

examples/bpk-component-price/stories.js
```diff
@@ -0,0 +1,31 @@
...
+  DefaultExample,
+  WithDescriptionExample,
+  WithSubtitleDescriptionExample,
+} from './examples';
+
+storiesOf('bpk-component-price', module)
+  .add('Default', DefaultExample)
+  .add('With description', WithDescriptionExample)
+  .add('With subtitle and description', WithSubtitleDescriptionExample)
+  .add('Visual test', WithSubtitleDescriptionExample);
```
jasonhuang-sky, line 31: Need to add all cases for the visual test

packages/bpk-component-price/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "bpk-component-price",
+  "version": "1.1.9",
```
jasonhuang-sky, line 3: It should be `1.0.0`

packages/bpk-component-price/src/BpkPrice.js
jasonhuang-sky: As @colorwin said [here](https://skyscanner.slack.com/archives/C03GU582NQP/p1654164733818369), it would be more flexible to use a param to control this layout, which would at least have the following benefits
1. More flexible usages
2. No need to add `BpkBreakpoint` as a dependency to reduce the package size
3. It will not support SSR very well, because `BpkBreakpoint` does not support it, that is, if we render this component on the server side, it will be re-rendered on the client side causing to flicker, but if we use param instead, we can use our [custom component](https://github.skyscannertools.net/hotels-web/hotels-website/blob/main/packages/client/src/components/Common/BpkBreakpoint/BpkBreakpointWrapper.tsx#L22) to solve this issue

jasonhuang-sky (changes requested):

examples/bpk-component-price/examples.js
colorwin: Thank you, Jason. Actually, in our usage. It won't be `/` now. It will be `a night` or `total` or `total stay` etc.
image
image
image

packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -0,0 +1,98 @@
...
+
+const getSubTitle = (subtitle, isSmall) =>
+  subtitle && (
+    <BpkText
+      className={getClassName('bpk-price__subtitle')}
+      textStyle={isSmall ? TEXT_STYLES.xs : TEXT_STYLES.sm}
+      tagName="span"
+    >
+      {subtitle}
+    </BpkText>
```
gert-janvercauteren, line 46: Should the subtitle use the textSecondary color?

packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -0,0 +1,98 @@
...
+
+const getDescription = (description, isSmall) =>
+  description && (
+    <BpkText
+      textStyle={isSmall ? TEXT_STYLES.xs : TEXT_STYLES.sm}
+      tagName="span"
+      className={!isSmall && getClassName('bpk-price__descriptionSpacing')}
+    >
+      {description}
+    </BpkText>
```
gert-janvercauteren, line 57: Should this also use the textSecondary color?

packages/bpk-component-price/src/BpkPrice.js
colorwin: I think so, nice caught, Jan.

packages/bpk-component-price/package.json
olliecurtis: Or this'll  need to be `0.0.1` as our release process will version it to v1

olliecurtis:
For the comment from @jasonhuang-sky 

> As @colorwin said [here](https://skyscanner.slack.com/archives/C03GU582NQP/p1654164733818369), it would be more flexible to use a param to control this layout, which would at least have the following benefits
> More flexible usages
> No need to add BpkBreakpoint as a dependency to reduce the package size
> It will not support SSR very well, because BpkBreakpoint does not support it, that is, if we render this component on the server side, it will be re-rendered on the client side causing to flicker, but if we use param instead, we can use our [custom component](https://github.skyscannertools.net/hotels-web/hotels-website/blob/main/packages/client/src/components/Common/BpkBreakpoint/BpkBreakpointWrapper.tsx#L22) to solve this issue

To respond to a couple of points here

> More flexible usages

Could you expand on this?

> No need to add BpkBreakpoint as a dependency to reduce the package size

This will be ok as we are moving away from individual packages to a single package so size will be not an issue due to them all be located together

> It will not support SSR very well, because BpkBreakpoint does not support it, that is, if we render this component on the server side, it will be re-rendered on the client side causing to flicker, but if we use param instead, we can use our [custom component](https://github.skyscannertools.net/hotels-web/hotels-website/blob/main/packages/client/src/components/Common/BpkBreakpoint/BpkBreakpointWrapper.tsx#L22) to solve this issue

This is a good point but could we use the CSS breakpoints and control it via styles instead, I think this will solve the concern here

olliecurtis (changes requested):

jasonhuang-sky:
> For the comment from @jasonhuang-sky
> 
> > As @colorwin said [here](https://skyscanner.slack.com/archives/C03GU582NQP/p1654164733818369), it would be more flexible to use a param to control this layout, which would at least have the following benefits
> > More flexible usages
> > No need to add BpkBreakpoint as a dependency to reduce the package size
> > It will not support SSR very well, because BpkBreakpoint does not support it, that is, if we render this component on the server side, it will be re-rendered on the client side causing to flicker, but if we use param instead, we can use our [custom component](https://github.skyscannertools.net/hotels-web/hotels-website/blob/main/packages/client/src/components/Common/BpkBreakpoint/BpkBreakpointWrapper.tsx#L22) to solve this issue
> 
> To respond to a couple of points here
> 
> > More flexible usages
> 
> Could you expand on this?
> 
> > No need to add BpkBreakpoint as a dependency to reduce the package size
> 
> This will be ok as we are moving away from individual packages to a single package so size will be not an issue due to them all be located together
> 
> > It will not support SSR very well, because BpkBreakpoint does not support it, that is, if we render this component on the server side, it will be re-rendered on the client side causing to flicker, but if we use param instead, we can use our [custom component](https://github.skyscannertools.net/hotels-web/hotels-website/blob/main/packages/client/src/components/Common/BpkBreakpoint/BpkBreakpointWrapper.tsx#L22) to solve this issue
> 
> This is a good point but could we use the CSS breakpoints and control it via styles instead, I think this will solve the concern here

> Could you expand on this?



> For the comment from @jasonhuang-sky
> 
> > As @colorwin said [here](https://skyscanner.slack.com/archives/C03GU582NQP/p1654164733818369), it would be more flexible to use a param to control this layout, which would at least have the following benefits
> > More flexible usages
> > No need to add BpkBreakpoint as a dependency to reduce the package size
> > It will not support SSR very well, because BpkBreakpoint does not support it, that is, if we render this component on the server side, it will be re-rendered on the client side causing to flicker, but if we use param instead, we can use our [custom component](https://github.skyscannertools.net/hotels-web/hotels-website/blob/main/packages/client/src/components/Common/BpkBreakpoint/BpkBreakpointWrapper.tsx#L22) to solve this issue
> 
> To respond to a couple of points here
> 
> > More flexible usages
> 
> Could you expand on this?
> 
> > No need to add BpkBreakpoint as a dependency to reduce the package size
> 
> This will be ok as we are moving away from individual packages to a single package so size will be not an issue due to them all be located together
> 
> > It will not support SSR very well, because BpkBreakpoint does not support it, that is, if we render this component on the server side, it will be re-rendered on the client side causing to flicker, but if we use param instead, we can use our [custom component](https://github.skyscannertools.net/hotels-web/hotels-website/blob/main/packages/client/src/components/Common/BpkBreakpoint/BpkBreakpointWrapper.tsx#L22) to solve this issue
> 
> This is a good point but could we use the CSS breakpoints and control it via styles instead, I think this will solve the concern here

> More flexible usages
In some cases, we do not display the corresponding layout directly according to the device, such as the card on the map, even though on the desktop, but because we don't want to cover the map too much, we will use the mobile layout, actually, the same thing happens on the `BpkRating` so it has a [size](https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-rating/src/BpkRating.js#L157) prop to control the layout and font-sizes

packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -0,0 +1,98 @@
...
+      )}
+      // $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
+      {...rest}
+    >
+      {getSubTitle(subtitle, isSmall)}
+      {conditionalWrap(
+        isSmall,
+        <Fragment>
+          {getTitle(title, isSmall)}
+          {getDescription(description, isSmall)}
```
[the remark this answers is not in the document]
colorwin: Make sense.

packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -0,0 +1,98 @@
...
+      className={getClassName(
+        'bpk-price',
+        isSmall && 'bpk-price--small',
+        className,
+      )}
+      // $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
+      {...rest}
+    >
+      {getSubTitle(subtitle, isSmall)}
+      {conditionalWrap(
```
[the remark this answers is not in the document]
colorwin: To implement the design, the desktop layout needs an extra div. And then, for the mobile layout, we have 2 methods: Method 1 doesn't need an extra wrap div while method 2 keeps the wrap div.

I think method 1 is more simple, so I make the conditional wrap.

image

packages/bpk-component-price/src/common-types.js
```diff
@@ -0,0 +1,48 @@
+/*
```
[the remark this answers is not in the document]
colorwin: Make sense.

packages/bpk-component-price/index.js
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* @flow strict */
+
+import BpkPrice from './src/BpkPrice';
+import { LAYOUTS, type Props } from './src/common-types';
+
+export type BpkPriceProps = Props;
```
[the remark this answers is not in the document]
colorwin: Make sense

packages/bpk-component-price/src/BpkPrice.js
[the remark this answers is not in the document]
jasonhuang-sky: Would be better to try to figure out a way to avoid this and use the same DOM structure between the two layouts and sizes

jasonhuang-sky (commented):
And seems there are something wrong in the storybook 
image

packages/bpk-component-price/src/BpkPrice.js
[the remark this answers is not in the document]
colorwin: Thank you. Have used the method to make the same DOM structure between `small` and `large` size

UNRELEASED.md
```diff
@@ -3,4 +3,9 @@
  - bpk-component-tooltip:
   - Migrated from custom `popper.js` version 1 to upstream version 2.
   - `popperModifiers` now accepts an array of objects instead of an object of options.
-    - Modifiers now follow a new format and existing versions will no longer function as expected. See https://popper.js.org/docs/v2/modifiers/ for more information.
\ No newline at end of file
+    - Modifiers now follow a new format and existing versions will no longer function as expected. See https://popper.js.org/docs/v2/modifiers/ for more information.
```
colorwin, line 6: Not sure why GIt marks this line as change, but their content is same.

jasonhuang-sky (approved):
LGTM

UNRELEASED.md
jasonhuang-sky: Guess it should be caused by the new lines

packages/bpk-component-price/src/BpkPrice-test.js
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+/* @flow strict */
+
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import BpkPrice, { SIZES } from './BpkPrice';
+
+jest.mock('bpk-component-breakpoint');
```
olliecurtis, line 25: Does this still require to be mocked here?

olliecurtis (changes requested):

olliecurtis (approved):

PR #2457, opened 2022-06-09
[KOA-5133]: Migrate calendar tests to RTL

anambl:
- Migrate all calendar tests from `enzyme` to `React Testing Library`
- Change `onMonthTransitionEnd` in `BpkCalendarGridTransition` from arrow function to manually binding as the method can't be found on prototype when using arrow functions and cannot be tested. There should be no performance difference according to [this issue](https://github.com/facebook/react/issues/9851) open with the React team.
- Fixed a bug in `Week` where `selectionType` prop wasn't always correctly set as it was comparing a `string` to a `date`

- [ ] `README.md`
- [ ] Tests


packages/bpk-component-calendar/src/Week.js
```diff
@@ -106,7 +106,9 @@ function getSelectionType(
 
   if (
     selectionConfiguration.type === CALENDAR_SELECTION_TYPE.single &&
-    selectionConfiguration.date === formatDateFull(date)
+    selectionConfiguration.date &&
+    (selectionConfiguration.date === formatDateFull(date) ||
+      formatDateFull(selectionConfiguration.date) === formatDateFull(date))
```
anambl, line 111: added this as currently in our documentation we use as an example for `formatDateFull` a function that takes a date and outputs a string with this format `EEEE, do MMMM yyyy`. `selectionConfiguration.date` should normally be a date object so then they would never be equal. I didn't want to change this directly in case there are consumers that are doing changes so that they are equal because in that case it could break for them, but I think this way if it's broken, it will get fixed. If it's not, it will continue to work

olliecurtis (approved):
Do you need to update the UNRELEASED file due to GridTransition and Week changing?

PR #2461, opened 2022-06-13
[NO JIRA]: Install missing dependencies

olliecurtis:
There were some dependencies that were not added to the single package during the migration, so this PR adds them in to ensure all dependencies are available.

- [ ] `README.md`
- [ ] Tests


anambl (approved):
LGTM! 🎉

PR #2462, opened 2022-06-14
St 4162 center graphic promo text

lkhughes:

- [x] `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
+++ b/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
@@ -152,6 +152,10 @@
 
       &--center {
         align-items: center;
+
+        > * {
+          margin-left: -0.75rem;
+        }
       }
 
       &--end {
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -5,5 +5,8 @@
 
 **Fixed:**
 
+- bpk-component-graphic-promotion
+  - Ensured text is centered correctly despite margins
+
 - `@skyscanner/backpack-web`
   - Added missing dependencies `@skyscanner/bpk-svgs`, `@react-google-maps/api`, `@popperjs/core` 


olliecurtis (approved):
LGTM!

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -152,6 +152,10 @@
 
       &--center {
         align-items: center;
+
+        > * {
+          margin-left: -0.75rem;
```
anambl, line 157: Could we be using a spacing token here instead? 🙂

anambl (changes requested):

anambl:
I'm just wondering what's the reason we chose that value? Not sure if it's something obvious that I just missed, but not sure why adding that margin centers it. If that is the only way of achieving this, it might be helpful to add a comment explaining it?

lkhughes:
> I'm just wondering what's the reason we chose that value for the left margin? Not sure if it's something obvious that I just missed, but not sure why adding that margin centers the text. If that is the only way of achieving this, it might be helpful to add a comment explaining it?

The margins on the container are uneven as set out in the [figma design](https://www.figma.com/file/hkDTmrGXgKbFerStrlgnrD/Homepages-v1-%C2%B7-specs?node-id=1252%3A8583)

As such auto-centering the text in the container leaves it 12px/0.75rem to the right of the actual center of the promo. There isn't a token for this size but I've adjusted the code to use half the value the 24px/1.5rem bpk-spacing-lg token which I hope is suitable.

anambl:
> > I'm just wondering what's the reason we chose that value for the left margin? Not sure if it's something obvious that I just missed, but not sure why adding that margin centers the text. If that is the only way of achieving this, it might be helpful to add a comment explaining it?
> 
> The margins on the container are uneven as set out in the [figma design](https://www.figma.com/file/uQ8LTX9gSU3a2J8zfjK68b/Graphic-Promos-2.0?node-id=497%3A5155)
> 
> As such auto-centering the text in the container leaves it 12px/0.75rem to the right of the actual center of the promo. There isn't a token for this size but I've adjusted the code to use half the value the 24px/1.5rem bpk-spacing-lg token which I hope is suitable.

Great, thanks for explaining! And yes, no worries, totally fine to use half `bpk-spacing-lg`
Will the difference be `0.75rem` on all devices then? As I believe margins are different for each device, for example there will be different margins on desktop vs tablet. Will this still work on tablet and other devices?

lkhughes:
> > > I'm just wondering what's the reason we chose that value for the left margin? Not sure if it's something obvious that I just missed, but not sure why adding that margin centers the text. If that is the only way of achieving this, it might be helpful to add a comment explaining it?
> > 
> > 
> > The margins on the container are uneven as set out in the [figma design](https://www.figma.com/file/uQ8LTX9gSU3a2J8zfjK68b/Graphic-Promos-2.0?node-id=497%3A5155)
> > As such auto-centering the text in the container leaves it 12px/0.75rem to the right of the actual center of the promo. There isn't a token for this size but I've adjusted the code to use half the value the 24px/1.5rem bpk-spacing-lg token which I hope is suitable.
> 
> Great, thanks for explaining! And yes, no worries, totally fine to use half `bpk-spacing-lg` Will the difference be `0.75rem` on all devices then? As I believe margins are different for each device, for example there will be different margins on desktop vs tablet. Will this still work on tablet and other devices?

Double checked and tested and there should be no issues with portrait/mobile as it's only set to affect the landscape orientation and there is no centre alignment for the mobile version. Also since it's now using a spacing token like the rest of the margins it should theoretically be relative. It's rendering correctly when I'm testing mobile/tablet/desktop in any case!

PR #2476, opened 2022-06-21
[ST-4403] adds a11y to floating alert

LouiseReid:
Updates bpk-component-floating-alert to be a11y compliant 

- [x] `README.md`
- [ ] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,3 @@
+**Fixed**
+- bpk-component-floating-notification
+  - adds a11y roles and attributes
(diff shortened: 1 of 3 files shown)


olliecurtis (commented):
I am wondering if we don't need to do this but actually invoke our BpkAriaLive component instead for this exact use case: https://backpack.github.io/components/aria-live?platform=web

LouiseReid:
> I am wondering if we don't need to do this but actually invoke our BpkAriaLive component instead for this exact use case:

Yep, that works. I didn't know about that so I'll close this off

PR #2479, opened 2022-06-21
[ST-4405] Fix accessibility issues in Graphic Promo.

angel3cu:
[ST-4405](https://gojira.skyscanner.net/browse/ST-4405)

Graphic Promo had accessibility issues with the button nested inside an `a` tag. We need both the button and the whole card to be clickable. For accessibility users, however, we treat the whole card as one item and use `aria-hidden` to hide its contents from the screen reader.

Unfortunately, an update to `axe` seems to have broken our current implementation. Axe throws accessibility errors, which we are fixing in this PR by making the card a `div` that uses the styles of Skyscanner's card component. The `div` is clickable and can be focused on using Tab; Enter and Space click the link

- [ ] ~`README.md`~
- [x] Tests
- [ ] ~Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)~

### DIFF
--- a/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
+++ b/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
@@ -99,18 +99,13 @@ const BpkGraphicPromo = (props: Props) => {
     textAlign,
   } = props;
 
-  const onClickWrapper = React.useCallback(
-    (event: React.MouseEvent<HTMLElement>) => {
-      event.stopPropagation();
-      onClick();
-    },
-    [onClick],
-  );
-  const onKeyWrapper = React.useCallback(
-    (event: React.KeyboardEvent<HTMLElement>) =>
-      isAccessibilityClick(event) && onClick(),
-    [onClick],
-  );
+  // FIXME: Use useCallback() here when React is updated.
+  const onClickWrapper = (event: React.MouseEvent<HTMLElement>) => {
+    event.stopPropagation();
+    onClick();
+  };
+  const onKeyWrapper = (event: React.KeyboardEvent<HTMLElement>) =>
+    isAccessibilityClick(event) && onClick();
 
   const cardClasses = getClassName(
     getCardClassName('bpk-card'),
--- a/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
+++ b/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
@@ -22,11 +22,13 @@ import React from 'react';
 import { cssModules } from 'bpk-react-utils';
 import BpkText from 'bpk-component-text';
 import BpkButton from 'bpk-component-button';
-import BpkCard from 'bpk-component-card';
 
-import STYLES from './BpkGraphicPromo.module.scss';
+import CARD_STYLES from '../../bpk-component-card/src/BpkCard.module.scss';
 
-const getClassName = cssModules(STYLES);
+import GP_STYLES from './BpkGraphicPromo.module.scss';
+
+const getClassName = cssModules(GP_STYLES);
+const getCardClassName = cssModules(CARD_STYLES);
 
 export const TEXT_ALIGN = {
   start: 'start',
@@ -94,7 +96,11 @@ const BpkGraphicPromo = (props: Props) => {
     onClick();
   };
 
-  const cardClasses = getClassName('bpk-graphic-promo', className);
+  const cardClasses = getClassName(
+    getCardClassName('bpk-card'),
+    'bpk-graphic-promo',
+    className,
+  );
   const containerClasses = getClassName(
     'bpk-graphic-promo__container',
     `bpk-graphic-promo__container--${textAlign}`,
@@ -104,12 +110,13 @@ const BpkGraphicPromo = (props: Props) => {
   const getTextClasses = (baseClass: string) =>
     getClassName(baseClass, `${baseClass}--${textAlign}`);
   return (
-    <BpkCard
+    <div
       className={cardClasses}
       style={style}
       onClick={onClickWrapper}
+      role="link"
       aria-label={constructAriaLabel(props)}
-      padded={false}
+      tabIndex={0}
     >
       <div className={containerClasses} aria-hidden>
         <div className={getTextClasses('bpk-graphic-promo__sponsor-content')}>
@@ -162,7 +169,7 @@ const BpkGraphicPromo = (props: Props) => {
           </BpkButton>
         </div>
       </div>
-    </BpkCard>
+    </div>
   );
 };
 
--- a/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
+++ b/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
@@ -23,9 +23,9 @@ import { cssModules } from 'bpk-react-utils';
 import BpkText from 'bpk-component-text';
 import BpkButton from 'bpk-component-button';
 
-import GP_STYLES from './BpkGraphicPromo.module.scss';
+import STYLES from './BpkGraphicPromo.module.scss';
 
-const getClassName = cssModules(GP_STYLES);
+const getClassName = cssModules(STYLES);
 
 const ACCESSIBILITY_KEYS = {
   Enter: 13 /* Enter */,
--- a/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
+++ b/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
@@ -100,17 +100,15 @@ const BpkGraphicPromo = (props: Props) => {
   } = props;
 
   const onClickWrapper = React.useCallback(
-    (
-      event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>,
-    ) => {
-      if (event.type === 'click') {
-        event.stopPropagation();
-      } else if (!isAccessibilityClick(event)) {
-        return;
-      }
-
+    (event: React.MouseEvent<HTMLElement>) => {
+      event.stopPropagation();
       onClick();
     },
+    [onClick],
+  );
+  const onKeyWrapper = React.useCallback(
+    (event: React.KeyboardEvent<HTMLElement>) =>
+      isAccessibilityClick(event) && onClick(),
     [onClick],
   );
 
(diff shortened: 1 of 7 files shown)


packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -22,11 +22,15 @@ import React from 'react';
...
-import STYLES from './BpkGraphicPromo.module.scss';
+import CARD_STYLES from '../../bpk-component-card/src/BpkCard.module.scss';
 
-const getClassName = cssModules(STYLES);
+import GP_STYLES from './BpkGraphicPromo.module.scss';
+
+const getClassName = cssModules(GP_STYLES);
+const getCardClassName = cssModules(CARD_STYLES);
+
+const ACCESSIBILITY_KEYS = [13, 32];
```
LouiseReid, line 33: nit: should we comment here that these are enter and space button keys?

LouiseReid (approved):
LGTM 👍

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
angel3cu: I had it below and then I extracted them and forgot. Good point, adding it right away.

olliecurtis: Can we turn this into an object instead similar to other components, this then avoids the need for a comment and lays out what they are

```suggestion
const ACCESSIBILITY_KEYS = {
  "key": 13,
  "otherKey": 32
};
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -89,12 +93,23 @@ const BpkGraphicPromo = (props: Props) => {
   } = props;
 
   // FIXME: Use useCallback() here when React is updated.
-  const onClickWrapper = (event: React.MouseEvent<HTMLElement>) => {
-    event.stopPropagation();
+  const onClickWrapper = (
```
olliecurtis, line 96: We can start to useCallback() here now so not sure if that will impact the changes here?

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -22,11 +22,15 @@ import React from 'react';
...
-import STYLES from './BpkGraphicPromo.module.scss';
+import CARD_STYLES from '../../bpk-component-card/src/BpkCard.module.scss';
 
-const getClassName = cssModules(STYLES);
+import GP_STYLES from './BpkGraphicPromo.module.scss';
+
+const getClassName = cssModules(GP_STYLES);
+const getCardClassName = cssModules(CARD_STYLES);
+
+const ACCESSIBILITY_KEYS = [13 /* Enter */, 32 /* Space */];
```
olliecurtis, line 33: Can we turn this into an object instead similar to other components, this then avoids the need for a comment and lays out what they are

Example: https://github.com/Skyscanner/backpack/blob/main/packages/bpk-react-utils/src/Portal.js#L28-L30


```suggestion
const ACCESSIBILITY_KEYS = {
  Enter: 13,
  Space: 32
};
```

olliecurtis (changes requested):

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
angel3cu: No impact, but happy to do this.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -88,13 +95,30 @@ const BpkGraphicPromo = (props: Props) => {
...
-  const onClickWrapper = (event: React.MouseEvent<HTMLElement>) => {
-    event.stopPropagation();
-    onClick();
-  };
+  const onClickWrapper = React.useCallback(
+    (
+      event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>,
+    ) => {
+      if (event.type === 'click') {
+        event.stopPropagation();
```
seraco, line 103: Are we sure we don't want to stop propagation in keyboard initiated events?

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -89,12 +93,23 @@ const BpkGraphicPromo = (props: Props) => {
...
 
   // FIXME: Use useCallback() here when React is updated.
-  const onClickWrapper = (event: React.MouseEvent<HTMLElement>) => {
-    event.stopPropagation();
+  const onClickWrapper = (
+    event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>,
+  ) => {
+    if (event.type === 'click') {
+      event.stopPropagation();
+    } else if (!ACCESSIBILITY_KEYS.includes(event.keyCode || event.which)) {
```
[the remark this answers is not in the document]
angel3cu: Sorted. I added key for new browsers, but kept `keyCode` and `which` for older ones.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
angel3cu: Sorted.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
angel3cu: I tried this and hooks don't work with unit tests for some reason. I will revert this.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -89,12 +93,23 @@ const BpkGraphicPromo = (props: Props) => {
   } = props;
 
   // FIXME: Use useCallback() here when React is updated.
-  const onClickWrapper = (event: React.MouseEvent<HTMLElement>) => {
-    event.stopPropagation();
+  const onClickWrapper = (
+    event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>,
+  ) => {
+    if (event.type === 'click') {
```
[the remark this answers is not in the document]
angel3cu: I just wanted to keep it in one place, but I agree it is better to keep things simple. Sorted! 😄

seraco (approved):
LGTM 💪🏻

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -103,13 +120,19 @@ const BpkGraphicPromo = (props: Props) => {
 
   const getTextClasses = (baseClass: string) =>
     getClassName(baseClass, `${baseClass}--${textAlign}`);
+
   return (
-    <BpkCard
```
olliecurtis, line 107: In thinking about this have you tried making the BpkCard `non-atomic` https://backpack.github.io/components/card?platform=web#non-atomic as this might be the solution that you need for this and not need too much of the work you are doing?

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -22,11 +22,22 @@ import React from 'react';
...
 
-const getClassName = cssModules(STYLES);
+import GP_STYLES from './BpkGraphicPromo.module.scss';
+
+const getClassName = cssModules(GP_STYLES);
+const getCardClassName = cssModules(CARD_STYLES);
+
+const ACCESSIBILITY_KEYS = {
+  Enter: 13 /* Enter */,
+  Space: 32 /* Space */,
```
olliecurtis, line 35: Have you tried this with VoiceOver on Safari? As I know that in order to select a element its a combination of keys and not just space, so not sure if just space gets detected or the event that gets triggered specifically

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -22,11 +22,15 @@ import React from 'react';
 import { cssModules } from 'bpk-react-utils';
 import BpkText from 'bpk-component-text';
 import BpkButton from 'bpk-component-button';
-import BpkCard from 'bpk-component-card';
 
-import STYLES from './BpkGraphicPromo.module.scss';
+import CARD_STYLES from '../../bpk-component-card/src/BpkCard.module.scss';
```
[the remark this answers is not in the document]
angel3cu: Cool, I'll cherry-pick the styles we need. In hindsight, using a BpkCard for this was probably a mistake.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
angel3cu: Thanks for bringing this up. Testing on Safari now. 🙂

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -22,11 +22,19 @@ import React from 'react';
 import { cssModules } from 'bpk-react-utils';
 import BpkText from 'bpk-component-text';
 import BpkButton from 'bpk-component-button';
-import BpkCard from 'bpk-component-card';
 
-import STYLES from './BpkGraphicPromo.module.scss';
+import GP_STYLES from './BpkGraphicPromo.module.scss';
```
olliecurtis, line 26: Can this be reverted back to `STYLES` to follow the format back as other components?

```suggestion
import STYLES from './BpkGraphicPromo.module.scss';
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js
```diff
@@ -22,11 +22,19 @@ import React from 'react';
...
 
-const getClassName = cssModules(STYLES);
+const getClassName = cssModules(GP_STYLES);
```
olliecurtis, line 28: Same as the above

```suggestion
const getClassName = cssModules(STYLES);
```

olliecurtis (changes requested):
Small naming convention change but otherwise all is great!

olliecurtis (approved):

PR #2491, opened 2022-07-04
[NO JIRA] remove deprecated spacing tokens

anambl:

- [ ] `README.md`
- [ ] Tests


olliecurtis (approved):
This is gonna trigger a release so not sure if we need to add a changelog entry?

anambl:
> This is gonna trigger a release so not sure if we need to add a changelog entry?

hmm good point, I'll add a changelog entry!

PR #2494, opened 2022-07-11
[ST-4398] Add trigger button to Floating Notification component Storybook

alakbarz:
# Changes

- Added a new prop `onExit` which takes in a function to execute after the exit animation of the floating notification component has finished
- Added "Trigger alert" button to the Storybook examples so that users can replay the animation

image

# Checklist

- [x] `README.md`
- [x] Tests
  - Follow up ticket [TICKET GOES HERE]

### DIFF
--- a/packages/bpk-component-floating-notification/src/BpkFloatingNotification-test.js
+++ b/packages/bpk-component-floating-notification/src/BpkFloatingNotification-test.js
@@ -104,4 +104,13 @@ describe('BpkFloatingNotification', () => {
 
     expect(onClick).toHaveBeenCalledTimes(1);
   });
+
+  it('should do function when unmounted (exited)', () => {
+    const onExit = jest.fn();
+    render(<BpkFloatingNotification onExit={onExit} {...props} />);
+
+    setTimeout(() => {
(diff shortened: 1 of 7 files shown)


packages/bpk-component-floating-notification/src/BpkFloatingNotification-test.js
```diff
@@ -104,4 +104,13 @@ describe('BpkFloatingNotification', () => {
...
     expect(onClick).toHaveBeenCalledTimes(1);
   });
+
+  it('should do function when unmounted (exited)', () => {
+    const onExit = jest.fn();
+    render(<BpkFloatingNotification onExit={onExit} {...props} />);
+
+    setTimeout(() => {
+      expect(onExit).toHaveBeenCalledTimes(1);
+    });
```
alakbarz, line 114: Without the `setTimeout()` this test fails. I beileve it's because we need to wait for the notification to disappear (it takes ~4800ms, 800ms for the in and out animations and it holds for 4 seconds), however it is odd that I don't have to specify a period of time to wait.

I'm open to suggestions on improving this, but it appears to work.

olliecurtis (approved):
LGTM!

PR #2505, opened 2022-07-18
[KOA-5424]: Fix for findDOMNode in horizontal nav

olliecurtis:
This PR removes the need for using `findDOMNode` which is deprecated in strict mode - it also when is use produces the following error: 'Unable to find node on an unmounted component.'

To do this we remove the need of ReactDOM.findDOMNode and utilise React refs instead

- [ ] `README.md`
- [x] Tests


paultan2021 (approved):
LGTM

frugoman (approved):
LGTM

PR #2508, opened 2022-07-26
[ST-4451]: Add contentId to Graphic Promotion component

jonlaw101:
This adds an optional property to the Graphic Promotion component, to attach an 'id' to the 'div' wrapping the main content, rather than the top level containing 'div'. This is to uniquely identify the main content, to be used for purposes outside of Backpack.

- [x] `README.md`
- [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,0 +1,4 @@
+**Added:**
+
+- `bpk-component-graphic-promotion`: `3.1.2` => `3.1.3`
+  - Added optional property `contentId`, to apply a HTML attribute `id` to the `div` containing the main content of the Graphic Promotion.
--- a/.spelling
+++ b/.spelling
@@ -286,6 +286,7 @@ collapseDelay
 color
 compareValues
 containerClassName
+contentId
 cornerStyle
 CORNER_STYLES.sm
 CORNER_STYLES.lg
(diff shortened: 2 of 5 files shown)


jonlaw101:
> Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)

Unsure if this change requires any meaningful documentation update? Happy to do so if I can be pointed to what would be changed.

UNRELEASED.md
```diff
@@ -0,0 +1,4 @@
+**Added:**
+
+- `bpk-component-graphic-promotion`: `3.1.2` => `3.1.3`
+  - Added optional property `contentId`, to apply a HTML attribute `id` to the `div` containing the main content of the Graphic Promotion.
```
olliecurtis, line 4: Out of interest what is the reasoning for naming the property `contentId` over just `id`? 

Given there is only one id applied it would make sense to name it as the default?

jonlaw101: Perhaps I'm overthinking it but my reasoning here is that my expectation for a `id` is that it would be applied at the top level, whereas this `contentId` is applied nested further in.

I'm happy to rename this to `id` if that works better though

jonlaw101:
Closing this PR in favour of #2511 , which is a branch, whereas this PR is from a fork

PR #2509, opened 2022-07-26
[ST-4102] Refactor `bpk-component-graphic-promotion` to use the standard breakpoints.

angel3cu:
This PR modifies the graphic promotion to use the new standard breakpoints from Backpack. Suggestions from design listed [here](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=14245%3A17584) were included.

**Preview Video:**

https://user-images.githubusercontent.com/65911077/181204443-f064c297-28e5-4c06-b5ab-a056184baac5.mov

**NOTE:** We may have to change the logo we use in Storybook because it is too large. Can we use the same logo as the designs [here](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=13148%3A15323)?

- [x] `README.md`
- [x] Tests


robaw:
Would it be possible to leave comments next to each CSS change that isn't related to the breakpoint swap, giving more context on what it's trying to achieve (what design outcome) and if you can, how the change achieves that. Right now it's a bit hard to join the 2 together.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,359 @@
...
 
-    &__sponsor-content {
-      margin-bottom: bpk-spacing-xl();
+  @include bpk-breakpoint-small-mobile {
+    @include aspect-ratio(133%);
+  }
+
+  &::after {
+    display: none;
+  }
```
jonlaw101, line 62: I see this was already present before this change, but would it be possible to explain what this does?

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,359 @@
...
+
+        /* adjusting for 1.5rem difference between margins */
+        > * {
+          @include bpk-margin-trailing(bpk-spacing-lg());
+        }
+      }
+
+      &--end {
+        align-items: end;
+      }
```
jonlaw101, line 90: Not 100% this suggestion would work but:

Would it be possible to split `start`, `middle` and `end` into their own rules, and combine them on the element such as `class="bpk-graphic-promo__container start"`

The idea here would that it may be possible to eliminate the duplication further down in this file? e.g. L100-116.

angel3cu: Backpack uses BEM, so that would go against the rules, I am afraid.

angel3cu: I really hate the duplication and the resetting overrides I had to make, but this is because of the way the new breakpoints are unfortunately. An alternative would be to use mixins, but then not everything can be a mixin, so I tried to put the line somewhere. 🙂

olliecurtis:
> We may have to change the logo we use in Storybook because it is too large. Can we use the same logo as the designs [here](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=13148%3A15323)?

This is fine, its just an example so doesn't matter in the grand scheme of things, I will check with our designer and if they really want it changed I can provide the other one, but this is just an example/minor detail so all good right now

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
olliecurtis: >Backpack uses BEM, so that would go against the rules, I am afraid.

This is correct!

Let me check what we may have done for other components there might be something else we can do here!

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,359 @@
...
+      @include bpk-subheading;
+    }
+
+    @include bpk-breakpoint-mobile {
+      @include bpk-body-longform;
+    }
+
+    @include bpk-breakpoint-small-mobile {
       @include bpk-body-default;
     }
```
olliecurtis, line 257: Here you can simplify this and only need to override when its at a breakpoint, so:

```
@include bpk-body-default
@include bpk-breakpoint-small-tablet {
  @include bpk-subheading;
}

@include bpk-breakpoint-mobile {
  @include bpk-body-longform;
}

```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,359 @@
...
+
+      &--center {
+        align-items: center;
+        text-align: center;
+      }
+
+      &--end {
+        align-items: start;
+        text-align: start;
       }
```
olliecurtis, line 300: Similar here, you can simplify this by only needing to define `--start`, `--center`, `--end` once and just use the breakpoints to override the width

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,359 @@
...
-      /* adjusting for 1.5rem different between margins */
-      &--center > * {
-        @include bpk-margin-trailing(bpk-spacing-lg());
-      }
+    @include bpk-breakpoint-small-mobile {
+      width: initial;
+      align-items: initial;
+      text-align: initial;
     }
+  }
```
olliecurtis, line 320: One thing it might be worth trying to clean this up (no idea if this will work in this case)

```
@include bpk-breakpoint-small-tablet,
@include bpk-breakpoint-mobile,
@include bpk-breakpoint-small-mobile {
 width: initial;
 align-items: initial;
 text-align: initial;
}
```

olliecurtis (changes requested):

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,359 @@
...
-    }
+  @include bpk-breakpoint-mobile {
+    @include aspect-ratio(133%);
+  }
 
-    &__sponsor-content {
-      margin-bottom: bpk-spacing-xl();
+  @include bpk-breakpoint-small-mobile {
+    @include aspect-ratio(133%);
+  }
```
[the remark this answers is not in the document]
angel3cu: Good point. I can delete the other two breakpoints. 🙂

UNRELEASED.md
```diff
@@ -0,0 +1,12 @@
+**Fixed**
```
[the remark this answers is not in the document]
angel3cu: 👍🏼

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
angel3cu: Sorted. 🙂

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
jonlaw101: I'm happy with this if there isn't a reasonable solution to this 👍

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
angel3cu: 👍🏼

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
angel3cu: Yeah, I tried this, but SCSS does not support it. We need to either write a complicated mixin with `if...else` or create a mixin for the 3 properties, then include it in each breakpoint. 😞

angel3cu: Given that it is 3 properties, I thought writing another mixin isn't worth it.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
angel3cu: I did the same as the other suggestions below. 🙂

angel3cu: I simplified the code a bit, based on the other suggestions. 🙂

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,359 @@
...
+      @include fill-parent;
+
+      &--inverted {
+        flex-direction: column-reverse;
       }
 
-      @include breakpoint-landscape-tablet {
-        @include bpk-heading-1;
+      > * {
+        margin: initial;
```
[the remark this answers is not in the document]
angel3cu: Good spot! This was because the selector for `bpk-margin-leading` and `bpk-margin-trailing` has two parts in RTL (`html[dir='rtl'] .bpk-graphic-promo__container`), which has a higher specificity than the selector I am using which only has one part (`.bpk-graphic-promo__container`).

I've modified the code to override the style using `bpk-margin-leading(0, true)`, which sets the leading margin to zero and resets the one opposite also to 0.

image

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,359 @@
...
+      align-items: initial;
+
+      @include fill-parent;
+
+      &--inverted {
+        flex-direction: column-reverse;
+      }
+
+      > * {
+        margin: initial;
```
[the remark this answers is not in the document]
angel3cu: Sorted. 😄

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,215 @@
...
+    }
+
+    @include bpk-breakpoint-small-tablet {
+      padding: bpk-spacing-xl();
+      justify-content: space-between;
+      align-items: initial;
 
-      @include breakpoint-landscape-tablet {
-        @include bpk-heading-1;
+      @include bpk-margin-leading(0, true);
```
olliecurtis, line 91: You shouldn't need to pass true as the value as it defaults to true :) 
```suggestion
      @include bpk-margin-leading(0);
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
```diff
@@ -33,227 +45,215 @@
...
+      @include fill-parent;
+
+      &--inverted {
+        flex-direction: column-reverse;
       }
 
-      &:not(:last-child) {
-        margin-bottom: bpk-spacing-md();
+      > * {
+        @include bpk-margin-leading(0, true);
```
olliecurtis, line 99: You shouldn't need to pass true as the value as it defaults to true :) 
```suggestion
      @include bpk-margin-leading(0);
```

olliecurtis (changes requested):

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
angel3cu: I just wanted to be explicit, but I removed it now. 🙂

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.module.scss
angel3cu: Removed.

olliecurtis (approved):

PR #2524, opened 2022-08-03
[KOA-4994] Upgrade React version to 17.0.2

anambl:
- Deleted the following deprecated components:
  - `bpk-component-navigation-stack`
  - `PortalV1` (in `bpk-react-utils`)
- Dropped support for `React` versions < 17
- Remove `enzyme` dependency as all tests are now using `React Testing Library`


UNRELEASED.md
```diff
@@ -0,0 +1,6 @@
+**Changed:**
+- Deleted the following deprecated components:
+  - `bpk-component-navigation-stack`
+  - `PortalV1` (in `bpk-react-utils`)
+
+- Dropped support for React versions < 17. No new functionality is introduced, but from this point forwards it should not be assumed React 16 will work.
```
olliecurtis, line 6: As this is a big jump, particularly as there are still some React < 16 we should probably mark this as a breaking change as its officially dropping support

anambl: Hm ok good point 👍

olliecurtis (approved):

PR #2531, opened 2022-08-09
[KOA-5027]: Migrate calendar from react-virtualized

olliecurtis:
Migrated the scrollable calendar from previously heavy react-virtualized library to much more lightweight library to aid in React 17 support and also reduce bundlesizes

- [ ] `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
+++ b/packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
@@ -21,12 +21,7 @@ import React from 'react';
 import { cssModules, deprecated } from 'bpk-react-utils';
 import { DateUtils, BpkCalendarGridPropTypes } from 'bpk-component-calendar';
 import { startOfDay, startOfMonth } from 'date-fns';
-import {
-  AutoSizer,
-  List,
-  CellMeasurer,
-  CellMeasurerCache,
-} from 'react-virtualized';
+import { VariableSizeList as List } from 'react-window';
 
 import STYLES from './BpkScrollableCalendarGridList.module.scss';
 import BpkScrollableCalendarGrid from './BpkScrollableCalendarGrid';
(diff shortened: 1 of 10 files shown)


packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -49,87 +52,121 @@ class BpkScrollableCalendarGridList extends React.Component {
...
-      cache,
+      monthItemHeights,
+      outerHeight: ESTIMATED_MONTH_ITEM_HEIGHT,
     };
   }
 
+  componentDidMount = () => {
+    this.setComponentHeight();
+    if (typeof document !== 'undefined') {
+      document.addEventListener('resize', this.setComponentHeight);
```
anambl, line 78: the event listener for `resize` will actually need to be attached on the `window`, not the `document`. I've seen this issue in the past 😅

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -49,87 +52,121 @@ class BpkScrollableCalendarGridList extends React.Component {
...
+      document.addEventListener('orientationchange', this.setComponentHeight);
+      document.addEventListener('fullscreenchange', this.setComponentHeight);
+    }
+  };
+
+  componentWillUnmount = () => {
+    if (typeof document !== 'undefined') {
+      document.removeEventListener('resize', this.setComponentHeight);
```
anambl, line 86: same as above -  event listener for resize will actually need to be attached on the `window`, not the `document`

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
```diff
@@ -49,87 +52,121 @@ class BpkScrollableCalendarGridList extends React.Component {
...
+      return BASE_MONTH_ITEM_HEIGHT + ROW_HEIGHT * rowCount;
     });
+
     this.state = {
       months,
-      cache,
+      monthItemHeights,
+      outerHeight: ESTIMATED_MONTH_ITEM_HEIGHT,
     };
   }
```
frugoman, line 73: could this be moved to a separate class/function/custom hook to be able to test it on isolation? and add a test to it

frugoman: I think I would also wrap the call to such a function in a useMemo hook, to cache the calculation when possible

olliecurtis: I will double check on this as I am reverting previous work we have done back to what it once was back a few months ago! 

So will be putting it all back as it was but will see what can be done to improve it

packages/bpk-component-scrollable-calendar/src/BpkScrollableCalendarGridList.js
olliecurtis: Ah I must've missed this when bringing back the old changes! But also this might not have been part of the current set so will see!

anambl (approved):

PR #2533, opened 2022-08-11
[KOA-5473] Update components to semantic tokens

anambl:
|Before|After|
|-|-|
|image|image|
|image|image|
|image|image|
|image|image|

- [ ] `README.md`
- [ ] Tests


.storybook/themeableAttributes.js
```diff
@@ -18,7 +18,6 @@
 
 /* @flow */
 
-import { themeAttributes as accordionThemeAttributes } from '../packages/bpk-component-accordion';
```
frugoman, line 21: is this change related to update semantic tokens?

packages/bpk-component-accordion/index.js
```diff
@@ -22,12 +22,10 @@ import BpkAccordion from './src/BpkAccordion';
...
 import withSingleItemAccordionState from './src/withSingleItemAccordionState';
 import withAccordionItemState from './src/withAccordionItemState';
-import themeAttributes from './src/themeAttributes';
 
 export {
   BpkAccordion,
   BpkAccordionItem,
   withSingleItemAccordionState,
   withAccordionItemState,
-  themeAttributes,
```
frugoman, line 32: same as above. is this needed/related?

packages/bpk-component-accordion/src/BpkAccordionItem.module.scss
```diff
@@ -84,27 +63,7 @@ $bpk-spacing-v2: true;
...
-        --bpk-accordion-hover-color,
-        $bpk-color-sky-blue-shade-03
-      );
-    }
-
-    &:active {
-      @include bpk-themeable-property(
-        fill,
-        --bpk-accordion-active-color,
-        $bpk-color-sky-blue-shade-03
```
frugoman, line 105: do we need to also remove this vars where they were declared? (maybe not, just to know)

.storybook/themeableAttributes.js
anambl: Not exactly. But we use these theming for `hover` and `active` states which we don't need in the accordion component as they will all have the same colours. So we can just get rid of them. It's more like cleaning up the code as we go

packages/bpk-component-accordion/index.js
anambl: yep same as above, just cleaning up as we don't need this anymore

packages/bpk-component-accordion/src/BpkAccordionItem.module.scss
anambl: same 😬

frugoman (approved):

PR #2535, opened 2022-08-12
Convert all graphic promotion files to TypeScript

dominicfraser:
This PR updates the Graphic Promotion component to TypeScript. There should be zero functional production changes.

It changes test and Storybook files alongside the implementation, and fixes errors in those files that TS showed.

- [x] `README.md`
- [x] Tests


packages/bpk-component-graphic-promotion/README.md
```diff
@@ -8,25 +8,25 @@ npm install bpk-component-graphic-promotion --save-dev
...
 ## Usage
 
-```js
+```tsx
 import React from 'react';
 import BpkGraphicPromo from 'bpk-component-graphic-promotion';
 
 export default () => (
   <BpkGraphicPromo
+    buttonText="Learn more"
```
dominicfraser, line 17: A bunch of these are just sorting alphabetically. It makes the diff a bit hard to read, but makes the existing code much easier to skim.

packages/bpk-component-graphic-promotion/index.ts
```diff
@@ -15,7 +15,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* @flow strict */
```
dominicfraser, line 18: 👋 🥾

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
```diff
@@ -0,0 +1,123 @@
...
+import BpkGraphicPromo, { TEXT_ALIGN } from './BpkGraphicPromo';
+import type { Props } from './BpkGraphicPromo';
+import STYLES from './BpkGraphicPromo.module.scss';
+
+const getClassName = cssModules(STYLES);
+
+const props: Props = {
+  buttonText: 'Learn more',
+  headline: 'Ride your wave',
+  invertVertically: false,
```
dominicfraser, line 32: `invertVertically` was missing, but is a required prop.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
```diff
@@ -0,0 +1,123 @@
...
+  },
+  subheading:
+    'Portugal and 6 more countries have just been added to the UK travel green list',
+  tagline: 'Tagline',
+  textAlign: TEXT_ALIGN.start,
+};
+
+describe('BpkGraphicPromo', () => {
+  beforeEach(() => {
+    (props.onClick as jest.MockedFunction<Props['onClick']>).mockReset();
```
dominicfraser, line 47: Start TS for mocked functions, otherwise TS will complain that `.mockReset` does not exist on `onClick`.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
```diff
@@ -0,0 +1,123 @@
...
+  });
+
+  it('should render correctly with all properties set', () => {
+    const { asFragment } = render(<BpkGraphicPromo {...props} />);
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
```
dominicfraser, line 55: There had been a test for

```tsx
  it('should support arbitrary props', () => {
    const { asFragment } = render(<BpkGraphicPromo testid="123" {...props} />);

  });
  ```
  
but this proved to be a lie. Looking at the snapshot we can see `testid` was never actually applied. Possibly from an old implementation? The big downside of snapshot tests, people very commonly just press 'update' and don't actually check the diff.

File: packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx [hunk 6]
```diff
@@ -0,0 +1,197 @@
...
+  subheading: Props['subheading'],
+  tagline: Props['tagline'],
+) => {
+  const text: string[] = [];
+  const addText = (value: string) => value && text.push(value);
+
+  if (sponsor) {
+    addText(sponsor.label);
+    addText(sponsor.altText);
+  } else if (tagline) {
```
dominicfraser, line 74: `tagline` is nullable, so TS showed we needed a conditional here.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  addText(headline);
+  addText(buttonText);
+
+  return text.join('. ');
+};
+
+const BpkGraphicPromo = ({
+  buttonText,
+  className = null,
```
dominicfraser, line 90: Stop using soon-to-be-deprecated `.defaultProps` and start using the recommended default arguments.

https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -0,0 +1,197 @@
...
+  onClick: () => void;
+  sponsor?: {
+    label: string;
+    logo: string;
+    altText: string;
+  } | null;
+  style?: {};
+  subheading?: string | null;
+  tagline?: string | null;
+  textAlign: typeof TEXT_ALIGN[keyof typeof TEXT_ALIGN];
```
dominicfraser, line 58: In our main codebase we have a util for this - where should we put TS utils?

```ts
export type Values<T> = T[keyof T];
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -0,0 +1,197 @@
...
+        sponsor,
+        subheading,
+        tagline,
+      )}
+      tabIndex={0}
+      onClick={onClickWrapper}
+      onKeyDown={onKeyWrapper}
+    >
+      <div
+        id={(contentId && contentId) || ''}
```
dominicfraser, line 139: `contentId` is nullable, add a conditional.

packages/bpk-component-graphic-promotion/src/accessibility-test.tsx
```diff
@@ -44,13 +42,14 @@ describe('BpkGraphicPromo accessibility tests', () => {
...
     const { container } = render(
       <BpkGraphicPromo
+        buttonText={buttonText}
         className="test"
-        tagline={tagline}
         headline={headline}
-        subheading={subheading}
-        sponsor={sponsor}
-        buttonText={buttonText}
+        invertVertically={false}
```
dominicfraser, line 48: Add missed `invertVertically` that TS makes us notice

olliecurtis:
>This PR depends on tooling updates in https://github.com/Skyscanner/backpack/pull/2527,

~Can we make tooling updates in a separate PR? Having it tightly coupled whilst doing work seems like its going to cause blocks~ Scratch that I am just in the process of raising a separate PR to solve this for both PRs

packages/bpk-component-graphic-promotion/README.md
```diff
@@ -35,17 +35,17 @@ export default () => (
...
 | onClick          | func              | true     | -             |
-| textAlign        | oneOf(TEXT_ALIGN) | true     | -             |
-| className        | string            | false    | null          |
-| style            | object            | false    | {}            |
-| tagline          | string            | false    | null          |
 | sponsor          | object            | false    | null          |
+| style            | object            | false    | {}            |
 | subheading       | string            | false    | null          |
+| tagline          | string            | false    | null          |
+| textAlign        | oneOf(TEXT_ALIGN) | true     | -             |
```
olliecurtis, line 48: These should remain below the required props - I know alphabetically makes sense but we found that - listing all the required props first and then optional ones next makes it easier at a glance for consumers

Within those groupings go crazy with alphabetical

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
olliecurtis: Might also be from the default boilerplate tests that get created when you run the new component script

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
olliecurtis: Can this not be just and actually it runs as an if statement if its available or otherwise fallback to the `''`?
 
```suggestion
        id={contentId || ''}
```

olliecurtis (changes requested):
Couple of comments.

Could you also add an entry to the UNRELEASED file?

packages/bpk-component-graphic-promotion/src/accessibility-test.tsx
dominicfraser: I think you're right, here in the JS we assign it a defaultValue: https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.js#L214

In the TS we also give it a default value. 

I'd made the Props type based on just whats in the docs, but the docs and the code don't match.

I'll update the docs for it to be optional to match the code - as I think it having a default of false semantically makes sense and its the docs that are wrong.

dominicfraser: https://github.com/Skyscanner/backpack/pull/2535/commits/a1e654929188f7acf9bace3e84f043fd8b1735dc

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: https://github.com/Skyscanner/backpack/pull/2535/commits/0bf909af62c69f32d0b708b4f479ea2b44a9f16b

packages/bpk-component-graphic-promotion/README.md
dominicfraser: TIL

https://github.com/Skyscanner/backpack/pull/2535/commits/d0b202e3885754c3676c70bed95308631d2c8f37

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -15,12 +15,13 @@
...
  */
-/* @flow strict */
 
-import PropTypes from 'prop-types';
 import React from 'react';
+// @ts-expect-error Untyped import. Not added as a stub declaration so consumers still have the suppression in their node_modules when shipped untranspiled.
 import { cssModules } from 'bpk-react-utils';
+// @ts-expect-error Untyped import. Not added as a stub declaration so consumers still have the suppression in their node_modules when shipped untranspiled.
 import BpkText from 'bpk-component-text';
+// @ts-expect-error Untyped import. Not added as a stub declaration so consumers still have the suppression in their node_modules when shipped untranspiled.
```
dominicfraser, line 24: Tested prior to this addition with `@skyscanner/backpack-web@9.2.0-beta.1`.

This proved:

We need to update the script that transforms from multi to single package to also target imports in TS files. This was addressed in https://github.com/Skyscanner/backpack/commit/2d795915283b71ad3aafa7fc7ce25eb15e220570.

It also showed that `npm start` via `react-scripts` typecheck runs also against non-transpiled node_modules.

When running with `"noImplicitAny": true` this causes a problem, as these imports error.

To fix this (in the interim before they are converted to TS themselves) that means either:

1. Consumers need to add lines like `declare module '*/bpk-component-text';` to their stub declaration files to account for the `../../bpk-component-text` import it will be transformed to
2. We can add ts-expect-error inline in each case. This is more ugly inside Backpack, but less work for consumers. The advantage of ts-expect-error over ts-ignore is that when the suppression is no longer needed it will generate an error, reminding us to clear it up.

I've gone for (2) here, feeling it lower effort for consumers while we incrementally convert all of Backpack over.

WDYT?

dominicfraser: cc: @kirstybryce

dominicfraser: Rather than a long inline comment, we could also have a shortened inline comment that references a decision doc?

dominicfraser: I think the context may be worth keeping around past this PR, so have put it in /decisions and updated so that each suppression directs people there. This should help people to understand why, and also if someone finds a better way make it easier for them to propose it.

dominicfraser: Tested in `@skyscanner/backpack-web@9.2.0-beta.3` with ts-expect-error

olliecurtis (commented):
One minor comment - I will also check with the team for alignment and agreement of the decision doc that has been added!

dangerfile.js
```diff
@@ -86,12 +86,12 @@ const componentSourceFilesModified = fileChanges.some(
...
 
 const snapshotsModified = fileChanges.some(
-  (filePath) => filePath.endsWith('.js.snap') || filePath.endsWith('.ts.snap'),
+  (filePath) => filePath.endsWith('.js.snap') || filePath.endsWith('.tsx.snap'),
 );
 
 if (componentSourceFilesModified && !snapshotsModified) {
   warn(
-    "Package source files (e.g. `packages/package-name/src/Component.js`) were updated, but snapshots weren't. Have you checked that the tests still pass?",
+    "Package source files (e.g. `packages/package-name/src/Component.tsx`) were updated, but snapshots weren't. Have you checked that the tests still pass?",
```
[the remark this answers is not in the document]
dominicfraser: My thinking was that as we know we're aiming for TS only this helps to keep that mentality in mind from step one (this PR) onwards? We're now in 'Flow and plain JS is legacy', matching the internal standard.

WDYT?

>  I will also check with the team for alignment and agreement of the decision doc that has been added!

Thanks!

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: and cc: @LexSwed

LexSwed: I love `// @ts-expect-error` with decision record link 😍

dangerfile.js
```diff
@@ -86,12 +86,12 @@ const componentSourceFilesModified = fileChanges.some(
 );
 
 const snapshotsModified = fileChanges.some(
-  (filePath) => filePath.endsWith('.js.snap') || filePath.endsWith('.ts.snap'),
+  (filePath) => filePath.endsWith('.js.snap') || filePath.endsWith('.tsx.snap'),
```
angel3cu, line 89: What is the convention for Backpack? I guess test files that take snapshot have to be `tsx` to render the component for snapshot, but just to quickly double check.

decisions/imports-ts-suppressions.md
```diff
@@ -0,0 +1,80 @@
...
+## TL;DR
+
+When using TypeScript in a file that imports from any untyped add a suppression like the below:
+
+```ts
+// @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
+import { cssModules } from 'bpk-react-utils';
+// @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
+import { helper } from 'some-external-untyped-package';
+```
```
angel3cu, line 12: Shall we add stubs in the type definitions like Banana instead?

decisions/imports-ts-suppressions.md
```diff
@@ -0,0 +1,80 @@
...
+4. Backpack uses inline types, rather than emitting a declaration file along with a transpiled output
+
+As Backpack is non-transpiled `backpack-react-scripts` (BRS) will transpile it at build time for consumers even though it is inside node modules. BRS also runs a type check at this time for any files it detects are using TypeScript, including in these non-transpiled node modules.
+
+With no action `npm run build` or `npm run start` fail on this type check due to `noImplicitAny` in consumers picking up paths inside of Backpack that have no type declarations.
+
+Two options exist here:
+
+1. Suppress inside of consumers
+2. Suppress inside of Backpack
```
angel3cu, line 38: +1

packages/bpk-component-graphic-promotion/README.md
```diff
@@ -35,17 +35,17 @@ export default () => (
...
-| contentId        | string            | false    | null          |
 | buttonText       | string            | true     | -             |
 | headline         | string            | true     | -             |
-| invertVertically | bool              | true     | -             |
 | onClick          | func              | true     | -             |
 | textAlign        | oneOf(TEXT_ALIGN) | true     | -             |
 | className        | string            | false    | null          |
-| style            | object            | false    | {}            |
-| tagline          | string            | false    | null          |
+| contentId        | string            | false    | null          |
```
angel3cu, line 43: Oh dear, we forgot to add the `contentId` to docs. Good catch. 🙂

packages/bpk-component-graphic-promotion/index.ts
angel3cu: 😂

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
```diff
@@ -15,35 +15,37 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* @flow strict */
 
 import React from 'react';
 import { fireEvent, render } from '@testing-library/react';
+// @ts-expect-error Untyped import. Not added as a stub declaration so consumers still have the suppression in their node_modules when shipped untranspiled.
```
angel3cu, line 21: Hi! I've read that above. Just to confirm, the stub declaration won't be shipped with the package, if the package is not transpiled?

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
angel3cu: For conciseness, you can use the following, since the exact signature of the mocked function is not needed:

```suggestion
    (props.onClick as jest.Mock).mockReset();
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -40,64 +41,66 @@ export const TEXT_ALIGN = {
...
-  buttonText: string,
-  onClick: () => void,
-  invertVertically: boolean,
-  textAlign: TEXT_ALIGN,
-  style: ?{},
+  className?: string | null;
+  contentId?: string | null;
+  tagline?: string | null;
+  headline: string;
+  subheading?: string | null;
```
angel3cu, line 51: I would say, let's remove the `| null`, to avoid confusion. One `null` type is bad enough, but two is a bit too much. 😅

angel3cu: Releasing in TypeScript will be a breaking change anyway, so I would opt for doing it right instead of Backwards compatibility. Also, we currently only use this in one place Banana within Skyscanner, so it won't need massive refactoring or anything.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -40,64 +41,66 @@ export const TEXT_ALIGN = {
...
+  sponsor?: {
+    label: string;
+    logo: string;
+    altText: string;
+  } | null;
```
angel3cu, line 56: Same as above. 🙂

dangerfile.js
dominicfraser: Good question.

ESLint TypeScript rules don't play well with JSX in `.ts` files, meaning we'd always use `.tsx` to ensure linting works without bugs - even though technically either work fine.

As you say, because of this snapshots, which are based on files with JSX in them, will be `.tsx`.

Ollie/Ana - are there any non-JSX snapshots used in Backpack? Do we need 

```
(filePath) => filePath.endsWith('.js.snap') || filePath.endsWith('.ts.snap') || filePath.endsWith('.tsx.snap'),
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -40,64 +41,66 @@ export const TEXT_ALIGN = {
...
+  buttonText: string;
+  onClick: () => void;
+  invertVertically?: boolean;
+  textAlign: typeof TEXT_ALIGN[keyof typeof TEXT_ALIGN];
```
angel3cu, line 60: 👍🏼

decisions/imports-ts-suppressions.md
dominicfraser: See the full doc - we cannot as these stubs are not picked up by consumers of Backpack, stopping `npm run start` / `build` working.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: Let's do this in a second functional change PR, keeping this one a like-for-like change. 

Currently it is nullable as we default it to `null`, hence the type reflecting it.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -40,64 +41,66 @@ export const TEXT_ALIGN = {
...
-  tagline,
-}: Props) => {
-  const text = [];
-  const addText = (value) => value && text.push(value);
+const constructAriaLabel = (
+  buttonText: Props['buttonText'],
+  headline: Props['headline'],
+  sponsor: Props['sponsor'],
+  subheading: Props['subheading'],
+  tagline: Props['tagline'],
```
angel3cu, line 69: Can we do:

```suggestion
  { buttonText, headline, sponsor, subheading, tagline }: Pick<Props, 'buttonText', 'headline', 'sponsor', 'subheading', 'tagline'>
```

Then. we could pass the whole `props` object rather than having to unbox it too. 🙂

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
angel3cu: Hmmm, `addText()` does a null check already. We need to sort out the type there instead.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -40,64 +41,66 @@ export const TEXT_ALIGN = {
...
+) => {
+  const text: string[] = [];
+  const addText = (value: string) => value && text.push(value);
```
angel3cu, line 72:
```suggestion
  const addText = (value?: string) => value && text.push(value);
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -40,64 +41,66 @@ export const TEXT_ALIGN = {
...
     addText(sponsor.altText);
-  } else {
+  } else if (tagline) {
     addText(tagline);
   }
+
   addText(headline);
-  addText(subheading);
+
+  if (subheading) {
```
angel3cu, line 83: Let's keep it in `addText()`, instead of putting the if everywhere.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
dominicfraser: We can ship them, it's just if TypeScript pays attention to them.

If there is an `index.d.ts` per `package.json` then TypeScript will pay attention, or if a `types` field is added to each `package.json`, but not otherwise. We can make it work this way, but it is a much more hidden implementation, for both maintainers and consumers. The proposed solution (temporarily suppressing inline) is a bit more ugly as its so visible, but the fact it is visible makes it clearer what is going on, and has the benefit that as soon as real types are added the suppression will error to remind us to clean up - not risking us stubbing modules that actually do have types without realising.

decisions/imports-ts-suppressions.md
dominicfraser: Continuing in https://github.com/Skyscanner/backpack/pull/2535#discussion_r952585158

dangerfile.js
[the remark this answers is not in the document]
olliecurtis: I think for this whilst this is the goal - if we are making some fixes or modifications before doing a full component TS migration it would still be a js file so would like to keep it from being strict to avoid people feeling like they need to make the whole component TS compatible for a small fix

dangerfile.js
olliecurtis: I think we would need to capture all three just in case off the top of my head to answer the question I am not sure, but for now lets just do this catch all and once we have migrated Backpack to TS they this will be re checked anyway

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
olliecurtis: >If there is an index.d.ts per package.json then TypeScript will pay attention, or if a types field is added to each package.json, but not otherwise.

This won't exist anymore when we stop publishing individual packages as all the package files per component will be gone!

dangerfile.js
dominicfraser: Agree 👍 

https://github.com/Skyscanner/backpack/pull/2535/commits/01057cdc2cdbb30415955ad96a9679a52857ddd0

UNRELEASED.md
```diff
@@ -1,6 +1,8 @@
 **Changed:**
 
-- bpk-component-scrollable-calendar:
+- `bpk-component-graphic-promotion` - converted to TypeScript
```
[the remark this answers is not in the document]
dominicfraser: https://github.com/Skyscanner/backpack/pull/2535/commits/876d515e5ff5781c6ba998280f383f64beb010f8

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
olliecurtis: >so I would opt for doing it right instead of Backwards compatibility

I would disagree with this and we should right now ALWAYS be backwards compatible as if folks don't have chance to make changes or upgrades we risk them falling behind on Backpack versions because its seen as a burden

olliecurtis (approved):

dangerfile.js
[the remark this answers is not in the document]
dominicfraser: https://github.com/Skyscanner/backpack/pull/2535/commits/d74777dedce3b7af687f9022afc072bf7470ecb7

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: Ditto 😛

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: Much nicer to just fix the type of addText since it already handles it, good spot.

https://github.com/Skyscanner/backpack/pull/2535/commits/378cfb846abfbbbabb2828cf46472edee6a310f1

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: https://github.com/Skyscanner/backpack/pull/2535/commits/378cfb846abfbbbabb2828cf46472edee6a310f1

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: I'd been going with the simplest Flow->TS switch for readability, but now we've gone through the implementation and made sure it is the same output we can tweak types to be more TS-like.

`Pick` is a good suggestion:

https://github.com/Skyscanner/backpack/pull/2535/commits/985717eff2579de0ade20506d4aad2eb065d40e6

Unboxing props makes for more readable-at-a-glance code inside of GitHub, so following the convention of destructuring in the function component args here and pass them individually, since we're only going one level deep.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -189,30 +198,4 @@ const BpkGraphicPromo = (props: Props) => {
   );
 };
 
-BpkGraphicPromo.propTypes = {
```
angel3cu, line 192: 👍🏼 🥾 😂

packages/bpk-component-graphic-promotion/src/__snapshots__/BpkGraphicPromo-test.tsx.snap
```diff
@@ -412,12 +372,14 @@ exports[`BpkGraphicPromo should support custom class names 1`] = `
   <div
     aria-label="Sponsored. Airline Name. Ride your wave. Portugal and 6 more countries have just been added to the UK travel green list. Learn more"
     class="bpk-card bpk-graphic-promo custom-classname"
+    id=""
```
angel3cu, line 375: Let's update the tests to use the content ID. I would also suggest we do not set the `id` when `contentId` is not there.

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -121,17 +124,23 @@ const BpkGraphicPromo = (props: Props) => {
     // clicks and key presses (Enter/Space) for the whole component, as described here:
     // https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets
     <div
-      id={contentId}
+      id={contentId || ''}
```
angel3cu, line 127:
```suggestion
      id={contentId}
```

Why are we setting this to an empty string when missing? 🙂

olliecurtis: I had asked the same here: https://github.com/Skyscanner/backpack/pull/2535#discussion_r949281964

Looks like a commit may have been reverted

alakbarz:
This has been a long time coming! Great to see that we're moving to TypeScript in Backpack.

Hopefully this makes the future creation of TS Backpack components easier too!

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
```diff
@@ -40,24 +41,24 @@ export const TEXT_ALIGN = {
   start: 'start',
   center: 'center',
   end: 'end',
-};
+} as const;
 
 export type Props = {
```
LexSwed, line 46: I think it was already discussed, so just confirming, the approach we take is exporting component props, instead of inferring them:
```tsx
import type { Props as BpkGraphicPromoProps } from 'bpk-component-graphic-promotion';
// vs
import { BpkGraphicPromo } from 'bpk-component-graphic-promotion';
type BpkGraphicPromoProps = React.ComponentProps<typeof BpkGraphicPromo>
```

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: TS knows contentId can be null, so requires a fallback.

Screenshot 2022-08-23 at 15 54 08

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo-test.tsx
dominicfraser: Knowing that sometimes we need it I like the more verbose form that allows others to easily copy it and quickly infer how to replace when they need to :)

packages/bpk-component-graphic-promotion/src/__snapshots__/BpkGraphicPromo-test.tsx.snap
dominicfraser: Continuing in https://github.com/Skyscanner/backpack/pull/2535#discussion_r952735241

olliecurtis (approved):

packages/bpk-component-graphic-promotion/src/BpkGraphicPromo.tsx
dominicfraser: This is a very good point.

This was kept as it is purely for similarity to the existing component API, with no deeper thinking.

The code snippet definitely works (just tested with the beta package to make sure in that assertion).

The downside is that every consumer needs to set up the type themselves, rather than it being provided. 

I can go either way on this - any preference @olliecurtis? (/Ana)

cc: @kirstybryce

packages/bpk-component-graphic-promotion/src/__snapshots__/BpkGraphicPromo-test.tsx.snap
dominicfraser: Add contentId to tests: https://github.com/Skyscanner/backpack/pull/2544

PR #2537, opened 2022-08-15
ARGG-211: Add script for transpilation

xalechez:

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   },
   "scripts": {
     "build": "gulp",
+    "build:transpiled": "./scripts/transpilation/transpiled-build.sh",
     "check-bpk-dependencies": "node scripts/npm/check-bpk-dependencies.js",
     "check-react-versions": "node scripts/npm/check-react-versions.js",
     "check-owners": "npm whoami && ( node scripts/npm/get-owners.js | grep -E $(npm whoami) ) && node scripts/npm/check-owners.js",
@@ -99,6 +100,7 @@
     "verbose": true
   },
   "devDependencies": {
+    "@babel/cli": "^7.18.10",
     "@babel/core": "^7.18.2",
     "@babel/plugin-proposal-class-properties": "^7.17.12",
     "@babel/plugin-proposal-object-rest-spread": "^7.18.0",
@@ -129,6 +131,10 @@
     "autoprefixer": "^9.4.7",
     "babel-jest": "^26.0.0",
     "babel-loader": "^8.0.0",
+    "babel-plugin-css-modules-transform": "^1.6.2",
+    "babel-plugin-module-resolver": "^4.1.0",
+    "babel-plugin-replace-imports": "^1.0.2",
+    "babel-plugin-inline-react-svg": "^2.0.1",
     "babel-plugin-require-context-hook": "^1.0.0",
     "bpk-mixins": "^34.0.5",
     "bpk-stylesheets": "^8.1.5",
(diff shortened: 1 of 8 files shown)


scripts/transpilation/transpiled-build.sh
```diff
@@ -0,0 +1,66 @@
...
+  then
+      cp -R "../../node_modules/d3-array" "./d3-array"
+      cp -R "../../node_modules/d3-color" "./d3-color"
+      cp -R "../../node_modules/d3-format" "./d3-format"
+      cp -R "../../node_modules/d3-interpolate" "./d3-interpolate"
+      cp -R "../../node_modules/d3-scale" "./d3-scale"
+      cp -R "../../node_modules/d3-time" "./d3-time"
+      cp -R "../../node_modules/d3-time-format" "./d3-time-format"
+      cp -R "../../node_modules/internmap" "./internmap"
+  fi
```
runmoore, line 38: what's the reason. behind doing something custom here for d3? could you add a comment explaining why?

scripts/transpilation/transpiled-build.sh
```diff
@@ -0,0 +1,66 @@
...
+
+  if [ "$d" == "bpk-component-flare/" ] || [ "$d" == "bpk-component-icon/" ]
+  then
+    npm run build
+  fi
```
runmoore, line 43: could you add an explanation explaining why these components need a special usecase

package.json
```diff
@@ -131,6 +133,10 @@
     "autoprefixer": "^9.4.7",
     "babel-jest": "^26.0.0",
     "babel-loader": "^8.0.0",
+    "babel-plugin-css-modules-transform": "^1.6.2",
+    "babel-plugin-module-resolver": "^4.1.0",
+    "babel-plugin-replace-imports": "^1.0.2",
```
runmoore, line 138: how are we using `babel-plugin-replace-imports`?

packages/package.json
```diff
@@ -32,6 +32,7 @@
     "d3-path": "^2.0.0",
     "d3-scale": "^4.0.2",
     "date-fns": "^2.21.1",
+    "internmap": "^2.0.3",
```
runmoore, line 35: what's the reason for declaring this now? It looks like we're moving it around in our build step, not actually using it. Is it a nested dependency

scripts/transpilation/transpiled-build.sh
xalechez: They have their own build scripts and guplfile. This allows to execute them before transpilation. It's mostly about copying some icons from node_modules. To my mind it's redundant for transpiled build as we have svg import plugin, but I left it here for now in order not to break anything

scripts/transpilation/transpiled-build.sh
xalechez: It allows us to transpile those deps. I faced issues with D3 minified build imports, so this script transpiles d3 modules and includes them into bundle. Not sure if it's the best way to solve this issue, open to discussion

scripts/transpilation/transpiled-build.sh
olliecurtis: We probably don't need to have it as part of this script as part of general running, setup and releasing we have this separately or you can even in the package.json file before running the script have it run our usual `npm run build`

I'd be wary about doing it here separately incase we change things, this will introduce a new level of change

olliecurtis: >To my mind it's redundant for transpiled build as we have svg import plugin, but I left it here for now in order not to break anything

We still need it even with a new plugin as we do some transformation of them before we use them to do some specific web thing for the icons which the plugin won't cover

scripts/transpilation/transpiled-build.sh
runmoore: ok, happy for now how we have it, we just need a comment in the code explaining this :)

PR #2538, opened 2022-08-18
[NO JIRA]: Extract TS updates to separate PR

olliecurtis:
This PR extracts the typescript specific changes from #2527 to a separate PR so that it can be merged ahead of the PR and would also unblock #2535 from waiting on the first PR

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,6 @@
 {
-  "parser": "babel-eslint",
   "extends": [
-    "@skyscanner/eslint-config-skyscanner",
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -1,7 +1,6 @@
 {
-  "parser": "babel-eslint",
```
olliecurtis, line 2: We don't need this anymore and projects in Skyscanner don't use this - alongside this `babel-eslint` is deprecated as well so removing it makes sense

PR #2544, opened 2022-08-23
KOA-5198: Tweak Graphic Promo tests

dominicfraser:
In https://github.com/Skyscanner/backpack/pull/2535 it was requested that we set the contentId in tests by default.

This PR does so, and ensures we still have a test for when the contentId is not present (it is optional, with a default of null currently).

- [x] `UNRELEASED.md` (See [How to write a good changelog entry](https://github.com/Skyscanner/backpack/blob/main/CHANGELOG_FORMAT.md)) - **this bundles into an existing merged but not released change**
- [x] `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-graphic-promotion/src/__snapshots__/BpkGraphicPromo-test.tsx.snap
+++ b/packages/bpk-component-graphic-promotion/src/__snapshots__/BpkGraphicPromo-test.tsx.snap
@@ -5,14 +5,14 @@ exports[`BpkGraphicPromo should display the tagline and no sponsor for non-spons
   <div
     aria-label="Tagline. Ride your wave. Portugal and 6 more countries have just been added to the UK travel green list. Learn more"
     class="bpk-card bpk-graphic-promo"
-    id=""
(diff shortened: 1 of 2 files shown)


packages/bpk-component-graphic-promotion/src/__snapshots__/BpkGraphicPromo-test.tsx.snap
```diff
@@ -108,11 +108,11 @@ exports[`BpkGraphicPromo should render correctly when centre aligned 1`] = `
   >
     <div
       aria-hidden="true"
-      class="bpk-graphic-promo__container bpk-graphic-promo__container--center"
+      class="bpk-graphic-promo__container bpk-graphic-promo__container--start"
```
dominicfraser, line 111: This looks like its changed, but actually its because this snapshot is now output in a different order (reasons unknown) under a new title (line 100) making the diff hard to track.

olliecurtis (approved):

PR #2548, opened 2022-08-27
[KOA-5173]: Convert bpk-component-aria-live to TypeScript

dominicfraser:
Converts bpk-component-aria-live to TypeScript. No functional changes are expected.

- [x] `README.md`
- [x] Tests


decisions/inexact-rest.md
```diff
@@ -0,0 +1,56 @@
...
+If using `...rest` parameter with the intention to allow any arbitrary prop to be passed then type using:
+
+```tsx
+type Props = {
+  namedPropOne: string;
+  namedPropTwo: boolean;
+  [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
+};
+```
+
```
dominicfraser, line 14: We've got a lot of 
```
// $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
```

in the codebase right now.

While adding the type as above means no suppression is necessary for TypeScript I do think it is worth keeping the inline comment, for the reasons given in the doc.

WDYT?

packages/bpk-component-aria-live/README.md
```diff
@@ -37,6 +37,6 @@ export default () => (
 
 `aria-relevant` and `aria-atomic` props can also be set.
 
-`aria-relevant` determines what sort of changes should be read out. By default it is `additions` but can be `removals` or `all`. [Read more about `aria-relevant` on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-relevant_attribute).
+`aria-relevant` determines what sort of changes should be read out. By default it is `text` but can be `additions`, `removals` or `all`. [Read more about `aria-relevant` on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-relevant).
```
dominicfraser, line 40: The MDN doc states that `text` is the default.

packages/bpk-component-aria-live/src/BpkAriaLive.tsx
```diff
@@ -29,38 +29,32 @@ export const POLITENESS_SETTINGS = {
...
-  );
-};
-
-BpkAriaLive.propTypes = {
-  className: PropTypes.string,
-  politenessSetting: PropTypes.oneOf(Object.keys(POLITENESS_SETTINGS)),
-  visible: PropTypes.bool,
-};
-
-BpkAriaLive.defaultProps = {
```
dominicfraser, line 60: Stop using soon-to-be-deprecated .defaultProps and start using the recommended default arguments.

https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components

decisions/inexact-rest.md
```diff
@@ -0,0 +1,56 @@
...
+## TL;DR
+
+If using `...rest` parameter with the intention to allow any arbitrary prop to be passed then type using:
+
+```tsx
+type Props = {
+  namedPropOne: string;
+  namedPropTwo: boolean;
+  [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
```
LexSwed, line 11: I would always default to stricter recommendation first, like mentioned below: `React.HTMLAttributes<HTMLDivElement>;`. I don't know all the cases, but IMO they are never arbitrary and spread to a specific element that has specific list of supported attributes.

decisions/inexact-rest.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+if then needing to override some of the built in type definitions do so by using `Omit`. In the example below React knows className can only be `string | undefined`, but we assert that it can also be `null` as in our use case we don't apply it directly to the `div`:
+
+```tsx
+type Props = {
+  namedPropOne: string;
+  namedPropTwo: boolean;
+  className?: string | null;
+} & Omit<React.HTMLAttributes<HTMLDivElement>, 'className'>;
```
LexSwed, line 49: Might be against some other best practices, but  I found that for me remembering `React.ComponentProps` is easy as it works for both, JSX tags and components: `React.ComponentProps<'div'>` and `React.ComponentProps<typeof BpkButton>`

PR #2556, opened 2022-08-30
[KOA-5470] Upgrade to Node 16

anambl:
- Upgrade Backpack to Node 16
  - Add `react-dom` as a peer dependency to each component (this is due to the fact that `npm` installs the latest version of a peer dependency instead of installing the version that aligns with the root dependency/other packages and therefore causes a conflict)
  - Remove `puppeteer` dependency
  - Upgrade `react-responsive` to 8.2.0 (to match the version used in the breakpoint component)
-  Add missing `react-table` dependency used by datatable component in single package

- [ ] `README.md`
- [ ] Tests


anambl:
> > Remove puppeteer dependency
> 
> Won't we need this as part of screenshot testing? This was why originally it was added for compatibility with Storybook Storyshots plugin that we use

I think we don't! We don't use it directly.
`puppeteer` is a peer dependency of `@storybook/addon-storyshots-puppeteer` and with npm versions > 7 it will be installed automatically so we don't need to install it! I removed it as the version we were using was different from the one required by `@storybook/addon-storyshots-puppeteer` 🤔

packages/package.json
```diff
@@ -39,13 +39,14 @@
...
     "prop-types": "^15.7.2",
     "react-autosuggest": "^9.4.3",
-    "react-responsive": "^6.1.2",
+    "react-responsive": "^8.2.0",
     "react-slider": "^1.3.1",
     "react-transition-group": "^2.5.3",
-    "react-virtualized": "^9.22.3",
-    "react-window": "^1.8.7"
+    "react-window": "^1.8.7",
+    "react-table": "^7.8.0"
```
olliecurtis, line 46: minor nitpick: Can you move this up above `react-transition-group` to keep it alphabetical?

olliecurtis (changes requested):

olliecurtis (approved):

PR #2558, opened 2022-08-31
[NO JIRA]: Revert recent TS changes

olliecurtis:

- [ ] `README.md`
- [x] Tests


UNRELEASED.md
```diff
@@ -0,0 +1,6 @@
+**Fixed:**
```
anambl, line 1: will it not be a breaking change for the teams which already started using our TS components? 🤔

packages/bpk-component-aria-live/src/BpkAriaLive.js
```diff
@@ -29,17 +29,17 @@ export const POLITENESS_SETTINGS = {
   off: 'off',
   polite: 'polite',
   assertive: 'assertive',
-} as const;
```
anambl, line 32: might need to comment this too since we commented out the rest of the TS code?

UNRELEASED.md
olliecurtis: This is a fix as putting TS in was breaking for consumers not using TS so this unblocks people

packages/bpk-component-aria-live/src/BpkAriaLive.js
olliecurtis: Good point!

packages/bpk-component-aria-live/src/BpkAriaLive.js
```diff
@@ -29,17 +29,17 @@ export const POLITENESS_SETTINGS = {
...
+//   typeof POLITENESS_SETTINGS[keyof typeof POLITENESS_SETTINGS];
 
 export type Props = {
-  children: ReactElement | string;
-  politenessSetting?: PolitenessSetting;
-  visible?: boolean;
-  className?: string | null;
-  [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
+  children: ReactElement | string,
+  politenessSetting: $Keys<typeof POLITENESS_SETTINGS>, // PolitenessSetting,
```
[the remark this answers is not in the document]
olliecurtis: This is just reverting this back the previous setting here: https://github.com/Skyscanner/backpack/pull/2548/files#diff-5d3f1232565254dfd1c5fa845669dfc7f88355532fedb2165d6c42ab2d52ea31L36

anambl (approved):
LGTM!

PR #2564, opened 2022-09-02
[KOA-5063]: Add new smaller switch, semantic tokens and type deprecation

olliecurtis:
New small variant added to Backpack 

Screenshot 2022-09-02 at 15 11 10
Screenshot 2022-09-02 at 15 11 15

Also in this PR:
- Migrated to semantic tokens
- Deprecated Event type switch as this is no longer a supported design pattern

- [x] `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-switch/src/BpkSwitch.module.scss
+++ b/packages/bpk-component-switch/src/BpkSwitch.module.scss
@@ -20,18 +20,14 @@
 
 $bpk-spacing-v2: true;
 
-$primary-on-color: $bpk-color-sky-blue;
-$event-on-color: $bpk-color-abisko;
-$off-color: $bpk-color-sky-gray-tint-06;
-$handle-color: $bpk-color-white;
-
 $height: bpk-spacing-xl();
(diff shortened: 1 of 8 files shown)


packages/bpk-component-switch/src/BpkSwitch.module.scss
```diff
@@ -43,12 +39,13 @@ $border-width: $bpk-one-pixel-rem * 2;
...
           --bpk-switch-checked-color,
-          $primary-on-color
+          $bpk-core-accent-day
         );
       }
 
+      // This is now deprecated and will be removed in the future.
       &--event {
-        background: $event-on-color;
+        background: $bpk-color-abisko;
```
anambl, line 48: should we still use this given it's deprecated?

olliecurtis: As there is no new colour mapping and event type deprecated we have to keep using this. Soon to be removed though once the one usage has been updated

anambl (approved):
LGTM!

PR #2573, opened 2022-09-08
[NO-JIRA] Fix broken list key prop in BpkContentCards

kirstybryce:
When using the `bpk-component-content-cards` component in [Banana](https://github.skyscannertools.net/dingo/banana/pull/5834/#issuecomment-844446) I noticed that there was an error in the console that `Each child in a list should have a unique "key" prop` and I realised we didn't set up the key prop correctly in the BpkContentCards component so this PR fixes that.

- [x] `README.md`
- [x] Tests


olliecurtis (approved):
LGTM

PR #2583, opened 2022-09-12
[KOA-5548] Update autosuggest to semantic colour tokens

anambl:
|Before|After|
|-|-|
|image|image|

- [ ] `README.md`
- [ ] Tests


olliecurtis (changes requested):
It looks like the packages/package.json file was updated but not the lockfile could you take a look at that?

anambl:
> It looks like the packages/package.json file was updated but not the lockfile could you take a look at that?

You're right, good catch! hadn't noticed that. Updated it now 😄

olliecurtis (approved):

PR #2598, opened 2022-09-28
[KOA-5591] Update tables to semantic colours

anambl:

- [x] `README.md`
- [ ] Tests


packages/bpk-component-datatable/src/BpkDataTable.module.scss
```diff
@@ -71,7 +71,7 @@ $bpk-border-size-datatable: $bpk-border-size-xl * 2;
     }
 
     &--selected {
-      background-color: $bpk-color-sky-gray-tint-06;
+      background-color: $bpk-canvas-contrast-day;
```
olliecurtis, line 74: Is selected to be contrast day? I see in the designs hover is contrast but selected is warning fill?

Screenshot 2022-09-28 at 12 15 45

olliecurtis (changes requested):

packages/bpk-component-datatable/src/BpkDataTable.module.scss
anambl: ohh yes, you are right!!! good catch! will update it now

olliecurtis (approved):

PR #2603, opened 2022-09-29
[KOA-5593]: Update Scrim colours

olliecurtis:

- [x] `README.md`
- [x] Tests


packages/bpk-scrim-utils/src/withScrim.js
```diff
@@ -46,7 +47,10 @@ const withScrim = (WrappedComponent) => {
     static propTypes = {
       getApplicationElement: PropTypes.func.isRequired,
       onClose: onClosePropType,
-      dark: PropTypes.bool,
+      dark: deprecated(
```
anambl, line 50: do we need to keep it as `deprecated` if we removed the logic behind it anyways so using it will not have any effect? 🤔

olliecurtis: True that and how it works the prop isn't being used so will update the PR

anambl (approved):
LGTM!

PR #2604, opened 2022-09-30
[BD-6195] contribute component - bpk page indicator

giriawu:
Figma link: https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=1763%3A18266

image

- [x] `README.md`
- [x] Tests
- [x] Docs (followed TK has been created)


UNRELEASED.md
```diff
@@ -8,3 +8,8 @@
 
 - bpk-component-accordion:
     - Remove redundant `aria-labeledby` tags in accordion item
+
+**Added**
+
+- bpk-component-page-indicator
+  - The page indicator is a new component that will ensure the consistency of the image gallery display indicator.
```
jasonhuang-sky, line 15:
```suggestion
  - The page indicator is a new component that will be used to display multiple swappable images.
```

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+```sh
+npm install bpk-component-page-indicator --save-dev
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkPageIndicator from 'bpk-component-code';
```
jasonhuang-sky, line 15:
```suggestion
import BpkPageIndicator from 'bpk-component-page-indicator';
```

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,38 @@
...
+  <BpkPageIndicator
+    currentItem={0}
+    totalItems={7}
+    updateItemCallback={(newIndex) => {
+      console.log('Slide to new index:', newIndex);
+    }}
+  />;
+```
+
+## Props
```
jasonhuang-sky, line 27: You missed the key required props and functions, otherwise how can we used it for displaying multiple swappable images?

jasonhuang-sky (changes requested):

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,190 @@
...
+);
+const DISPLAYED_TOTAL = 5;
+const START_SCROLL_INDEX = Math.floor(DISPLAYED_TOTAL / 2);
+
+export type Props = {
+  className: ?string,
+  showNav?: boolean,
+  currentItem: number,
+  totalItems: number,
+  updateItemCallback?: Function,
```
Supremeyh, line 44: should be like `updateItemCallback: ?() => void,`

packages/bpk-component-page-indicator/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "bpk-component-page-indicator",
+  "version": "2.1.1",
```
olliecurtis, line 3: Can you set this to 0.0.1 just to ease up when we release?

packages/bpk-component-page-indicator/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "description": "Backpack page-indicator component.",
+  "license": "Apache-2.0",
+  "private": true,
```
olliecurtis, line 6: Can you also remove this line?

(These arent your fault but more out automation not updating when creating new components)

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,190 @@
...
+      newItem = totalItems - 1;
+    }
+    if (index >= totalItems) {
+      newItem = 0;
+    }
+    updateItemCallback(newItem);
+  };
+
+  const handleKeyboardEvent = (e, index) => {
+    if (e.key === KEYS.ENTER || e.key === KEYS.SPACE || e.key === ' ') {
```
olliecurtis, line 81: What does `e.key === '  '` represent here?

olliecurtis (changes requested):

packages/bpk-component-page-indicator/README.md
giriawu: Updated, thank you!

examples/bpk-component-page-indicator/examples.js
```diff
@@ -0,0 +1,67 @@
...
+  );
+};
+
+const DefaultExample = () => (
+  <PageIndicatorContainer id="default-example" totalItems={7} />
+);
+
+const ThreePagesExample = () => (
+  <PageIndicatorContainer id="no-scroll-example" totalItems={3} />
+);
```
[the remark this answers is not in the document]
giriawu: Updated, thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Updated, thank you!!

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+| Property            | PropType | Required | Default Value |
+|---------------------|----------|----------|---------------|
+| currentItem         | number   | true     | -             |
+| totalItems          | number   | true     | -             |
```
jasonhuang-sky, line 32: We can use `currentIndex` and `totalBullets` instead

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,38 @@
...
+| className           | string   | false    | null          |
+| itemClassName       | string   | false    | null          |
+| activeItemClassName | string   | false    | null          |
+| ariaLabel           | string   | false    | 'Go to Slide' |
+| showNav             | bool     | false    | false         |
+| updateItemCallback  | func     | false    | null          |
```
jasonhuang-sky, line 38: Would be better to update to `onClick: (event, index, direction)` instead, so that we can use it to response the three actions, clicking on the bullet or clicking on the left and right buttons

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,190 @@
...
+      {showNav && (
+        <BpkButton
+          iconOnly
+          link
+          onClick={() => {
+            updateItem(currentItem - 1);
+          }}
+        >
+          <AlignedLeftArrowIcon />
+          <span className="visually-hidden">Prev</span>
```
jasonhuang-sky, line 113: Seems we can just set the `aria-label` on the button instead of this span here, eg: `aria-label="Previous Slide"`

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,190 @@
...
+      {showNav && (
+        <BpkButton
+          iconOnly
+          link
+          onClick={() => {
+            updateItem(currentItem + 1);
+          }}
+        >
+          <AlignedRightArrowIcon />
+          <span className="visually-hidden">Next</span>
```
jasonhuang-sky, line 163: same here

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
jasonhuang-sky: And seems that it's easy to put the prev and next button together, we can just use css to make them `space-between` for bullets

jasonhuang-sky: We can also consider wrapping a component `NavButton` for the prev and next buttons

jasonhuang-sky: The button component needs a `disabled` property to ensure it cannot be left clickable at `currentIndex=0`, the same for right click event

packages/bpk-component-page-indicator/README.md
jasonhuang-sky: And I don't think we need `itemClassName` and `activeItemClassName`, because we don't expect the bullet's style to change, that's one reason why it's wrapped as a backpck component, giving users a consistent product experience, which of course includes its look visual

packages/bpk-component-page-indicator/index.js
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* @flow strict */
+
+import BpkPageIndicator, {
+  type Props as BpkPageIndicatorProps,
+} from './src/BpkPageIndicator';
+
+export type { BpkPageIndicatorProps };
```
jasonhuang-sky, line 24: Why export it?

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,190 @@
...
+                      ? Math.min(
+                          currentItem - START_SCROLL_INDEX,
+                          totalItems - DISPLAYED_TOTAL,
+                        )
+                      : 0,
+                }
+              : undefined
+          }
+        >
+          {bulletArray.map((index) => (
```
jasonhuang-sky, line 133:
```suggestion
          {[...Array(totalbullets)].map((val, index) => (
```

packages/bpk-component-page-indicator/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org/"
+  },
+  "dependencies": {
+    "bpk-component-button": "^7.1.2",
+    "bpk-component-icon": "^10.1.2",
+    "bpk-mixins": "^32.0.2",
+    "bpk-react-utils": "^7.1.0",
+    "prop-types": "^15.5.8",
+    "react-swipeable": "^7.0.0"
```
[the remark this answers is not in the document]
giriawu: As this component is no longer required for swipe function, we will be removing this library. Thank you for your comment!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: This is for the onclick function can also be triggered when the user press on a space.
I don't know why just adding key.spacebar didn't seem to work, so I took a page from [here](https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-split-input/src/BpkSplitInput.js#L165) and added `e.key === ' '` to determine

UNRELEASED.md
giriawu: Have updated, thank you!

packages/bpk-component-page-indicator/README.md
giriawu: Have removed the swipeable library, thank you!

packages/bpk-component-page-indicator/package.json
giriawu: Have remove this line, thank you!

packages/bpk-component-page-indicator/README.md
giriawu: Have changed name from currentItem -> currentIndex, and totalItems -> totalBullets, thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Have removed empty span and replace it with new aria-label, thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Have updated, thank you!

packages/bpk-component-page-indicator/index.js
giriawu: I found that props are exported in many other components even though they don't use them in the backpack project.🤔 So I exported them as well.

packages/bpk-component-page-indicator/README.md
giriawu: Have replaced updateItemCallback with onClick and add direction(on prev button -> -1, on next button -> 1, on bullets -> 0)

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: I try to wrap them into the same NavButton, but could not find a good way to overlap this component and page indicator. This is what it looks like after I have combine the `NavButton` and set the position of the page indicator to `absolute` and added a partial margin. It looks like the alignment is not as standard as with flex+center.

packages/bpk-component-page-indicator/src/accessibility-test.js
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react';
+import { render } from '@testing-library/react';
+import { axe } from 'jest-axe';
+
+import BpkPageIndicator from './BpkPageIndicator';
+
+describe('BpkPageIndicator accessibility tests', () => {
+  it('should not have programmatically-detectable accessibility issues', async () => {
+    const { container } = render(
+      <BpkPageIndicator currentItem={0} totalItems={7} showNav />,
```
jasonhuang-sky, line 29: This test case is passed, but you have passed wrong props, interesting, can you make sure that it's executed?

.spelling
```diff
@@ -349,7 +349,6 @@ inputClassName
 inputRef
 isIphone
 isOpen
-itemClassName
```
jasonhuang-sky, line 352: Nice catch, seems it doesn't use to anywhere

.spelling
```diff
@@ -900,3 +900,7 @@ INPUT_TYPES.number
 inputLength
 onSubmit
 otp
+currentIndex
+totalBullets
+showNav
+updateItemCallback
```
jasonhuang-sky, line 906: It should be removed

UNRELEASED.md
```diff
@@ -0,0 +1,4 @@
+**Added**
+
+- bpk-component-page-indicator
+  - The page indicator is a new component that will be used to display multiple swappable images.
```
jasonhuang-sky, line 4: Sorry, you have to update it again since this component doesn't have swappable functionality

packages/bpk-component-page-indicator/README.md
jasonhuang-sky: Then you need to update some usages as well

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,163 @@
...
+          }}
+          aria-label="Previous Slide"
+          disabled={currentIndex === 0}
+        >
+          <AlignedLeftArrowIcon />
+        </BpkButton>
+      )}
+      <div className={getClassName('bpk-page-indicator__container')}>
+        <div
+          className={getClassName('bpk-page-indicator__bulletsContainer')}
```
jasonhuang-sky, line 96:
```suggestion
          className={getClassName('bpk-page-indicator__bullets-container')}
```

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,163 @@
...
+                          totalBullets - DISPLAYED_TOTAL,
+                        )
+                      : 0,
+                }
+              : undefined
+          }
+        >
+          {[...Array(totalBullets)].map((val, index) => (
+            <div
+              role="button"
```
jasonhuang-sky, line 113:
```suggestion
              role="button"
              key={`bullet-${index}`}

```

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,163 @@
...
+                onClick(e, index, DIRECTIONS.BULLETS);
+              }}
+              onKeyDown={(e) => {
+                handleKeyboardEvent(e, index);
+              }}
+              className={getClassName(
+                'bpk-page-indicator__bullet',
+                index === currentIndex && 'bpk-page-indicator__bullet--active',
+              )}
+              aria-label={`${ariaLabel} ${index + 1}`}
```
jasonhuang-sky, line 125:
```suggestion
              aria-label={`${ariaLabel} ${index + 1}`}
              aria-pressed={currentIndex === index ? 'true' : 'false'}
```

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,163 @@
...
+                          currentIndex - START_SCROLL_INDEX,
+                          totalBullets - DISPLAYED_TOTAL,
+                        )
+                      : 0,
+                }
+              : undefined
+          }
+        >
+          {[...Array(totalBullets)].map((val, index) => (
+            <div
```
jasonhuang-sky, line 112: Maybe we can consider just using `button` directly here?

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Have extract a `NavButton` component to wrap the nav button. Thank you for your comment!

packages/bpk-component-page-indicator/src/accessibility-test.js
giriawu: Have updated, thank you!

.spelling
giriawu: Updated, thank you!!

UNRELEASED.md
giriawu: Have updated a new instrument, hope to be reviewed again! Thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Updated, thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Updated, thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Added this, thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Updated, thank you!!

jasonhuang-sky (approved):

olliecurtis (changes requested):

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+## Props
+
+| Property     | PropType | Required | Default Value |
+|--------------|----------|----------|---------------|
+| currentIndex | number   | true     | -             |
+| totalBullets | number   | true     | -             |
+| className    | string   | false    | null          |
+| ariaLabel    | string   | false    | 'Go to Slide' |
```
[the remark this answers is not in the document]
giriawu: I've looked at the design of some of the other components of backpack and I think we could really design `ariaLabel` as a required property.

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,157 @@
...
+  const handleKeyboardEvent = (e, index) => {
+    if (e.key === KEYS.ENTER || e.key === KEYS.SPACE || e.key === ' ') {
+      e.preventDefault();
+      onClick(e, index, DIRECTIONS.BULLETS);
+    }
+  };
+
+  return (
+    // $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
+    <div className={className} {...rest}>
```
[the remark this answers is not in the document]
giriawu: Have removed `...rest`, thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.module.scss
```diff
@@ -0,0 +1,86 @@
...
+    width: $dot-size;
+    height: $dot-size;
+    padding: 0;
+    border: none;
+    border-radius: 50%;
+    background-color: $bpk-line-day;
+    margin-inline: $dot-margin;
+
+    &--active {
+      background-color: $bpk-marcomms-grey-40;
```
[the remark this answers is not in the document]
giriawu: Have changed it into `$bpk-text-secondary-day` thank you!

packages/bpk-component-page-indicator/src/BpkPageIndicator.module.scss
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  &__dark {
+    background-color: $bpk-private-button-link-on-dark-pressed-foreground-night;
+
+    &--active {
+      background-color: $bpk-color-white;
```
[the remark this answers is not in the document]
giriawu: Have changed it into `$bpk-text-on-dark-day`, thank you!

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,39 @@
+# bpk-component-page-indicator
+
+> Backpack example component.
+
+## Installation
+
+```sh
+npm install bpk-component-page-indicator --save-dev
```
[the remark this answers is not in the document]
giriawu: Have updated, thank you!

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+## Usage
+
+```js
+import React from 'react';
+import BpkPageIndicator from 'bpk-component-page-indicator';
```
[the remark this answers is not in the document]
giriawu: Have updated, thank you!

packages/bpk-component-page-indicator/README.md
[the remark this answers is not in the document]
giriawu: Have updated, thank you!

packages/bpk-component-page-indicator/src/NavButton.js
```diff
@@ -0,0 +1,80 @@
...
+      iconOnly
+      link
+      onClick={(e) => {
+        if (prev) {
+          onClick(e, currentIndex - 1, DIRECTIONS.PREV);
+        } else {
+          onClick(e, currentIndex + 1, DIRECTIONS.NEXT);
+        }
+      }}
+      aria-label={prev ? 'Previous Slide' : 'Next Slide'}
```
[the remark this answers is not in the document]
giriawu: Have export `bulletLabel`, `prevNavLabel`, `nextNavLabel` to add aria-label for bullets and navButtons.  Thank you for your comment!

packages/bpk-component-page-indicator/src/BpkPageIndicator.module.scss
```diff
@@ -0,0 +1,86 @@
...
+    background-color: $bpk-line-day;
+    margin-inline: $dot-margin;
+
+    &--active {
+      background-color: $bpk-marcomms-grey-40;
+    }
+  }
+
+  &__dark {
+    background-color: $bpk-private-button-link-on-dark-pressed-foreground-night;
```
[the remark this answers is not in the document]
giriawu: Have changed it into `$bpk-line-on-dark-day`, thank you!!

packages/bpk-component-page-indicator/src/NavButton.js
```diff
@@ -0,0 +1,81 @@
...
+import BpkButton from '../../bpk-component-button';
+import LeftArrowIcon from '../../bpk-component-icon/lg/chevron-left';
+import RightArrowIcon from '../../bpk-component-icon/lg/chevron-right';
+
+type Props = {
+  prev?: boolean,
+  disabled?: boolean,
+  currentIndex: number,
+  onClick: ?() => void,
+  navButtonLabel: string,
```
jasonhuang-sky, line 33: Would be better to just use `ariaLabel` instead

packages/bpk-component-page-indicator/src/NavButton.js
```diff
@@ -0,0 +1,81 @@
...
+      {prev ? <AlignedLeftArrowIcon /> : <AlignedRightArrowIcon />}
+    </BpkButton>
+  );
+};
+
+NavButton.propTypes = {
+  prev: PropTypes.bool,
+  disabled: PropTypes.bool,
+  currentIndex: PropTypes.number.isRequired,
+  onClick: PropTypes.func,
```
jasonhuang-sky, line 72: `ariaLabel` also needs to be added here

packages/bpk-component-page-indicator/src/BpkPageIndicator-test.js
```diff
@@ -0,0 +1,59 @@
...
+
+  it('should support showNav attribute', () => {
+    const { asFragment } = render(<BpkPageIndicator {...props} showNav />);
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should support arbitrary props', () => {
+    const { asFragment } = render(<BpkPageIndicator {...props} testid="123" />);
+    expect(asFragment()).toMatchSnapshot();
+  });
```
jasonhuang-sky, line 58: Seems it doesn't support arbitrary props anymore

examples/bpk-component-page-indicator/examples.js
```diff
@@ -0,0 +1,110 @@
...
+    style={{ width: imageWidth, height: imageHeight }}
+    imageStyle={{
+      backgroundRepeat: 'no-repeat',
+      backgroundSize: 'cover',
+      backgroundPosition: '50% 50%',
+    }}
+    src={image}
+  >
+    <PageIndicatorContainer
+      id="over-image-example"
```
jasonhuang-sky, line 71: `id` also needs to be removed

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,162 @@
...
+            }
+          >
+            {[...Array(totalBullets)].map((val, index) => (
+              <button
+                type="button"
+                tabIndex={0}
+                onClick={(e) => {
+                  onClick(e, index, DIRECTIONS.BULLETS);
+                }}
+                onKeyDown={(e) => {
```
jasonhuang-sky, line 110: Since you have changed this tag from `div` to `button` so no need `tabIndex` and `onKeyDown` anymore

packages/bpk-component-page-indicator/src/NavButton.js
```diff
@@ -0,0 +1,81 @@
...
+  currentIndex: number,
+  onClick: ?() => void,
+  navButtonLabel: string,
+};
+
+export const DIRECTIONS = {
+  PREV: -1,
+  BULLETS: 0,
+  NEXT: 1,
+};
```
jasonhuang-sky, line 40:
```suggestion
export const DIRECTIONS = {
  PREV: 'PREV',
  BULLETS: 'BULLETS',
  NEXT: 'NEXT',
};
```

packages/bpk-component-page-indicator/src/NavButton.js
giriawu: Have changed into `ariaLabel`, thank you for your patient review!

packages/bpk-component-page-indicator/src/NavButton.js
giriawu: Have added as required, thank you for your review!

packages/bpk-component-page-indicator/src/BpkPageIndicator-test.js
giriawu: Right, have removed this test, thank you!

examples/bpk-component-page-indicator/examples.js
giriawu: Have removed `id`, thank you for your patient review!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Right, have removed these props, thank you for your patient review!

packages/bpk-component-page-indicator/src/NavButton.js
giriawu: Have updated, thank you for your patient review!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,157 @@
...
+const START_SCROLL_INDEX = Math.floor(DISPLAYED_TOTAL / 2);
+
+export type Props = {
+  className: ?string,
+  showNav: ?boolean,
+  currentIndex: number,
+  totalBullets: number,
+  onClick: ?() => void,
+  ariaLabel: ?string,
+  dark: ?boolean,
```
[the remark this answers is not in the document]
olliecurtis: Would we be able to change it to OverImage so it matches the figma naming? (The naming on the left in the above screenshot)

As we want to reserve the use of `dark` so that it doesn't give the impression of dark mode which at this time we don't support but may in the future

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,143 @@
...
+          disabled={currentIndex === totalBullets - 1}
+          ariaLabel={nextNavLabel}
+        />
+      )}
+    </div>
+  </div>
+);
+
+BpkPageIndicator.propTypes = {
+  bulletLabel: PropTypes.string.isRequired,
```
olliecurtis, line 125: Would you be able to use indicatorLabel instead of bulletLabel just so it clears any potential confusion on what that is and fits with the component?

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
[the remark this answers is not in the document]
giriawu: Have updated this props from `dark` into `overImage`, and changed from `bullet` into `indicator`
`bulletLabel` -> `indicatorLabel`
`totalBullets` -> `totalIndicators`
`__bullet` -> `__indicator`

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Have updated and changed from `bullet` into `indicator`
`bulletLabel` -> `indicatorLabel`
`totalBullets` -> `totalIndicators`
`__bullet` -> `__indicator`
Thank you very much!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,144 @@
...
+  className: ?string,
+  showNav: ?boolean,
+  currentIndex: number,
+  totalIndicators: number,
+  onClick: ?() => void,
+  indicatorLabel: string,
+  prevNavLabel: string,
+  nextNavLabel: string,
+  overImage: ?boolean,
+};
```
olliecurtis, line 43: I spoke to the team to align on the API so that we can replicate the same for apps and a good idea came up

Could we change this from being a boolean and switch to using a prop called `style`, which would then be an enum of two styles `'default'` and `'overImage'`?

This makes it expandable for the future and moves away from booleans to set the style

giriawu: Hi I've exported the props `style` with two styles - `default` and `overImage`. Thank you for your comment!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,144 @@
...
+              : undefined
+          }
+        >
+          {[...Array(totalIndicators)].map((val, index) => (
+            <button
+              type="button"
+              onClick={(e) => {
+                onClick(e, index, DIRECTIONS.INDICATORS);
+              }}
+              className={getClassName(
```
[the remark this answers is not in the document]
giriawu: I've simplified logic here and adjust scss file accordingly. Thank you!!

.spelling
```diff
@@ -900,3 +900,11 @@ INPUT_TYPES.number
...
 onSubmit
 otp
+currentIndex
+totalIndicators
+showNav
+indicatorLabel
+prevNavLabel
+nextNavLabel
+overImage
+default
```
jasonhuang-sky, line 910: This is a correct word, I think it shouldn't be here

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,44 @@
...
+| Property        | PropType | Required | Default Value |
+|-----------------|----------|----------|---------------|
+| currentIndex    | number   | true     | -             |
+| totalIndicators | number   | true     | -             |
+| indicatorLabel  | string   | true     | -             |
+| prevNavLabel    | string   | true     | -             |
+| nextNavLabel    | string   | true     | -             |
+| className       | string   | false    | null          |
+| showNav         | bool     | false    | false         |
+| style           | object   | false    | 'default'     |
```
jasonhuang-sky, line 42:
```suggestion
| style           | oneOf(INDICATOR_STYLE_TYPE)   | false    | INDICATOR_STYLE_TYPE.default    |
```

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,146 @@
...
+  </div>
+);
+
+BpkPageIndicator.propTypes = {
+  indicatorLabel: PropTypes.string.isRequired,
+  prevNavLabel: PropTypes.string.isRequired,
+  nextNavLabel: PropTypes.string.isRequired,
+  currentIndex: PropTypes.number.isRequired,
+  totalIndicators: PropTypes.number.isRequired,
+  style: PropTypes.oneOfType(Object.values(INDICATOR_STYLE_TYPE)),
```
jasonhuang-sky, line 133: I think it would be more common to use `Object.keys`

packages/bpk-component-page-indicator/README.md
```diff
@@ -0,0 +1,44 @@
+# bpk-component-page-indicator
+
+> Backpack example component.
```
anambl, line 3: very small thing from me if you could change this to page-indicator instead of example 😄

.spelling
giriawu: Have removed this, thank you very much!

packages/bpk-component-page-indicator/README.md
giriawu: Have updated, thank you very much!

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Have updated `$Values `-> `$Keys`, `Object.values` ->` Object.keys`. Thank you for your patient review!

packages/bpk-component-page-indicator/README.md
giriawu: Have updated this line, thank you so much!

olliecurtis:
Would you be able to check the overImage example? For some reason I am not seeing the indicators when I open this

giriawu:
> Would you be able to check the overImage example? For some reason I am not seeing the indicators when I open this

This does seem a bit strange, in visual test again it is displayed normally. I have tried to republish.

jasonhuang-sky:
> > Would you be able to check the overImage example? For some reason I am not seeing the indicators when I open this
> 
> This does seem a bit strange, in visual test example it is displayed normally. I have tried to republish.

Could it be missed because the indicators are at the bottom of the image instead of the middle?
image

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
LexSwed: `style` can be confused with native `style` property, maybe align it with the `Badge` or `Chip` API naming, using `type`? Or start here aligning all components to `variant` prop name for representation variations (because f.e. `button` should have native `type` prop passed through) 👀

packages/bpk-component-page-indicator/src/BpkPageIndicator.module.scss
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  &__indicator {
+    display: inline-block;
+    width: $dot-size;
+    height: $dot-size;
+    padding: 0;
+    border: none;
+    border-radius: 50%;
+    margin-inline: $dot-margin;
```
LexSwed, line 71: I'm not against using it here, but last time we checked CSS logical properties we decided to wait until Safari 14< usage drops: https://github.skyscannertools.net/tech-initiatives/production-standards/pull/48
I think it's fine using it here since it keeps the gallery fully functional, just not as pretty for Safari<14 users

packages/bpk-component-page-indicator/src/NavButton.js
```diff
@@ -0,0 +1,82 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { withButtonAlignment, withRtlSupport } from '../../bpk-component-icon';
+import BpkButton from '../../bpk-component-button';
+import LeftArrowIcon from '../../bpk-component-icon/lg/chevron-left';
+import RightArrowIcon from '../../bpk-component-icon/lg/chevron-right';
+
+type Props = {
+  prev?: boolean,
```
LexSwed, line 29: I hope I do not move the conversations back to what already was discussed and decided upon 😅  But still, I think `prev: boolean` is pretty confusing here even for internal Indicators API. Since we already have directions, we could re-use the keys, I think this is easier to justify:

```tsx
<NavButton direction={DIRECTIONS.PREV} onClick={() => onClick(e, currentIndex - 1, DIRECTIONS.PREV)} />
<Indicators />
<NavButton direction={DIRECTIONS.NEXT} onClick={() => onClick(e, currentIndex + 1, DIRECTIONS.NEXT)} />
```

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,146 @@
...
+                onClick(e, index, DIRECTIONS.INDICATORS);
+              }}
+              className={getClassName(
+                'bpk-page-indicator__indicator',
+                `bpk-page-indicator__indicator--${style}`,
+                index === currentIndex &&
+                  `bpk-page-indicator__indicator--active-${style}`,
+              )}
+              aria-label={`${indicatorLabel} ${index + 1}`}
+              aria-pressed={currentIndex === index ? 'true' : 'false'}
```
LexSwed, line 108: I think `aria-pressed` should only be used for toggle buttons (selected/unselected states). Maybe https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current fits better?

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -0,0 +1,146 @@
...
+                          totalIndicators - DISPLAYED_TOTAL,
+                        )
+                      : 0,
+                }
+              : undefined
+          }
+        >
+          {[...Array(totalIndicators)].map((val, index) => (
+            <button
+              type="button"
```
LexSwed, line 97: Just wondering, is it a requirement from design for the indicators to be clickable? They seem to be quite small clicking targets, so I guess it's for keyboard navigation only?

LexSwed:
We will want to re-use this component in the [image gallery](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=6889%3A38191), but there are some differences:
- it uses [Non-finite indicators](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=1763%3A18255) (`transform: scale` applied): 
image
- the indicators are not clickable, as they are small targets and only used on mobile. My main concern with making them `tabbable` with keyboard is that keyboard users would need to Tab through potentially over 40+ indicators. In these cases we could implement it like Radio Group ([example](https://www.radix-ui.com/docs/primitives/components/radio-group) or Tabs: there's only 1 tabbable element – active one – so that users **Tab In** to the indicators rail and **Tab Out** of it with single Tab. The navigation between indicators is than handled by the arrow keys.

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
olliecurtis: Let me take this back to the team to align the API again.

This could prove useful using this to block the native prop as the component has been set and designed so you should not be looking to change the styles either 🙂

packages/bpk-component-page-indicator/src/BpkPageIndicator.module.scss
giriawu: Thank you! I think your concerns and suggestions are very good, I did some cursory searches and about <0.8% of users are using safari <14, so I think we should be OK with this solution for now and we'll optimize it in the future when we have a better solution.

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Because it is considered that they will also be used on the desktop, and there are many places in projects where there is a need to click, so it is set here to be clickable. 
This also provides better accessibility and allows users to select the page number you want to go to within a certain range via the keyboard.

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
giriawu: Thanks for the review! After consideration, I think aria-current is used for indicators more often. Considering that the `bpk-page-indicator` is often used in combination with other components, we still set it to `true` and `false` to have more versatility.

packages/bpk-component-page-indicator/src/NavButton.js
olliecurtis: +1 and good catch, using booleans is bad and we will be moving away from using bools to specify and using enums to control

giriawu:
> We will want to re-use this component in the [image gallery](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=6889%3A38191), but there are some differences:
> 
> * it uses [Non-finite indicators](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=1763%3A18255) (`transform: scale` applied):
>   image
> * the indicators are not clickable, as they are small targets and only used on mobile. My main concern with making them `tabbable` with keyboard is that keyboard users would need to Tab through potentially over 40+ indicators. In these cases we could implement it like Radio Group ([example](https://www.radix-ui.com/docs/primitives/components/radio-group) or Tabs: there's only 1 tabbable element – active one – so that users **Tab In** to the indicators rail and **Tab Out** of it with single Tab. The navigation between indicators is than handled by the arrow keys.

Hi, thanks for your comment! Yes we had considered the swipe case, but considering that the `bpk-page-indicator` may only be used as a display/click/indicator when combined with other components, we wanted to keep it as clear as possible and leave the swipe tap in/out functionality to be done in the outside swipe container. 
As for your question about the maximum number of indicators, This does have an important role to play in certain application scenarios. I think we can consider following up with the designer to provide requirements for the next planning step.

giriawu:
> We will want to re-use this component in the [image gallery](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=6889%3A38191), but there are some differences:
> 
> * it uses [Non-finite indicators](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=1763%3A18255) (`transform: scale` applied):
>   image
> * the indicators are not clickable, as they are small targets and only used on mobile. My main concern with making them `tabbable` with keyboard is that keyboard users would need to Tab through potentially over 40+ indicators. In these cases we could implement it like Radio Group ([example](https://www.radix-ui.com/docs/primitives/components/radio-group) or Tabs: there's only 1 tabbable element – active one – so that users **Tab In** to the indicators rail and **Tab Out** of it with single Tab. The navigation between indicators is than handled by the arrow keys.

Hi, first of all, thanks for your review! I think your advice is good. But my consideration is that the indicator should have the role of navigation as a whole, in addition to the logic of its own Radio state. It should by default have semantics such as page-turning, because in practice it is often used in combination with multiple tab pages, or images. 
Regarding component interaction, I think we can design more function and improvement in the future with the backpack team and designers and engineers in the following step.
Thanks again for your comment!

packages/bpk-component-page-indicator/src/NavButton.js
giriawu: Thank you for your patient review, have updated!

LexSwed:
Love the PR, thank you @giriawu ❤️

giriawu:
I'm not sure why this happens. But after I merge master, it tells me that I must update the snapshot in the `bpk-component-content-cards`, otherwise, the visual test will fail. The error log is here: https://github.com/Skyscanner/backpack/actions/runs/3317969295/jobs/5481397876
So I update it in this PR.

anambl:
> I'm not sure why this happens. But after I merge master, it tells me that I must update the snapshot in the `bpk-component-content-cards`, otherwise, the visual test will fail. The error log is here: https://github.com/Skyscanner/backpack/actions/runs/3317969295/jobs/5481397876 So I update it in this PR.

Hi @giriawu sorry about this, I'm having a look, but I can't access the uploaded artifacts not sure why 😢 does that work for you? there shouldn't be any differences between the screenshots though

giriawu:
The CI failed again. I've rerun `npm run storybook:dist` and `npm run jest:visual-tests:update` locally, but it cannot pass the visual test. I've passed once, but the CI snapshot about` bpk-component-content-cards` changed again🤔
Maybe I will try to revert the change later.

olliecurtis (approved):
LGTM!

anambl:
> I've rerun `npm run storybook:dist` and `npm run jest:visual-tests:update` locally, but it cannot pass the visual test. I've passed once, but the CI snapshot about` bpk-component-content-cards` changed again🤔

I see it passed on CI, does it fail locally? 🤔

PR #2610, opened 2022-10-11
[KOA-5480] Update buttons to semantic tokens and adopt new colour palette

anambl:

- [ ] `README.md`
- [ ] Tests


UNRELEASED.md
```diff
@@ -1,10 +1,16 @@
...
+- `@skyscanner/backpack-web`:
+  - Update `bpk-scrim-utils` component to semantic colours
+  - Deprecated `dark` property in `bpk-scrim-utils` component as we only support one scrim style and this property could never be applied.
 
 **Fixed:**
 
-- bpk-component-accordion:
-    - Remove redundant `aria-labeledby` tags in accordion item
+- `@skyscanner/backpack-web`:
+  - Remove redundant `aria-labeledby` tags in accordion item in accordion component
```
olliecurtis, line 16: If we are changing this to be the single package, then what we should do is above all the changes/headings we should just put `@skyscanner/backpack-web` at the top

So like

```
`@skyscanner/backpack-web`

  **Breaking:**
    - Switch all components to use new colour palette.
   
    `BpkButton`
      - Update buttons to semantic colour tokens.
    
  **Changed:**
    `scrim-utils`
       - Update `bpk-scrim-utils` component to semantic colours
       - Deprecated `dark` property in `bpk-scrim-utils` component as we only support one scrim style and this property could never be applied.

  **Fixed:**
    `BpkAccordion`
      - Remove redundant `aria-labeledby` tags in accordion item in accordion component

```

olliecurtis (changes requested):
Just a small formatting query now we are only shipping single package

UNRELEASED.md
anambl: I think it's ok to add `backpack-web` at the top, but we have a few components that are exporting multiple components one of them being the button which will also export `BpkButtonPrimary` etc. so I don't think that `BpkButton` works. I do think though that we need something to make it very easy to understand what component it refers to. I'm thinking we could still add `bpk-component-button` just won't add a version next to it 🤔

olliecurtis: Sounds good to me! The other way we could do it is whatever the component name is after the `bpk-compoment-` similar to iOS and Android.

Either works but we just need to move to only noting the single package once

anambl: Good point, I see iOS is using `Backpack/Button` we could do that!

PR #2614, opened 2022-10-11
[KOA-4964]: Migrate away from Lerna

olliecurtis:
This PR migrates away from individual package publishing to just the single package with the removal of Lerna. The changes are as follows:

- [Commit 1 - Removal of Lerna](https://github.com/Skyscanner/backpack/pull/2614/commits/16118b9072bcd4f25a5ef0ef43863e83e76a46c3) 
  - This change removes all the build scripts that utilise Lerna, and the files related to cleanup
  - Also as now Lerna isn't available to run the gulp tasks inside the flare and icons component, I have moved this up to the root of backpack as this is only needed by us and not consumers anyway, so moved it so it can use the `gulp-cli` natively and also follows best practices from gulp on having many split tasks. See [gulp docs](https://gulpjs.com/docs/en/getting-started/javascript-and-gulpfiles#splitting-a-gulpfile) if you are interested.
  - Removed old scripts that were part of the old release process that were harnessed by the individual publish
- [Commit 2 - Update imports to be local](https://github.com/Skyscanner/backpack/pull/2614/commits/f8ae78c4c8ab5f886dfab0f9a4e6126952505ac8) - this is how we publish the single package by changing all the imports of bpk components to be local references rather than npm modules local to each package.
  - I wanted to do this in a separate PR as its the bulk of the changes but it would cause the build to fail as lerna was no longer available to install dependencies local to each component, so had to do it here to avoid an outage 

## Next steps
Post this PR there will be another couple of cleanup PRs to solidify our deployment
- Remove package and lock files from inside each `bpk-component` as these are no longer needed in a single package world, as there is just one root file in `packages`
- Cleanup further some of the release scripts to make it even smoother and open the doors for CI releases


package.json
```diff
@@ -154,9 +152,9 @@
     "gulp-rename": "^2.0.0",
     "gulp-svgmin": "^2.2.0",
     "husky": "^7.0.0",
+    "inquirer": "^9.1.3",
     "jest": "^26.0.1",
     "jest-axe": "^4.1.0",
-    "lerna": "^5.5.1",
```
olliecurtis, line 159: 👋

zgeor:
Can you please change the name of the `gulpfile.js` folder to something more apparent? Consider adding `.d` or using a different name altogether.

olliecurtis:
> Can you please change the name of the `gulpfile.js` folder to something more apparent? Consider adding `.d` or using a different name altogether.

This naming has been created as per the gulp docs which it requires to be, see here: https://gulpjs.com/docs/en/getting-started/javascript-and-gulpfiles#splitting-a-gulpfile

zgeor (approved):

PR #2620, opened 2022-10-17
[NO JIRA]: Upgrade actions to the latest

olliecurtis:
Our CI currently sends the following warning during builds and this is due to our GH actions being outdated, so this PR solved that.

Screenshot 2022-10-17 at 11 16 45


marianeum (approved):
I'd also recommend adding github actions to the dependabot config - we do that on Android if you wanna have a look

PR #2628, opened 2022-10-24
Updating READMEs since solely single package publishing

olliecurtis:
Now we have shifted to only publishing the single package, we need to update the READMEs across the board to reflect this change.

This PR does this.

Root package and packages/package.json file - we shift to :

- Noting just the single version badge at the top of the file
- Update installation instructions to just be the single package
- Remove links to each npm package badges as we no longer publish to them
- Also added a missing `bpk-animate-height` link as it was never there

For each component README:
- Update the installation step to point to the main repo (as we do for Backpack React Native)
- Update our example code for imports to point at the single package import

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 > Backpack is a collection of design resources, reusable components and guidelines for creating Skyscanner's products.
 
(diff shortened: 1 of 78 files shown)


marianeum (approved):

olliecurtis:
Overriding with merge as CI won't run due to skip ci being used as this is documentation only

PR #2629, opened 2022-10-25
Remove unrequired props from select

mungodewar:

- [ ] `README.md`
- [ ] Tests

Not too sure on the docs update. Let me know what I should do as the web select docs looks not complete: https://github.com/Skyscanner/backpack-docs/blob/main/docs/src/pages/components/Select/WebSelect.mdx

This is the pattern that causes a warning, but this usage works just fine and there is no reason to provide a `value` prop (and a `onChange` prop + introduce state)

Screenshot 2022-10-25 at 10 56 55


packages/bpk-component-select/README.md
```diff
@@ -34,8 +34,8 @@ export default () => (
 | ------------              | --------   | -------- | ------------- |
 | id                        | string     | true     | -             |
 | name                      | string     | true     | -             |
-| value                     | string     | true     | -             |
-| onChange                  | func       | true     | -             |
+| value                     | string     | false    | use [defaultValue](https://reactjs.org/docs/dom-elements.html#value)|
+| onChange                  | func       | false    | undefined     |
```
olliecurtis, line 38: If these are no longer required, we will probably need to modify the examples above to not include these as they are optional.

olliecurtis (changes requested):
Can you also add any entry to the UNRELEASED file alongside the semvarlevel (major, minor or patch) for the change?

It would also be good to have the description reflect at a high level the motivation for the change as this seems like a significant contract change which would be consumer breaking

olliecurtis:
> Not too sure on the docs update. Let me know what I should do as the web select docs looks not complete: https://github.com/Skyscanner/backpack-docs/blob/main/docs/src/pages/components/Select/WebSelect.mdx

No need to worry about the docs in this case as its not a whole new example and gets bumped automatically later

mungodewar:
Closing in favor of: https://github.com/Skyscanner/backpack/pull/2630

PR #2630, opened 2022-10-25
Remove unrequired props from select

mungodewar:

- [ ] `README.md`
- [ ] Tests

Not too sure on the docs update. Let me know what I should do as the web select docs looks not complete: https://github.com/Skyscanner/backpack-docs/blob/main/docs/src/pages/components/Select/WebSelect.mdx

This is the pattern that causes a warning, but this usage works just fine and there is no reason to provide a `value` prop (and a `onChange` prop + introduce state)

Screenshot 2022-10-25 at 10 56 55


UNRELEASED.md
```diff
@@ -1,6 +1,2 @@
-`@skyscanner/backpack-web`: `14.1.0 => 14.2.0`
-
-**Changed:**
-
-- `BpkFieldset`
-  - Updated spacing between label and input field.
\ No newline at end of file
+**Fixed:**
+  - `bpk-component-select` - updated prop types to allow for BpkSelect to be used in a uncontrolled manner.
```
mungodewar: I'm not quite sure what is going on here... AFAIK UNRELEASED should be empty on `main`?

anambl: yep, I think at the time you created the PR it was probably after some changes were merged and before they were released! I'll resolve the conflict so that the CI can run 🙂

packages/bpk-component-select/README.md
```diff
@@ -26,14 +26,33 @@ export default () => (
 );
 ```
 
+## Usage (uncontrolled)
+
+```js
+import React from 'react';
+import BpkSelect from 'bpk-component-select';
```
anambl, line 33:
```suggestion
import BpkSelect from '@skyscanner/backpack-web/bpk-component-select';
```
just a small correction here of the path

anambl (approved):
LGTM!

PR #2649, opened 2022-12-04
[BD-6473] Contribute carousel component to backpack

Supremeyh:
##  Context
Subject: Contribute carousel component to backpack
Ticket:  https://gojira.skyscanner.net/browse/BD-6473
UX:  [figma](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=8335%3A38592&t=Emn2whMWM2Wk6Dw0-0)
Visual link: 

## Why Are We Making This Change?
- We' re going to contribute carousel component in inventory card. The carousel is new component which allows a series of content to cycle horizontally. It contains both an drag and on click indicators gesture to progress left and right. At the end of theses of images, it returns to the first image.

## Change Description
 - Add new `bpk-component-carousel` componennt which allows a series of content to cycle horizontally
 - It has following functionality and props
    - The `className` to override class and style in `Carousel`
    - The `accessibilityLabel` support customize the aria-label of the frame container of the carousel
    - The `infinite` to control infinite or repeat mode
    - The `showIndicator` to control indicators

### Screenshots

| has indicators                        | no indicators                        |
|  |  |

- [x] `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-carousel/README.md
+++ b/packages/bpk-component-carousel/README.md
@@ -1,0 +1,32 @@
+# bpk-component-carousel
+
+> Backpack example component.
+
+## Installation
+
+Check the main [Readme](https://github.com/skyscanner/backpack#usage) for a complete installation guide.
+
+## Usage
+
+```js
+import React from 'react';
+import BpkCarousel from '@skyscanner/backpack-web/bpk-component-carousel';
+
+export default () => (
+  <BpkCarousel accessibilityLabel="This is a carousel">
+    <span>slide 1</span>
+    <span>slide 2</span>
+    <span>slide 3</span>
+  </BpkCarousel>
+);
+```
+
+## Props
+
+| Property           | PropType | Required | Default Value |
+| ------------------ | -------- | -------- | ------------- |
+| className          | string   | false    | null          |
+| children           | node     | true     | null          |
+| accessibilityLabel | false    | false    | -             |
+| infinite           | false    | false    | true          |
+| showIndicator      | false    | false    | true          |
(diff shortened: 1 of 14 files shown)


packages/bpk-component-carousel/README.md
```diff
@@ -0,0 +1,32 @@
...
+);
+```
+
+## Props
+
+| Property           | PropType | Required | Default Value |
+| ------------------ | -------- | -------- | ------------- |
+| className          | string   | false    | null          |
+| children           | node     | true     | null          |
+| accessibilityLabel | false    | false    | -             |
```
giriawu, line 30: How about to set it as `required`? Refer [BpkSelectableChip](https://backpack.github.io/components/chip?platform=web)  and [BpkPageIndicator](https://backpack.github.io/components/page-indicator?platform=web).

packages/bpk-component-carousel/src/BpkCarousel.js
```diff
@@ -0,0 +1,88 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* @flow strict */
+
+import PropTypes from 'prop-types';
+import React, { type Node } from 'react';
+import Carousel from 'nuka-carousel';
```
jasonhuang-sky, line 22: We should reuse the [bpk-component-page-indicator](https://github.com/Skyscanner/backpack/tree/main/packages/bpk-component-page-indicator) component in this one, so we only need the swipeable function, but this library has a complete implementation, including the design of indicators. Even though we can get the same design through style coverage, but in order to maintain the consistency of the design, we always need extra efforts in subsequent maintenance. I think this is not acceptable

As we talked about before, I am still think that [react-swipeable](https://github.com/FormidableLabs/react-swipeable) is the best dependency library for this component

jasonhuang-sky (changes requested):

PR #2656, opened 2022-12-05
[NO JIRA]: Remove format-package from project

olliecurtis:
Remove `format-package` as its not something we regularly use and as the last published changelog was for v5 and now they operate at v7 we cannot know what is in there without going through in depth the source code.

So due to this given the small benefit it gives, we can safely remove it.


anambl (approved):
LGTM!

PR #2658, opened 2022-12-07
[NO JIRA]: Update content cards examples

olliecurtis:
Updates examples to follow more realistic example as per the Figma and Design guidence see below what it looks like now with better content example

Screenshot 2022-12-07 at 15 15 11

- [ ] `README.md`
- [ ] Tests


anambl (approved):
LGTM!

PR #2659, opened 2022-12-08
[KOA-5698] Replace the use of deprecated tokens

anambl:

- [ ] `README.md`
- [ ] Tests


packages/bpk-component-content-cards/src/BpkContentCard.module.scss
```diff
@@ -67,7 +67,7 @@ $bpk-spacing-v2: true;
   }
 
   &--headline {
-    margin-bottom: $bpk-heading-margin-bottom;
+    margin-bottom: bpk-spacing-sm();
```
anambl, line 70: This is actually a different token now, but it was previously using the wrong value so I have fixed that to match with the one in Figma

olliecurtis (approved):

PR #2665, opened 2022-12-14
[KOA-5770]: Remove BpkContentContainer

olliecurtis:
Removed the Backpack Content Container. This component was originally created when building the Backpack docssite years ago but has since shifted and will be going away in 2023.

So removing this to cleanup our components/code and advising people to use BpkText which does the same thing and can apply in a similar way

- [x] `README.md`
- [ ] Tests


anambl (approved):
LGTM! 🎉

PR #2671, opened 2022-12-19
[NO JIRA]: Update disabled input cursor for the BpkCheckbox and BpkRadio components

akinniyi-akinpelu:
Updating the cursor of the disabled input in the `BpkCheckbox` and `BpkRadio` components to use the `not-allowed` style, as having it on the parent `<label>` wasn't enough for full coverage of the element when hovering over it, though because the label has the style, simply inheriting the value is enough to solve the issue.

[Context and video](https://skyscanner.slack.com/archives/C0JHPDSSU/p1671114670988229?thread_ts=1671114550.335799&cid=C0JHPDSSU) with the before and after example of the change. You can also see the issue in the [Backpack docs](https://backpack.github.io/components/checkbox/?platform=web) for the components.

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -29,3 +29,6 @@
 
   - `BpkCheckbox`
     - Update disabled input cursor to inherit `not-allowed` style
+
+  - `BpkRadio`
+    - Update disabled input cursor to inherit `not-allowed` style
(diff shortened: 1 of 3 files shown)


anambl (approved):
LGTM!  Thank you for the contribution! 🎉 🥳

PR #2677, opened 2023-01-03
Fix bpk switch deprecated type presence

oleksii-ziabkin:
This deprecated 'type' property has a default value set which causes a lot of warnings even in the case when 'type' property is not set in the code.
The idea is to remove default value from prototype and move it to the code to prevent these warnings.

- [ ] `README.md`
- [ ] Tests


anambl (approved):
LGTM! Thank you for the fix! 🙇

PR #2687, opened 2023-01-11
[LOOM-582] Remove display: none attribute from AnimateHeight

jamie-newbigging:
This PR removes all instances of the `style` attribute `display` being set in the `AnimateHeight` package

Previously `display: none` was being set by `AnimateHeight` when an element was being hidden (i.e `height` being set to 0)

Due to this, Banana were using a custom `Accordion` component with its own `AnimateHeight` package, so that `display: none` wasn't being set on hidden accordion elements in the footer for SEO reasons

With this change Banana will no longer need to use a custom implementation of the `Accordion` component and they can fall in line with the Backpack component

- [x] `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-animate-height/src/AnimateHeight.js
+++ b/packages/bpk-animate-height/src/AnimateHeight.js
@@ -45,19 +45,12 @@ class AnimateHeight extends Component {
     };
   }
 
-  componentDidMount() {
-    if (this.contentElement && this.props.height === 0) {
-      this.contentElement.style.display = 'none';
-    }
-  }
-
   UNSAFE_componentWillReceiveProps(nextProps) {
     const { height } = this.props;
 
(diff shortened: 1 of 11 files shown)


packages/bpk-animate-height/src/AnimateHeight.js
```diff
@@ -45,19 +45,12 @@ class AnimateHeight extends Component {
     };
   }
 
-  componentDidMount() {
-    if (this.contentElement && this.props.height === 0) {
-      this.contentElement.style.display = 'none';
```
anambl, line 50: was this tested with screen readers? As I believe elements with `display: none` are hidden from screen readers so removing this might mean it's now read?

jamie-newbigging:
Closed due to lack of benefits from the change

PR #2693, opened 2023-01-20
Fix BpkPagination throwing a warning caused by incorrect BpkButton prop

adamm9996:
`primary` is not a valid `BpkButton` field so I switched it to `primaryOnLight`. This was spamming the console wherever we use a `BpkPagination` component and it was making our danger checks fail, preventing us from merging PRs.

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-pagination/src/BpkPaginationPage.js
+++ b/packages/bpk-component-pagination/src/BpkPaginationPage.js
@@ -37,7 +37,7 @@ const BpkPaginationPage = (props) => {
   return (
     <BpkButton
       primaryOnDark={!isSelected}
-      primary={isSelected}
+      primaryOnLight={isSelected}
       onClick={onSelect}
       className={classNames.join(' ')}
       aria-label={pageLabel(page, isSelected)}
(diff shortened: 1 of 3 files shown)


packages/bpk-component-pagination/src/BpkPaginationPage.js
```diff
@@ -37,7 +37,7 @@ const BpkPaginationPage = (props) => {
   return (
     <BpkButton
       primaryOnDark={!isSelected}
-      primary={isSelected}
+      primaryOnLight={isSelected}
```
anambl, line 40: this works fine here visually as on web we only support light mode, but the primary and primaryOnLight are 2 different styles, the only difference being on dark mode
I think we can just remove this line completely as if you don't pass one of the button types as a prop it will return by default the primary button

anambl:
thank you for the fix!! 😄 could you also add an entry to the UNRELEASED file please? 🙂

sogame:
Hi @anambl, no rush, but do you think this could be merged? Just to avoid conflicts in `unreleased.md` and `.spelling` 😅

anambl (approved):
LGTM!

anambl:
> Hi @anambl, no rush, but do you think this could be merged? Just to avoid conflicts in `unreleased.md` and `.spelling` 😅

Hi @sogame, just approved now, very sorry for the delay!! will merge now and release later today or tomorrow! 🙂

PR #2713, opened 2023-02-06
[NO JIRA] Remove unused dependencies

anambl:
- Removed `danger-plugin-toolbox` and `flow-bin` as we are not using them
- Created a JIRA ticket to re-enable the disabled eslint rules. That's due to the fact that a minor bump to stylelint-config-skyscanner triggered a major bump to eslint-config-skyscanner which caused linting errors blocking CI.


.eslintrc
```diff
@@ -85,6 +85,9 @@
     "prettier/prettier": "error",
     "react/jsx-filename-extension": "off",
     "import/no-extraneous-dependencies": "off",
+    "no-restricted-imports": "off",
+    "no-unused-vars": "off",
+    "@typescript-eslint/consistent-type-imports": "off",
```
olliecurtis, line 90: Can we add a comment here to explain the reason for disabling with a todo or a ticket number to reference for future enablement of these?

package.json
```diff
@@ -109,7 +109,7 @@
     "@babel/register": "^7.18.9",
     "@skyscanner/bpk-foundations-web": "^13.6.0",
     "@skyscanner/bpk-svgs": "^16.1.0",
-    "@skyscanner/stylelint-config-skyscanner": "^6.0.2",
+    "@skyscanner/stylelint-config-skyscanner": "^6.1.0",
```
olliecurtis, line 112: This version will now throw a warning that it is deprecated and recommend to bump to 7.0.0 (which includes the same changes just correctly semver)

https://github.com/Skyscanner/stylelint-config-skyscanner/blob/main/CHANGELOG.md#610-deprecated

olliecurtis (changes requested):

olliecurtis (approved):

PR #2718, opened 2023-02-08
[NO JIRA] Fix datatable warnings

anambl:

- [ ] `README.md`
- [ ] Tests


packages/bpk-component-datatable/src/BpkDataTable.js
```diff
@@ -41,7 +41,13 @@ const BpkDataTable = (props: Props) => {
     children,
     className,
     defaultColumnSortIndex,
+    headerClassName,
+    headerHeight,
     height,
+    onRowClick,
+    rowClassName,
```
olliecurtis, line 48: Now this is explicitly called out should we add this to the README for the component?

olliecurtis (changes requested):

packages/bpk-component-datatable/src/BpkDataTable.js
anambl: ah this was one of those props that we never explicitly supported, but was part of the API because of the library that we were using before and when I checked the usage some projects were using it

olliecurtis (approved):

PR #2720, opened 2023-02-09
[KOA-5850] Make atomic cards accessible to keyboard users

anambl:

- [ ] `README.md`
- [ ] Tests


packages/bpk-component-card/src/BpkCard.js
```diff
@@ -78,15 +75,17 @@ const BpkCard = (props: Props) => {
     );
   }
 
-  // If the card is atomic, we need to enable keyboard focus and provide an appropriate role.
   if (atomic) {
-    atomicProps.tabIndex = 0;
-    atomicProps.role = 'button';
+    return (
+      <button type="button" className={classNames} {...rest}>
```
olliecurtis, line 80: If the html is already set to a button we shouldn't need to specify type="button" or are you doing this to protect from being overriden with `rest`?

anambl: No, but removing it throws a lint error. I read about it and it's actually a good practice to add a type because browsers have different default types!

olliecurtis (approved):

PR #2723, opened 2023-02-13
[KOA-5890]: Update Price API

olliecurtis:
This PR updates the Price component API to use a more clear set of parameters and aligns with our Figma designs and mobile platform API.

It makes the following changes:

  - `title` has been renamed to `price`.
  - `description` has been renamed to `trailingText`.
  - `subtitle` has been renamed to `leadingText`.
  - Added `previousPrice` property for displaying a previous price.

- [x] `README.md`
- [x] Tests


marianeum (approved):

packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -54,35 +63,60 @@ const BpkPrice = (props: Props) => {
...
-          className={getClassName('bpk-price__subtitle')}
-          textStyle={isSmall ? TEXT_STYLES.xs : TEXT_STYLES.sm}
-          tagName="span"
-        >
-          {subtitle}
-        </BpkText>
-      )}
+      <div
+        className={getClassName(
+          previousPrice && 'bpk-price__leadingText',
```
anambl, line 68: any reason we cannot use kebab case here? 🙂

olliecurtis: You are right, I fully blame copilot here as it wrote this for me 😆

examples/bpk-component-price/stories.js
```diff
@@ -17,44 +17,41 @@
  */
 
 import {
-  SmallLeftExample,
```
anambl, line 20: why were all these renamed? If I remember correctly, we use `Example` at the end of the name for all example components

olliecurtis: This was due to some of them getting really long and I noticed we don't fully follow this pattern all the way through with all our examples so I thought this was ok, especially with an example name of

`SmallLeftWithTrailingAndLeadingText`

Happy to change it though if you would like and we want to make it official all stories should change to this naming

olliecurtis: Although now I do notice I switched it around the naming of the examples and the stories, so let me switch that back to add them at the end.

They will still be renamed but not remove `Example`

packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -54,35 +63,60 @@ const BpkPrice = (props: Props) => {
...
+          </BpkText>
+        )}
+
+        {leadingText && (
+          <BpkText
+            textStyle={isSmall ? TEXT_STYLES.xs : TEXT_STYLES.sm}
+            tagName="span"
+          >
+            {leadingText}
+          </BpkText>
```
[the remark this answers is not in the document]
olliecurtis: Probably not worth it because of it creates an extra layer that will just wrap our Text component in the same way. So would then increase overhead and maintenance/complexity to the system. So to me doesn't make sense to just move this same logic to somewhere else.

Which then we would have to do massive change across our components to follow this pattern as if we introduce it once we have to introduce it everywhere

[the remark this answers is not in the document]
frugoman: Agree!

frugoman (approved):

anambl (approved):
LGTM!

PR #2727, opened 2023-02-14
[KOA-5904] Remove React imports

anambl:
React no longer needs to be imported when using JSX.
The default React import `import React from 'react';` will be removed in a future React release - https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports

First 3 commits of the PR were done using the React script for changing the imports. The following commits include manual changes as due to some errors the scripts didn't pick up the changes.

- [ ] `README.md`
- [ ] Tests


.storybook/babel.config.js
```diff
@@ -41,7 +41,7 @@ module.exports = {
         },
       },
     ],
-    '@babel/preset-react',
+    ['@babel/preset-react', { runtime: 'automatic' }],
```
olliecurtis, line 44: Small thing, could you add a comment above this just to clarify why we have this turned on just for future maintenance

babel.config.js
```diff
@@ -33,7 +33,7 @@ module.exports = {
         },
       },
     ],
-    '@babel/preset-react',
+    ['@babel/preset-react', { runtime: 'automatic' }],
```
olliecurtis, line 36: Same here

examples/bpk-animate-height/examples.js
```diff
@@ -18,7 +18,8 @@
 
 /* @flow strict */
 
-import React, { Component, type Node } from 'react';
+import { Component } from 'react';
+import type { Node } from 'react';
```
olliecurtis, line 22: Can we not keep these combined as before and just remove the React import?

I know this was auto completed by the script but wondering if we can keep the change and the number of imports low. This will be a similar comment throughout so keeping to this one thread.

If it doesn't work then that is fine just want to check first

olliecurtis (changes requested):
We should probably document this in the UNRELEASED as we will need to pop this into the changelog as it will pop a release.

Hopefully this shouldn't impact consumers but might be worth keeping it in mind and maybe we test it in a simple BRS app

examples/bpk-animate-height/examples.js
anambl: Yep so this was done automatically by the scripts. I did try to see if it's possible to keep them combined and it doesn't seem to work. Also had a look at other team's PRs and it seems like that's how it is used in other repos in the company too

olliecurtis (approved):

PR #2730, opened 2023-02-16
[BD-6622] Add BpkCardWrapper component

Supremeyh:
##  Context
According to [backpack design](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=22572%3A28385&t=S5Vk5LFZTG6ellLC-0), and some [discussion](https://skyscanner.slack.com/archives/C04G18ZK2TD/p1672987004954479) in slack, we're going to add new `BpkCardWrapper` component which is used for card wrapper component.

## Change Description
 - Add `BpkCardWrapper` component.
    - The Property `backgroundColor` is string type and required which set for background color of header and box shadow.
    - The Property `card` is node type for card content.
    - The Property `header` is node type for header content.
    - The Property `className` is string type override for the component.

### Screenshots

| Card Wrapper Example          | Divided Card Wrapper Example  |
|  |  |

| With background color Example          |
| ----------------------------- |
|  |

- [x] `README.md`
- [x] Tests


packages/bpk-component-card/src/BpkCardWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* @flow strict */
+
+import PropTypes from 'prop-types';
+import React, { type Node } from 'react';
+import { borderSizeLg } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
```
anambl, line 23: this is not used anywhere so it can be removed

packages/bpk-component-card/src/BpkCardWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+
+  return (
+    <div
+      className={classNames}
+      style={{
+        boxShadow: `0 0 0 2px ${backgroundColor}`,
+        '--background-color': backgroundColor,
+      }}
+    >
+      <section
```
anambl, line 51: what's the reason we're using `section` rather than `div` here (and below for the card as well)?

packages/bpk-component-card/src/BpkCardWrapper.js
```diff
@@ -0,0 +1,75 @@
...
+const BpkCardWrapper = (props: Props) => {
+  const { backgroundColor, card, className, header } = props;
+
+  const classNames = getClassName('bpk-card-wrapper', className);
+
+  return (
+    <div
+      className={classNames}
+      style={{
+        boxShadow: `0 0 0 2px ${backgroundColor}`,
```
anambl, line 47: ~might be good to use the same method as below and set the boxShadow in CSS rather than doing it here?~ ah wait just realised we do so I guess we don't need this?
also wondering if we don't already have a border mixin which supports this size?

packages/bpk-component-card/src/BpkCardWrapper.js
Supremeyh: Yeah, it's not used and redundant here.

packages/bpk-component-card/src/BpkCardWrapper.js
Supremeyh: I think the [section](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section)  represents a standalone section and might be semantic, and too much `div` looks not very good. Anyway the `div` also works here.

packages/bpk-component-card/src/BpkCardWrapper.js
Supremeyh: Agree with you, move it into scss might would be better. For border mixins, i'd prefer use spacing mixins as mixed `$bpk-border-size-lg` and `bpk-spacing-sm()` here seems not quite pretty? On the other hand, we need this for card wrapper box shadow effect.

packages/bpk-component-card/src/BpkCardWrapper.js
anambl: The reason I'm asking is because looking at the Mozilla definition it doesn't seem like it could apply here so wasn't sure if there was a specific reason to use it 🙂 
```The <section> [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a generic standalone section of a document, which doesn't have a more specific semantic element to represent it. Sections should always have a heading, with very few exceptions.```

File: packages/bpk-component-card/src/BpkCardWrapper.js (hunk 3 above)
Comment (Supremeyh) [reply packages/bpk-component-card/src/BpkCardWrapper.js]: It seems would result in some problem `box-shadow ` always comes in front of `@include` due to pre commit check and make box shadow not work.

Comment (Supremeyh) [reply packages/bpk-component-card/src/BpkCardWrapper.js]: Oh, i realize it can use `@include bpk-border-lg`, it works now.

File: packages/bpk-component-card/src/BpkCardWrapper.module.scss [hunk 4]
```diff
@@ -19,8 +19,9 @@
 @import '~bpk-mixins/index.scss';
 
 .bpk-card-wrapper {
-  @include bpk-border-radius-md;
   @include bpk-card;
+  @include bpk-border-radius-md;
+  @include bpk-border-lg(var(--background-color), '');
```
anambl, line 24: Looking at the Figma design now it actually seems like the shadow shouldn't be the colour of the wrapper, but the default shadows we support in Backpack

Supremeyh: Thanks for helping confirmation on this. I have remove shadow effect when hover and use default shadow instead.

packages/bpk-component-card/src/BpkCardWrapper.module.scss
```diff
@@ -0,0 +1,65 @@
...
+
+    &::before,
+    &::after {
+      position: absolute;
+      top: 0;
+      content: ' ';
+      z-index: 2;
+      width: bpk-spacing-lg();
+      height: bpk-spacing-lg();
+      border-radius: 50%;
```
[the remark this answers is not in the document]
Supremeyh: Make sense, has updated to using `bpk-border-radius-md` instead.

packages/bpk-component-card/src/BpkCardWrapper-test.js
```diff
@@ -0,0 +1,71 @@
...
+        className="custom-classname"
+        header={message}
+        card={longMessage}
+        backgroundColor={coreAccentDay}
+      />,
+    );
+    const results = await axe(container);
+
+    expect(results).toHaveNoViolations();
+  });
```
olliecurtis, line 70: We usually put these Accessibility tests in a file `accessibility-test.js`. Could you move this to that file?

as a `describe('BpkCardWrapper accessibility tests')`

packages/bpk-component-card/src/BpkCardWrapper.js
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <div
+      className={classNames}
+      style={{
+        '--background-color': backgroundColor,
+      }}
+    >
+      <div
+        style={{ backgroundColor }}
```
olliecurtis, line 50: Do we need to do this or can we in the css just set

`background-colour: inherit` to then just take from the higher div automatically

packages/bpk-component-card/src/BpkCardWrapper.js
```diff
@@ -0,0 +1,71 @@
...
+      >
+        {header}
+      </div>
+      <div className={getClassName('bpk-card-wrapper--content')}>{card}</div>
+    </div>
+  );
+};
+
+BpkCardWrapper.propTypes = {
+  backgroundColor: PropTypes.string.isRequired,
```
olliecurtis, line 61: Wondering if this should not be required but optional and the default colour should be `coreAccentDay`, might be worth us checking with design on this. To see if we would always have this

olliecurtis (changes requested):

packages/bpk-component-card/README.md
```diff
@@ -46,6 +48,17 @@ export default () => (
...
 ```
 
+### BpkCardWrapper
+
+```js
+import { BpkCardWrapper } from '@skyscanner/backpack-web/bpk-component-card';
+import { coreAccentDay } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
+
+export default () => (
+  <BpkCardWrapper header={'foo'} card={'bar'} backgroundColor={coreAccentDay} />
```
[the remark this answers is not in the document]
Supremeyh: Sure thing, has updated to node type

packages/bpk-component-card/src/BpkCardWrapper-test.js
Supremeyh: Thanks, has this case moved into `accessibility-test.js`

packages/bpk-component-card/src/BpkCardWrapper.js
Supremeyh: Make sense, has moved this logic into scss. For the backgroundColor, i'd prefer still put under `&--header` as put under `bpk-card-wrapper` would override by pre commit check and make backgroundColor not work.

packages/bpk-component-card/src/BpkCardWrapper.js
Supremeyh: Basically, i follow the [design definition](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=22572%3A28385&t=S5Vk5LFZTG6ellLC-0) for BPK Card Wrapper API. I' ll check with designer whether it's fine to change to optional and set default `coreAccentDay` colour.

packages/bpk-component-card/src/BpkCardWrapper.module.scss
```diff
@@ -0,0 +1,61 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '~bpk-mixins/index.scss';
+
+.bpk-card-wrapper {
+  @include bpk-card;
+  @include bpk-border-radius-md;
```
anambl, line 23: this won't be needed if we use the `bpk-card` mixin as that's already included in there!

packages/bpk-component-card/src/BpkCardWrapper.module.scss
```diff
@@ -0,0 +1,61 @@
...
+  @include bpk-border-lg(var(--background-color), '');
+
+  &--header {
+    border-radius: $bpk-border-radius-sm() $bpk-border-radius-sm 0 0;
```
anambl, line 27: I think there's an extra set of parenthesis for the top left border radius?

packages/bpk-component-card/src/BpkCardWrapper.module.scss
```diff
@@ -0,0 +1,61 @@
...
+    &::before,
+    &::after {
+      position: absolute;
+      top: 0;
+      content: ' ';
+      z-index: 2;
+      width: bpk-spacing-lg();
+      height: bpk-spacing-lg();
+      background-color: transparent;
+      box-shadow: bpk-spacing-base() 0 0 0 var(--background-color);
```
anambl, line 45: why is this line needed? I noticed it adds a bit of extra shadow to the card on bottom left and right corners when hovering over the card

packages/bpk-component-card/src/BpkCardWrapper.js
Supremeyh: Some discussion goes [here](https://skyscanner.slack.com/archives/C04G18ZK2TD/p1676860243721029) in slack.

packages/bpk-component-card/src/BpkCardWrapper.module.scss
Supremeyh: yeah, it's no need here

packages/bpk-component-card/src/BpkCardWrapper.module.scss
Supremeyh: nice catch, it's  extra set of parenthesis

packages/bpk-component-card/src/BpkCardWrapper.module.scss
Supremeyh: It due to there also need background color in the corner between header and card.
image

Supremeyh: For extra shadow when hover the card, i guess it comes from content radius, has removed it.

packages/bpk-component-card/src/BpkCardWrapper.js
Supremeyh: Confirmed with Adam again, we final decided backgroundColor keep required. See [here](https://skyscanner.slack.com/archives/C04G18ZK2TD/p1676950931734259)

anambl (approved):
LGTM! Thank you for the contribution! 👏 🎉 🥳

olliecurtis (approved):

PR #2733, opened 2023-02-20
Fix rtl issue for switch component

viktor-yang:
There is a rtl issue on the switch component, this PR is to fix it.

Before:
image
After:
image

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-switch/src/BpkSwitch.module.scss
+++ b/packages/bpk-component-switch/src/BpkSwitch.module.scss
@@ -54,13 +54,8 @@ $border-width: $bpk-one-pixel-rem * 2;
   }
 
   &__label {
-    margin-right: bpk-spacing-md();
     line-height: $height;
-
-    @include bpk-rtl {
-      margin-right: 0;
-      margin-left: bpk-spacing-md();
-    }
+    margin-inline-end: bpk-spacing-md();
   }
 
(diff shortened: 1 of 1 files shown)


giriawu (approved):

packages/bpk-component-switch/src/BpkSwitch.module.scss
```diff
@@ -56,6 +56,11 @@ $border-width: $bpk-one-pixel-rem * 2;
   &__label {
     margin-right: bpk-spacing-md();
     line-height: $height;
+
+    @include bpk-rtl {
+      margin-right: 0;
```
dominicfraser, line 61: Could we use `margin-inline-start` here?

cc: @LexSwed

dominicfraser: (We haven't decided on its usage in Backpack, but in main web are moving away from bpk-rtl to towards native support where possible)

olliecurtis: I think this would be safe to use here yes!

viktor-yang: Make sense of that, thank you both.

olliecurtis (approved):

PR #2738, opened 2023-02-20
[KOA-5057]: Remove width and height properties

olliecurtis:

- [x] `README.md`
- [x] Tests


packages/bpk-component-image/src/BpkBackgroundImage.js
```diff
@@ -55,15 +52,7 @@ class BpkBackgroundImage extends Component<BpkBackgroundImageProps> {
   onBackgroundImageLoad: () => mixed;
 
   static propTypes = {
-    aspectRatio: widthHeightAspectRatioPropType,
```
olliecurtis, line 58: Now that we have removed `width` and `height` we no longer need to have a custom prop type for `aspectRatio` and can set it to be a number and set it to be required properly now.

PR #2743, opened 2023-02-21
[LOOM-640] - setting up Percy for Backpack GitHub actions

kacperlsky:
**Description:**
Adding the Percy Storybook snapshots integrated with Github Actions as a new visual testing platform.

- [-] `UNRELEASED.md` (See [How to write a good changelog entry](https://github.com/Skyscanner/backpack/blob/main/CHANGELOG_FORMAT.md))
- [-] `README.md`
- [-] Tests
- [-] Storybook examples created/updated

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -137,7 +137,7 @@ jobs:
           key: ${{ env.BUILD_CACHE_NAME }}-${{ hashFiles('packages/**') }}
 
       - name: Percy Test
-        run: percy storybook ./dist-storybook
+        run: npx percy storybook ./dist-storybook
 
 
   Danger:
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -137,7 +137,8 @@ jobs:
           key: ${{ env.BUILD_CACHE_NAME }}-${{ hashFiles('packages/**') }}
 
       - name: Percy Test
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -90,6 +90,11 @@ jobs:
               body: "Visual regression tests failed 😢. You can download the failure diffs from the 'Artifacts' section of the failed CI run. To update the tests, run `npm run jest:visual-tests:update` locally."
             })
 
+      - name: Percy Test
+        run: npx percy storybook ./dist-storybook
```
olliecurtis, line 94: Worth noting here by using npx it will install the latest from npm rather than using the version specified locally, so could in the future cause any issues if say percy/cli released a major that we need to change.

One thing I would recommend is like when testing locally adding an npm command in the package file to run the installed version :)

.github/workflows/ci.yml
```diff
@@ -90,6 +90,11 @@ jobs:
               body: "Visual regression tests failed 😢. You can download the failure diffs from the 'Artifacts' section of the failed CI run. To update the tests, run `npm run jest:visual-tests:update` locally."
             })
 
+      - name: Percy Test
```
olliecurtis, line 93: We will probably also want this to not run on dependabot PRs and not forks - to avoid busting limits and prevent failures so will need to add under here:

```
if: github.repository == github.event.pull_request.head.repo.full_name && github.actor != 'dependabot[bot]'
```

olliecurtis (approved):

PR #2748, opened 2023-02-21
[KOA-5055] Remove deprecated calendar props

anambl:

- [x] `README.md`
- [ ] Tests


packages/bpk-component-calendar/src/BpkCalendarGrid.js
```diff
@@ -139,8 +137,6 @@ class BpkCalendarGrid extends Component {
               markToday={markToday}
               markOutsideDays={markOutsideDays}
               selectionConfiguration={selectionConfiguration}
-              selectionEnd={selectionEnd}
-              selectionStart={selectionStart}
```
olliecurtis, line 143: Is it right these got removed even though they aren't marked as deprecated?

olliecurtis (changes requested):

packages/bpk-component-calendar/src/BpkCalendarGrid.js
anambl: Yes, I think it was just something we accidentally forgot, but we did add it to the Changelog - https://github.com/Skyscanner/backpack/blob/main/CHANGELOG.md#2021-11-15
They're also not actually used anywhere in the Week component anyways so it has no effect!

packages/bpk-component-calendar/src/Week.js
```diff
@@ -379,44 +379,25 @@ Week.propTypes = {
...
   cellClassName: PropTypes.string,
   onDateClick: PropTypes.func,
   onDateKeyDown: PropTypes.func,
   selectionConfiguration: CustomPropTypes.SelectionConfiguration,
-  selectedDate: deprecated(
-    PropTypes.instanceOf(Date),
-    'Use selectionConfiguration to set selectedDate',
-  ),
-  selectionEnd: PropTypes.instanceOf(Date),
-  selectionStart: PropTypes.instanceOf(Date),
```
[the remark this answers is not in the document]
anambl: Yep, these 2 props are not used at all in the component - we don't specify them when destructuring the `props` and we also don't pass it on through `rest` or something similar as we don't do that in this component

olliecurtis (approved):

PR #2750, opened 2023-02-24
[ST-4754] Remove the problematic component from Calendar.

angel3cu:
While testing the calendar on Percy, we identified legacy components for displaying weekdays and the current month that are causing trouble. The components do not display on the screen and are hidden from screen readers under normal circumstances. However, they contain invalid HTML which confuses Percy, leading to irrational screenshots.

See [here](https://percy.io/4ea9fae5/flights-search-controls/builds/25407558/changed/1416658585?browser=chrome&browser_ids=34&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1280&widths=359%2C1280) for the issue that this PR is going to fix.

We decided to remove these components completely since they do not seem to serve any purpose and they do more harm than good.

I tested the accessibility of the calendar and the behaviour hasn't changed from the original.

https://user-images.githubusercontent.com/65911077/221234561-a3033b34-e52c-44a7-a56b-8c69a545be2e.mov

- [x] `README.md`
- [x] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -1,3 +1,11 @@
 **Changed:**
 
 `bpk-component-button`: Cleaned up the onClick functionality to remove a wrapper that is no longer necessary
+
+**Fixed:**
+
+`BpkCalendarGrid`
+
+- Updated to remove a hidden component that was causing visual testing issues.
+- Removed `daysOfWeek` and `formatMonth` from the props, since they are no
+  longer needed.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,7 +2,7 @@
 
 `bpk-component-button`: Cleaned up the onClick functionality to remove a wrapper that is no longer necessary
 
-**Fixed:**
+**Breaking:**
 
 `BpkCalendarGrid`
 
--- a/packages/bpk-component-calendar/src/BpkCalendarGrid.js
+++ b/packages/bpk-component-calendar/src/BpkCalendarGrid.js
@@ -21,14 +21,12 @@ import { Component } from 'react';
 
 import { cssModules, isDeviceIos } from '../../bpk-react-utils';
 
-import BpkCalendarGridHeader from './BpkCalendarGridHeader';
 import Week from './Week';
 import {
   addMonths,
   formatIsoDate,
   getCalendarMonthWeeks,
   isSameMonth,
-  orderDaysOfWeek,
 } from './date-utils';
 import CustomPropTypes, { CALENDAR_SELECTION_TYPE } from './custom-proptypes';
 import STYLES from './BpkCalendarGrid.module.scss';
@@ -58,24 +56,18 @@ class BpkCalendarGrid extends Component {
         props.month,
         props.weekStartsOn,
       ),
-      daysOfWeek: orderDaysOfWeek(props.daysOfWeek, props.weekStartsOn),
     };
   }
 
   UNSAFE_componentWillReceiveProps(nextProps) {
     // We cache expensive calculations (and identities) in state
     if (
-      nextProps.daysOfWeek !== this.props.daysOfWeek ||
       !isSameMonth(nextProps.month, this.props.month) ||
       nextProps.weekStartsOn !== this.props.weekStartsOn
     ) {
       this.setState({
         calendarMonthWeeks: getCalendarMonthWeeks(
           nextProps.month,
-          nextProps.weekStartsOn,
-        ),
-        daysOfWeek: orderDaysOfWeek(
-          nextProps.daysOfWeek,
           nextProps.weekStartsOn,
         ),
       });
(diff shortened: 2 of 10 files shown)


packages/bpk-component-calendar/src/BpkCalendarGrid.js
```diff
@@ -74,10 +70,6 @@ class BpkCalendarGrid extends Component {
           nextProps.month,
           nextProps.weekStartsOn,
         ),
-        daysOfWeek: orderDaysOfWeek(
```
olliecurtis, line 77: This has been removed but is a key part of the calendar for providing information on the calendar (see prop description) for different styles/localised info on the week days are we confident in removing it won't impact consumers:

https://github.com/Skyscanner/backpack/tree/main/packages/bpk-component-calendar#daysofweek

packages/bpk-component-calendar/src/BpkCalendarGrid.js
```diff
@@ -58,14 +56,12 @@ class BpkCalendarGrid extends Component {
         props.month,
         props.weekStartsOn,
       ),
-      daysOfWeek: orderDaysOfWeek(props.daysOfWeek, props.weekStartsOn),
```
olliecurtis, line 61: Same here, especially for localised here where some countries the first day of the week is not the same as UK

packages/bpk-component-calendar/src/BpkCalendarGrid.js
```diff
@@ -91,7 +83,6 @@ class BpkCalendarGrid extends Component {
       dateProps,
       focusedDate,
       formatDateFull,
-      formatMonth,
```
olliecurtis, line 94: This has been removed but still is mentioned in the README is this intentional?

olliecurtis (changes requested):
Few points from me. Also can you:

- Update UNRELEASED to mention the changes here?
- How does these changes work with screenreaders with: Scrollable calendar and Datepicker?

packages/bpk-component-calendar/src/BpkCalendarGrid.js
angel3cu: Yes, I believe so. In the grid, it is only used by the components that were removed, so without them it is flagged as unused variable. I double checked the code and indeed it is set in the constructor, but never really used.

packages/bpk-component-calendar/src/BpkCalendarGrid.js
angel3cu: I don't believe it is used in the rest of the grid.

packages/bpk-component-calendar/src/BpkCalendarGrid.js
angel3cu: Apologies, I will double check the README. It is the same case as before, it was used by the hidden components and not used anymore in the grid. Note that the rest of the calendar still uses this, but not the grid component.

angel3cu: Double checked the README. I left the occurrences in `Calendar` and `CalendarNav`, as well as the explanation of this property, because it is still used in these two components. The reference from `CalendarGrid` have already been removed in my original commit. 😄

anambl (commented):
LGTM! Only thing is to add an entry to the UNRELEASED.md file to indicate what has changed for consumers :)

angel3cu:
@anambl @olliecurtis Updated `UNRELEASED.md`. 😄

anambl (approved):
LGTM! 🎉

UNRELEASED.md
```diff
@@ -1,3 +1,11 @@
 **Changed:**
 
 `bpk-component-button`: Cleaned up the onClick functionality to remove a wrapper that is no longer necessary
+
+**Fixed:**
```
olliecurtis, line 5: Based on the change and we are removing props - therefore the contract I would say can you mark this as breaking? Thanks

olliecurtis (changes requested):

UNRELEASED.md
anambl: I think this change doesn't affect anyone as the bit which was removed didn't have any effect as far as we can see 🤔. Also the props which were removed didn't have any effect either, but given consumers still pass them (as they are marked as required), might be worth flagging so that they can be removed.

angel3cu: The change breaks the API (removing properties), but given that we don't have any typechecks in the released library, it will be backwards compatible with existing consumers. If they pass these two props in, then they will simply be ignored. I have marked this as breaking anyway, for clarity. Thanks both! 😄

olliecurtis (approved):

angel3cu:
Thank you both for reviewing and merging this. 😄
Which release are the changes planned to be in? 🙂

PR #2757, opened 2023-02-27
[KOA-2617]: Update device detection utils to support latest UA string

olliecurtis:
Google has recently deprecated the use of `navigator.platform` for determining a device being used so moving towards the newer favoured `navigator.userAgent`

- [ ] `README.md`
- [x] Tests


anambl (approved):
LGTM!

PR #2763, opened 2023-03-08
SEARCHEXP-873: Do not set height to 0 on re-size

robaw:
There is a bug with the BpkScrollableCalendarGridList where if you hide it (ie. `display: none`) and a re-size event is fired, the `setComponentHeight()` fn will be called and the BpkScrollableCalendarGridList will copy the `outerDiv` height of 0. If another re-size event does not occur, when the calendar is no longer hidden (ie. NOT `display: none`) then the calendar will still have a height of 0, making it invisible. 

- [ ] `README.md`
- [ ] Tests

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,3 +16,6 @@
 
 - `BpkThemeToggle`:
     - Fix theme toggle component to use correct tokens
+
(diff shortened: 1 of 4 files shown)


anambl (approved):
LGTM!

PR #2764, opened 2023-03-09
[KOA-5959] Automate Supernova docs publish

anambl:

- [ ] Component `README.md`
- [ ] Tests


scripts/check-readme-changes
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env node
```
olliecurtis, line 1: Don't forget to add the licence header under this line :)

olliecurtis (changes requested):

examples/bpk-component-datepicker/examples.js
```diff
@@ -429,128 +429,116 @@ class MultipleInputDatepicker extends Component {
 }
 
 const DefaultExample = () => (
-  <div id="application-element">
```
anambl, line 432: This is actually unrelated to this PR, but thought I'd fix these examples quickly as due to the fact that the id is `application-element`, we cannot render datepicker on ios due to the logic [here](https://github.com/Skyscanner/backpack/blob/main/packages/bpk-scrim-utils/src/withScrim.js#L73) which adds `display: none` to the application element.

.github/workflows/main.yml
```diff
@@ -49,6 +49,21 @@ jobs:
           external_repository: backpack/storybook
           publish_branch: main
 
+  SupernovaPublish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
```
anambl, line 57: this is needed in order to make the previous commit available

scripts/check-readme-changes
```diff
@@ -0,0 +1,36 @@
...
+    isChanged = true;
+    return;
+  }
+
+  isChanged = false;
+  return;
+};
+
+init().then(
+  isChanged => console.log(isChanged)
```
anambl, line 35: Github actions won't see the script output simply by returning it so we need to console log it

olliecurtis (approved):

.github/workflows/main.yml
```diff
@@ -49,6 +49,21 @@ jobs:
           external_repository: backpack/storybook
           publish_branch: main
 
+  SupernovaPublish:
```
marianeum, line 52: a few comments that are all somewhat related:
- looking at the script it just looks at the latest commit - if there's multiple commits how will that work? 
- we currently have concurrency enabled - ie previous runs will be cancelled when another push happens. 
- should this run as part of main or as part of release? otherwise we may publish things before they get released. i can see this working as part of release and _always_ publish (rather than check if any readmes were changed), rather than being on every commit to main

olliecurtis: I guess would we say that if we fix an issue with just the docs we need to publish a new release each time? 

That would be my thinking as to why we would run it on main each time and not on release as this means we would have emit a patch each time?

anambl: > looking at the script it just looks at the latest commit - if there's multiple commits how will that work?
It only runs on main so the latest commit should be a merge commit or all the commits combined into one commit so I think it should work

> we currently have concurrency enabled - ie previous runs will be cancelled when another push happens.
Again, this only runs on main so it shouldn't be a problem? 😄 unless I misunderstood your question 😅 

> should this run as part of main or as part of release? otherwise we may publish things before they get released.
I've thought about this, but I feel like there shouldn't be a long time between merging something to main and doing a release anyways. I don't have a very strong preference, but I'm thinking for example what if you only fix a README file? Or in the case of apps what if you update the screenshots? The readme will be then showing the latest screenshots, but the docs won't so there will always be a bit of an inconsistency either way?

marianeum: The release vs main is a tricky question as you both point out there's valid reasons for why we might want to run it on main - it's probably ok to cater for these cases, but we generally wait until after we publish to update docs at the moment to avoid confusion.

> latest commit should be a merge commit or all the commits combined into one commit
i don't remember if we enable all three options, but there's also the third rebase option where you'd have multiple commits. either way (and also relating to the currency issue) - i wonder if it's worthwhile doing the readme check or just publish. or maybe we just ALSO publish on releases to make sure we're definitely covered there.
the concurrency is also enabled on main - if you merge another PR this run will be stopped

anambl: I started writing this answer by saying I'm still slightly leaning towards running it on main, but by the time I finished writing all my thoughts I changed my mind 😄 

I think in the (kind of unlikely) event we only update a readme and a release is not needed, we could just simply do the publish manually from the Supernova UI.

> there's also the third rebase option

We don't have that enabled on this repo, but that's a good point. I'm actually not too sure if it's enabled on iOS/Android. But if we run this on release, it won't be a problem.

I think it makes sense to remove the readme check if we run this on release as it's likely we'd had made changes to the component that require a docs update.

anambl: We don't have an automated release on web yet (likely will be done next week or the week after). Might be worth doing this on web after that (creating a ticket and linking it to the release ticket), but we can enable it on iOS and Android now. What do you think?

marianeum: yep makes sense!

anambl:
Closing as will do this after setting up automated release

PR #2787, opened 2023-03-20
[BD-6775] Update BpkPrice previousPrice separator

Supremeyh:

- [ ] Component `README.md`
- [ ] Tests


olliecurtis (commented):
Would you be able to provide an example of this as it should already be following this correctly as seen in the following example which doesn't have a previous price: https://backpack.github.io/storybook/?path=/story/bpk-component-price--small-with-leading-and-trailing-text

Supremeyh:
> Would you be able to provide an example of this as it should already be following this correctly as seen in the following example which doesn't have a previous price: https://backpack.github.io/storybook/?path=/story/bpk-component-price--small-with-leading-and-trailing-text

Sure thing, when it has `leadingText` without `previousPrice`, no `leadingText` and `previousPrice`, or without `previousPrice`,  they should works fine both these three conditions. However, when it has `previousPrice` without `leadingText`, it would also has separator `·`. Since we may need line-through style, so i guess we may cannot use `leadingText` to replace `previousPrice`. Bellow is an example.
image

jasonhuang-sky (approved):

giriawu (approved):

packages/bpk-component-price/README.md
```diff
@@ -9,27 +9,31 @@ Check the main [Readme](https://github.com/skyscanner/backpack#usage) for a comp
...
   <BpkPrice
     size={SIZES.large}
     align={ALIGNS.left}
-    subtitle="£209"
-    title="£1,830"
-    description="a night"
+    price="£1,209"
+    previousPrice="£1,830"
+    leadingText="App only deal"
+    trailingText="a night"
```
Supremeyh, line 24: The usage in README is outdated, update it by the way.

olliecurtis (approved):

PR #2795, opened 2023-03-30
[KOA-5926]: Update BpkNavigationBar to align with Figma

olliecurtis:
PR to align with Figma, changes the default out of the box colour from being Blue to Surface Default (white). API wise this is not breaking but changes visually. However from our checking the blue is always overriden to be the Surface Default so shouldn't cause issues on the consumer side

Screenshot 2023-03-30 at 14 16 14

- [ ] Component `README.md`
- [x] Tests


arifhossain512 (approved):

frugoman (approved):
Looks good!

PR #2801, opened 2023-04-03
LOOM-631: Add forwardRef to BpkLink

runmoore:
Wrapping the component allows us to pass through a ref prop to a `<BpkLink>` component and have the ref attach to the `<a>` tag.

This allows us to expose the DOM node to the parent and is useful - for example, to allow focusing it.

Using `forwardRef` instead of creating a custom prop like `innerRef` as it gives us a predictable API instead of something custom which might change depending on the component.

- [ ] Component `README.md`
- [ ] Tests


mungodewar (approved):
LGTM, we'll need to include this in the component docs too.

olliecurtis (approved):

olliecurtis (approved):

PR #2803, opened 2023-04-04
[IRN-3778] Add BpkModalV2

ByAnthony:
# Why is the change being made?
As part of some successful experimentation we did in dBook platform, we want to introduce the Dialog Element to Backpack. This is a powerful solution for accessibility within a modal: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog;

- [ ] Component `README.md`
- [ ] Tests


olliecurtis (changes requested):
I would suggest rather than make a whole new component and have the version name in the import, we should just export it straight out of the bpk-component-dialog and have it just export `BpkDialogV2` this falls into similar how we have done it in the past and what we've done on our apps

PR #2808, opened 2023-04-10
[SEARCHEXP-1712] Stop propagation of events for Portal.

angel3cu:
Events used to bubble up from the portal into the parent component. Due to the way React works, events do not bubble up into the component in the DOM where the portal is rendered (i.e. render target), but instead they bubble up into the parent component in the shadow DOM, that is the component where you added the `Portal` in your React code. To avoid confusion and bugs caused by this behaviour, we have prevented mouse and keyboard events from bubbling up outside the portal.

- [x] Component `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-react-utils/README.md
+++ b/packages/bpk-react-utils/README.md
@@ -26,30 +26,28 @@ class MyComponent extends Component {
     this.state = {
       isOpen: false,
     };
-
   }
 
   onOpen = () => {
     this.setState({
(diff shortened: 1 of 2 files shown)


frugoman (dismissed):

angel3cu:
Updating the snapshots to fix the failed unit tests. 🙂

olliecurtis (approved):

PR #2810, opened 2023-04-13
[KOA-5053]: Transpilation of Backpack

olliecurtis:
The purpose of this PR is to move towards a transpiled Backpack and library. This PR does this.

The way in which it operates is the following:
- New transpilation commands have been added to `package.json`
  - "transpile" - the overall command for all the steps for transpilation
  - "transpile:clean" - removes old instances of the transpiled package for a clean start
  - "transpile:js" - this uses babel to transpile Backpack JS files into esmodules
  - "transform:js" - this updates the Sass file imports to use the compiled CSS imports
  - "transpile:generate-package-json" - copies the `package.json` from `packages/` for use in publishing
  - "transpile:copy-readme" - copies the `README` from `packages/` for use in publishing
  - "transpile:sass" - compiles Sass files into CSS
  - "copy-css" - copies the compiled CSS into the dist folder

- We also make use of `backpack-node-sass` for the purpose of compiling the Sass to CSS as it allows for parallel threads to speed up builds. (Will eventually use dart-sass).

**Note:** A follow PR will be made to change the build process in order to publish from the new `dist` folder that is created during transpilation


babel.config.js
```diff
@@ -42,8 +42,42 @@ module.exports = {
...
+        'inline-react-svg',
+      ],
+      ignore: [
+        (fileName) =>
+          !!(
+            fileName.includes('-test') ||
+            fileName.includes('-Test') ||
+            fileName.includes('dist') ||
+            fileName.includes('dist-storybook') ||
+            fileName.includes('examples') ||
```
anambl, line 76: we can just remove this as it doesn't seem to be working? or does it and we just need to remove them manually first?

anambl (approved):
LGTM!

PR #2815, opened 2023-04-18
[KOA-5173] Migrate aria live component to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests


.eslintrc
```diff
@@ -110,12 +110,7 @@
         "allowRequiredDefaults": true
       }
     ],
-    "react/require-default-props": [
-      "error",
-      {
-        "forbidDefaultForRequired": false
-      }
-    ],
+    "react/require-default-props": "off",
```
olliecurtis, line 113: We might not need this rule anymore at all as our linter handles this for us here: https://github.com/Skyscanner/eslint-config-skyscanner/blob/main/index.js#L44-L53

packages/bpk-component-aria-live/src/BpkAriaLive.tsx
```diff
@@ -29,18 +29,17 @@ export const POLITENESS_SETTINGS = {
...
-  children: ReactElement | string,
-  politenessSetting?: $Keys<typeof POLITENESS_SETTINGS>, // PolitenessSetting,
-  visible?: boolean,
-  className?: string | null,
-  [rest: string]: any, // Inexact rest. See decisions/inexact-rest.md
+  children: ReactElement | string;
+  politenessSetting?: PolitenessSetting;
+  visible?: boolean;
+  className?: string | null;
+  [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
```
olliecurtis, line 42: Wonder now we are using TS if we still need the comment here, as the Inexact rest was from when we were using flow and to avoid the need for flow to error every time

olliecurtis (changes requested):

packages/bpk-component-aria-live/src/BpkAriaLive.tsx
olliecurtis: Ignore me, I see this doc and comment was added exactly for TS!

.eslintrc
anambl: I had actually tried removing it and got a lot of lint errors so added it back 😢 
`forbidDefaultForRequired: true` we are actually using defaults for some required props.. not too sure why. But we can probably clean it up as we go and once all components have been migrated to TS we can remove it? 🤔

olliecurtis: Yeah let's just make sure we have a follow up task in the backlog to do that once we have migrated all :) and as we go through each component let's make sure no required props have any defaults to ensure we align with correct approaches

olliecurtis (approved):

PR #2819, opened 2023-04-24
[KOA-5198]: Migrate graphic promo to TS

olliecurtis:
Migrating graphic promo to typescript

- [ ] Component `README.md`
- [x] Tests


packages/bpk-component-graphic-promotion/index.d.ts
anambl, line 1: Is this right? or should it be just index.ts?

olliecurtis: Huh interesting on my local code it does have content in it and is needed for the types for using in other projects as without it fails to compile

anambl (approved):

PR #2824, opened 2023-04-24
[NO JIRA]: Ensure TS is included in transpiled package

olliecurtis:
This PR ensures that the typescript types files that we create as part of moving to TS are included in the transpiled package bundle.

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -71,7 +71,8 @@ module.exports = {
           !!(
             fileName.includes('-test') ||
             fileName.includes('-Test') ||
-            /\/node_modules\//.test(fileName)
+            /\/node_modules\//.test(fileName) ||
+            fileName.includes('.d.ts')
           ),
       ],
     },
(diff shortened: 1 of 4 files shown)


babel.config.js
```diff
@@ -71,7 +71,8 @@ module.exports = {
           !!(
             fileName.includes('-test') ||
             fileName.includes('-Test') ||
-            /\/node_modules\//.test(fileName)
+            /\/node_modules\//.test(fileName) ||
+            fileName.includes('.d.ts')
```
olliecurtis, line 75: We don't want to transpile the types files to TS as otherwise that defeats the point of having them

scripts/transpilation/copy-types.js
olliecurtis, line 1: There is probably a neater way to do this than a whole script but the babel cli is complex and we want to unblock PRs and start shipping so this will do for now and maybe even the permenant solution

anambl (approved):

PR #2830, opened 2023-04-28
[KOA-5539] Migrate input component to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-input/src/common-types.ts
anambl, line 1: There's an issue here - `PropsWithClearButtonMode` this should only be set if the `clearButtonMode` is not `never`

packages/bpk-component-input/src/BpkInput.tsx
```diff
@@ -70,7 +64,7 @@ class BpkInput extends Component<Props, State> {
     } = this.props;
 
     // Used as a ref for focussing the input when cleared.
-    let ref = null;
+    let ref: HTMLElement | null = null;
```
LouiseReid, line 67: Should this be HTMLInputElement?

LouiseReid: Which is why you then use the type assertion within there 👍

packages/bpk-component-input/src/BpkInput.tsx
```diff
@@ -156,12 +150,13 @@ class BpkInput extends Component<Props, State> {
...
             label={clearButtonLabel || ''}
             onMouseDown={onMouseDown}
-            onClick={(e) => {
+            onClick={(e: SyntheticEvent<HTMLButtonElement>) => {
               if (ref) {
                 ref.focus();
               }
               if (onClear) {
-                e.target.name = name;
+                const target = e.target as HTMLButtonElement;
```
LouiseReid, line 158: We've just recently stripped out all type assertions in banana. There's quite a few articles like this as to why you should avoid using them - https://netbasal.com/why-you-should-avoid-using-type-assertion-in-typescript-e44e8217cae5

In this instance could you do this instead?

```
              if (onClear) {
                if (e.target instanceof HTMLButtonElement) {
                  const { target } = e;
                  target.name = name;
                }
                onClear(e);
                this.setState({ persistClearButton: false });
              }

File: packages/bpk-component-input/src/BpkClearButton.tsx [hunk 3]
```diff
@@ -30,14 +31,19 @@ const getClassName = cssModules(STYLES);
 const ClearButtonIcon = withButtonAlignment(ClearIcon);
 
 type Props = {
```
LexSwed, line 33: I would offer 
```suggestion
interface Props extends ComponentProps<'button'> {
```
since you allow to pass through `...rest` props. Then, you don't need to manually specify `onClick`, `className` (and leave it `undefined`, without defaulting to `null`)

LexSwed: No strong opinion for `type` vs `interface` for me, `banana` is happy with `type`, but there are some [hints here](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/basic_type_example/#types-or-interfaces). 

For me, in component libraries, you often need to extend and override some types. With `type`, TypeScript will silently merge the types in whatever way it thinks is best, with `interface` it will complain if handling of the same properties is not clear. [Short example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UDeAoK6qkgLigcgCMBXYYAewDs8UBfFFLaAIQWTXUdwGdgAnASwoBzANy16jKAGFWLAGSwxKAMaUeUYqUoAFXmTBdcMxKgyZwEXHgAmXAIa271OikHAIvAGZ3l0GATYzTnxNcipxVwp3Lx9mANMOC24+QVEItw9vX2kAiAAPdwpbWAIAGigmeJdVCnVQnT0DJiMAk3ZEnHxHR2d6IA)

packages/bpk-component-input/src/BpkClearButton.tsx
```diff
@@ -30,14 +31,19 @@ const getClassName = cssModules(STYLES);
...
-  className: ?string,
+  label: string;
+  onClick: (event: SyntheticEvent<HTMLButtonElement>) => void | null;
+  className?: string | null;
+  [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
 };
 
-const BpkClearButton = (props: Props) => {
+const BpkClearButton = ({
+  className = null,
```
LexSwed, line 41: I understand it's there for historical reasons (default value for optional prop), but now it should not required, so can be avoided to save some bites? 👀

packages/bpk-component-input/src/BpkInput.tsx
anambl: Ah interesting, the point stated in the article makes sense.
I think in this case though it won't really make a difference as the event target will always be an HTMLButtonElement and neither the type nor the actual target are something we can change?

packages/bpk-component-input/src/BpkClearButton.tsx
anambl: Good point and thanks for the example, yeah I think that makes sense - will change it to an interface!

packages/bpk-component-input/src/BpkInput.tsx
anambl: I was completely wrong, it's not actually used for both 😅 changed it to HTMLInputElement!

olliecurtis (approved):

PR #2837, opened 2023-05-09
[KOA-5945]: Button v2 rewrite

olliecurtis:
We have wanted to rewrite this component for some time to improve the API and align this with our mobile platforms, so we have finally done it.

In order to allow for a smoother transition and not one large breaking change, we have created a `BpkButtonV2` (written in TS) in order for folks to migrate across to a new button.

We have created a migration guide for consumers to understand the changes but the main changes we made are:

- Moved the button types from a boolean for each variant to an enum of `BUTTON_TYPES`
- Moved the `large` property to use an enum of sizes `SIZE_TYPES` for the new `size` property.

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-button/README.md
```diff
@@ -11,20 +11,20 @@ Check the main [Readme](https://github.com/skyscanner/backpack#usage) for a comp
 ```js
 import { withButtonAlignment, withRtlSupport } from '@skyscanner/backpack-web/bpk-component-icon';
 import ArrowIcon from '@skyscanner/backpack-web/bpk-component-icon/sm/long-arrow-right';
-import BpkButton from '@skyscanner/backpack-web/bpk-component-button';
+import BpkButtonV2, { BUTTON_TYPES, SIZE_TYPES } from '@skyscanner/backpack-web/bpk-component-button';
```
anambl, line 14: I guess this should be a named import as for now it won't be the default 🤔

packages/bpk-component-button/src/BpkButtonV2/common-types.tsx
```diff
@@ -0,0 +1,54 @@
...
+  link: 'link',
+  linkOnDark: 'link-on-dark',
+} as const;
+
+export const SIZE_TYPES = {
+  small: 'small',
+  large: 'large',
+} as const;
+
+export type ButtonType = typeof BUTTON_TYPES[keyof typeof BUTTON_TYPES];
```
anambl, line 38: I think this would actually have to be the values, not the keys of BUTTON_TYPES as the class names are using hyphens

packages/bpk-component-button/docs/migrating-from-v1-to-v2.md
```diff
@@ -31,48 +27,32 @@ class Toggle extends React.Component {
 ### Replacement:
 
 ```
-import BpkCheckbox from 'bpk-component-checkbox';
+import BpkButtonV2, { BUTTON_TYPES } from 'bpk-component-button';
```
anambl, line 30: Same here - would need to be a named import?

packages/bpk-component-button/src/BpkButtonV2/BpkButtonBase.module.scss
anambl, line 1: We could just rename this file to `BpkButton`?

anambl (commented):
🎉 🎉 This is so much better!! no longer 50 files for one component! 😄 have we also looked at the loading button component and wether we could merge it into the main button component? Could also be something we do separately next week.

packages/bpk-component-button/src/BpkButtonV2/common-types.tsx
olliecurtis: Good point this was just me copying from the Badges

olliecurtis: I did a check of this in both Banana and online, this is correct for getting the values to be expected for the type :)

anambl (approved):
LGTM!

PR #2838, opened 2023-05-10
Fix BpkDialog type

runmoore:
### getApplicationElement

`getApplicationElement` is designed per the README to be used with `getElementById`.

`getElementById` returns null if the element can't be found

image

### dialogRef

`dialogRef` is an internal prop that isn't part of the public API, but is listed as required.

image

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-dialog/src/common-types.ts
+++ b/packages/bpk-component-dialog/src/common-types.ts
@@ -28,8 +28,8 @@ export type DialogInnerProps = {
   ariaLabel: string;
   id: string;
   children: ReactNode;
-  dialogRef: (ref: HTMLElement) => void;
-  getApplicationElement: () => HTMLElement;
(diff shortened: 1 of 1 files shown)


packages/bpk-component-dialog/src/common-types.ts
```diff
@@ -28,8 +28,8 @@ export type DialogInnerProps = {
   ariaLabel: string;
   id: string;
   children: ReactNode;
-  dialogRef: (ref: HTMLElement) => void;
-  getApplicationElement: () => HTMLElement;
+  dialogRef?: (ref: HTMLElement) => void;
```
runmoore, line 31: having this optional isn't a great solution as it's part of an internal API but exposed publicly.

But without changing all the type structure here there isn't much better option

anambl (approved):

PR #2839, opened 2023-05-11
[BD-6906] Support mapId

viktor-yang:
To support vector map and custom style on the google cloud platform, we need to support the mapId options on the Backpack map

[MapOptions](https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions) interface
mapId doc: https://developers.google.com/maps/documentation/get-map-id

- [x] Component `README.md`
- [ ] Tests


packages/bpk-component-map/src/BpkMap.js
```diff
@@ -131,6 +132,8 @@ const BpkMap = (props: Props) => {
       }
       zoom={zoom}
       options={{
+        // If you want to set style by mapId, you should remove the styles property: https://developers.google.com/maps/documentation/get-map-id#javascript
+        mapId,
```
anambl, line 136: Is it supposed to be part of the `options` object prop? Looking at their docs, it seems like it's passed directly as a prop to the Map component - https://developers.google.com/maps/documentation/get-map-id#add-a-map-id-to-your-app

viktor-yang: Yeah, but that is the usage for the native while we use it by the @react-google-maps/api package at https://github.com/Skyscanner/backpack/pull/2839/files#diff-6d96123f2f0a2023e2b9304718d92a28a46810359b1b73bc54bf2500b4113c29R24

And we can its types  as  at google.maps.MapOptions 
https://github.com/JustFly1984/react-google-maps-api/blob/develop/packages/react-google-maps-api/src/GoogleMap.tsx#L72

And we can see the google.maps.MapOptions interface at https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions

And I have tested it on the development, it works fine.

anambl (approved):
LGTM!

PR #2845, opened 2023-05-17
[KOA-5176] Migrate banner alert to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests


.eslintrc
```diff
@@ -21,6 +21,7 @@
       }
     }
   },
+  "ignorePatterns": ["packages/**/*.d.ts"],
```
olliecurtis, line 24: We won't need this anymore because of the `eslintignore` change in #2857 right?

packages/bpk-component-banner-alert/src/BpkBannerAlert.tsx
```diff
@@ -15,21 +15,19 @@
...
+  icon: null,
 };
 
-const BpkBannerAlert = (props: Props) => <BpkBannerAlertInner {...props} />;
-
-BpkBannerAlert.propTypes = COMMON_PROP_TYPES;
-BpkBannerAlert.defaultProps = COMMON_DEFAULT_PROPS;
+const BpkBannerAlert = (props: CommonProps) => (
+  <BpkBannerAlertInner {...defaultProps} {...props} />
+);
```
olliecurtis, line 31: Should we be passing through the `defaultProps` here instead of defining them as we have in other components as part of the props: CommonProps definition?

Only reason I question is because I have a worry that due to some weirdness that React could do at runtime if someone looks to override these default values it could accidently be they get applied after the consumer `props` values so causing a potential bug of being applied in the wrong order and not giving the expected outcome

.eslintrc
anambl: yep! I'll remove that!!

packages/bpk-component-banner-alert/src/BpkBannerAlert.tsx
anambl: oh interesting, is that like a known bug or something? the order in which props are passed should definitely matter. I only wrote it this way because didn't want to have to pass each prop individually given it's just prop drilling and nothing in addition

olliecurtis (approved):

PR #2846, opened 2023-05-18
[NO JIRA] Fix transpilation scripts to move css and types files

anambl:

- [ ] Component `README.md`
- [ ] Tests


gert-janvercauteren (approved):
Wondering if we can find a more stable way of doing this in the (near) future.
For now it's better to unblock people upgrading to the latest version of Backpack!

Awesome job making this fix 💪

anambl:
> Wondering if we can find a more stable way of doing this in the (near) future. For now it's better to unblock people upgrading to the latest version of Backpack!
> 
> Awesome job making this fix 💪

Yes, definitely! I'm setting up a reminder to chat with @olliecurtis about this when he's back as I want to also confirm this is the structure we want for any new V2 components. For now this will be ok as we don't have any other "nested" components and if we were to add any new ones, people would have hopefully read the guide so it will have the same structure 😄

PR #2848, opened 2023-05-22
[KOA-5186] Migrate chip component to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-chip/src/BpkDismissibleChip.tsx
```diff
@@ -16,31 +16,34 @@
...
 
-const BpkDismissibleChip = (props: Props) => {
-  const { className, type, ...rest } = props;
+const BpkDismissibleChip = ({
+  className,
+  type = CHIP_TYPES.default,
+  leadingAccessoryView = null,
+  ...rest
+}: Props) => {
+  console.log(type);
```
olliecurtis, line 45: Should this be here?

packages/bpk-component-chip/src/BpkDismissibleChip.tsx
```diff
@@ -16,31 +16,34 @@
...
-  COMMON_PROP_TYPES,
-  COMMON_DEFAULT_PROPS,
-  type CommonProps,
-} from './commonTypes';
+import type { CommonProps } from './commonTypes';
+import { CHIP_TYPES, COMMON_DEFAULT_PROPS } from './commonTypes';
 import STYLES from './BpkSelectableChip.module.scss';
 
-export type Props = CommonProps;
+export interface Props extends CommonProps {}
```
olliecurtis, line 35: Wondering if we need this given it looks like we don't add anything additional to the `Props`?

So it could just be we don't need to export from here and also below we can just replace `Props` with `CommonProps`

packages/bpk-component-chip/src/BpkDropdownChip.tsx
```diff
@@ -16,36 +16,35 @@
...
-    />
-  );
-};
-
-BpkDropdownChip.propTypes = COMMON_PROP_TYPES;
-BpkDropdownChip.defaultProps = COMMON_DEFAULT_PROPS;
+import type { CommonProps } from './commonTypes';
+import { CHIP_TYPES } from './commonTypes';
+
+export interface Props extends CommonProps {}
```
olliecurtis, line 31: Same thing here as above it doesn't seem relevant/required given we don't add to it

packages/bpk-component-chip/src/commonTypes.ts
```diff
@@ -16,28 +16,29 @@
...
-  type: $Values<typeof CHIP_TYPES>,
-  className: ?string,
-  leadingAccessoryView: ?Node,
-};
+// onClick and children are already part of the button props, but we need to specify that they are required
+interface ButtonProps extends Omit<ComponentProps<'button'>, 'type'> {
+  onClick: (event: SyntheticEvent<HTMLButtonElement>) => void | null;
+  children: ReactNode | string;
+  type?: typeof CHIP_TYPES[keyof typeof CHIP_TYPES]; // this is different from the native button type
+}
```
olliecurtis, line 33: How come we need to declare this interface and do this instead of just having these inside the `CommonProps`?

packages/bpk-component-chip/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import BpkSelectableChip, {
+  type Props as SelectableProps,
+} from './src/BpkSelectableChip';
+import BpkDismissibleChip, {
+  type Props as DismissibleProps,
```
olliecurtis, line 23: Based on the comments I left around not further enhacning the existing props inside this could then be omitted from here and we could just import from the `commonTypes` file the `CommonProps` and set them down below on the types

packages/bpk-component-chip/src/BpkDismissibleChip.tsx
anambl: oh God.. good catch! left it accidentally while I was checking something

packages/bpk-component-chip/src/BpkDismissibleChip.tsx
anambl: Hmm alright, good idea!

olliecurtis (approved):

PR #2854, opened 2023-05-24
[SEARCHEXP-1394] New AutoSuggest Component

LouiseReid:
### Why has this change been made?

We've been running with a different [base autosuggest library](https://www.downshift-js.com/use-combobox/) in banana since the new homepage launched and would like to add this in backpack so we can:
- install the component from backpack
- share the more accessible component with the rest of skyscanner

### What are the changes?
The new component has been implemented as a V2 with a migration guide. It has been made to match the existing implementation as close as possible. Some small changes have been made though which are breaking changes.

- it requires a ref to be passed into it instead of passing the ref back out via a callback. BpkAutosuggest is now a forwardRef as is BpkInput with the ref being appended via downshifts `getInputProps` function. 
- some signature differences. Now we're not passing pack out the event object the signature of `onSuggestionSelected` has changed to reflect this. `onSuggestionFetchRequested` has also changed to just take a value rather than an object containing the value
- consumers can no longer pass an `onChange` or `value` down with inputProps due to how the component controls its own internal state. Consumers can use the `onInputValueChange` prop which will return the new value and type of change for any logging required from this interaction. A `defaultValue` prop can be used to pass down the initial on load value for the input
- we'll now have greater control over the markup due to the hook just passing out the aria props whilst the skeleton of the component is constructed by us
- new a11y props - `getA11yResultsMessage`. We implement it in banana like this and other microsites could use the same translation strings. 

```
    const getA11yResultsMessage = (resultCount: number) => {
      if (fetchingState === 'resolved') {
        if (resultCount > 0) {
          return translatePlural(
            isDesktop
              ? translate('aria_autosuggest_results_single')
              : translate('aria_autosuggest_results_mobile_single'),
            isDesktop
              ? translate('aria_autosuggest_results_plural', { number: resultCount })
              : translate('aria_autosuggest_results_mobile_plural', { number: resultCount }),
            resultCount,
          );
        }
        return translate('aria_autosuggest_results_no-results');
      }
      return '';
    };
```
- We also require an `ariaLabels` prop. Only banana will pass through labels for label and clearButton but all should pass through a value for resultsList (axe test fails without this value)
- updates all autosuggest files to typescript

### Summary of all props and what they are used for
| Prop                           | Description                                                    |Banana specific                                                  |
|--------------------------------|----------------------------------------------------------------| -------------
| `ariaLabels`                   | Object containing various ARIA labels for accessibility        |
| `getSuggestionValue`           | Function that returns the display value for a suggestion        |
| `id`                           | ID for the autosuggest component                               |
| `inputProps`                   | Props for the HTML input element                                |
| `onSuggestionsClearRequested`  | Callback function called when suggestions are cleared           |
| `onSuggestionSelected`         | Callback function called when a suggestion is selected          |
| `onSuggestionsFetchRequested`  | Callback function called when suggestions are fetched           |
| `renderSuggestion`             | Function that renders a suggestion item                         |
| `suggestions`                  | Array of suggestions or sections                                          |
| `defaultValue`                 | Default value for the input field                              |
| `enterKeyHint`                 | String for mobile keyboard enter key                     |    x
| `getA11yResultsMessage`        | Function that returns an accessible message for the results     |
| `withLabel`                     | Boolean indicating if input is to be rendered wrapped in a label                         | x
| `isDesktop`                    | Boolean indicating if it is running on a desktop environment    | x
| `onClick`                      | Callback function called when the input is clicked         | x - we use this in banana for destination quicklinks rendering
| `onLoad`                       | Callback function called when the component is loaded           |
| `renderBesideInput`            | Function that renders content beside the input field            | x - for rendering different labels
| `showClear`                    | Boolean indicating if the clear button should be shown          | x
| `theme`                        | Object containing theme customisation for the component        |
| `highlightFirstSuggestion`     | Boolean indicating if the first suggestion should be highlighted|
| `shouldRenderSuggestions`      | Function that determines if suggestions should be rendered      |
| `multiSection`                 | Boolean indicating if suggestions are grouped into sections     |
| `getSectionSuggestions`        | Function that returns an array of suggestions for a section     |
| `renderSectionTitle`           | Function that renders a title for a section of suggestions      |
| `alwaysRenderSuggestions`      | Boolean indicating if suggestions should always be rendered when input loses focus    |
| `onInputValueChange`           | Callback function called when the input value changes           |

### Testing
This component has been tested by updating the component within the node_modules within banana and flight search controls.

- Draft PR of changes required within FSC to use the new component - https://github.skyscannertools.net/web/flights-search-controls/pull/858

- [x] Component `README.md`
- [x] Tests


packages/bpk-component-autosuggest/src/BpkAutosuggest.module.scss
```diff
@@ -153,4 +153,14 @@ $arrow-size: bpk-spacing-sm() * 3;
       @include bpk-caption;
     }
   }
+
+  &__section-title {
+    padding: bpk-spacing-lg() 0 $bpk-spacing-xs $bpk-spacing-md;
```
anambl, line 158: I think `bpk-spacing-xs` is deprecated, also we use the spacing functions so that it uses the new grid, can you replace it with whatever is the equivalent of these 2 tokens?

packages/bpk-component-autosuggest/src/BpkAutosuggest.module.scss
```diff
@@ -153,4 +153,14 @@ $arrow-size: bpk-spacing-sm() * 3;
...
+    background-color: $bpk-color-white;
+    color: $bpk-text-secondary-day;
+
+    @include bpk-text;
+    @include bpk-text-sm;
```
anambl, line 163: `bpk-text--bold` is deprecated, can you use instead the typography style equivalent to sm + bold?

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.tsx
```diff
@@ -16,31 +16,34 @@
...
+import type { ReactNode, FunctionComponent, SVGProps } from 'react';
 
+// @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
 import { cssModules } from '../../bpk-react-utils';
 
 import STYLES from './BpkAutosuggest.module.scss';
 
 const getClassName = cssModules(STYLES);
 
 type Props = {
```
anambl, line 28: I think we'd need to add support for `rest` to this type or extend the section html element interface (if there is one)?

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.tsx
```diff
@@ -16,31 +16,34 @@
...
+  subHeading?: ReactNode;
+  tertiaryLabel?: string;
+  icon?: FunctionComponent<SVGProps<SVGSVGElement>>;
+  indent?: boolean;
+  className?: string;
 };
 
-const BpkSuggestion = (props: Props) => {
+const BpkSuggestion: FunctionComponent<Props> = ({
+  className,
```
anambl, line 38: also noticed default values are missing, is that intentional?

anambl (commented):
Only other thing, related to what we discussed on Slack, would be good to move this into a BpkAutosuggestV2 component. I think for BpkInput the only difference is the ref so it's not a big enough change to justify creating a new component, but for autosuggest it seems like it's quite a big change both from a dev perspective as well as consumer 🤔 And would also be useful to add a migration guide inside a docs folder in the bpk-component-autosuggest folder

anambl:
Left a comment in one of the components, but I saw the default props are missing in all components as well as the rest props in the defined type

packages/bpk-component-autosuggest/src/BpkAutosuggest.tsx
```diff
@@ -0,0 +1,408 @@
...
+      }
+
+      if (!isOpen && inputValue.length) {
+        onSuggestionsFetchRequested(inputValue);
+        openMenu();
+      } else {
+        onClick?.();
+      }
+
+      // Desktop destination autosuggest lives on the homepage and is "loaded/interacted with" via clicking on it
```
anambl, line 239: Not sure I understand this, might be useful to have a chat separately, but I'm also thinking whether we could use something else other than `isBanana` for this boolean. Why is there a special behaviour for when the component is render in banana?

packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.tsx
LouiseReid: Yea its intentional. In the js file they are all nulls and false so passing the default values has no purpose.

```
BpkSuggestion.defaultProps = {
  subHeading: null,
  tertiaryLabel: null,
  icon: null,
  indent: false,
  className: null,
};
```
With typescript for the optional props setting defaultProps as non values and not passing any values equates to the same thing.

I think we would only need this if the default prop had a value, like we're doing in BpkAutosuggest with theme

packages/bpk-component-autosuggest/src/BpkAutosuggest.tsx
LouiseReid: Yea you're right the `isBanana` flag use was wrong here. I'll add a couple of tests to help with the understanding.

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
```diff
@@ -0,0 +1,406 @@
...
+  suggestionsList: getClassName('bpk-autosuggest__suggestions-list'),
+  suggestion: getClassName('bpk-autosuggest__suggestion-item'),
+  suggestionHighlighted: getClassName(
+    'bpk-autosuggest__suggestion-item--highlighted',
+  ),
+  sectionContainer: getClassName('bpk-autosuggest__section-container'),
+  sectionTitle: getClassName('bpk-autosuggest__section-title'),
+};
+
+const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
```
anambl, line 115: Just to check as there's a lot of new props - were these all (except for those mentioned in the migration guide already) already supported by the `react-autosuggest` library?

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
```diff
@@ -0,0 +1,406 @@
...
+          >
+            {showSuggestions && renderList()}
+          </ul>
+        </div>
+      </div>
+    );
+  },
+);
+
+BpkAutosuggest.defaultProps = {
```
anambl, line 402: we've stopped using this in BPK as it will be deprecated for functional components, you can instead destructure the props and set the default values using arguments?

packages/bpk-component-autosuggestV2/docs/migrating-from-v1-to-v2.md
```diff
@@ -0,0 +1,101 @@
...
+  ref={this.autosuggestRef}>
+
+```
+* aria strings are required:
+    * `getA11yResultsMessage` is a callback which returns the number of results being rendered. This should be used to pass translations for single, multiple and no results returned
+    * `ariaLabels` requires a `resultsList` string which will be appended to the `<ul>`
+
+* `renderInputComponent` is no longer a prop. The component renders a BpkInput component, any extension or modification of this can be done via the `inputProps` prop which is passed into the input or by extending or overriding the styling
+
+**Extending**
```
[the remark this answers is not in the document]
LouiseReid: Extending and overriding the themes. I'll make that clearer and also add another before and replacement section for how to handle no  `renderInputComponent` prop

packages/bpk-component-autosuggestV2/docs/migrating-from-v1-to-v2.md
```diff
@@ -0,0 +1,101 @@
...
+this.autosuggestRef = React.createRef<HTMLInputElement>();
+<BpkAutosuggest
+  ref={this.autosuggestRef}>
+
+```
+* aria strings are required:
+    * `getA11yResultsMessage` is a callback which returns the number of results being rendered. This should be used to pass translations for single, multiple and no results returned
+    * `ariaLabels` requires a `resultsList` string which will be appended to the `<ul>`
+
+* `renderInputComponent` is no longer a prop. The component renders a BpkInput component, any extension or modification of this can be done via the `inputProps` prop which is passed into the input or by extending or overriding the styling
```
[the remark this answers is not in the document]
LouiseReid: Its used within flights search controls. I'd looked at it as part of this draft pr to test the new component which is where I found we could do away with it - https://github.skyscannertools.net/web/flights-search-controls/pull/858

packages/bpk-component-autosuggestV2/docs/migrating-from-v1-to-v2.md
```diff
@@ -0,0 +1,101 @@
...
+
+**Extending**
+
+```js
+this.theme = {
+  ...BpkAutosuggest.defaultProps.theme,
+  input: `${BpkAutosuggest.defaultProps.theme.input} ${STYLES['fsc-location-input']} `,
```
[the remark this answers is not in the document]
LouiseReid: This is also used throughout the microsites quite a bit - https://github.skyscannertools.net/search?q=BpkAutosuggest.defaultProps.theme&type=code

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
```diff
@@ -0,0 +1,406 @@
...
+      renderSectionTitle,
+      renderSuggestion,
+      shouldRenderSuggestions,
+      showClear,
+      suggestions,
+      theme: customTheme,
+    },
+    forwardedRef,
+  ) => {
+    const theme = { ...defaultTheme, ...customTheme };
```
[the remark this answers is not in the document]
LouiseReid: We need this for bananas custom theming but also to support [these use cases](https://github.com/Skyscanner/backpack/pull/2854#discussion_r1210310436)

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
LouiseReid: Yea. I went through all the main BpkAutosuggest implementations in skyscanner to find what props were used and passed through.

Slingshot uses alwaysRenderSuggestions - this keeps the list open when the input looses focus.

Carhire and hotels use highlightFirstSuggestion, shouldRenderSuggestions, multiSection and the getters associated

The banana needs are marked up in the description

packages/bpk-component-autosuggestV2/docs/migrating-from-v1-to-v2.md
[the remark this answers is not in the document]
anambl: Even if it is used, it should not have any effect as we set the `renderInputComponent` in BPK and you cannot override it - https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-autosuggest/src/BpkAutosuggest.js#L59

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
LouiseReid: We need this here for the consumers to override the theming [here](https://github.com/Skyscanner/backpack/pull/2854#discussion_r1210310436)

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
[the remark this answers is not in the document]
anambl: We shouldn't need any custom theming for banana (or any other project) unless any BPK designer has a agreed to it beforehand. Supporting custom theming like this would open too many possibilities for consumers to override the BPK design (with the assumption that we currently do not support this in the autosuggest component)
We do have custom theming in some components, but that was for different use cases i.e. partner branding and brand refresh

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
anambl: amazing! I don't like that some of our bpk components refer to an external lib's docs because that opens the API WAY too much

packages/bpk-component-input/src/BpkInput.tsx
```diff
@@ -149,16 +137,16 @@ class BpkInput extends Component<Props, State> {
             label={clearButtonLabel || ''}
             onMouseDown={onMouseDown}
             onClick={(e) => {
-              if (ref) {
-                ref.focus();
-              }
               if (onClear) {
+                if (ref) {
```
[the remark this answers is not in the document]
anambl: I meant that the ref focus was moved inside the if. So previously it would still be focused even if there was no `onClear`, but yeah you're right, this should be good!

packages/bpk-component-autosuggestV2/docs/migrating-from-v1-to-v2.md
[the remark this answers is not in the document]
LouiseReid: Exposing the default props was allowing users to override it. I can remove this though. I only added it when I found flights search controls was taking the default suggestionsList styling and adding to it - https://github.skyscannertools.net/web/flights-search-controls/blob/main/src/components/LocationSelector/LocationSelector.tsx#L299

packages/bpk-component-autosuggestV2/docs/migrating-from-v1-to-v2.md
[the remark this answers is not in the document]
LouiseReid: I think given it is being used a migration guide away from using it is needed. Its not a prop on the new component so it can't be used in future but we don't want to block migrations by a consumer not knowing what to do with this now non existent prop

packages/bpk-component-autosuggestV2/docs/migrating-from-v1-to-v2.md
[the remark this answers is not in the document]
anambl: Oh you're right, I didn't notice the `theme` prop hidden there at the bottom of the API table - https://github.com/moroshko/react-autosuggest#theme-prop 😞 no, you're totally right to have added given it's being used, I just didn't realise you could override this because I didn't see the prop in the table. That's why I hate opening the API like this 🙃
I do still think we should remove this though given it's only 6 projects using it and we don't want people to have custom designs so we can add this in the migration guide as removed!

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
anambl: I meant the defaultProps in functional components, but yes also the theming 🙂

examples/bpk-component-autosuggestV2/examples.js
```diff
@@ -0,0 +1,243 @@
...
+  value: string,
+  suggestions: Array<any>,
+};
+
+type Props = {
+  hanzi: boolean,
+  includeIcon: boolean,
+  includeSubheading: boolean,
+  includeTertiaryLabel: boolean,
+  isBanana: boolean,
```
olliecurtis, line 141: We can't have a business specific property in our OSS component so this will need to be removed.

This also doesn't align with Backpack standards as this now makes it business logic and specific usecase which neither of these are to be present in Backpack due to the nature and reusability of components

olliecurtis (changes requested):

packages/bpk-component-autosuggestV2/src/BpkAutosuggest.tsx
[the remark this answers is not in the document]
anambl: That's interesting as the focus ring for me is blue, I think it depends on the browser?
Anyways, the idea would be that we should align the designs and not need a custom design for banana. As I said, it's ok to have different designs between mobile and desktop, but not between banana or any other specific project and bpk.
So we should probably be going back to the BPK designers and your designer to make sure the designs are aligned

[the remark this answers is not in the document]
olliecurtis: >focus ring is a different colour with larger border radius

This is something we are already working on it Backpack to add better focus indicators to our components across the business as part of A11y

>label lives inside input container

It's painful that currently Adam the Backpack designer is on vacation right now as I rememeber there are conversations ongoing about the inputs right now and how they should be.

Why does this have to be through a theme rather than configuration to the `BpkInput` component that can then be passed through.

>clear button is bigger

Similar to the above this doesn't align with our current components in Figma for the Backpack library so we need to check and sort this in design before we can just drop this into the component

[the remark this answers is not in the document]
LouiseReid: > Why does this have to be through a theme rather than configuration to the BpkInput component that can then be passed through.

The alternative is banana passing through its own element to render in place of the default BpkInput. It was found it was quite messy as it was reliant on useCombobox attributes being passed back out to banana. The theme and isBanana props felt the cleanest way.

I can go back to the drawing board and try something different though.

anambl:
Hey one more thing - could you move the bpk-autosuggest-v2 folder inside the main autosuggest one so that it can be imported from the same place? You can have a look at the button component to see the folder structure

LouiseReid:
I'll close this PR for the time being.

[Decision doc](https://skyscanner.atlassian.net/wiki/spaces/SEARCHEXPTRIBE/pages/619464489/BpkAutosuggestV2+decision+doc) on how we move forward with it and the discussions in this thread point heavily towards a design realign.

We can pick this back up, or this can be used as an example of how to use downshift in the future when we're at a point where all consumers style and required suggestions list styles are in figma and handled by backpack

PR #2858, opened 2023-05-25
[KOA-6117] Migrate scrim utils to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-scrim-utils/src/scroll-utils.tsx
```diff
@@ -31,7 +31,7 @@ const getScrollBarWidth = () => {
   const window = getWindow();
   const body = getBodyElement();
 
-  if (body === null && window === null) {
+  if (body === null || window === null) {
```
olliecurtis, line 34: Could you confirm why we have needed to make this change here to the logic?
As this seems a significant change to the statement which could impact consumers/usage of the scrim utils

packages/bpk-scrim-utils/src/withScrim.tsx
```diff
@@ -138,9 +144,9 @@ const withScrim = (WrappedComponent) => {
 
       return (
         <div className={classNames.join(' ')}>
-          <BpkScrim onClose={closeOnScrimClick ? onClose : null} />
+          <BpkScrim onClose={closeOnScrimClick ? onClose : undefined} />
           <WrappedComponent
-            {...rest}
+            {...(rest as P)}
```
olliecurtis, line 149: Should we be defining `rest` within our `Props` object or is this due to it being a class component that it is done this way?

olliecurtis (changes requested):
Do we need to add the `.d.ts`files for this component?

packages/bpk-scrim-utils/src/withScrim.tsx
anambl: this is something specific to hocs actually, and I believe the reason you need to cast it like this is actually due to a bug in TS - https://github.com/Microsoft/TypeScript/issues/28938
But I did notice I believe a mistake here so let me fix that quickly 😬

olliecurtis (approved):

PR #2867, opened 2023-06-01
[KOA-6115] Migrate modal to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-modal/index.ts
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { BpkModalV2 } from './src/BpkModalV2/BpkModal';
+import themeAttributes from './src/themeAttributes';
+import BpkModal from './src/BpkModal';
+import type { Props } from './src/BpkModal';
+// @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
+import { propTypes, defaultProps } from './src/legacy-prop-types';
```
olliecurtis, line 24: Just checking why are we keeping the legacy props? 

Am I right in assuming this is because there is another component using these right now?

packages/bpk-component-modal/src/BpkModalInner.tsx
```diff
@@ -91,14 +104,14 @@ const BpkModalInner = (props: Props) => {
       transitionTimeout={300}
     >
       <section
-        id={props.id}
-        tabIndex="-1"
+        id={rest.id}
```
olliecurtis, line 107: Do we not need to define this above instead of pulling it out of `rest`? Or is this how required props are supposed to work?

packages/bpk-component-modal/src/BpkModalInner.tsx
```diff
@@ -91,14 +104,14 @@ const BpkModalInner = (props: Props) => {
...
+        tabIndex={-1}
         role="dialog"
-        aria-labelledby={props.showHeader ? headingId : null}
+        aria-labelledby={showHeader ? headingId : undefined}
         className={classNames.join(' ')}
-        ref={props.dialogRef}
+        ref={rest.dialogRef}
```
olliecurtis, line 112: Do we not need to define this above instead of pulling it out of `rest`? Or is this how required props are supposed to work?

packages/bpk-component-modal/src/BpkModalInner.tsx
```diff
@@ -108,71 +121,33 @@ const BpkModalInner = (props: Props) => {
...
+                    label={closeLabel}
+                    onClick={onClose}
                   />
                 )
               }
             />
           </header>
         )}
-        <div className={contentClassNames.join(' ')}>{props.children}</div>
+        <div className={contentClassNames.join(' ')}>{rest.children}</div>
```
olliecurtis, line 147: Do we not need to define this above instead of pulling it out of `rest`? Or is this how required props are supposed to work?

packages/bpk-component-modal/index.ts
anambl: It's actually because we export them and they're also used by our examples and didn't want to make this into a breaking change. There's a few components where we export the prop types and default props, but I thought we could remove them once we migrate our examples?

packages/bpk-component-modal/src/BpkModalInner.tsx
anambl: no, we can pull them out that's ok!

olliecurtis (approved):

ByAnthony:
Hi @anambl 👋 It looks like the title is broken on the modal. I had a quick look but can't figure it the root cause:

Screenshot 2023-06-07 at 21 42 33

PR #2869, opened 2023-06-02
[NO-JIRA] Update BpkModalV2

ByAnthony:
# Why has this change been made?

Following this [PR](https://github.com/Skyscanner/backpack/pull/2809), we introduced a new version to bpk-modal-component. After testing it in our frontend service, we have come up with a few updates. 

# What changed?

- Introduce a `showHeader` prop: this gives the option to hide the entire header, including close button.
- Change background color of the modal to the old one.

# Checklist

- [x] Component `README.md`
- [x] Tests


packages/bpk-component-modal/src/BpkModalV2/BpKModal-test.tsx
```diff
@@ -91,15 +91,37 @@ describe('BpkModalV2', () => {
...
+        <BpkModalV2 {...props}>
           <div>Content</div>
         </BpkModalV2>,
       );
 
-      expect(screen.queryByRole('Dialog Element')).not.toBeInTheDocument();
-      expect(document.getElementsByClassName('bpk-close-button')).toBeTruthy();
+      expect(screen.queryByText('MyTitle')).not.toBeInTheDocument();
+      expect(screen.getByTitle('bpk-modal-button-close')).toBeInTheDocument();
+    });
```
olliecurtis, line 114: I am not too sure about this test as you are already passing it in via the `props` object but the test just checks if a string that is not in the rendered output, but a title would still be there as the prop with a string exists.

Could you check this one out?

olliecurtis (changes requested):

packages/bpk-component-modal/src/BpkModalV2/BpKModal-test.tsx
ByAnthony: Oh, I messed up this test and the previous one which checks that the title is rendered. I'll change both to reflect better what is really happening 👍

packages/bpk-component-modal/src/BpkModalV2/BpkModal.tsx
```diff
@@ -38,10 +38,39 @@ export type Props = {
   noFullScreenOnMobile?: boolean;
   onClose: () => void | null;
   padded?: boolean;
+  showHeader: boolean;
```
anambl, line 41: I don't think this should be a required prop (also looking at the description of the PR) 🤔 shouldn't it have a default value which is `true` in which case this prop would simply introduce the option of hiding the header?

examples/bpk-component-modal-v2/examples.tsx
```diff
@@ -227,6 +228,12 @@ const NoFullScreenOnMobileNoTitleExample = () => (
   </ModalContainer>
 );
 
+const NoHeaderExample = () => (
+  <ModalContainer showHeader={false}>
+    This is a default modal. You can put anything you want in here.
```
anambl, line 233: is this the default modal? 😄 the default modal would be when `showHeader` is `true`, no?

packages/bpk-component-modal/src/BpkModalV2/accessibility-test.tsx
```diff
@@ -30,6 +30,7 @@ describe('BpkModal accessibility tests', () => {
         closeLabel="bpk-modal-button-close"
         isOpen
         onClose={jest.fn()}
+        showHeader
```
anambl, line 33: small thing, but this shouldn't be needed anymore as it will be true by default

olliecurtis (approved):

PR #2876, opened 2023-06-07
[KOA-6114] Migrate scrollable calendar to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-calendar/index.d.ts
+++ b/packages/bpk-component-calendar/index.d.ts
@@ -18,13 +18,15 @@
 
 import BpkCalendarContainer, { withCalendarState } from './src/BpkCalendarContainer';
 import BpkCalendarGrid, { BpkCalendarGridWithTransition } from './src/BpkCalendarGrid';
+import type { Props as BpkCalendarGridProps } from './src/BpkCalendarGrid';
 import BpkCalendarGridHeader from './src/BpkCalendarGridHeader';
 import BpkCalendarNav from './src/BpkCalendarNav';
 import BpkCalendarDate, { ROW_TYPES, SELECTION_TYPES } from './src/BpkCalendarDate';
+import type { Props as BpkCalendarDateProps } from './src/BpkCalendarDate';
 import composeCalendar from './src/composeCalendar';
 import { CALENDAR_SELECTION_TYPE } from './src/custom-proptypes';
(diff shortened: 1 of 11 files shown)


examples/bpk-component-calendar/examples-components.js
```diff
@@ -289,12 +289,10 @@ class CalendarContainer extends Component {
 }
 
 CalendarContainer.propTypes = {
-  selectTodaysDate: PropTypes.bool,
```
anambl, line 292: I deleted this from everywhere in the code because it doesn't seem to be an actual prop except for the example components, but even then it's never actually used!

packages/bpk-component-calendar/src/BpkCalendarContainer.tsx
```diff
@@ -42,18 +42,18 @@ import {
 } from './date-utils';
 
 type Props = {
-  fixedWidth: boolean;
-  maxDate: Date;
-  minDate: Date;
+  fixedWidth?: boolean;
```
anambl, line 45: have finally figured out how to handle the optional props which have default values 🙂
A bit confusing as I had read and seen examples where it says you don't need to make them optional (e.g. https://stackoverflow.com/questions/37282159/default-property-value-in-react-component-using-typescript) I think maybe this doesn't work as well with HOCs and nested components, but next best thing I found was marking them optional and using the null assertion type operator (ONLY if there is a default prop) - https://stackoverflow.com/questions/45867568/typescript-and-react-defaultprops-are-still-being-treated-as-possibly-undefined

packages/bpk-component-scrollable-calendar/src/utils.ts
```diff
@@ -31,16 +31,16 @@ const getMonthsArray = (startDate, count) => {
...
+  months: Date[],
+  weekStartsOn: 0 | 1 | 2 | 3 | 4 | 5 | 6,
+  columnCount: number,
+  rowHeight: number,
+  baseMonthItemHeight: number,
+): number[] => {
   const monthItemHeights = months.map((month) => {
     const firstDayOffset = (month.getDay() + 7 - weekStartsOn) % 7;
     const monthLength = DateUtils.daysInMonth(
-      month.getYear(),
```
anambl, line 43: this is deprecated, but using `getFullYear` will give the same results in this case

olliecurtis (approved):

PR #2878, opened 2023-06-08
[KOA-5191] Migrate datepicker component to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-banner-alert/src/BpkBannerAlert.tsx
```diff
@@ -19,13 +19,6 @@
 import type { CommonProps } from './common-types';
 import BpkBannerAlertInner from './BpkBannerAlertInner';
 
-const defaultProps = {
-  animateOnEnter: false,
-  animateOnLeave: false,
-  show: true,
-  icon: null,
-};
-
```
olliecurtis, line 28: Why has this changed in this PR?

anambl: ah I completely forgot about this, was meant to add a comment to explain. It's just something I noticed from when I migrated the banner alert to TS. You'd requested a change to pass each prop one by one instead of using the `{...defaultProps}` and after making the change, I forgot to remove this so this is actually not used at all so it's just cleaning up!

olliecurtis: Could you add this as a separate PR, just to keep this one scoped to this change, in the case we need to revert the datepicker changes or have a better traceability for the changes?

olliecurtis (approved):

PR #2882, opened 2023-06-13
[KOA-5537] Migrate image to TS

anambl:

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-image/src/withLazyLoading.tsx
```diff
@@ -130,7 +114,9 @@ export default function withLazyLoading(
             return supportsPassiveOption;
           },
         });
+        // @ts-expect-error TODO: make this TS compliant
```
anambl, line 117: changing this causes the unit tests to fail so it needs a bit of a bigger investigation. Added a ts exception for now

packages/bpk-component-image/src/withLazyLoading.tsx
```diff
@@ -130,7 +114,9 @@ export default function withLazyLoading(
...
         window.addEventListener('test', null, opts);
+        // @ts-expect-error TODO: make this TS compliant
```
anambl, line 119: changing this causes the unit tests to fail so it needs a bit of a bigger investigation. Added a ts exception for now

packages/bpk-component-image/src/BpkBackgroundImage.tsx
```diff
@@ -15,55 +15,39 @@
...
-    loading: PropTypes.bool,
-    onLoad: PropTypes.func,
-    style: PropTypes.object, // eslint-disable-line react/forbid-prop-types
-    imageStyle: PropTypes.object, // eslint-disable-line react/forbid-prop-types
-  };
+  trackImg?: HTMLImageElement | null;
 
   static defaultProps = {
-    className: null,
+    className: '',
```
olliecurtis, line 50: How do we do this on our other TS components? As this will now apply a blank `""` to the class each time the component is rendered

packages/bpk-component-image/src/withLazyLoading.tsx
```diff
@@ -15,79 +15,63 @@
...
-      style: PropTypes.object, // eslint-disable-line react/forbid-prop-types
-      className: PropTypes.string,
-    };
+    placeholderReference?: string;
 
     static defaultProps = {
-      style: null,
-      className: null,
+      style: {},
+      className: '',
```
olliecurtis, line 51: Same as above here

packages/bpk-component-image/src/withLazyLoading.tsx
olliecurtis: Maybe we should check this as part of this work as it means that something must be wrong and we don't want to risk releasing a component that has a bug into it after we move to TS

packages/bpk-component-image/src/withLazyLoading.tsx
olliecurtis: Same here

packages/bpk-component-image/src/BpkBackgroundImage.tsx
anambl: Yeah, I think we generally set `null` as default, but in this case because it's actually passed directly to the `div` class attribute, TS will complain because it expects a string or undefined.
We generally set it as `null` because we almost always use the `getClassName` util and not pass it directly to an element

packages/bpk-component-image/src/withLazyLoading.tsx
olliecurtis: Let's have a chat about this one as I do worry about us introducting a bug by doing workarounds and doesn't follow great coding practice to work around it to make tests just pass

anambl: It's not introducing a bug though, at most there's already a bug present which we're not fixing in this PR 😞 but yes, let's have a chat!

anambl: ok I fixed it 😅 turns out that it would always throw an error because of the args in `addEventListener`/`removeEventListener`!

olliecurtis (approved):

PR #2892, opened 2023-06-20
[LOOM-772]: remove the display.none fix - replace with existing scroll fixes

mungodewar:
### Context:

Recently a bug was discovered when releasing changes to the cookie-banner. This caused an incident that is documented here: https://skyscanner.atlassian.net/wiki/spaces/OOPS/pages/629716381/INC-629+-+Cookie+Banner+crashing+pages+on+iOS

The root cause of this incident was because of the `applicationElement.style.display = 'none';`, this line was being executed at the correct moment, however, the component rendering the scrim was _inside_ the `applicationElement` this resulted in a blank screen for iOS users. 

### Change:
This PR moves away from the `display: none` workaround and instead utilises already existing alternatives in the codebase. This causes no behavioural change and is far safer given that no element is being actively hidden from the user. 

### Test:
This has been extensively tested against various problematic iOS versions using simulators. Things to highlight: 
- inner scroll on dialog works in all cases
- scroll position restored on closing of dialog (sometimes required to not have storybook UI in place for this to work)
- background impossible to scroll

## iOS 16.2 and iOS 12.4
https://github.com/Skyscanner/backpack/assets/89925955/7626e4c6-8c3f-4ebb-b81c-35bee9c8a704

## iOS 11 - large dialog example
https://github.com/Skyscanner/backpack/assets/89925955/708bf7c3-90c5-4579-b65d-2f34e94c2bfa

## iOS 11- small dialog example
https://github.com/Skyscanner/backpack/assets/89925955/537131b1-3c9f-4d29-83f7-ff0c1958626f


olliecurtis (approved):
I would mark this as a patch vs a minor just as this is fixing an issue rather than adding functionality

mungodewar:
> I would mark this as a patch vs a minor just as this is fixing an issue rather than adding functionality

It will change the visuals on iOS devices. Before the background content will not be shown and with this change it will be. I'm happy with either but it was for this reason that I considered it to be `minor`.

PR #2899, opened 2023-06-26
[KOA-6135] Update contribution guidelines

anambl:

- [ ] Component `README.md`
- [ ] Tests


CONTRIBUTING.md
```diff
@@ -33,99 +30,111 @@ If you use [nvm](https://github.com/creationix/nvm) or [nave](https://github.com
...
+#### Contribute breaking changes
 
-If you add a new file of mixins, for example for a new *atom*, make sure you add the file to the imports in `_index.scss`.
+Anytime a change could break existing applications it's considered a breaking change. To make upgrading Backpack easier for consumers, breaking changes should follow a deprecation cycle.
 
-### React component
+In most cases, it is recommended to create a V2 component and provide a migration guide. If your breaking change either
+- requires consumers to make multiple modifications to their code to adopt the changes (e.g. renaming or removing multiple props of an API), OR
+- involves a significant alteration of the structure or API of a component in a way that keeping both APIs within one component may make the code unreadable (e.g. rewriting a component to reduce the number of files from 20 to 5)
+then you should contribute your changes as a separate V2 component.
```
olliecurtis, line 121: What guidence do we follow if its a visually breaking change or even wanting to experiment with a new design?

For the latter of that question we should link down to the experimentation but also have a chat as a squad about testing with visual changes too

CONTRIBUTING.md
```diff
@@ -33,99 +30,111 @@ If you use [nvm](https://github.com/creationix/nvm) or [nave](https://github.com
...
 
-**All new components should be written in Typescript.**
+Once the old component has been removed from the codebase and is no longer in use, the Koala team will run automated scripts to make the new component the default.
```
olliecurtis, line 123: >the Koala team will run automated scripts to make the new component the default.

Do we have these or even a plan for this? I would right now exclude it as its not a true reflection of what the current state is and this document should be what is possible now :)

CONTRIBUTING.md
```diff
@@ -33,99 +30,111 @@ If you use [nvm](https://github.com/creationix/nvm) or [nave](https://github.com
...
 
-Use `bpk-component-boilerplate` to create a new skeleton React component. Once this is created, use existing components for code style inspiration.
+If your breaking change doesn't require consumers to make many modifications or its usages are limited (around 10-20 usages across Skyscanner web repositories), and keeping both versions in one component doesn't affect the readability of it, you will not need to create a separate component.
```
olliecurtis, line 125: 10-20 usages seems like a lot so wondering if we should be quoting such a high number, this is what would prompt me to make a separate component as that is a massive impact

olliecurtis (changes requested):

CONTRIBUTING.md
```diff
@@ -33,99 +30,111 @@ If you use [nvm](https://github.com/creationix/nvm) or [nave](https://github.com
...
-### Install dependencies
+2. Install npm dependencies
 
-Run `npm install` to install dependencies from npm.
+```sh
+npm install
+```
 
-### Build the code
+3. Build icons
```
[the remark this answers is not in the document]
anambl: It isn't? 🤔 I think it is just SVGs looking at where we use gulp

CONTRIBUTING.md
```diff
@@ -33,99 +30,111 @@ If you use [nvm](https://github.com/creationix/nvm) or [nave](https://github.com
...
-## Adding icons
+Before you start writing code, we recommend familiarising yourself with the engineering convetions and squad decisions which are kept in the [decisions folder](/decisions).
 
-If you want to add icons, please discuss them with us first.
+### React components
 
-Once they're signed off, you can [raise a request](https://bit.ly/backpack-request) and attach the SVG files. If you're feeling heroic and want to make the PR yourself, just copy the correctly named SVG files to the `lg` and `sm` directories in [`@skyscanner/bpk-svgs/src/icons/`](https://github.com/Skyscanner/backpack-foundations/tree/main/packages/bpk-svgs/src/icons/) and then run `npm run build`.
+> **All Backpack components are written in Typescript or are being gradually migrated to Typescript.**
+>
+> As we're in the process of migrating all Backpack components to Typescript, we kindly request that engineers contributing changes to an existing component also migrate the respective component.
```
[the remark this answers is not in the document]
anambl: > Maybe want to clarify here that if its not already in TS.

What do you mean?

> Also do we need to call out about the generation of the .d.ts files?

Hmm... this paragraph is more of a temporary note while we migrate the components. We have the danger warning for the .d.ts files

CONTRIBUTING.md
```diff
@@ -33,99 +30,111 @@ If you use [nvm](https://github.com/creationix/nvm) or [nave](https://github.com
...
+- [bpk-component-chip](./packages/bpk-component-chip/index.ts)
+- [bpk-component-button](./packages/bpk-component-button/src/BpkButtonV2/BpkButton.tsx)
 
-Figma is the preferred format for non-technical folks. We’d appreciate if you could provide an exact match of your component in Figma format together with folders for each state e.g. disabled, expanded etc.
+#### CSS
 
-### Tokens
+We use [CSS Modules](https://github.com/css-modules/css-modules) along with [BEM](http://getbem.com/) to prevent collisions and accidental overwrites in CSS.
 
-Any visual CSS parameters of the component, such as *color, margins, paddings* etc. should not live as magic numbers in the component code, but as **tokens** in the [`@skyscanner/bpk-foundations-web`](https://github.com/Skyscanner/backpack-foundations/tree/main/packages/bpk-foundations-web) package.
```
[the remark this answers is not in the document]
anambl: The only reason I removed it is because we don't have the tokens in this repo anymore 😄 I was also hesitant because even though they don't live in this repo anymore, it's still relevant as this is where they would be used so I'll add it back since you mentioned it too 👍

CONTRIBUTING.md
```diff
@@ -33,99 +30,111 @@ If you use [nvm](https://github.com/creationix/nvm) or [nave](https://github.com
...
-> You should probably not touch `aliases.json`, as our colour palette or grid rarely changes.
+If you want to add a new component:
 
-### Sass mixins
+1. Use `bpk-component-boilerplate` to create a new skeleton React component
+2. It's unlikely you will need to create a class component. If you are in doubt, have a look at the React guidelines for component and see if you can use a function component instead: [React component guidelines](https://react.dev/reference/react/Component)
+3. Create stories - each component has a set of stories living under `examples/bpk-component-{name}/stories.ts`. Stories should cover most visual variants of a component. Read more about Storybook stories [here](https://storybook.js.org/docs/react/writing-stories/introduction)
+4. Create tests
+    - Visual regression tests - Each UI component's stories should also include a story that begins with the name `VisualTest` - these will then be picked up by Percy to run on CI
+    - Unit tests - Unit tests live in the same folder with the component's code and rely on `jest` and `React Testing Library`
```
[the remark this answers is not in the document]
anambl: You're right

CONTRIBUTING.md
anambl: We don't have the scripts yet, but I definitely remember agreeing on this 😬  as otherwise we'd be creating extra work for our consumers

CONTRIBUTING.md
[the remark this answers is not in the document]
olliecurtis: It is just svgs but they aren't scoped to just the icons its part of the flare too which is not an icon :)

CONTRIBUTING.md
marianeum: I would leave this out for now as well, as it'll also depend on the component - we won't be able to do it for everything, depending on the changes

CONTRIBUTING.md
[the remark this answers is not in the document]
olliecurtis: >What do you mean?

Reading it right now makes it sound like they have may have to do it for those already there which I know it looks obvious but make it more explicit so we make clear distinctions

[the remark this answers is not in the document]
olliecurtis: >Hmm... this paragraph is more of a temporary note while we migrate the components. We have the danger warning for the .d.ts files

We have it as a danger rule but that is a long cycle before they get that reminder (as it comes via CI) so having it part of the docs whilst folks are doing things and following this means it is less likely to get missed. We shouldn't just let it come late in the feedback cycle

CONTRIBUTING.md
olliecurtis: >I think the same guidance should apply as the goal is for consumers to not have to make too many changes to adopt the newer version AND to ensure the component API and code is not too complex to cause confusion to devs

In which case we should call this out. That was the main idea behind my point we don't actually say it follows this so can lead to people doing things differently as it doen't explicitly mention

CONTRIBUTING.md
olliecurtis: Yeah but right now we should in this document be reflecting what the current state is right now so we should have it removed until we do/have the process in place :)

CONTRIBUTING.md
olliecurtis: It's still a large impact especially if its across multiple projects, yes the numbers are more significant but its still a lot of places that people will have to plan to adopt

CONTRIBUTING.md
```diff
@@ -33,99 +30,119 @@ If you use [nvm](https://github.com/creationix/nvm) or [nave](https://github.com
...
-Run `npm install` to install dependencies from npm.
+```sh
+npm start
+```
 
-### Build the code
+## Write your code
 
-Backpack's code depends on some things that must be built first, such as icon fonts, SVGs and tokens.
+Before you start writing code, we recommend familiarising yourself with the engineering conventions and squad decisions which are kept in the [decisions folder](/decisions).
```
olliecurtis, line 76: Should we also add here to check out the guidence on the docs site too for getting started with Backpack?

https://www.skyscanner.design/latest/getting-started/about-backpack-pN209Wjo

anambl: Hmm... feel like that page is quite generic and about the DS in general. We do hvae the principles https://www.skyscanner.design/latest/getting-started/backpack-in-code/principles-MmgbZH40 page though that could be useful to link 🤔 although they're not web only

marianeum: I think them not being web only is fine. The about backpack site is a good start, though - it also links to the engineering guide etc which is helpful

marianeum (approved):
looks good - just a small comment around the initial link

olliecurtis (approved):

PR #2901, opened 2023-07-03
[KOA-5703]: Bump css-loader and postcss-loader

olliecurtis:
Bumps the following dependencies:
- `css-loader`
- `postcss-loader`

Makes required changes to webpack config for storybook to align with latest changes.

- [ ] Component `README.md`
- [ ] Tests


.storybook/webpack.config.js
```diff
@@ -57,15 +57,19 @@ module.exports = ({ config }) => {
         options: {
           importLoaders: 1,
           modules: useCssModules,
-          localIdentName: '[local]-[hash:base64:5]',
+          // localIdentName: '[local]-[hash:base64:5]',
```
olliecurtis, line 60: One TODO: check if we still need the localIndentName. It works fine without it right now as can see it operates but would be good to check why it was added/needed

.storybook/webpack.config.js
```diff
@@ -23,9 +23,8 @@ const sassFunctions = require('bpk-mixins/sass-functions');
 
 const postCssPlugins = require('../scripts/webpack/postCssPlugins');
 
-const { BPK_TOKENS, ENABLE_CSS_MODULES } = process.env;
```
anambl, line 26: What was the purpose of ENABLE_CSS_MODULES var? Also wondering if we use BPK_TOKENS as well

olliecurtis: It looks to be super old configuration which going through our docs and setup we never actually use anywhere so looks to be an irrelevant variable.

It also doesn't line up and work with the configuration we need to provide in the newer `css-loader` so that we can specify our `localIndentName` as it would either be a boolean or an object, which we always need to provide the `localIndentName`, `modules` would always need to be an object :) 

Hope that helps, let me know if not

anambl (approved):

PR #2908, opened 2023-07-10
Koa 5884 migrating packages

olliecurtis:
Migrating `bpk-mixins` and `bpk-stylesheets` to Backpack Web

The benefit of doing so are:

For Backpack itself
- No need to manage version bumps between foundations and this library so keep toil down and easier to make/test change to mixins
- Keeps consolidation together as usually `@skyscanner/backpack-web` and `bpk-mixins`, and even `bpk-stylesheets` exist together so benefits to consuming projects too

For consumers
- Less dependencies to manage as all bundled into one place
- Less duplication of versions as now all served from the central point of `@skyscanner/backpack-web`, instead of having to install separately and the risk of having a version defined of mixins/stylesheets whilst Backpack web had older versions defined.

- [x] Component `README.md`
- [x] Tests


examples/bpk-component-aria-live/examples.module.scss
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-@import '~bpk-mixins/index.scss';
+@import '../../packages/bpk-mixins/index.scss';
```
anambl, line 19: I guess all of these imports were automatically updated by the code editor? 😄

.gitignore
```diff
@@ -3,17 +3,15 @@
 .vscode
 .sass-cache
 /dist
+/dist-sassdoc
 /dist-storybook
+/dist-sassdoc
```
anambl, line 8: this is now added twice 😄

anambl (approved):
LGTM!

PR #2913, opened 2023-07-12
[KOA-6040] Bump autoprefixer

anambl:
Tested with https://github.com/Skyscanner/backpack/pull/2908 too and all runs fine


marianeum (approved):
I assume the warnings are fine?

gert-janvercauteren (approved):

PR #2943, opened 2023-08-01
BpkButton: Improve onClick type to Mouse Event

runmoore:
### Description

`SyntheticEvent` is react's most generic kind of event, but specifying a `MouseEvent` we get [more properties](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6be113476bef803465cd7e1a96a14a96154351f3/types/react/index.d.ts#L1321-L1341). It also aligns with [intellisense of vscode](https://stackoverflow.com/questions/59753118/what-is-the-correct-type-for-react-click-event)

image

For keyboard users, onClick is still fired, but just some properties of the event aren't populated, but they do exist.

If we wanted to responed to keyboard events, we would use the onKeyDown prop

### Checks

- [ ] Component `README.md`
- [ ] Tests


olliecurtis (approved):
LGTM!

Thanks for the contribution

PR #2959, opened 2023-08-08
Scripts for transpilation of mixins and stylesheets

olliecurtis:
As part of migrating packages to the single package, it turns out that the transpilation process was broken for these packages and would therefore not work when we ship and then proceed to use it.

To avoid confusing the already approved PR and making reviews harder, I have split this out to this PR to be merged into the original then we can ship it.

**Key Notes**
- Moved `bpk-storybook-utils` out of packages
  - As these were introduced a long time ago for the old docs site
  - We don't publish them to consumers (check .npmignore)
  - They are no solely used by our Storybook examples so makes it cleaner to process

- Added new scripts to ensure that the correct files are moved to the transpiled package but NOT transpiled as otherwise they can't be used in consumer code, due to them being Sass

- Simplfied up the transpilation scripts to make them easier to digest when reading the code  for these steps rather than big one line of commands


package.json
```diff
@@ -16,7 +16,8 @@
     "registry": "https://registry.npmjs.org/"
   },
   "scripts": {
-    "build": "gulp && npm run build:stylesheets",
+    "build": "gulp && npm run build:sass && npm run build:stylesheets",
+    "build:sass": "backpack-node-sass --prefixComment=\"$(cat ./license.txt)\" && rm packages/bpk-stylesheets/index.css",
```
anambl, line 20: I'm a bit confused, what is this cmd supposed to do and why are we removing the stylesheets index.css?

scripts/transpilation/copy-utils.js
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+const fs = require("fs");
+const { execSync } = require('child_process');
+
+// eslint-disable-next-line no-console
+console.log('Copying bpk-mixins...');
+
+// if(!fs.existsSync('dist/bpk-mixins')) {
```
anambl, line 25: do we need this?

package.json
olliecurtis: So this command, which we changed from just the transpiled process to regular build is: 

- To compile the Sass files to CSS but does it in a parallel manner for speed vs one file by one file.
- For legal copyright appends the licence header to each file

For the removal of the file, we don't want this file transpiled because it will be unusable and don't want systems to attempt use as it generates links to local machine directories for Normalise CSS.

This command was updated based on what we used to have in Backpack before we moved these to foundations - see here for reference: https://github.com/Skyscanner/backpack/pull/2381/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L21

scripts/transpilation/copy-utils.js
olliecurtis: Ooop good catch!

scripts/transpilation/copy-utils.js
```diff
@@ -0,0 +1,54 @@
...
+
+
+// eslint-disable-next-line no-console
+console.log('Copying bpk-stylesheets files...');
+
+if(!fs.existsSync('dist/bpk-stylesheets')) {
+  execSync(`mkdir dist/bpk-stylesheets`);
+}
+
+const styleFiles = execSync('find packages/bpk-stylesheets -name "*.scss" -o -name "*.js" -o -name "*.css" | grep -v  "webpack.config.babel.js" | grep -v "build.js"')  .toString()
```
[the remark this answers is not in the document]
olliecurtis: Good catch, my linter has been a bit weird in VSCode so is missing a few things!

package.json
olliecurtis: For reference this was where it was before in the transpilation process: https://github.com/Skyscanner/backpack/pull/2959/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L57

scripts/transpilation/copy-utils.js
olliecurtis: The copy command already handle this processing and creating it if its not there, otherwise it makes nested folder which is bad!

anambl: Oh I thought the `cp` command would fail if dest folder doesn't exist 🤔

scripts/transpilation/copy-utils.js
```diff
@@ -0,0 +1,54 @@
...
+
+// eslint-disable-next-line no-console
+console.log('bpk-mixins copied.  👍');
+
+
+// eslint-disable-next-line no-console
+console.log('Copying bpk-stylesheets files...');
+
+if(!fs.existsSync('dist/bpk-stylesheets')) {
+  execSync(`mkdir dist/bpk-stylesheets`);
```
anambl, line 36: https://github.com/Skyscanner/backpack/pull/2959/files/6ec072c445f11a53f7c362df4245f24b055d59fb#r1287264390 in that case can we also remove this bit?

olliecurtis: Good point let me check this one!

olliecurtis: So the reason it needs to be there and not for the mixins is due to the mixins being inexistance at the time of copy because of the sass-functions.js file gets babeled over.

Whereas the stylesheets one is not getting babel transpiled so the folders never exist so needs to be copied. If I remove that makedir command it just crashes that the folder doesn’t exist so doesn’t complete the process, so why it remians for one and not the other

Screenshot 2023-08-08 at 16 45 13

anambl (approved):

PR #2967, opened 2023-08-14
[KOA-6248] Add optional label to nudger component

anambl:
### TL;DR
- New props `title`, `subtitle`, `icon` to add a label to the nudger component.
- Removed `BpkNudgerRow` component. This should not impact any projects in Skyscanner as it is a new component and we found no usages of it.

### Migrate to using the built-in nudger label
If you are using BpkNudger together with a label, consider passing the label content as props to inherit the accessible structure and styling of the component :accessibility: ✨
To migrate, remove the label component and instead pass the `title` and `subtitle` and `icon` props accordingly.

Checkout the [README.md](https://github.com/Skyscanner/backpack/tree/main/packages/bpk-component-nudger#bpk-component-nudger) for more details on the props.

```
<div>
  <BpkLabel id="passenger-label" htmlFor={id} className={labelClassName}>
      Number of passengers
  </BpkLabel>
  <BpkNudger
      aria-labelledby="passenger-label"
      id={id}
      min={1}
      max={10}
      value={value}
      onChange={handleChange}
      decreaseButtonLabel="Remove passenger"
      increaseButtonLabel="Add passenger"
      buttonType={buttonType}
   />
</div>
```
to

```
<div>
  <BpkNudger
      id={id}
      min={1}
      max={10}
      value={value}
      onChange={handleChange}
      decreaseButtonLabel="Remove passenger"
      increaseButtonLabel="Add passenger"
      title="Adults"
      subtitle="Aged 16+"
      icon={<AlignedAccountIcon />}
    />
</div>
```

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-nudger/README.md
```diff
@@ -129,22 +127,25 @@ class App extends Component {
 
 ### BpkConfigurableNudger
```
gert-janvercauteren, line 128: Should we have 2 prop tables? 
1 for one with label, one for without? 

This way we can show the title is required etc.

anambl: I'm afraid that'd make it look like there's 2 separate components (as that is the only time when we have more than 1 api table) and that's not the case here, but maybe can add a comment in the api table that title is required if you want to show a label?

frugoman (approved):

PR #2975, opened 2023-08-25
Correct README.md instructions

olliecurtis:
This PR corrects the usage instructions for `bpk-stylesheets` as it's main recommended/best use is to import via JS but the docs were confusing and pointing at CSS

### DIFF
--- a/packages/bpk-stylesheets/README.md
+++ b/packages/bpk-stylesheets/README.md
@@ -11,13 +11,13 @@ Check the main [Readme](https://github.com/skyscanner/backpack#usage) for a comp
 Within your Javascript (usually the entrypoint or top-most JS file)
 
 ```js
-import 'bpk-stylesheets';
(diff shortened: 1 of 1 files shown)


packages/bpk-stylesheets/README.md
```diff
@@ -8,14 +8,16 @@ Check the main [Readme](https://github.com/skyscanner/backpack#usage) for a comp
 
 ## Usage
 
-```scss
-@import '~@skyscanner/backpack-web/bpk-stylesheets';
+Within your Javascript (usually the entrypoint or top-most JS file)
+
+```js
+import 'bpk-stylesheets';
```
anambl, line 14: don't we need the `@skyscanner/backpack-web` prefix?

olliecurtis: 🤦yes you are right! Good catch

anambl (approved):

PR #2985, opened 2023-09-01
[no-jira]: Fix spacing tokens that are no longer available

olliecurtis:
When we removed the old spacing tokens from Backpack Foundations, I hadn't realised that some tokens are still in use from the JS file in code and they have not been provided for some time/working.

This PR fixes this by setting the values the component requires to keep the same design as previously expected.

- [ ] Component `README.md`
- [x] Tests


anambl:
Wondering if we shouldn't actually maybe re-introduce the spacing tokens for JS in foundations? Might be worth talking to design about this 🤔 as I think that they are meant to be used for padding. Although anyone can add them up together so in the end you can use any value you want....
Also wondering what's the current usage of it? 🤔

packages/bpk-component-barchart/src/__snapshots__/BpkChartAxis-test.js.snap
```diff
@@ -435,7 +435,7 @@ exports[`BpkChartAxis X should render with label 1`] = `
         class="bpk-chart__axis-label"
         text-anchor="middle"
         x="60"
-        y="40"
+        y="34"
```
anambl, line 438: Why have these change? 🤔

olliecurtis: Let me check into this as you are right it shouldn't have and might make sense with some of the changes in this PR now

olliecurtis: Turns out these should be the correct values and what happened is it was actually broken 😆

anambl (approved):

PR #2988, opened 2023-09-01
[KOA-6230] Datatable uses rem values

anambl:
Data table should use values in rem so that it can scale.
- Add warning messages when value in px (i.e. no unit provided) is passed to data table. This will be removed in the future and only rem values will be allowed.
- Convert px value to rem

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-datatable/src/BpkDataTable.js
+++ b/packages/bpk-component-datatable/src/BpkDataTable.js
@@ -28,7 +28,7 @@ import STYLES from './BpkDataTable.module.scss';
 import type { Props } from './common-types';
 import { SORT_DIRECTION_TYPES } from './sort-types';
 import BpkDataTableHeader from './BpkDataTableHeader';
-import { getColumns } from './utils';
+import { pxToRem, getColumns } from './utils';
 
 const getClassName = cssModules(STYLES);
 
@@ -139,7 +139,7 @@ const BpkDataTable = (props: Props) => {
   return (
     <div
       {...getTableProps({
-        style: { width, height },
+        style: { width: pxToRem(width), height: pxToRem(height) },
         className: classNames,
       })}
       {...restOfProps}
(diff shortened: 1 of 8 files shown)


packages/bpk-component-datatable/src/utils.js
```diff
@@ -17,10 +17,21 @@
  */
 /* @flow strict */
 
-// TODO: Remove these functions once we want to change the API to match the react-table library API
+const ROOT_FONT_SIZE_PX = 16;
+
+export const pxToRem = (value: string | number) => {
+  let parsed = value;
+
+  if (typeof value === 'number') {
```
olliecurtis, line 25: What happens if someone doesn't pass a number and actually parses the string `16px` would this attempt to just parse as is?

packages/bpk-component-datatable/src/BpkDataTable.js
olliecurtis, line 1: Mad one but as we are touching this component and making changes should we also be converting to TS whilst we are touching it, so we can cut through some of our TS remaining migrations?

olliecurtis (commented):
Should we put something in the README to hightlight the change in expectation of the props for consumers to know of the change?

packages/bpk-component-datatable/src/utils.js
anambl: Hmm yes, it wouldn't be parsed to `rem` if someone passes a string. Let me see, I'll change this slightly

anambl:
> Should we put something in the README to hightlight the change in expectation of the props for consumers to know of the change?

Good point!!

packages/bpk-component-datatable/src/BpkDataTable.js
olliecurtis: Probably a safer approach to deprecate and messaging than just going in and breaking with no notice. Sounds like the safer approach

olliecurtis (approved):

PR #2995, opened 2023-09-06
[NO JIRA]: Correct types for datepicker and dialog

olliecurtis:
This PR corrects some types that we had some issues with when we moved to TS.
- Corrects Datepicker input type to accept a ReactElement vs a shared prop across multiple components
- Improves readability of README for dialog to be clearer on the `onClose` prop
- Adds a runtime type check to ensure that the correct types and combinations are passed
- Makes the Dialog `onClose` prop optional for when dismissible is true

- [x] Component `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-datepicker/src/BpkDatepicker.d.ts
+++ b/packages/bpk-component-datepicker/src/BpkDatepicker.d.ts
@@ -17,6 +17,7 @@
  */
 
 import { Component } from 'react';
+import type { ReactElement } from 'react';
 import type { DaysOfWeek, ReactComponent, SelectionConfiguration } from '../../bpk-component-calendar';
 type Props = {
     changeMonthLabel: string;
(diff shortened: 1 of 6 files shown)


packages/bpk-component-datepicker/src/BpkDatepicker.d.ts
olliecurtis, line 1: I am slightly worried about the changes in this file as this suddenly seems like a lot of changes?

Looks like a lot of reformatting of order, but I do see there is an added `Omit<` (L64) but also some addition of `globalThis.` which seems suspicious

anambl: Ohhh... that is a bit weird indeed. Not sure why it adds the `Omit` to the type of default props 😓 Although looking at the props https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-calendar/src/BpkCalendarContainer.tsx#L353-L388  I think it does make sense as it is overwriting those ommited props

For the `globalThis` I think we don't need to import JSX?

examples/bpk-animate-height/stories.js
anambl, line 1: I think this is accidental? 😄

olliecurtis: Whoops yes!

anambl (approved):

PR #3013, opened 2023-09-25
feat: add iconMarkerDefaultSelectedFillColor

LuoTuxiu:

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-map/src/BpkIconMarkerBackground.module.scss
+++ b/packages/bpk-component-map/src/BpkIconMarkerBackground.module.scss
@@ -42,8 +42,11 @@
   }
 
   &--selected {
-    fill: $bpk-surface-default-day;
-
+    @include bpk-themeable-property(
+      fill,
+      --bpk-icon-marker-default-selected-fill-color,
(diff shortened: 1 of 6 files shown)


packages/bpk-component-map/src/BpkIconMarkerBackground.module.scss
```diff
@@ -42,8 +42,11 @@
   }
 
   &--selected {
-    fill: $bpk-surface-default-day;
-
+    @include bpk-themeable-property(
+      fill,
+      --bpk-icon-marker-default-selected-fill-color,
+      $bpk-surface-default-day
+    );
```
olliecurtis, line 49: We won't need to make this a themable colour and `bpk-surface-default-day` will be fine :)

olliecurtis (changes requested):
Looking at the design we shouldn't be using theming for this and just setting the colour. Theming was only tooling for partner branding so shouldn't be used now or overriden outside of the component.

Feel free to reach out in slack to chat more about this

PR #3018, opened 2023-09-27
[NO JIRA]: New Skyscanner Focus Indicator

olliecurtis:
Brings the new focus indictor to Skyscanner! We want to improve the exprience for all our travellers and we achive this through providing a focus indicator that is stand out and not clashing with our brand

We provide a mixin for the focus indicator and also utilise this in our `bpk-stylesheets` which is the first package loaded when in use so applies globally by default

Screenshot 2023-09-27 at 13 45 57

We also in this PR make the focus for button links to be rounder like out buttons so that it is not too harsh when in use

Screenshot 2023-09-27 at 13 46 06

- [ ] Component `README.md`
- [x] Tests


anambl (approved):
😍 LGTM!

PR #3019, opened 2023-09-28
[KOA-6045] Add API tables to components

anambl:
This PR introduces automatically generated API tables for all components (existing, but also any new components) through Storybook.
It's the 3rd out of 3 PRs to enable automatically generated API tables for all BPK components.
Since API tables are now in Storybook, we removed them from the README.md files.

This PR also:
- moves props descriptions from Readme.md to the API table (see screenshot for example)
- fixes `BpkList` prop types
- fixes `BpkDismissibleChip` prop types
- fixes `BpkNavigationBar` prop types
- changes the `BpkMobileScrollContainer` props from static (to align with all of our other BPK components, but also to ensure react-docgen picks them up).
- fixes `BpkDataTable` prop types
- fixes `BpkDrawer` prop types
- fixes `BpkNudger` prop types
- changes the `BpkPopover` props from static (to align with all of our other BPK components, but also to ensure react-docgen picks them up).
- changes the `BpkProgress` props from static (to align with all of our other BPK components, but also to ensure react-docgen picks them up).


packages/bpk-component-button/src/BpkButtonV2/BpkButton.tsx
```diff
@@ -23,7 +23,8 @@ import COMMON_STYLES from './BpkButton.module.scss';
 
 const getCommonClassName = cssModules(COMMON_STYLES);
 
-const BpkButtonV2 = ({
+// eslint-disable-next-line import/prefer-default-export
+export const BpkButtonV2 = ({
```
anambl, line 27: for some reason the problem with react docgen was the way the button was exported.. I changed it to a named export because that's what we do for modalv2 and shouldn't matter as in the end the export to consumers stays the same

packages/bpk-component-map/src/BpkPriceMarkerV2/BpkPriceMarker.tsx
```diff
@@ -47,7 +47,7 @@ type Props = {
   status?: Status;
 };
 
-const BpkPriceMarkerV2 = (props: Props) => {
+export const BpkPriceMarkerV2 = (props: Props) => {
```
anambl, line 50: for some reason the problem with react docgen was the way the button was exported.. I changed it to a named export because that's what we do for modalv2 and shouldn't matter as in the end the export to consumers stays the same

examples/bpk-animate-height/stories.js
```diff
@@ -17,10 +17,13 @@
  */
 
 /* @flow strict */
```
olliecurtis, line 19: Should we remove the flow here?

anambl: Hmm not sure why we have flow in the stories files tbh, we do still use flow, but not in the stories 😅 
I guess we could remove them since we're already making a change to every single story file 🤔

anambl: done!

olliecurtis (approved):

PR #3022, opened 2023-10-02
[KOA-6249] Add automated visual tests with zoom enabled

anambl:
This PR introduces automated visual tests for all BPK components with font size set at 32px which is 200% of the base Skyscanner font size. This is to ensure any changes to components do not break accessibility requirements and components still scale as expected.

- [ ] Component `README.md`
- [ ] Tests


.storybook/preview.tsx
```diff
@@ -61,6 +69,9 @@ const preview: Preview = {
         </>
       )
     },
+  },
+  args: {
+    fontSize: '16px'
```
olliecurtis, line 74: Do we want to use `fontSizeBase` from our tokens for this?

examples/bpk-component-accordion/stories.js
```diff
@@ -73,3 +73,8 @@ export const WithoutDivider = SingleItemExampleWithoutDivider;
 
 export const VisualTest = SingleItemExample;
 export const VisualTestOnDark = WithDarkBackgroundExample;
+
+export const VisualTestWithZoom = VisualTest.bind({});
+VisualTestWithZoom.args = {
+  fontSize: '32px'
```
olliecurtis, line 79: Might we want with this to possibly make it easier to control vs individually change the font size (should we need to) do it where we use one of our font tokens instead that way we can always use something that could change across the board?

arnaugm: not sure if this is what you mean here but I think a variable would be more flexible if in the future this value needs change

.storybook/preview.tsx
arnaugm: why this is 16 and in the stories 32? is because this represents the default size and the others are with the double size?

examples/bpk-component-accordion/stories.js
anambl: I'm not sure I understand the comment/s 😄
But the value will not need to change in the future, unless the WCAG change so that instead of 200% zoom we need to support something else. We could use 2 *`fontSizeBase` so that it's more obvious it represents 200% zoom?

.storybook/preview.tsx
anambl: yes, exactly. 16 is the default root font size for browsers. The 32 in stories is only used for those stories that are meant to be testing components with 200% zoom.

anambl: @olliecurtis It's not exactly 100% linked to our Skyscanner typography system. Here the 16px is meant to represent the default browser font size which is a separate thing. Although I don't think either will ever change 😄

examples/bpk-component-accordion/stories.js
olliecurtis: It might make it clear to use values at least from our system so its clear its a number we came up with and self documenting, just from past experience of just having a number of question raised were 'how did we get to this number?' or 'can it be documented how we got to this 'magic number'?'

.storybook/preview.tsx
anambl: It's not representative of our base size though, it's representative of the standard browser size which is 16px.
I think all our font sizes are in rem too so technically our base size is not 16px, it's 1rem which with most browser settings ins 16px

examples/bpk-component-accordion/stories.js
anambl: I don't really want to link it to our system, nor add a comment for every story that uses it, let me think how to document this one 🤔 🧐

olliecurtis (approved):

PR #3025, opened 2023-10-03
ST-5229: Add support for matching breakpoints in SSR

robaw:
BpkBreakpoint is not compatible with Skyscanner's SSR solution (via backpack-react-scripts). This is down to an incompatibility with the underlying library `react-responsive`. The actual functionality of BpkBreakpoint in SSR is also not supported, meaning that if a React Component can be both CSR and SSR, then you can't use BpkBreakpoint.

With this change, it acknowledges whether you want a breakpoint to match against a server-side render, or not, allowing the consumer to choose which breakpoint to apply.

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-breakpoint/src/BpkBreakpoint.tsx
+++ b/packages/bpk-component-breakpoint/src/BpkBreakpoint.tsx
@@ -34,6 +34,13 @@ const BREAKPOINTS = {
   DESKTOP_ONLY: breakpoints.breakpointQueryDesktopOnly,
 } as const;
 
+const canUseDOM = () =>
+  !!(
+    typeof window !== 'undefined' &&
+    window.document &&
+    window.document.createElement
+  );
+
 type Props = {
   /**
    * The content to render when the breakpoint matches.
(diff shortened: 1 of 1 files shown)


scripts/jest/setup.js
```diff
@@ -20,4 +20,14 @@ import 'jest-axe/extend-expect';
 import 'raf/polyfill';
 import registerRequireContextHook from 'babel-plugin-require-context-hook/register';
 
+// The below is a workaround to the problem were calling resetModules causes react to be required twice.
+// further details can be found here: https://github.com/jestjs/jest/issues/8987#issuecomment-584898030
+let mockActualReact;
+jest.doMock('react', () => {
```
anambl, line 26: Ah nice! I remember seeing a similar issue in the past with `isolateModules` so we do the same in BpkModalV2 https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-modal/src/BpkModalV2/BpKModal-test.tsx#L27
I guess it should be ok to do this across all tests, we could probably remove the one in modal too?

robaw: Yup, let me know if you want me to add that to this PR

anambl: Yes, please if you don't mind! 😄

robaw: I've tried to remove the instances of `isolateModule` in the test you referenced, which most of the tests are happy with but some are failing so I think it's going to require a bit more time for someone to work out the nuances of that test file.

anambl (approved):

PR #3047, opened 2023-10-24
[DTO-4858] Create BpkBadgeV2

Iain530:
Introducing an experimental component `BpkBadgeV2` which:
- increases the label size to 14pt (Footnote) for better legibility
- Uses less prominent colour so the badges can be used as progressive disclosure on Flights inventory cards.

| Before | After |
|--------|--------|
| image | image |

- [x] Component `README.md`
- [x] Tests


packages/bpk-component-badge/src/BpkBadgeV2/BpkBadge.tsx
```diff
@@ -0,0 +1,57 @@
...
+}: Props) => {
+  const classNames = getClassName(
+    'bpk-badge-v2',
+    badgeTypeClassNames[type],
+    docked === 'right' && 'bpk-badge-v2--docked-right',
+    docked === 'left' && 'bpk-badge-v2--docked-left',
+    centered && 'bpk-badge-v2--centered',
+    className,
+  );
+
```
Iain530, line 53: The figma says that use of the success, warning and critical types without an icon is now deprecated - though we don't include the icon in the badge so I wasn't sure how to implement this as a warning - or if we should be.

packages/bpk-mixins/src/mixins/_badges-v2.scss
```diff
@@ -0,0 +1,236 @@
...
+/// @require {mixin} bpk-badge
+///
+/// @example scss
+///   .selector {
+///     @include bpk-badge-v2();
+///     @include bpk-badge-v2--normal();
+///   }
+
+@mixin bpk-badge-v2--normal {
+  background-color: $bpk-canvas-contrast-day;
```
Iain530, line 120: Figma says new token to be made, can't see where and this colour already exists - let me know if there's somewhere else I should be adding this.

image

packages/bpk-mixins/src/mixins/_badges-v2.scss
```diff
@@ -0,0 +1,236 @@
...
+///
+/// @example scss
+///   .selector {
+///     @include bpk-badge-v2();
+///     @include bpk-badge-v2--centered();
+///   }
+
+@mixin bpk-badge-v2--centered {
+  vertical-align: text-bottom;
+}
```
Iain530, line 55: Some of these have no changes but I thought it was a bit cleaner to have a whole v2 file that can be swapped in if the experiment is successful - let me know if you'd prefer less duplication though.

packages/bpk-mixins/src/mixins/_badges-v2.scss
olliecurtis: Hello 👋 so our colours exist and any to be made here in our Backpack Foundations repositiory https://github.com/Skyscanner/backpack-foundations/

packages/bpk-component-badge/src/BpkBadgeV2/BpkBadge.module.scss
```diff
@@ -0,0 +1,67 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '../../../bpk-mixins/index.scss';
+
+.bpk-badge-v2 {
```
olliecurtis, line 21: Given the way this has been created and will hopefully superseed the current version of the badge to make life easier you could just have all that is named here be `bpk-badge` as it won't conflict given the file is localised to just eh V2 component

packages/bpk-mixins/src/mixins/_badges-v2.scss
```diff
@@ -0,0 +1,236 @@
...
+////
+
+/// Standard badge.
+///
+/// @example scss
+///   .selector {
+///     @include bpk-badge-v2();
+///   }
+
+@mixin bpk-badge-v2 {
```
olliecurtis, line 33: Similar to the above you could just keep this to `bpk-badge` and instead of then exporting it to the index file, just import this file into the scss of the V2 badge code

packages/bpk-mixins/_index.scss
```diff
@@ -18,6 +18,7 @@
 
 @import 'src/bonds.scss';
 @import 'src/mixins/badges.scss';
+@import 'src/mixins/badges-v2.scss';
```
olliecurtis, line 21: Instead of doing this here because we are going for the V2 component we can just do this import inside the `BpkBadge.module.scss`

packages/bpk-mixins/src/mixins/_badges-v2.scss
Iain530: Thanks! I've now pulled in the new version and using the variable 🎉

Iain530:
Percy changes LGTM, not sure what's happening with the maps component but looks broken on main too 😕

olliecurtis:
> Percy changes LGTM, not sure what's happening with the maps component but looks broken on main too 😕

Yeah the maps one is super flaky so will be marking it as approved as its something we are aware of

PR #3056, opened 2023-10-30
ARGG-816: Removed encodebase64 function

xalechez:
After removing all the usages, we no longer need to keep this function and use it in Storybook. We also no longer require node-sass as a peerDependency

- [ ] Component `README.md`
- [ ] Tests


olliecurtis (changes requested):
We will also need to remove `node-sass` in the root `package.json` too just for a clean cut?

As I know backpack-node-sass will already bring its own version

.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org
```
xalechez, line 1: Added this to stop confusing registries locally 😆

olliecurtis:
Closing as being brought as part of #3103

PR #3059, opened 2023-11-03
[BD-8468][BpkCarousel] new component for mobile

R1CC1M:
Added carousel component for mobile screens which displays images which can be swiped to move onto the next image. 

Screenshot 2023-11-03 at 13 15 26

- [x] Component `README.md`
- [x] Tests


packages/bpk-component-carousel/src/BpkCarouselSlidesIndicator.tsx
olliecurtis, line 1: Rather than build our own we already have the Indicator for this here: https://www.skyscanner.design/latest/components/page-indicator/web-4FOtKGmR


Which is the component the Figma design uses

Screenshot 2023-11-03 at 10 01 36

packages/bpk-component-carousel/src/accessibility-test.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+import { render } from '@testing-library/react';
+import { axe } from 'jest-axe';
+
+import BpkCarousel from './BpkCarousel';
+
+const DemoImages = () => (
+  <img src="https://url/WxH.jpg" alt='' />
```
olliecurtis, line 25: Do we need an alt tag?

packages/bpk-component-carousel/src/types.ts
olliecurtis, line 1: You will need to add the licence header here

packages/bpk-component-carousel/src/utils.tsx
olliecurtis, line 1: You will need to add the licence header here

packages/bpk-component-carousel/src/BpkCarouselContainer.tsx
olliecurtis, line 1: You will need to add the licence header here

packages/bpk-component-carousel/src/BpkCarouselContainer.module.scss
olliecurtis, line 1: You will need to add the licence header here

olliecurtis (changes requested):

examples/bpk-component-carousel/stories.js
```diff
@@ -0,0 +1,29 @@
...
+
+import DefaultExample from './example'
+
+export default {
+  title: 'bpk-component-carousel',
+  component: BpkCarousel,
+};
+
+
+export const Default = DefaultExample;
```
olliecurtis, line 29: Will you be able to add a visual test example which can use the Default Example

```suggestion
export const VisualTest = DefaultExample;
export const VisualTestWithZoom = VisualTest.bind({});
VisualTestWithZoom.args = {
  zoomEnabled: true
};
```

packages/bpk-component-carousel/README.md
```diff
@@ -0,0 +1,25 @@
...
+> Backpack carousel component.
+
+## Installation
+
+Check the main [Readme](https://github.com/skyscanner/backpack#usage) for a complete installation guide.
+
+## Usage
+
+```tsx
+import BpkBoilerplate from '@skyscanner/backpack-web/bpk-component-carousel';
```
olliecurtis, line 12:
```suggestion
import BpkCarousel from '@skyscanner/backpack-web/bpk-component-carousel';
```

olliecurtis (changes requested):
Couple of minor changes and otherwise this looks good to me

olliecurtis (approved):

olliecurtis (commented):
Just going to run an A11y test to make sure that VO and TB answer as expected

packages/bpk-component-carousel/src/BpkCarouselContainer.tsx
```diff
@@ -0,0 +1,105 @@
...
+          }}
+        />
+      ))}
+      <BpkCarouselImage
+        image={images[0]}
+        index={0}
+        ref={(el) => {
+          observeCycleScroll(el);
+          observeImageChange(el);
+        }}
```
olliecurtis, line 99: I think this needs to be before the logic to map each of the images? 
As it seems to always have the first selected image as the last one when it renders which is this expected behaviour?

JafferN: Ricci is away on AL now, but it seems to be in order for the scroll snap to create an infinite loop.

olliecurtis:
Completed A11y testing on this component and with a screen reader exhibits behaviour that causes focus to go through more items than available and read outs to not be as correct for users with A11y requirements

anambl:
@R1CC1M @JafferN is there any work being done currently on this or should I close the ticket? 🙂

olliecurtis (changes requested):
Some changes suggested, will also try this with A11y tests to make sure there is no strange behaviour

packages/bpk-component-carousel/src/BpkCarousel.module.scss
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@use '../../bpk-mixins/index.scss';
+
+.bpk-carousel-slides-indicator {
+  position: relative;
+  right: 0;
+  bottom: 1.5rem;
```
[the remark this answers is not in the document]
KathyWang0208: No, forgot to change it.

packages/bpk-component-carousel/src/BpkCarousel.tsx
```diff
@@ -0,0 +1,68 @@
...
+        images={images}
+        onVisible={updateShownImageIndex}
+        imagesRef={imagesRef}
+        onImageChanged={onImageChanged}
+      />
+      <BpkPageIndicator
+        currentIndex={shownImageIndex}
+        totalIndicators={images.length}
+        // eslint-disable-next-line @skyscanner/rules/forbid-component-props
+        className={getClassName('bpk-carousel-slides-indicator', indicatorClassName)}
```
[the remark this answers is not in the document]
KathyWang0208: using `getClassName('bpk-carousel-slides-indicator'` to change the location of the page indicator on image, as `variant` in `BpkPageIndicator` is used to set the `background-color` of the dot, if removing it, we can't see it in Carousel, like
image

Adding one more `indicatorClassName` is used to adjust the bottom's value, because bottom's value in `bpk-carousel-slides-indicator` does not meet the hotel funnel style.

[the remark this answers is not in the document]
olliecurtis: Thanks! This could indicate a potential issue then with the indicator component as this should have been built to work over images already, so we should probably look at this as part of this work to make sure it works as expected, as eventually className props will be removed from components to avoid CSS non-determinism issues

So we will need to fix the component to make this work :)

packages/bpk-component-carousel/src/BpkCarouselImage.module.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.bpk-carousel-image {
+  display: inline-flex;
+  width: 100%;
+  min-width: 0;
+  height: 100%;
+  min-height: 0;
+  justify-content: center;
+  scroll-snap-align: start;
+  scroll-snap-stop: always;
```
olliecurtis, line 27: Can we check these in Safari as according to https://caniuse.com/?search=scroll-snap-stop it seems it isn't supported in Safari 14 which is our current lowest version supported

olliecurtis:
Can we check the TalkBack and VoiceOver on this component?

Doing a check it looks like initially to work but when you get to the last image it then goes through the indicator component individually and you have to cycle through the whole component which doesn't have any descriptions to them or are selectable like the page indicator component

packages/bpk-component-carousel/src/BpkCarouselImage.module.scss
KathyWang0208: Yeah, I tried to remove `scroll-snap-stop` setting, the carousel doesn't break, could we leave it here to cover most of users to have a good experience?

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -57,91 +57,114 @@ const BpkPageIndicator = ({
   showNav,
   totalIndicators,
   variant,
-}: Props) => (
-  // $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
-  <div className={className}>
+}: Props) => {
+  const isOverImage = variant === VARIANT.overImage;
+  const isInteractive = !!onClick;
```
olliecurtis, line 62: Could you clarify or add a comment as the purpose of this?

olliecurtis:
Doing a quick test for interaction andTalkBack test on Android, it looks like when the component first loads image one is last and the last indicator dot is highlighted, so when you scroll though its like it starts on the last element and then when you get the last element the first dot is highlighted.

Wondering if this is related to this question about logic I had in the past is in the reverse order? https://github.com/Skyscanner/backpack/pull/3059/files#r1382019852

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -36,19 +36,19 @@ export const VARIANT = {
...
   showNav: ?boolean,
   currentIndex: number,
   totalIndicators: number,
   onClick: ?() => void,
-  indicatorLabel: string,
-  prevNavLabel: string,
-  nextNavLabel: string,
+  indicatorLabel: ?string,
+  prevNavLabel: ?string,
+  nextNavLabel: ?string,
```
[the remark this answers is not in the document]
KathyWang0208: In carousel, we don't need to pass the onClick function to `BpkPageIndicator`, and these label are used to render the button(only onClick is available, the button is rendered), so here I changed them as optional.

But I also worried about what you said, so I reverted the changes, and set the default en label in carousel(they won't be used in `BpkPageIndicator`, so they don't introduce the localization issues I think).

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
```diff
@@ -36,19 +36,19 @@ export const VARIANT = {
 };
 
 export type Props = {
-  className: ?string,
```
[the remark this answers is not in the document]
KathyWang0208: I was hesitant to get rid of it because a recent rule is not to recommend `className`.
Based on your comment, I am confident to revert the change, thank you.

packages/bpk-component-page-indicator/src/BpkPageIndicator.js
KathyWang0208: For `isOverImage`, what I did this change before I was thinking this component should include the according to `css` when customer pass `VARIANT.overImage ` and in order to be compatible with the custom bottom spacing, customer need pass one more prop `bottom`, it seems too many changes for this existing component, so I planed to move this css for `overImage` to `BpkCarousel`, what do you think about it?

For `isInteractive`, I will add some comments.

KathyWang0208:
> Doing a quick test for interaction andTalkBack test on Android, it looks like when the component first loads image one is last and the last indicator dot is highlighted, so when you scroll though its like it starts on the last element and then when you get the last element the first dot is highlighted.
> 
> Wondering if this is related to this question about logic I had in the past is in the reverse order? https://github.com/Skyscanner/backpack/pull/3059/files#r1382019852

Yes, in order for the scroll snap to create an infinite loop, render extra two images.

olliecurtis (approved):
LGTM!

Thank you

PR #3060, opened 2023-11-03
[ARGG-851] Fix withScrimmedPortal component for deterministic order of execution

arnaugm:
Provide an `isPortalReady` prop to notify when the component inside the portal is ready to be used. This may be necessary to interact with the content of the component in a `useEffect` hook, for example to set the focus on mount.

- [x] Component `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-scrim-utils/README.md
+++ b/packages/bpk-scrim-utils/README.md
@@ -18,3 +18,29 @@ const Box = props => (
 
 const BoxWithScrim = withScrim(Box);
 ```
+
+The version using a [React portal](https://react.dev/reference/react-dom/createPortal) renders the wrapped component in a different part of the DOM. It also provides an `isPortalReady` prop to notify when the component inside the portal is ready to be used. This may be necessary to interact with the content of the component in a `useEffect` hook, for example to set the focus on mount.
+
+```js
+import { withScrimmedPortal } from '@skyscanner/backpack-web/bpk-scrim-utils';
+
+const Box = props => {
+  const dialogRef = useRef(null);
+  const { isPortalReady, onClose } = props;
+
+  useEffect(() => {
+    if (isPortalReady) {
+      dialogRef.current?.focus();
+    }
+  }, [isPortalReady]);
+
+  return (
+    <div>
+      <BpkButton ref={dialogRef} onClick={onClose}>Close</BpkButton>
+      Hello in a portal
+    </div>
+  );
+};
+
+const BoxWithScrimmedPortal = withScrimmedPortal(Box);
+```
--- a/packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
+++ b/packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
@@ -42,13 +42,13 @@ const getPortalElement = (target: (() => HTMLElement | null) | null | undefined)
 
 const withScrimmedPortal = (WrappedComponent: ComponentType<ScrimProps>) => {
     const Scrimmed = withScrim(WrappedComponent);
-    let portalElement = document.body;
 
     const ScrimmedComponent = ({ renderTarget, ...rest}: Props) => {
+        const portalElement = getPortalElement(renderTarget);
+
         const [isPortalReady, setIsPortalReady] = useState(false);
 
         useEffect(() => {
-            portalElement = getPortalElement(renderTarget);
             setIsPortalReady(true);
           }, []);
 
--- a/packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
+++ b/packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
@@ -17,7 +17,7 @@
  */
 
 import type { ComponentType} from 'react';
-import { useEffect, useRef } from 'react';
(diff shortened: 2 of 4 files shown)


packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
```diff
@@ -17,56 +17,51 @@
...
-    
+
     if (document.body) {
         return document.body;
     }
     throw new Error('Render target and fallback unavailable');
 }
 
-const withScrimmedPortal = (WrappedComponent: ComponentType<ScrimProps>) => {
+const withScrimmedPortal = (WrappedComponent: ComponentType<Props>) => {
```
arnaugm, line 44: not sure about this change, is correct? if not TS was complaining

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
```diff
@@ -17,56 +17,51 @@
...
         if (!node.current) {
-            node.current = document.createElement('div');            
+            node.current = document.createElement('div');
         }
 
         useEffect(() => {
-            const portalElement = getPortalElement(renderTarget);
-            
-            if (node.current) {
-                portalElement.appendChild(node.current);
```
arnaugm, line 57: this is not necessary any more since the state update already triggers the render

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
olliecurtis: Looks right to me :)

olliecurtis (approved):

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
```diff
@@ -17,56 +17,51 @@
...
+    let portalElement = document.body;
 
     const ScrimmedComponent = ({ renderTarget, ...rest}: Props) => {
+        const [isPortalReady, setIsPortalReady] = useState(false);
+
         const node = useRef<HTMLDivElement | null>(null);
 
         if (!node.current) {
-            node.current = document.createElement('div');            
+            node.current = document.createElement('div');
```
anambl, line 54: Is this still used anywhere?

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
```diff
@@ -17,56 +17,51 @@
...
 const getPortalElement = (target: (() => HTMLElement | null) | null | undefined) => {
     const portalElement = target && typeof target === 'function' ? target() : null;
-    
+
     if (portalElement) {
         return portalElement;
     }
-    
+
     if (document.body) {
```
anambl, line 38: I guess we don't need this bit anymore.. or we can restrucutre the code a little bit as the default will be `document.body` anyways?

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
anambl: Hmmmm 🤔 I think `ScrimProps` should've been ok as `Props` just has `renderTarget` in addition, but the wrapped component won't accept `renderTarget`

arnaugm: true, changed, it doesn't seem to complain now

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
arnaugm: no, missed that one

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
arnaugm: this one makes sense I think, it makes very clear the purpose of the `getPortalElement` method

packages/bpk-scrim-utils/README.md
```diff
@@ -17,4 +17,30 @@ const Box = props => (
...
-```
\ No newline at end of file
+```
+
+The version using a [React portal](https://react.dev/reference/react-dom/createPortal) renders the wrapped component in a different part of the DOM. It also provides an `isPortalReady` prop to notify when the component inside the portal is ready to be used. This may be necessary to interact with the content of the component in a `useEffect` hook, for example to set the focus on mount.
+
+```js
+import { withScrimmedPortal } from '@skyscanner/backpack-web/bpk-scrim-utils';
+
+const Box = props => {
```
arnaugm, line 27: example for the new component

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
anambl: but we also set it in line 45 to `document.body`

anambl: so maybe we can call `getPortalElement` in L45 to set the default value?

arnaugm: adjusted now

anambl (approved):

PR #3085, opened 2023-11-15
[ARGG-876] Add support for withScrimmedPortal with SSR

anambl:

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
+++ b/packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
@@ -19,6 +19,7 @@
 import '@testing-library/jest-dom';
 import { render, within, screen } from '@testing-library/react';
 import { useEffect, useState } from 'react';
+import { renderToString } from 'react-dom/server';
 
 import withScrimmedPortal from './withScrimmedPortal';
 import type { Props } from './withScrimmedPortal';
@@ -104,5 +105,49 @@ describe('withScrimmedPortal', () => {
     );
 
     expect(screen.getByText('Wrapped Component / portal is not ready yet / portal is now ready')).toBeInTheDocument();
+  });
+
+  it('renders the wrapped component inside a portal correctly when runOnServer is false (default)', () => {
+    const DialogContent = () => <div>Dialog content</div>;
+    const ScrimmedComponent = withScrimmedPortal(DialogContent);
+
+    render(
+      <div id="pagewrap">
+        <div> Content hidden from AT</div>
+        <ScrimmedComponent
+          getApplicationElement={() => document.getElementById('pagewrap')}
+        />
+      </div>
+    );
+    expect(document.body).toMatchSnapshot();
+  });
+
+  it('renders the wrapped component inside a portal correctly when runOnServer is true', () => {
+    const DialogContent = () => <div>Dialog content</div>;
+    const ScrimmedComponent = withScrimmedPortal(DialogContent);
+
+    render(
+      <div id="pagewrap">
+        <div> Content hidden from AT</div>
+        <ScrimmedComponent
+          getApplicationElement={() => document.getElementById('pagewrap')}
+        />
+      </div>
+    );
+    expect(document.body).toMatchSnapshot();
+  });
+
+  it('throws an error when runOnServer is false and component is rendered on the server', () => {
+    const DialogContent = () => <div>Dialog content blablablabla</div>;
+    const ScrimmedComponent = withScrimmedPortal(DialogContent);
+
+    expect(() => renderToString(
+      <div id="pagewrap">
+        <div> Content hidden from AT</div>
+        <ScrimmedComponent
+          getApplicationElement={() => document.getElementById('pagewrap')}
+        />
+      </div>
+    )).toThrow();
   });
 });
--- a/packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
+++ b/packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
@@ -21,10 +21,15 @@ import { useState, useEffect } from 'react';
 import { createPortal } from 'react-dom';
 
 import withScrim from './withScrim';
+import BpkScrim from './BpkScrim';
 import type { Props as ScrimProps } from './withScrim';
 
 export type Props = ScrimProps & {
     renderTarget?: (() => HTMLElement | null) | null;
+    /**
+     * If true, the component will run on both server and client.
+     */
+    runOnServer?: boolean;
 };
 
 const getPortalElement = (target: (() => HTMLElement | null) | null | undefined) => {
@@ -37,22 +42,34 @@ const getPortalElement = (target: (() => HTMLElement | null) | null | undefined)
     if (document.body) {
         return document.body;
     }
-    throw new Error('Render target and fallback unavailable');
+    throw new Error('Render target and fallback unavailable. The ScrimmedComponent uses a portal making it unsuitable for SSR. If you are trying to render on SSR, please pass the `runOnServer` prop.');
 }
 
 const withScrimmedPortal = (WrappedComponent: ComponentType<ScrimProps>) => {
     const Scrimmed = withScrim(WrappedComponent);
 
-    const ScrimmedComponent = ({ renderTarget, ...rest}: Props) => {
-        const portalElement = getPortalElement(renderTarget);
-
+    const ScrimmedComponent = ({ renderTarget, runOnServer = false, ...rest}: Props) => {
         const [isPortalReady, setIsPortalReady] = useState(false);
 
         useEffect(() => {
             setIsPortalReady(true);
           }, []);
 
-        return createPortal(<Scrimmed {...rest} isPortalReady={isPortalReady} />, portalElement);
+        /**
+         * The following code runs only on the client.
+         * If consumer passes runOnServer as false (meaning they want to render the component on client only), we render the component as normal.
+         * If consumer passes runOnServer as true (meaning they want to render the component on server and client), we render the component as normal once the component has been mounted.
+         * This is to ensure the snapshotted markup (initial render before the component has been mounted) is the same on both server and client.
+         */
+        if (!runOnServer || isPortalReady) {
+            const portalElement = getPortalElement(renderTarget);
+            return createPortal(<Scrimmed {...rest} isPortalReady={isPortalReady} />, portalElement);
+        }
+
+        /**
+         * The following code will run on both server and on the intial render on the client when consumers opts for both SSR and CSR i.e. by passing the `runOnServer` prop.
+         */
+        return <BpkScrim />;
     }
 
     return ScrimmedComponent;
--- a/packages/bpk-scrim-utils/README.md
+++ b/packages/bpk-scrim-utils/README.md
@@ -21,6 +21,8 @@ const BoxWithScrim = withScrim(Box);
 
 The version using a [React portal](https://react.dev/reference/react-dom/createPortal) renders the wrapped component in a different part of the DOM. It also provides an `isPortalReady` prop to notify when the component inside the portal is ready to be used. This may be necessary to interact with the content of the component in a `useEffect` hook, for example to set the focus on mount.
 
+The `withScrimmedPortal` runs by default only on client due to the use of the portal. However, it accepts the `runOnSSR` prop which, when passed, renders a scrim on the server to block users from interacting with the page and making it evident that the page is not interactive.
+
 ```js
 import { withScrimmedPortal } from '@skyscanner/backpack-web/bpk-scrim-utils';
 
--- a/packages/bpk-scrim-utils/src/__snapshots__/withScrimmedPortal-test.tsx.snap
+++ b/packages/bpk-scrim-utils/src/__snapshots__/withScrimmedPortal-test.tsx.snap
@@ -1,5 +1,61 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`withScrimmedPortal renders the wrapped component inside a portal correctly when runOnServer is false (default) 1`] = `
+<body
+  style="overflow: hidden;"
+>
+  <div>
+    <div
+      aria-hidden="true"
+      id="pagewrap"
+    >
+      <div>
+         Content hidden from AT
+      </div>
+    </div>
+  </div>
+  <div
+    class="bpk-scrim-content"
+  >
+    <div
+      class="bpk-scrim"
+      role="presentation"
+    />
+    <div>
+      Dialog content
+    </div>
+  </div>
+</body>
+`;
+
+exports[`withScrimmedPortal renders the wrapped component inside a portal correctly when runOnServer is true 1`] = `
+<body
+  style="overflow: hidden;"
+>
+  <div>
+    <div
+      aria-hidden="true"
+      id="pagewrap"
+    >
+      <div>
+         Content hidden from AT
+      </div>
+    </div>
+  </div>
+  <div
+    class="bpk-scrim-content"
+  >
+    <div
+      class="bpk-scrim"
+      role="presentation"
+    />
+    <div>
+      Dialog content
+    </div>
+  </div>
+</body>
+`;
+
 exports[`withScrimmedPortal renders the wrapped component inside a portal correctly with fallback to document.body 1`] = `
 <body
   style="overflow: hidden;"


packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
```diff
@@ -37,22 +42,34 @@ const getPortalElement = (target: (() => HTMLElement | null) | null | undefined)
...
           }, []);
 
-        return createPortal(<Scrimmed {...rest} isPortalReady={isPortalReady} />, portalElement);
+        /**
+         * The following code runs only on the client.
+         * If consumer passes runOnServer as false (meaning they want to render the component on client only), we render the component as normal.
+         * If consumer passes runOnServer as true (meaning they want to render the component on server and client), we render the component as normal once the component has been mounted.
+         * This is to ensure the snapshotted markup (initial render before the component has been mounted) is the same on both server and client.
+         */
+        if (!runOnServer || isPortalReady) {
```
runmoore, line 64: Would we expect a consumer to be aware of where the component is being rendered (server or browser) and act accordingly to pass a prop in? If so, this is a code smell to me.

Modern web frameworks like Remix have no real concept of "running on server" or "running on browser". The expectation is all components are executed in both environments and it's common practice for components to be required to be self aware of the where they are being rendered, and thus act accordingly.

What's the reason we don't want this to be self aware?

anambl: > Modern web frameworks like Remix have no real concept of "running on server" or "running on browser". The expectation is all components are executed in both environments

I didn't know that, but I think that's still ok as then you just pass the `runOnServer` prop which means it runs on both (not just server)? we can make that the default instead (this component is not used anywhere, so we have the freedom of making any change to it).

> and it's common practice for components to be required to be self aware of the where they are being rendered, and thus act accordingly.

How can we tell if a component is running on server?
I guess my thoughts here were that consumers can choose whether they want to render on both and use the `BpkScrim` on server, OR just render it on client.

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
```diff
@@ -105,4 +106,48 @@ describe('withScrimmedPortal', () => {
...
+
+    render(
+      <div id="pagewrap">
+        <div> Content hidden from AT</div>
+        <ScrimmedComponent
+          getApplicationElement={() => document.getElementById('pagewrap')}
+        />
+      </div>
+    );
+    expect(document.body).toMatchSnapshot();
```
arnaugm, line 122: can we use more specific expects? snapshots tend to offer a poor testing experience in the long term https://medium.com/@sapegin/whats-wrong-with-snapshot-tests-37fbe20dfe8e

anambl: Hmm, I agree sometimes the snapshot tests are pretty useless, but in this case it really is exactly the tree that we want to check, but I guess we can change it to something like - element X is a descendant of pagewrap, element Y is not.

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
arnaugm, line 64: I'm a bit confused about this, what really decides if we enter here is the `isPortalReady`, what's the purpose of `runOnServer` exactly?

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
```diff
@@ -105,4 +106,48 @@ describe('withScrimmedPortal', () => {
...
+  });
+
+  it('renders the wrapped component inside a portal correctly when runOnServer is true', () => {
+    const DialogContent = () => <div>Dialog content</div>;
+    const ScrimmedComponent = withScrimmedPortal(DialogContent);
+
+    render(
+      <div id="pagewrap">
+        <div> Content hidden from AT</div>
+        <ScrimmedComponent
```
arnaugm, line 132: where is the `runOnServer` prop?

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
anambl: > what's the purpose of runOnServer exactly?

Because if you want to render this component only on client side, then you'll just run this code directly.
However, if you want to render it on both client and server, the behaviour will be slightly different on client - on the first render it will be rendering the `BpkScrim`, and then once it's been mounted it will be rendering the portal.
That is because if you want to render on both server and client, the initial render needs to be exactly the same, otherwise we get a warning from React and might run into some unexpected issues. But if we only render on client, we don't need to care about it so we just render it normally

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
anambl: Oh... :/
I think I just tabbed on the suggestion from Copilot and missed that it didn't add the prop 😅

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
runmoore: Agree with Arnau here, snapshot testing is against our advice, see [our documentation](https://web-documentation.skyscannertools.net/testing/unit-testing-granular-details.html#snapshot-testing):

> TLDR: Avoid using snapshots for testing components where possible
>
> Good tests encode the developers intent, but snapshot tests lack this. Inevitably any change to the component breaks the snapshot, so engineers become habituated to blindly accepting changes to snapshots without verifying whether the change is desired or not. This is particularly bad for large snapshots.

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
runmoore: > I didn't know that, but I think that's still ok as then you just pass the runOnServer prop which means it runs on both (not just server)?

Then the naming here should be improved, `runOnServer` suggests only the server.

But it still raises the question, do we _need_ flexibility here? Why can't this component decide what should happen depending on the environment?  

> How can we tell if a component is running on server?

common practice is to check for the presence of window, but it depends what the requirements are. If we need the `document` to be available, we should check for that.

```js
if (typeof window === 'undefined')
```

arnaugm: totally agree with Chris and probably that's the source of my confusion in the other comment

anambl: > Then the naming here should be improved, runOnServer suggests only the server.

I thought we always run on either just client, or both client and server so it's clear that this will run on both (although also added a comment above to explain it)

> But it still raises the question, do we need flexibility here? Why can't this component decide what should happen depending on the environment?

The component does decide what should happen depending on the env, I just didn't want to force the `BpkScrim` on the initial render on the client unless required i.e. to match the server initial render also thinking whether maybe in the future we want to render something else on the server, not the `BpkScrim` 

Maybe this comment helps here https://github.com/Skyscanner/backpack/pull/3085/files#r1395443998

anambl: I think this is mostly for any project that only runs on the client. Is that not common? Do all our microsites run on both client and server?

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
anambl: I've removed the snapshot tests and added new tests

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
runmoore: >  Do all our microsites run on both client and server completely? 

There's a big mix. Banana is CSR only. Falcon and hotels do SSR. OCs SSR the HTML and deliver it when requested.

runmoore: Thanks for the comment @anambl, I think I understand the challenge you want to solve. This double render feels sub-optimal, wondering if this is a recommended approach from somewhere?

What's the specific warning you get? Is it a prod warning or just local dev?

anambl: I think it's actually okay as the `BpkScrim` is rendered as part of the `withScrim` anyways so all we do is add some new things once the component has been mounted 🤔
Most issues I've found will recommend in these cases to basically only return something in the component if it has been mounted 😅 so something like this https://github.com/vercel/next.js/discussions/17443#discussioncomment-637879

The warning is `Warning: Expected server HTML to contain a matching <div> in <div>`

runmoore: > I think it's actually okay

I don't doubt that it's _okay_ :) my concern is the double render is emitting a big code smell to me and I can imagine this run around will be hard to understand for maintainers in months and years to come

One question on this comment:

>  if you want to render it on both client and server, the behaviour will be slightly different on client - on the first render it will be rendering the BpkScrim

What's the difference between what's expected to be rendered on the server and in the client?

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
```diff
@@ -108,44 +110,47 @@ describe('withScrimmedPortal', () => {
...
           getApplicationElement={() => document.getElementById('pagewrap')}
+          runOnServer
         />
       </div>
     );
-    expect(document.body).toMatchSnapshot();
+    expect(document.body).toContainElement(screen.getByTestId('dialog-content'));
   });
 
   it('throws an error when runOnServer is false and component is rendered on the server', () => {
```
arnaugm, line 144: could we also have the equivalent test when `runOnsServer` is true?

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
```diff
@@ -108,44 +110,47 @@ describe('withScrimmedPortal', () => {
   });
 
   it('renders the wrapped component inside a portal correctly when runOnServer is false (default)', () => {
-    const DialogContent = () => <div>Dialog content</div>;
+    const DialogContent = () => <div data-testid="dialog-content">Dialog content</div>;
```
arnaugm, line 113: Reading the tests it more evident that the prop name is confusing, we say runOnServer and we are checking that it renders on client, maybe we could use `ssrCompatible` or similar. Still if we could make the component self aware of the environment and avoid passing the prop it would be even better.

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
runmoore: Did you look at just [suppressing the warning](https://react.dev/reference/react-dom/hydrate#suppressing-unavoidable-hydration-mismatch-errors)? 

It's an expected difference, so as long as we add code comments explaining the suppression I think we've covered the requirement here?

runmoore: > I'm a little bit confused about what you mean by double render

image

We're doing this, but in a weird round about way to get around a react warning that we can supress

>  I don't think this counts as an expected difference 

By expected, I mean the component is set up to render something different on the server and client. It doesn't matter what that difference is, the DOM is different, and react knows it's different because it's throwing the warning.

We're going round in circles here, and I think it's clear me and @arnaugm have reservations about this aproach.

To try and get a resolution quickly, is there a reason you don't want to or can't suppress the warning?

anambl: > We're doing this, but in a weird round about way to get around a react warning that we can supress

Why do you think it's in a weird round about way? I think it is exactly what the example is doing. The only difference is that we have this extra prop which consumers can pass if they render their component **only** on the client, but it's not that important and I can remove it if it causes confusion.

> By expected, I mean the component is set up to render something different on the server and client. 

In that sense it is expected, but the React team advises suppressing the warning as the last resort and their example actually makes sense because they're using timestamps which of course will be different. Our use case is different imo. The recommended approach is to render the same content on the initial render on server and client (which I had linked [above](https://react.dev/reference/react-dom/hydrate#handling-different-client-and-server-content)). There's a few articles about this, this one specifically compares these exact 2 solutions: https://blog.hao.dev/render-client-side-only-component-in-next-js

> To try and get a resolution quickly, is there a reason you don't want to or can't suppress the warning?

I am definitely in favour of _not_ suppressing any warnings, but I do agree we need to get a resolution quickly so if the preferred solution from both of you is to suppress it and no one else has strong feelings about it then it's fine.

runmoore: What does this PR look like if we remove the `runOnServer` prop but keep the `useEffect` on mount? 

Does react show the warning?

anambl: > Does react show the warning?

No, it does not.

So the purpose of the `runOnServer` prop (which I agree it needs to have a different name) was simply to make the default of this component be client-only as I think that makes most sense given it's using a portal. As you said earlier, banana is CSR only, but there's also projects that do conditional rendering and render some components CSR only so by having this prop by default the component will be working on client only.
But, if you want to run it on both client and server, you can pass this prop so on server you'll be getting the scrim first. The prop has nothing to do with the React warning. I've added it because I wanted this component to be client-only by default, but if it's rendered in a project that uses SSR, you can use it there too and you'll be getting the scrim on the server. So if we remove the prop, on CSR-only projects it will still be rendering the scrim first and then the component which doesn't really make much sense given we only need to do it if we hydrate, but that's not a big issue as it wouldn't even be visible to the user so I'm ok with removing it if it causes confusion

anambl: Ah, yes, that'll work too, we don't need the prop

runmoore: > The prop has nothing to do with the React warning

Thanks for this clarification, I was under the impression the prop and react warning were linked.

> if we remove the prop, on CSR-only projects it will still be rendering the scrim first and then the component which doesn't really make much sense given we only need to do it if we hydrate, but that's not a big issue as it wouldn't even be visible to the user so I'm ok with removing it if it causes confusion

I understand your concern here, but I think it's a case of premature optimisation. As you say users won't notice this, but by adding this prop for CSR-only is introducing a lot of confusing logic (as evidenced by the comments on this PR). If we're comparing these two concerns, my preference is the one with the better DX.

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
```diff
@@ -37,22 +42,34 @@ const getPortalElement = (target: (() => HTMLElement | null) | null | undefined)
...
 
-    const ScrimmedComponent = ({ renderTarget, ...rest}: Props) => {
-        const portalElement = getPortalElement(renderTarget);
-
+    const ScrimmedComponent = ({ renderTarget, runOnServer = false, ...rest}: Props) => {
         const [isPortalReady, setIsPortalReady] = useState(false);
 
         useEffect(() => {
             setIsPortalReady(true);
           }, []);
```
mungodewar, line 56: Given all the discussion regarding double render and the general strategy being taken here, I recommend linking to the react docs: https://react.dev/reference/react-dom/hydrate#handling-different-client-and-server-content and popping some of the context from the comments directly into code.

packages/bpk-scrim-utils/src/withScrimmedPortal.tsx
anambl: I think like Mungo suggested above, the prop can just be replaced with checking if `window` is defined so we can achieve the same and remove the prop

anambl: I've removed the prop, although just to clarify - the point of the `runOnServer` prop was to set the expectation whether the component will run only on client only or if it'll run on both. I think using the `window` object helps us determine whether at run time it's running on server or on client, but I think achieving the same as the prop (i.e. if the intent is for the component to run on both client and server, or just on client so that for client-only projects we don't need to first render the scrim) makes it confusing as when it's running on client we'd need to check if it's already ran on server 🤔

runmoore: > we'd need to check if it's already ran on server 

Can you explain what you mean by this? I don't quite follow

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
```diff
@@ -26,27 +26,27 @@ import type { Props } from './withScrimmedPortal';
...
       </div>
     );
-    expect(document.body).toMatchSnapshot();
+    expect(document.body).toContainElement(screen.getByTestId('dialog-content'));
   });
 
   it('renders the wrapped component inside a portal with renderTarget provided', () => {
-    const WrappedComponent = () => <div>Wrapped Component</div>;
-    const ScrimmedComponent = withScrimmedPortal(WrappedComponent);
+    const DialogContent = () => <div data-testid="dialog-content">Dialog content</div>;
```
arnaugm, line 44: what's the difference between this test and the next one?

arnaugm, line 44: what's the difference between this test and the next one?

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
```diff
@@ -17,7 +17,7 @@
  */
 
 import '@testing-library/jest-dom';
-import { render, within, screen } from '@testing-library/react';
+import { render, within, screen, isInaccessible } from '@testing-library/react';
```
arnaugm, line 20: is `isInaccessible` used somewhere?

anambl: Ohh no, I'd added it, but then removed it
I'll remove this, I thought the linting would do that 😅

packages/bpk-scrim-utils/src/withScrimmedPortal-test.tsx
anambl: hmm, yeah, you're right. Nothing really, I'll remove it.

runmoore (approved):
@anambl and I had a chat on zoom. I'm happy with the implementation that is currently in the PR, so approving on he basis that @arnaugm also gives his approval too :)

PR #3093, opened 2023-11-20
Fix bpk-stylesheets :global transpilation

olliecurtis:
This PR fixes up the output that is generated for bpk-stylesheets to remove the `:global` from CSS for compatibility from browsers.

It turns out when we migrated from foundations to the single package and did a lot of the webpack plugin version bumps of some major versions, it broke the output and didn't strip out browser unreadable code. This change fixes this!


mungodewar (approved):
LGTM, `:global` gone from output as expected.

PR #3094, opened 2023-11-20
changed typing declarations to match `common-types.ts` file

sambocharov:
Changed typing declarations in `common-types.d.ts` to match `common-types.ts`

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-info-banner/src/common-types.d.ts
+++ b/packages/bpk-component-info-banner/src/common-types.d.ts
@@ -19,26 +19,32 @@
 import type { FunctionComponent, ReactNode } from 'react';
 
 export declare const ALERT_TYPES: {
-  readonly PRIMARY: 'primary';
-  readonly SUCCESS: 'success';
-  readonly WARN: 'warn';
-  readonly ERROR: 'error';
-  readonly NEUTRAL: 'neutral';
+  readonly SUCCESS: 'success',
+  readonly WARNING: 'warning',
+  readonly ERROR: 'error',
+  readonly INFO: 'info',
+};
+export declare const STYLE_TYPES: {
+  readonly DEFAULT: 'default',
+  readonly ON_CONTRAST: 'onContrast',
 };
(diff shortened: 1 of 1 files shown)


packages/bpk-component-info-banner/src/common-types.d.ts
```diff
@@ -19,26 +19,32 @@
...
+};
+export declare const STYLE_TYPES: {
+  readonly DEFAULT: 'default',
+  readonly ON_CONTRAST: 'onContrast',
 };
 export type AlertTypeValue = (typeof ALERT_TYPES)[keyof typeof ALERT_TYPES];
+export type StyleTypeValue = (typeof STYLE_TYPES)[keyof typeof STYLE_TYPES];
 export type CommonProps = {
-  type: AlertTypeValue;
+  type?: AlertTypeValue;
```
olliecurtis, line 34: This value would always have a value right as we set a default in the Inner Banner should would never need to be optional here?

sambocharov: I hope I get it right, but won't `BpkInfoBanner` complain about typing if we'll remove optional here? I mean, we may not pass to `BpkInfoBanner` a `type` prop and it should still work and pass typechecking. Since `BpkInfoBanner` has `CommonProps` as type for props, if we'll make `type` required it would fail. According to [figma](https://www.figma.com/file/MYDR66UzVOdVLLxO0IWhIA/branch/xIcjYtdQ0EvRLkvfM6vH3D/BPK-Info-Banner?type=design&node-id=404-2367&mode=design&t=7KhcV4UQeSRiZaOR-0), if I get it right, it has a default value which we specify in `BpkInfoBannerInner`, as you mentioned.

olliecurtis: Ah yeah you are right my mistake here, apologies!

olliecurtis (approved):

PR #3102, opened 2023-11-29
Wrap styles of BpkLink and bpk-stylesheets in layers

Sybsw:
A test of using @layer in css to fix upstream a CSS specificity bug in footer

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-stylesheets/index.scss
```diff
@@ -16,65 +16,77 @@
...
+      -webkit-font-smoothing: antialiased;
+    }
   }
 }
 
-:focus-visible {
-  @include bpk-focus-indicator;
-}
+@layer utility {
+  /* stylelint-disable-next-line no-duplicate-at-import-rules, no-invalid-position-at-import-rule */
```
Sybsw, line 61: Had to disable two lint rules here, mixins needs imported twice, each for the two layers.
This could be solved by combining theme and utility into one layer

olliecurtis: I was going to say everything in this file should come under one layer that I would see as `base` as this is core across everything and acts as the base the system then builds on

packages/bpk-stylesheets/index.scss
```diff
@@ -16,65 +16,67 @@
  * limitations under the License.
  */
 
-@import '~normalize.css';
-@import '../bpk-mixins/index.scss';
+@layer base {
```
mungodewar, line 19: I suggest that before this line we define all the layers upfront, something like: 

`@layer base, components, overrides`;

olliecurtis:
Superseeded by #3111

PR #3103, opened 2023-11-30
[NOJIRA]: Migrate to sass-embedded

xalechez:
Please look through commits one-by-one

* Created modern-compatible mixins that are autogenerated using the [official migrator](https://sass-lang.com/documentation/cli/migrator)
* Added all generated content to gitignore
* Replaced imports with new mixins
* Applied [official migrator](https://sass-lang.com/documentation/cli/migrator) to all components
* Switched from node-sass to sass-embedded

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-datatable/src/BpkDataTable.module.scss
```diff
@@ -20,22 +20,6 @@
 
 $bpk-border-size-datatable: $bpk-border-size-xl * 2;
 
-@mixin bpk-datatable-border-left() {
-  @include _bpk-border-left(
```
xalechez, line 24: Private mixins are not supported in sass-embedded

packages/bpk-mixins/src/mixins/_badges.scss
```diff
@@ -17,7 +17,9 @@
  */
 
 @import '../bonds.scss';
+@import 'radii.scss';
```
xalechez, line 20: Missing imports, sass-embedded is more strict with them

scripts/scss/generate-v2.sh
```diff
@@ -0,0 +1,38 @@
...
+
+printf "✅  Done! \n \n"
+
+printf "⚙️ Running sass-migrator...\n"
+
+
+# Migrate all other files to new syntax
+printf "👉 Applying new module resolution...\n"
+
+sass-migrator --load-path=node_modules module packages/bpk-mixins-next/src/**/*.scss
```
xalechez, line 43: Applying official sass migrator to generate modern mixins

scripts/scss/upgrade.sh
```diff
@@ -0,0 +1,14 @@
+set -e
```
xalechez, line 17: script used to migrate components

packages/bpk-stylesheets/base.css
olliecurtis, line 1: This one will be good for us to keep, so that we can track to ensure before we have 'on demand' that nothing is broken here on transpile out

packages/package.json
```diff
@@ -46,9 +46,22 @@
     "react-window": "^1.8.7"
   },
   "peerDependencies": {
-    "node-sass": ">=4.12.0",
+    "node-sass": ">= 7",
+    "sass": "*",
+    "sass-embedded": "*",
```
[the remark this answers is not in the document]
xalechez: I think we can add smth like "< 2", however we should be compatible with v2 once they release it as we're switching to the new API in this PR

packages/bpk-mixins/src/mixins/_borders.scss
```diff
@@ -264,3 +264,19 @@
 @mixin bpk-border-left-xl($color, $inset: inset) {
   @include _bpk-border-left($bpk-border-size-xl, $color, $inset);
 }
+
+@mixin bpk-datatable-border-left($bpk-border-size-datatable) {
```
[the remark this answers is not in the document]
anambl: Hmm. I don't think I'd include them in the package to be exported given they're not meant to be used outside BPK and were just a way to avoid duplicating code in the datatable component 🤔

anambl:
Looks great!! A few questions:
- What is the reason we named it as `bpk-mixins-next`? I'm also wondering if we could document this as I'm slightly concerned at some point in the future people are going to be confused looking into the codebase for a folder named `bpk-mixins-next` or will try to import from `bpk-mixins`?
- How long are we planning to support the node-sass version of bpk-mixins?
- What are the plans for communicating this change and the existing of migration scripts?
- Have we tested these changes in BPK in another project? I'm assuming there's no conflicts between node-sass and sass-embedded so even if someone continues to use node-sass in their project, the fact that we use sass-embedded in our components should be fine?

packages/bpk-mixins/src/mixins/_borders.scss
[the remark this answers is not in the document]
xalechez: Moved back to the component and made two private mixins public

anambl:
> > What is the reason we named it as bpk-mixins-next?
> 
> We want to keep both versions as we don't want node-sass consumers to stop receiving the updates, that's why we have two versions. I've updated the readme file of bpk-mixins to reflect both ways to use them
> 
> > How long are we planning to support the node-sass version of bpk-mixins?
> 
> No idea at the moment, probably long enough 😆
> 
> > Have we tested these changes in BPK in another project? I'm assuming there's no conflicts between node-sass and sass-embedded so even if someone continues to use node-sass in their project, the fact that we use sass-embedded in our components should be fine?
> 
> For consumers there're no changes as Backpack components are transpiled and old bpk-mixins folder is unchanged

> We want to keep both versions as we don't want node-sass consumers to stop receiving the updates, that's why we have two versions. I've updated the readme file of bpk-mixins to reflect both ways to use them

That makes total sense. I was just wondering what the reason is behind called it `next` 😄

> No idea at the moment, probably long enough 😆

Is the plan to rename it to `bpk-mixins` once we get rid of the node-sass version? I would also say it would be good to plan the timeline for removing it completely as we have a history in Backpack of deprecating and not removing things and then we end up with a lot of code that it's not meant to be supported 😢

> For consumers there're no changes as Backpack components are transpiled and old bpk-mixins folder is unchanged

I just meant that you need to have both `sass-embedded` and `node-sass` installed 🤔

xalechez:
> I was just wondering what the reason is behind called it next
Not sure. Naming things is one of the two hardest problems in programming 😆 

> Is the plan to rename it to bpk-mixins once we get rid of the node-sass version?
Yes

> I would also say it would be good to plan the timeline for removing it completely as we have a history in Backpack of deprecating and not removing things and then we end up with a lot of code that it's not meant to be supported 😢

Agree, but not sure if we have a strict timeline as it's our own initiative

> I just meant that you need to have both sass-embedded and node-sass installed 🤔

No, if you see peerDependencies and peerDependenciesMeta, consumers should choose only one sass compiler. With Webpack sass-loader will prioritise one based on its internal rules (sass-embedded > sass > node-sass). Tools like vite will prioritise dart sass by default

olliecurtis (dismissed):

anambl:
>  > I was just wondering what the reason is behind called it next

> Not sure. Naming things is one of the two hardest problems in programming 😆

It makes me think of the `next` framework 😅 I don't think it's a blocker, so as long as both you and Ollie are ok with the naming then fine. Personally, I'd prefer something like `v2` which is what we generally use I think

> > I would also say it would be good to plan the timeline for removing it completely as we have a history in Backpack of deprecating and not removing things and then we end up with a lot of code that it's not meant to be supported 😢

> Agree, but not sure if we have a strict timeline as it's our own initiative

I think that's fine. We can still estimate how much time we'd want to give people to migrate.

packages/bpk-stylesheets/base.css
```diff
@@ -16,5 +16,4 @@
...
...it-input-placeholder{color:inherit;opacity:.54}::-webkit-file-upload-button{-webkit-appearance:button;font:inherit}}@layer bpk-base{@keyframes bpk-keyframe-spin{100%{transform:rotate(1turn)}}html{font-size:100%;box-sizing:border-box}*{box-sizing:inherit}*::before,*::after{box-sizing:inherit}body{color:#161616;font-family:"Skyscanner Relative",-apple-system,BlinkMacSystemFont,"Roboto","Oxygen","Ubuntu","Cantarell","Fira Sans","Droid Sans","Helvetica Neue",sans-serif;font-size:1rem;line-height:1.3rem}body &.scaffold-font-size{font-size:13px}body &.enable-font-smoothing{-webkit-font-smoothing:antialiased}:focus-visible{outline:.125rem solid #0062e3 !important;outline-offset:.125rem !important}.hidden,.hide{display:none !important}.visuallyhidden,.visually-hidden{position:absolute;width:1px;height:1px;margin:-1px;padding:0;border:0;overflow:hidden;clip:rect(0 0 0 0)}.visuallyhidden.focusable:active,.visuallyhidden.focusable:focus,.visually-hidden.focusable:active,.visually-hidden.focusable:focus{position:static;width:auto;height:auto;margin:0;overflow:visible;clip:auto}.invisible{visibility:hidden}.clearfix::before,.clearfix::after{content:"";display:table}.clearfix::after{clear:both}}
```
olliecurtis, line 19: Taking a look at the diff between this file (it being a painful thing), it seems as though there looks like there are some changes to the output rendering of the normalize selectors (see below), is this expected and will this have an impact on the end rendering?

Screenshot 2024-01-18 at 15 54 27

xalechez: I think it's kinda ok based on [this issue](https://github.com/sass/sass/issues/2845) and Percy screenshots being identical

olliecurtis: Looks like that issue tells a good story there and how its interpreted.

I checked with storybook and it looks to still work with and without the quotes as mentioned!

olliecurtis (approved):

olliecurtis (approved):

CONTRIBUTING.md
```diff
@@ -105,6 +109,18 @@ We use [CSS Modules](https://github.com/css-modules/css-modules) along with [BEM
...
 When creating (S)CSS files, follow the CSS Module naming convention by using the `.module.(s)css` extension.
 
+When creating or modifying SCSS files, follow these rules
+
+1. Use Modern SASS API
+   * Prefer `@use` instead of `@import`
+   * Prefer `math.div($a, $b)` instead of `$a / $b`. Add `@use sass:math` statement to the top of your file to make this function available
+   * Read more about [@use rule](https://sass-lang.com/documentation/at-rules/use/) and [SASS math functions](https://sass-lang.com/documentation/modules/math/)
+2. Use only what you need
+   * Instead of blank import of all mixins, import them on demand. E.g. if you need only colour tokens, add `@use '../unstable_bpk-mixins/bonds'` statement only
```
runmoore, line 119: what does the word "bond" mean in this context?

xalechez: That's the name of the file re-exporting tokens. Maybe @olliecurtis knows more about the historic reasons for this naming?

olliecurtis: This naming sadly predates me at Skyscanner and that original naming knowledge has left Skyscanner. Never really questioned it in the past and just left it as is as there was maybe a reason to it

runmoore: Something to think about then as this is now going from an internal API, to an external one proliferated across all microsites.

We either need a reason why "bond" is used, or think about restructuring so that the files are intuitive for consumers to use

xalechez: Maybe just rename to tokens?

anambl (approved):

PR #3110, opened 2023-12-06
[DTO-4949][BpkNavigationBar] Create on-dark variant of BpkNavigationBar & Add Typescript

Iain530:
### Description

[Figma link](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?type=design&node-id=4522-2521&mode=design&t=L7paVF2hOl5PXKk6-0)

- Added prop `barStyle` to BpkNavigationBar, BpkNavigationBarButtonLink, BpkNavigationBarIconButton
- Added `on-dark` style
- Migrated BpkNavigationBar to typescript

image

- [x] Component `README.md`
- [x] Tests


Iain530:
Build currently failing due to missing dependency, will be fixed after #3109 is merged

Iain530:
Percy changes look as expected 👍

packages/bpk-component-navigation-bar/src/themeAttributes.js
```diff
@@ -21,13 +21,22 @@
 export default [
   'navigationBarBackgroundColor',
   'navigationBarTitleColor',
+  'navigationBarOnDarkBackgroundColor',
```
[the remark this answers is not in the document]
anambl: Although just to double check with @olliecurtis who's been in the team for longer than I have - is that right, Ollie?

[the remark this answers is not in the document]
Iain530: Ah yeah I did wonder if they were ever used, should I remove all of the themable parts for the on-dark variant in that case?

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
```diff
@@ -29,22 +29,30 @@ import STYLES from './BpkNavigationBar.module.scss';
 
 const getClassNames = cssModules(STYLES);
 
+export const BAR_STYLES = {
+  'default': 'default',
+  onDark: 'on-dark',
+};
+export type BarStyle = $Values<BAR_STYLES>
```
anambl, line 36: Do you think you could convert it into TS as we are moving away from flow? We've added this bit in the contribution guidelines https://github.com/Skyscanner/backpack/blob/main/CONTRIBUTING.md#typescript but maybe it's not too visible 😢

packages/bpk-component-modal/src/__snapshots__/BpkModalInner-test.tsx.snap
anambl, line 1: Hmm this also means it's going to break snapshots likely for users of navBar, but I guess that's fine, we can add a note maybe in the changelog 🤔

packages/bpk-component-navigation-bar/src/themeAttributes.js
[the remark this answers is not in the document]
anambl: Yes, that would be great@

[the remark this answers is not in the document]
Iain530: Removed 🧹

packages/bpk-component-navigation-bar/src/BpkNavigationBar.js
Iain530: Migrated to typescript, does this mean it will need a major version change or still a minor?

anambl: I think it's ok to keep it as minor 🤔 Looking at the changelog, I see we've generally released these as minor.

packages/bpk-component-navigation-bar/src/BpkNavigationBar-test.tsx
```diff
@@ -38,22 +38,14 @@ describe('BpkNavigationBar', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
-  it('should render correctly with arbitrary props', () => {
-    const { asFragment } = render(
-      <BpkNavigationBar id="test" title="test" testid="arbitrary value" />,
-    );
-
-    expect(asFragment()).toMatchSnapshot();
-  });
```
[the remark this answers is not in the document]
anambl: Hmm, we actually have something in our decision log for this - https://github.com/Skyscanner/backpack/blob/main/decisions/inexact-rest.md
I'm not too sure what the reason is behind supporting arbitrary props in NavBar, but don't think we need to worry about it in this PR and we should keep the test and modify the type to accept arbitrary props 🤔

packages/bpk-component-navigation-bar/src/BpkNavigationBarButtonLink.tsx
```diff
@@ -16,33 +16,37 @@
...
+import { BAR_STYLES, type BarStyle } from './BpkNavigationBar';
 
 const getClassName = cssModules(STYLES);
 
-export type Props = {
-  children: Node,
-  onClick: (event: SyntheticEvent<>) => void,
-  className: ?string,
-};
+export interface Props extends ComponentProps<BpkButtonLink> {
```
anambl, line 31: Same here, it seems like we accept `rest` so we can add that to the type

packages/bpk-component-navigation-bar/src/BpkNavigationBarIconButton.tsx
```diff
@@ -16,31 +16,35 @@
...
+// @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
 import BpkIconButton from '../../bpk-component-close-button';
 import { cssModules } from '../../bpk-react-utils';
 
 import STYLES from './BpkNavigationBarIconButton.module.scss';
+import { BAR_STYLES, type BarStyle } from './BpkNavigationBar';
 
 const getClassName = cssModules(STYLES);
 
 export type Props = {
```
anambl, line 31: Same here, it seems like we accept `rest` so we can add that to the type

packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
```diff
@@ -60,49 +66,62 @@ const BpkNavigationBar = (props: Props) => {
...
-  title: PropTypes.node.isRequired,
+  title: PropTypes.oneOf([PropTypes.node, PropTypes.string]).isRequired,
   className: PropTypes.string,
   leadingButton: PropTypes.element,
   trailingButton: PropTypes.element,
   sticky: PropTypes.bool,
+  barStyle: PropTypes.oneOf(Object.values(BAR_STYLES)),
 };
 
 BpkNavigationBar.defaultProps = {
```
anambl, line 119: We can set the default props directly when destructuring the props rather than using this. We should also be getting rid of the propTypes now we're using TS, but since we export them it will be a breaking change so maybe I wouldn't add it to this PR. We do have an incoming breaking change though so maybe we can bundle it in the same release with that change in a separate PR?

packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
```diff
@@ -60,49 +66,62 @@ const BpkNavigationBar = (props: Props) => {
...
+        cloneWithClasses(
+          trailingButton,
+          'bpk-navigation-bar__trailing-item',
+          `bpk-navigation-bar__trailing-item--${barStyle}`,
+        )}
     </nav>
   );
 };
 
 BpkNavigationBar.propTypes = {
```
anambl, line 109: We should also be getting rid of the propTypes now we're using TS, but since we export them it will be a breaking change so maybe I wouldn't add it to this PR. We do have an incoming breaking change though so maybe we can bundle it in the same release with that change in a separate PR?

packages/bpk-component-navigation-bar/src/BpkNavigationBar-test.tsx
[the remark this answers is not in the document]
Iain530: Ah that makes sense, re-added now

packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
Iain530: Added the default values to the destructuring expect the nulls

packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
Iain530: Sounds like a good plan, I've updated propTypes with the barStyle prop for now, and will raise another PR after this one to remove them completely

anambl (approved):
LGTM!

PR #3111, opened 2023-12-08
[bpk-stylesheets][ARGG-892]: Adding CSS layers to central styles

olliecurtis:
This PR creates the setup for CSS @layer in Backpack that will be utilised to set the specificity of styles that are used by Backpack and consumers. The idea here is that with the @layer definition and wrapping the baseCSS in @layer we set the order of layers that would be lower in specifity order before classNames provided to components.

The first line in `packages/bpk-stylesheets/index.scss` refers to setting the order in which layers take precedence over so would be 

- `base` first, then `component` (which would be any component defined layers) and then `utility` css
- In the next change we then wrap the base core standardisation CSS inside the `base` layer as this is the one we always want to load CSS for first before anything else.

In following PRs we will be looking to apply the `@layer component` to each of our Backpack components to allow them to be loaded next in the chain.

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-stylesheets/index.scss
```diff
@@ -16,65 +16,69 @@
...
-@import '~normalize.css';
-@import '../bpk-mixins/index.scss';
+@layer base, component, utility;
 
-// stylelint-disable-next-line selector-max-type
-html {
-  font-size: $bpk-font-size-root;
-  box-sizing: border-box;
-}
+@layer base {
```
Sybsw, line 21: Might be better to have normalize in its own first 'normalize' or 'reset' layer as that seems to be the tradition in learning materials I've seen

olliecurtis:
Closing in favour of #3119

PR #3119, opened 2023-12-11
Wrapping components in CSS Cascade Layers

Sybsw:
LOOM-1101 - Adding layers around bpk components
- Wrapped bpktext scss in 'component' layer
- Most changes are exactly the same, wrapping all the scss file inside a 'component' layer

Designed to be merged after https://github.com/Skyscanner/backpack/pull/3111

- [ ] Component `README.md`
- [ ] Tests


Sybsw:
Closed https://github.com/Skyscanner/backpack/pull/3111 as its changes are included here

packages/bpk-component-accordion/src/BpkAccordion.module.scss
```diff
@@ -16,12 +16,14 @@
  * limitations under the License.
  */
 
-@import '../../bpk-mixins/index.scss';
+@layer component {
+  @import '../../bpk-mixins/index.scss';
```
mungodewar, line 20: Screenshot 2024-01-05 at 16 43 12

Is there a reason that we need `bpk-mixins/index.scss` inside the @layer statement. This should already be wrapped in a layer higher in the stylesheet. Namely `base`.

packages/bpk-stylesheets/index.scss
```diff
@@ -16,65 +16,72 @@
  * limitations under the License.
  */
 
-@import '~normalize.css';
-@import '../bpk-mixins/index.scss';
+@layer reset, base, component, utility;
```
dominicfraser, line 19: To prevent naming clashes in consumers who also use named layers could these all have the `bpk-` prefix? The latter 3 are highly likely to be names a consumer is/wishes to use.

I know this is WIP, just commenting ahead of names potentially becoming baked in and harder to swap out.

packages/bpk-component-accordion/src/BpkAccordion.module.scss
Sybsw: We can't do that because of scope but from recent discussions we have observed that mixins can be declared fine above, outside of the layers

packages/bpk-stylesheets/index.scss
Sybsw: We've recently encountered race-condition bugs here related to the repeated use of the name 'component' so yes I would agree that something like this would enable further protection in case the consumer uses these names. We did discuss that as an option ages ago but it didn't seem very consequential at the time.

olliecurtis: Thanks for the comment @dominicfraser let us take this away and chat about it.

> The latter 3 are highly likely to be names a consumer is/wishes to use.

This could also be influenced by the guidance we provide too of possible avoid these, as we will want to make sure whatever we use/provide to folks is consistent across the product too instead of just leaving it to consumers to come up with namings

Sybsw: Have added 'bpk-' prefix

packages/bpk-component-accordion/src/BpkAccordion.module.scss
Sybsw: Correction on this for anyone viewing later, we found that we can actually import the mixins outside of the layers and we will make this change in the near future

olliecurtis (approved):
LGTM! 👍

PR #3120, opened 2023-12-13
[BpkAutosuggest]: Fixing autosuggest component

olliecurtis:
This change fixes the name of the exported component of the `BpkAutosuggestSuggestion` component, which currently causes our docs to crash as this component can't be found, See here: https://www.skyscanner.design/latest/components/autosuggest/web-1dnVX8RX

This is a strange one of how this has been working for years without modification as nothing has recently changed with this component and the naming hasn't been an issue but there is a mismatch in namings which causes storybook to not render the story. See here: https://backpack.github.io/storybook 

This fix then aligns all the naming as per our docs and fixes the error with storybook not rendering.

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
+++ b/packages/bpk-component-autosuggest/src/BpkAutosuggestSuggestion.js
@@ -36,7 +36,7 @@ type Props = {
   tertiaryLabel: ?string,
 };
 
-const BpkSuggestion = (props: Props) => {
+const BpkAutosuggestSuggestion = (props: Props) => {
   const classNames = [getClassName('bpk-autosuggest__suggestion')];
   const { className, icon, indent, subHeading, tertiaryLabel, value, ...rest } =
     props;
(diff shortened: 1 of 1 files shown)


anambl (commented):
What 🤯 😱 I'm very confused. It's not a named export, why would this not work 😮

olliecurtis:
> What 🤯 😱 I'm very confused. It's not a named export, why would this not work 😮

Same here @anambl its been working fine for a long time with no changes but if you go to the deployed storybook or even our docs site for the autosuggest you will see its not there 😱 and checking in the dev tools it says the error is because it can't find the component because the names are different!

anambl:
> > What 🤯 😱 I'm very confused. It's not a named export, why would this not work 😮
> 
> Same here @anambl its been working fine for a long time with no changes but if you go to the deployed storybook or even our docs site for the autosuggest you will see its not there 😱 and checking in the dev tools it says the error is because it can't find the component because the names are different!

My guess is that it's something to do with Storybook and their build as otherwise it wouldn't make any sense..

anambl (approved):

PR #3122, opened 2023-12-15
[DTO-4949][BpkNavigationBar] Remove prop-types from navbar

Iain530:
Follow up to #3110 - removing the `propTypes` now we have typescript in the nav bar. Separate PR as we initially thought this was a breaking change, but turns out it's not

- [x] Component `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
+++ b/packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
@@ -18,7 +18,6 @@
 
 import type { ReactElement, ReactNode } from 'react';
 import { cloneElement } from 'react';
-import PropTypes from 'prop-types';
 
 import { cssModules } from '../../bpk-react-utils';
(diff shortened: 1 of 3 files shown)


anambl:
Are they not imported and exported in the index file?

anambl:
> > Are they not imported and exported in the index file?
> 
> It seems to be the component props that are exported in the index file:
> 
> https://github.com/Skyscanner/backpack/blob/b2f042331ae29a89d772eb5cd536762ce0e79f5f/packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx#L35-L44

Ohhhhh you're right!! I totally missed that. In that case I don't actually think it's a breaking change? 😅

Iain530:
> Ohhhhh you're right!! I totally missed that. In that case I don't actually think it's a breaking change? 😅

Oh nice! I've updated the labels in that case 😄

anambl (approved):

PR #3142, opened 2024-01-04
[DTO-5004][BpkInfoBanner] Remove Alert Role

Iain530:
Currently when an info banner is render on a page, the screenreader will immediately switch to reading out the title text. This happens because we treat the info banner as an alert for screen readers. [The documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role) says the alert role shouldn’t be used for elements rendered on page load. Since this component is used more as a styling container than an alert dialog it should only be read in the normal page flow.

### Bug caused by alert role (sound on)

https://github.com/Skyscanner/backpack/assets/22656572/e665fc0f-cfd4-440d-bd97-96b61fa4cbf9

- [ ] Component `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
+++ b/packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
@@ -45,7 +45,6 @@ import type {
   CommonProps,
   OnDismissHandler,
   OnExpandToggleHandler,
-  StyleTypeValue,
   ExpandableBannerAction,
 } from './common-types';
 import { ALERT_TYPES, STYLE_TYPES } from './common-types';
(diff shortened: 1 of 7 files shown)


packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
```diff
@@ -189,7 +188,7 @@ const BpkInfoBannerInner = ({
       show={show}
       {...rest}
     >
-      <section className={sectionClassNames} role="alert">
+      <section className={sectionClassNames} role="presentation">
```
Iain530, line 191: `presentation` seemed like the most sensible value to me here as it removes it as a landmark which this probably. These banners are probably to be used as part of a section of the page and not contain a whole section themselves? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

olliecurtis (approved):

PR #3155, opened 2024-01-08
[ARGG-794][BpkDataTable] Remove deprecated props in BpkDataTable and migrate to TS

anambl:
Migration guide:
https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-datatable/docs/migrating.md

- [ ] Component `README.md`
- [x] Tests


examples/bpk-component-datatable/examples.tsx
```diff
@@ -74,8 +73,8 @@ const complexRows = [
 const onRowClick = (row) => alert(JSON.stringify(row));
 
 const sortFunc = (rowA, rowB) => {
-  const deskA = rowA.values.seat.desk;
-  const deskB = rowB.values.seat.desk;
+  const deskA = rowA.seat.desk;
```
anambl, line 76: This was a bit of a bug as the row object was the one coming from `react-table` which had a lot of unnecessary properties on it. I've created a function to abstract away the `react-table` API and the row will be just the object that the consumer passes

packages/bpk-component-datatable/src/BpkDataTable.tsx
```diff
@@ -122,17 +116,17 @@ const BpkDataTable = (props: Props) => {
     );
 
   const onRowClicked = (index: number) => {
-    if (rowSelected === index) {
-      updateRowSelected(undefined);
-    } else {
-      updateRowSelected(index);
-    }
     if (onRowClick !== undefined) {
+      if (rowSelected === index) {
```
anambl, line 120: moved this inside the `if` as previously it was triggered even if `onRowClick` was undefined which meant it still changed the colour to show it as selected

packages/bpk-component-datatable/src/BpkDataTable.tsx
```diff
@@ -15,20 +15,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* @flow strict */
 
-import PropTypes from 'prop-types';
 import type { KeyboardEvent } from 'react';
 import { useMemo, useState } from 'react';
+// @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
```
anambl, line 21: `react-table` is not typed. There are some community types, but they're not very accurate and are causing issues so went with `any` type whenever there's something from `react-table`

anambl:
The Percy snapshots for mobile view changed slightly. That's because the `react-table` column has some default properties as well one of which is `minWidth` set to 0. The BpkDataTable is not at all optimised for mobile which is where this could be an issue (it was an issue before as well as even though there was no `minWidth`, the headers were no longer aligned with the columns). We could just set the `minWidth` to undefined instead, but I don't think that's an improvement so I'd leave it as is. The other default values from column are fine and don't impact the component

olliecurtis (approved):

PR #3157, opened 2024-01-10
Update boilerplate to remove className, update contribution and add danger check

olliecurtis:

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,7 +92,7 @@ We also ship the type declaration files for all TypeScript components to ensure 
 Our current supported React version is 17.0.2, please be mindful when using React features that may not yet be supported.
 
 #### Design
-If you'd like to contribute a change to a React component, please first reach out to Koala team on Slack to discuss and agree on the proposed change. Make sure to add to your message the design (Figma file) and include examples for each state e.g. disabled, expanded etc. if applicable.
+If you'd like to contribute a change to a React component, please first reach out to the backpack team on Slack to discuss and agree on the proposed change. Make sure to add to your message the design (Figma file) and include examples for each state e.g. disabled, expanded etc. if applicable.
 
 #### Example components
 Look at existing components for code style inspiration. Here are some good examples to follow:
@@ -111,6 +111,7 @@ If you want to add a new component:
 
 1. Use `bpk-component-boilerplate` to create a new skeleton React component
 2. Our components where possible are written as function components, familiarise yourself using [React component guidelines](https://react.dev/reference/react/Component) for more guidance
+    - **For new components we restrict the use of `className` and `style` props to avoid allowing overwriting the component's styles and to ensure consistency across our product.**
 3. Create stories - each component has a set of stories living under `examples/bpk-component-{name}/stories.ts`. Stories should cover most visual variants of a component. Read more about Storybook stories [here](https://storybook.js.org/docs/react/writing-stories/introduction)
 4. Create tests
     - Visual regression tests - Each UI component's stories should also include a story that begins with the name `VisualTest` - these will then be picked up by Percy to run on CI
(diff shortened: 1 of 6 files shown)


packages/bpk-component-boilerplate/src/__snapshots__/BpkBoilerplate-test.tsx.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`BpkBoilerplate should render correctly 1`] = `
+exports[`BpkBoilerplate should not support custom class names 1`] = `
```
runmoore, line 3: what part of this test asserts we don't support className?

olliecurtis: Currently its just this snapshot, as that is how most of our tests are setup for in Backpack, something that as a future initative we need to invest the time into doing more assertions for tests.

However here this is just the boilerplate so it's not necessarily key as all the tests here and written in the test file are subject to change when someone generates a new component and modifies towards the component they build

olliecurtis: @runmoore have made some improvements to the tests now to not rely on snapshot testing for such a check :)

CONTRIBUTING.md
```diff
@@ -92,7 +92,7 @@ We also ship the type declaration files for all TypeScript components to ensure
 Our current supported React version is 17.0.2, please be mindful when using React features that may not yet be supported.
 
 #### Design
-If you'd like to contribute a change to a React component, please first reach out to Koala team on Slack to discuss and agree on the proposed change. Make sure to add to your message the design (Figma file) and include examples for each state e.g. disabled, expanded etc. if applicable.
+If you'd like to contribute a change to a React component, please first reach out to the backpack team on Slack to discuss and agree on the proposed change. Make sure to add to your message the design (Figma file) and include examples for each state e.g. disabled, expanded etc. if applicable.
```
soulcheck, line 95: Does contacting backpack via slack also apply to OSS contributors?

olliecurtis: No this is done via issues, but here we don't need to call this out because for design related, our designers aren't on OSS so this part is not applicable here

packages/bpk-component-boilerplate/src/BpkBoilerplate.tsx
```diff
@@ -23,17 +23,17 @@ import STYLES from './BpkBoilerplate.module.scss';
...
 
 export type Props = {
-  className?: string | null;
   [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
 };
-const BpkBoilerplate = ({ className = null, ...rest }: Props) => {
-  const classNames = getClassName('bpk-boilerplate', className);
+const BpkBoilerplate = ({ ...rest }: Props) => {
+
+  // eslint-disable-next-line no-param-reassign
```
soulcheck, line 30: I think you could avoid it by destructuring `className` separately from `rest`, but I guess `delete` is more explicit.

CONTRIBUTING.md
soulcheck: Makes sense.

packages/bpk-component-boilerplate/src/BpkBoilerplate.tsx
olliecurtis: Yeah we did used to have that but then its encouraging use or showing it could be available whereas here we want to prohibit it completely

soulcheck: As in 

```
{ className, ...rest }: Props
```

olliecurtis: We want to avoid showing it could be in the props and used, so doing it explicitly with force shows if someone tries to build with it later on it would through an IDE error

olliecurtis: If we drop `className` it can still be passed via `rest` so it could still be included, and this forces it out completely so that it can't be spread via `...rest`

soulcheck (approved):

PR #3178, opened 2024-01-17
[bpk-stylehseets]: Correcting base line height to be in line with typogrpahy system

olliecurtis:
When we did typography changes a year or so ago, we missed in the base `bpk-stylesheets` updating the line-height to our default line-height for our font which is `1.5rem`.

This PR fixes this! This may cause some layout adjustments by a couple of pixels but would cause percy to fail so it would be debatable if we should make it `breaking` for the previous reasons or `patch` because it fixes something we failed to do.

**Note: This change only affects those that aren't using Backpack styles or components for text, so providing that we are utilising Backpack properly this should have a lower blast radius than EVERYTHING**

- [ ] Component `README.md`
- [x] Tests


olliecurtis:
Pausing for now!

PR #3194, opened 2024-01-22
[BpkInput]: Updating input HoC for keyboard events

olliecurtis:
This PR fixes an issue with the `withOpenEvents` HoC in the `BpkInput` whereby the input could not be opened with either the `Return` or `Space` keys. 

This is due to using old logic whereby `keyCode` has been deprecated from the Keyboard event object and so should use `key` instead. Further reading [MDN KeyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)

Also fixes having `keyUp` when space is clicked to trigger the events as `keyUp` is not triggered when pressed but `keyDown`.

- [ ] Component `README.md`
- [x] Tests


examples/bpk-component-datepicker/examples.js
```diff
@@ -629,7 +629,6 @@ const DefaultVisualExample = () => (
       }}
       minDate={new Date(2020, 3, 1)}
       initiallyFocusedDate={new Date(2020, 3, 19)}
-      isOpen
```
olliecurtis, line 632: Removed this, so the calendar doesn't open by default and allows us to utilise the openEvents fix with a keyboard and improve A11y testing

packages/bpk-component-input/src/withOpenEvents.tsx
```diff
@@ -168,14 +167,7 @@ const withOpenEvents = <P extends object>(InputComponent: ComponentType<P>) => {
...
         onClick: withEventHandler(onOpen, onClick),
-        onKeyDown: withEventHandler(
-          handleKeyEvent(KEYCODES.ENTER, onOpen),
-          onKeyDown,
-        ),
-        onKeyUp: withEventHandler(
-          handleKeyEvent(KEYCODES.SPACEBAR, onOpen),
-          onKeyUp,
-        ),
+        onKeyDown: withEventHandler(handleKeyEvent(onOpen), onKeyDown),
```
soulcheck, line 170: Why does it complain about types? KeyboardEvent should be a subtype of UIEvent.

examples/bpk-component-datepicker/examples.js
gert-janvercauteren: can we re-add this back? 
We plan to run an experiment where we test the focus behaviour :D

olliecurtis: This is just for the examples so doesn't affect production and is good for us to do for our examples

packages/bpk-component-input/src/withOpenEvents.tsx
olliecurtis: I don't know why it is complaining about this, searching online isn't helpful for the typing of this and just only points at what KeyboardEvent is

packages/bpk-component-input/src/withOpenEvents.tsx
```diff
@@ -67,15 +67,14 @@ type EventHandlers = {
 type InputProps = ComponentProps<'input'> &
   Omit<EventHandlers, 'readOnly' | 'aria-readonly'>;
 
-const handleKeyEvent =
-  (keyCode: number, callback?: () => void) => (e: UIEvent) => {
-    if (e instanceof KeyboardEvent && e.keyCode === keyCode) {
```
olliecurtis, line 72: e was never coming back as an instanceof Keyboard event even though it is a subtype of UIEvent so mean it was always evaluating false and never running this code further!

packages/bpk-component-input/src/withOpenEvents.tsx
Sybsw: Discussed with @olliecurtis , this looks like the best solution

Sybsw (approved):

PR #3198, opened 2024-01-25
[DTO-5100] BpkChipGroup

Iain530:
[Figma link](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=30119%3A37736&mode=dev)

image

image

The BpkChipGroup is used to group a list of chips ([bpk-component-chip](https://github.com/Skyscanner/backpack/tree/main/packages/bpk-component-chip)) in either a wrapped layout or rail (single row with scroll. See image above or Figma or all examples.

## Changes

- Created new component BpkChipGroup

- [x] Component `README.md`
- [x] Tests


packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,195 @@
...
+  leadingAccessoryView?: ReactNode;
+  className?: string;
+  [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
+};
+
+export type ChipItem = {
+  component?: (props: BpkSelectableChipProps) => ReactElement;
+  onClick?: (selected: boolean, index: number) => void;
+  selected?: boolean;
+} & SingleSelectChipItem;
```
Iain530, line 53: The shape of this is a bit different to the figma at the moment but I've tried to keep it as close as possible: https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=30119%3A42328&mode=dev

Currently differs by:
- `icon` replaced by `leadingAccessoryView` as this is the prop in BpkSelectableChipProps, although this could be renamed to icon for this prop if preferred.
- `component` instead of `type` but could be made to use an enum to choose from `selectable`, `dropdown` and `dismissable` chips.

Since these just passed through to a BpkChip component then a `...rest` param seemed sensible at the time, although may not be neccessary now I'm looking back.

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,195 @@
...
+    >
+      {text}
+    </Component>
+  );
+
+  // TODO: Fix/remove the scroll indicator when style=CHIP_TYPES.onImage
+  // TODO: Fix shadows + focus indicators being cutoff when type = rail
+  const wrapRailInScroll = (children: ReactElement) =>
+    type === CHIP_GROUP_TYPES.rail ? (
+      <BpkMobileScrollContainer
```
Iain530, line 129: This BpkMobileScrollContainer seems like a nice way of making horizontal scrolls but it also causes a couple of issues in this case:

1. It adds this scrolling indicator which is nice as long as you've got the right background colour. So far I have kept it white for default chip style, and matched the on dark color to our contrast background color. However this would really restrict the backgrounds this can be used on. I'm thinking the simple solution would be to have a prop to disable these? 
image


2. The focus indicator gets cutoff. I've spent some time trying to get this working as expected but without much luck. We'll need to spend more time to solve this. We see the same problem affecting all of the `onImage` chips as they use a shadow border.
image
image

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,195 @@
...
+      {type === CHIP_GROUP_TYPES.rail && (
+        <BpkBreakpoint query={BREAKPOINTS.ABOVE_TABLET}>
+          <Nudger chipStyle={style} scrollContainerRef={scrollContainerRef} />
+        </BpkBreakpoint>
+      )}
+    </div>
+  );
+};
+
+export const BpkChipGroupState = ({ chips, ...rest }: ChipGroupProps) => {
```
Iain530, line 174: The figma didn't mention that we'd expose stateful versions of these components but I needed them for testing anyway so they're here. If we'd prefer not to expose these then they can easily be moved to the example stories.

packages/bpk-component-chip-group/src/Nudger.tsx
```diff
@@ -0,0 +1,104 @@
...
+  scrollContainerRef: MutableRefObject<HTMLElement | null>;
+  leading?: boolean;
+}
+
+
+const AlignedLeftArrowIcon = withButtonAlignment(ArrowLeft);
+const AlignedRightArrowIcon = withButtonAlignment(ArrowRight);
+
+// TODO: how many px to scroll on click?
+const SCROLL_DISTANCE = 100;
```
Iain530, line 54: Who would be best to talk to for guidance on this value? Should it be a prop?

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
```diff
@@ -0,0 +1,90 @@
...
+  // TODO: Should this be a new chip component type without text?
+  .bpk-sticky-chip {
+    padding-inline-end: 0;
+    padding-inline-start: bpk-spacing-md();
+
+    // TODO: better way to do this? Hide all contents of the chip except the leading icon.
+    span:not(:nth-of-type(1)) {
+      display: none;
+    }
+  }
```
Iain530, line 69: This was needed to implement the sticky chip only showing text on mobile. It seems a little hacky and made me think a new type of chip for this would be better. What do you think?

image

mungodewar:
Following on the conversation from slack: https://skyscanner.slack.com/archives/C0JHPDSSU/p1706274673682399?thread_ts=1706259605.819169&cid=C0JHPDSSU

I’m thinking that the composable pattern is better, as the prop is a hard coupling, however, we could make it looser by the `ChipGroup` rendering children and managing `Chip` props via something like context. 

eg there would be no need for the custom stickyChip code, the composition would be enough I think? 

```
<BpkChipGroup>
    <BpkChip sticky />
    {myOtherChips.map(chip => <BpkChip {...chip} />)}
</BpkChipGroup>
```

The dismissed example becomes a lot more concise: 

```
  return (
    <BpkChipGroup>
        <BpkChip disabled />
        <BpkChip dismissable />
        <BpkDropdownChip >
        <BpkChip />
        <BpkChip selected/>
    </BpkChipGroup>
  );
```

Semantically, the `BpkChipGroup` manages a group of `BpkChip`s. This reflects other web apis like `<select>` and `<option>`. 

I can think of future requirements/designs that "clumps" chips together or seperate them. Having this api allows easy extensibility & no change to `BpkChipGroup` would be required: 

```
  return (
    <BpkChipGroup>
        <BpkChip disabled />
        <BpkChip dismissable />
        <BpkChipSpacer />
        <BpkDropdownChip >
        <BpkChip />
        <BpkChip selected/>
    </BpkChipGroup>
  );
```
or

```
  return (
    <BpkChipGroup>
        <BpkChipSpacer >
             <BpkChip disabled />
             <BpkChip dismissable />
        </BpkChipSpacer >
        <BpkDropdownChip >
        <BpkChip />
        <BpkChip selected/>
    </BpkChipGroup>
  );
```

mungodewar:
There are a few different types of BpkChipGroup, could you give me a TLDR on when a consumer would use which one and why, are some of these just internal maybe? 

Screenshot 2024-01-26 at 14 31 19
Screenshot 2024-01-26 at 14 31 21

mungodewar:
I'm interested in `onItemClick` prop, is there a reason that we don't want that on all the different version of BpkChipGroup?

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
mungodewar: Feels a little weird, I'm wondering on an a11y perspective if this is problematic at all? Feels like an icon only chip or something like that 🤷

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,195 @@
...
+  type: ChipGroupType;
+  className?: string | null;
+  style?: ChipStyleType;
+};
+
+export type ChipGroupProps = {
+  chips: ChipItem[];
+  stickyChip?: ChipItem;
+  _isSingleSelect?: boolean; // TODO: is there a better way for internal only props?
+} & CommonProps;
```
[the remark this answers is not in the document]
mungodewar: This component is very similar to https://react-spectrum.adobe.com/react-spectrum/ActionGroup.html drawing inspiration from them they have a `selectionMode` prop. That feels like what `_isSingleSelect` is describing, however, gives a more semantic meaning.

mungodewar:
Finding: https://react-spectrum.adobe.com/react-spectrum/ActionGroup.html & how similar it is to this component, it might be worth taking a look and doing a contrast and compare to their solution - particularly around the whole api approach and how they manage it.

packages/bpk-component-chip-group/src/Nudger.tsx
mungodewar: Design potentially - having a static value might mean that _really_ long chips are not supported as well. 

I could imagine a desire to have a nudge bring the "next" chip into view? https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView maybe?

Iain530: `scrollIntoView` would be really nice, I'm struggling to think of a way to determine which chip to call it on though - you'd need some way to calculate the next chip that is not fully visible in the container. This might be possible but may not be very clean - we can spend a little time exploring it at least.

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
Iain530: I've created this ticket to add an icon only chip: https://skyscanner.atlassian.net/browse/DTO-5126

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
Iain530: These issues will be addressed by [DTO-5125](https://skyscanner.atlassian.net/browse/DTO-5125). From design: the total height should be 36px giving a 2px padding on the top and bottom which will make the shadows and hopefully this indicator visible

Iain530:
> Following on the conversation from slack: [skyscanner.slack.com/archives/C0JHPDSSU/p1706274673682399?thread_ts=1706259605.819169&cid=C0JHPDSSU](https://skyscanner.slack.com/archives/C0JHPDSSU/p1706274673682399?thread_ts=1706259605.819169&cid=C0JHPDSSU)
> 
> I’m thinking that the composable pattern is better, as the prop is a hard coupling, however, we could make it looser by the `ChipGroup` rendering children and managing `Chip` props via something like context.
> 
> eg there would be no need for the custom stickyChip code, the composition would be enough I think?

This conversation continued on slack a bit, the main reason for having the ChipItems is to give us more control over the BpkChip components which I think we lose if they become composable?

See:
- https://skyscanner.slack.com/archives/C0JHPDSSU/p1706276915820979?thread_ts=1706259605.819169&cid=C0JHPDSSU
- https://skyscanner.slack.com/archives/C0JHPDSSU/p1706542501214379?thread_ts=1706259605.819169&cid=C0JHPDSSU

I understand this would mean it's harder to migrate to using this component though so if there's a potential middle ground where we still can have control over the props we need but with this pattern that might be ideal. Although I don't know if that's possible?

Iain530:
> There are a few different types of BpkChipGroup, could you give me a TLDR on when a consumer would use which one and why, are some of these just internal maybe?

Yes, so there are two main types `MultiSelect` and `SingleSelect`:
- `MultiSelect` means that any number of chips can be selected
- `SingleSelect` means that only a single chip can be selected (controlled through the `selectedIndex` prop)

These two components are not stateful so a consumer should use these if they want to manage the chip states themselves.

Then there are also stateful versions of both of these which wrap the stateless ones and make them stateful, simplifying the job of the consumer if their use case is simple. The stateful ones weren't included in the design but I needed to write them for testing anyway so it seemed like a good idea to include them. If we'd rather not though I can take them out or refactor them to use a wrapping pattern like this: https://github.com/Skyscanner/backpack/blob/773b1e534e33a0a924b1ef28efc53cf09cc07cec/packages/bpk-component-info-banner/src/withBannerAlertState.d.ts#L25

The [storybook](https://backpack.github.io/storybook-prs/3198) contains examples of both single and multi select (stateful).

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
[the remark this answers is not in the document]
Iain530: I've renamed this to `ariaMultiselectable` for now as this id all the prop controls, the actual selection mode depends on which component is used.

packages/bpk-component-chip-group/src/Nudger.tsx
Iain530: I've dug into this a bit today, it would be possible to support this by using IntersectionObservers on each chip to track which ones are visible then scroll the next one into view, however it does add a lot of complexity to the solution and currently could cause issues if the number of chips change. I've parked it for now but happy to discuss further.

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
```diff
@@ -0,0 +1,78 @@
...
+}
+
+.bpk-chip-group-scroller-indicator {
+  &::before,
+  &::after {
+    display: none;
+  }
+}
```
Iain530, line 78: This will soon become the default behaviour: https://github.com/Skyscanner/backpack/pull/3275

packages/bpk-component-chip-group/src/Nudger.tsx
Iain530: After catching up with Adam Wilson, we've decided on a static value of 150px as this feels pretty good and keeps it simple.

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,213 @@
...
+  hidden?: boolean;
+} & SingleSelectChipItem;
+
+export type CommonProps = {
+  ariaLabel?: string;
+  type?: ChipGroupType;
+  chipStyle?: ChipStyleType;
+  // only required when type = rail
+  leadingNudgerLabel?: string;
+  trailingNudgerLabel?: string;
```
Iain530, line 78: This doesn't feel ideal as it can lead to these props not being populated easily. Either:
1. These props are required, but they are not needed for `wrap` chip groups which will be a little annoying for consumers.
2. Add a warning console log when type = `rail` and these are empty - see this pattern in other components but still no guarantee they'll be populated.
3. Refactor the interface and make "conditional" props [like this](https://dev.to/maissenayed/conditional-react-props-with-typescript-43lg). This ensures props are set correctly but at the const of a more complicated interface (and being tricky to implement).

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,213 @@
...
+        leadingIndicatorClassName={scrollContainerIndicator}
+        trailingIndicatorClassName={scrollContainerIndicator}
+      >
+        {children}
+      </BpkMobileScrollContainer>
+    ) : children;
+
+  return (
+    <div className={containerClassnames}>
+      {type === CHIP_GROUP_TYPES.rail && (
```
mungodewar, line 158: `type === CHIP_GROUP_TYPES.rail` there are a lot of these statements. Have you considered if the different types of `BpkChipGroup` are more different to one another than they are similar? 

Basically, did you look at having:

```tsx
const BpkChipGroup = ({type, ...rest}) => {
    if(type === "rail"){
        return <RailChipGroup {...rest}/>
   }

    return <DefaultChipGroup {...rest} />
}

```

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,213 @@
...
+  );
+
+  const stickyChipContainerClassnames = getClassName(
+    'bpk-sticky-chip-container',
+    `bpk-sticky-chip-container--${chipStyle}`,
+  );
+
+  const scrollContainerIndicator = getClassName('bpk-chip-group-scroller-indicator');
+
+  const renderChipItem = ({
```
mungodewar, line 114: Is there a reason that this needs to be inline within `BpkChipGroup`. Looks like we can add `chipStyle` to the props and extract it from the definition. 

The reason that we might want to do this is that as written a new function `renderChipItem` will be created on each render of `BpkChipGroup` and we can just avoid it. 

This could also allow us to turn `renderChipItem` into a component which could make the code more readable? Certainly less surprising seeing a component than a function _that happens to return JSX_, if you see what I mean.

mungodewar (commented):
I'm still looking at the PR, but I thought that I would share what I've got so far.

examples/bpk-component-chip-group/examples.module.css
olliecurtis, line 1: This file is no longer required into Backpack as its autogenerated at build time so can safely be removed

examples/bpk-component-chip-group/stories.ts
```diff
@@ -0,0 +1,64 @@
...
+} from './examples';
+
+export default {
+  title: 'bpk-component-chip-group',
+  component: BpkChipGroup,
+  subcomponents: {
+    BpkChipGroupSingleSelect,
+    BpkChipGroupState,
+    BpkChipGroupSingleSelectState,
+    // TODO: can we show the shape of ChipItem here?
```
olliecurtis, line 46: Is this still required?

packages/bpk-component-chip-group/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  CHIP_COMPONENT,
+} from './src/BpkChipGroup';
+import BpkChipGroupSingleSelect, {
+  type SingleSelectProps,
+  type SingleSelectStateProps,
+  BpkChipGroupSingleSelectState,
+} from './src/BpkChipGroupSingleSelect';
+
+export type { ChipGroupProps, SingleSelectProps, SingleSelectStateProps, ChipItem, SingleSelectChipItem };
+export { BpkChipGroupState, CHIP_COMPONENT, CHIP_GROUP_TYPES, BpkChipGroupSingleSelect, BpkChipGroupSingleSelectState};
```
olliecurtis, line 34: nit: Should these be prettierfied to be a new line per export?

packages/bpk-component-chip-group/README.md
```diff
@@ -0,0 +1,155 @@
...
+    ariaLabel="Select a city"
+  />
+);
+```
+
+
+## Props
+
+[//]: # (TODO: this update to web tab)
+Check out the full list of props on Skyscanner's [design system documentation website](https://www.skyscanner.design/latest/components/chip-group/compose-LwGOKNct).
```
olliecurtis, line 155: I've now made this available

```suggestion
Check out the full list of props on Skyscanner's [design system documentation website](https://www.skyscanner.design/latest/components/chip-group/web-4eQsMvYv).
```

packages/bpk-component-chip-group/src/BpkChipGroup-test.tsx
```diff
@@ -0,0 +1,247 @@
...
+      text: 'Florence',
+    },
+    {
+      text: 'Stockholm',
+    }
+  ];
+
+  it('should render correctly with type = rail', () => {
+    const { asFragment } = render(<BpkChipGroup {...defaultProps} chips={chips} type={CHIP_GROUP_TYPES.rail} />);
+    expect(asFragment()).toMatchSnapshot();
```
olliecurtis, line 57: Snapshots are known to not always be reliable or good for testing, can we make a switch out from this to do some other render check for reliability?

packages/bpk-component-chip-group/src/BpkChipGroup.module.css
olliecurtis, line 1: This file will no longer be needed as its now generated at shipping time :)

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
olliecurtis: This is now merged so you can rebase latest main to achieve this :)

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,218 @@
...
+export type ChipItem = {
+  component?: ChipComponentType;
+  onClick?: (selected: boolean, index: number) => void;
+  selected?: boolean;
+  hidden?: boolean;
+} & SingleSelectChipItem;
+
+export type CommonProps = {
+  label?: string;
+  ariaLabel?: string;
```
olliecurtis, line 75: Is there a reason this is optional? Instead of having ariaLabel as required?

packages/bpk-component-chip-group/src/BpkChipGroupSingleSelect-test.tsx
```diff
@@ -0,0 +1,228 @@
...
+    const { asFragment } = render(
+      <BpkChipGroupSingleSelect
+        chips={chips}
+        type={CHIP_GROUP_TYPES.rail}
+        ariaLabel="Filter cities"
+        leadingNudgerLabel="Scroll back"
+        trailingNudgerLabel="Scroll forward"
+      />
+    );
+    expect(asFragment()).toMatchSnapshot();
```
olliecurtis, line 64: Same comment as before

packages/bpk-component-chip-group/src/Nudger.module.css
olliecurtis, line 1: File not required

olliecurtis (changes requested):

packages/bpk-component-chip-group/src/BpkChipGroup-test.tsx
Iain530: Makes sense, I have now refactored them 👍

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
Iain530: This is only needed if the consumer wants an addional visually hidden label - I think this might not be always neccessary for example when using a sticky chip that contains the same text.

packages/bpk-component-chip-group/index.ts
Iain530: Should the linter automatically do this?

examples/bpk-component-chip-group/stories.ts
olliecurtis: Ah gotcha! This should be possible to do, as we have the documentation part which we can add subcomponents to! Let me check if for types it is possible but should be as we have it for other components!

packages/bpk-component-chip-group/index.ts
olliecurtis: Not automatically no, currently we just run `npm run lint` vs the fix command

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
olliecurtis: Maybe we just need to make it clear in the documentation that this is how that works. I will also check as I think we can use comments to add a description that will show in the props table

Iain530: Nice, that would be good!

anambl:
Hey guys what's the status of this? What needs to happen to unblock this? 😄 @Iain530 @olliecurtis @mungodewar

olliecurtis:
> Hey guys what's the status of this? What needs to happen to unblock this? 😄 @Iain530 @olliecurtis @mungodewar

I need to come and review now that this has been updated since my last review :)

Iain530:
> > Hey guys what's the status of this? What needs to happen to unblock this? 😄 @Iain530 @olliecurtis @mungodewar
> 
> I need to come and review now that this has been updated since my last review :)

Thanks @olliecurtis! I think we should have addressed everything except the documentation questions around storybook/props comments.

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+  const stickyChipContainerClassnames = getClassName(
+    'bpk-sticky-chip-container',
+    `bpk-sticky-chip-container--${chipStyle}`,
+  );
+
+  return (
+    <>
+      <BpkBreakpoint query={BREAKPOINTS.ABOVE_TABLET}>
+        <Nudger leading ariaLabel={leadingNudgerLabel} chipStyle={chipStyle} scrollContainerRef={scrollContainerRef} />
```
olliecurtis, line 173: The aria label is required in the Nudger but is marked as optional here and we would pass through an empty string which would cause A11y issues? 

This looks to conflict with the APIs, so we should align them - preference on making it required here instead of passing in a blank string

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,254 @@
...
+      >
+        <ChipGroupContent ariaLabel={ariaLabel}
+          ariaMultiselectable={ariaMultiselectable}
+          chipGroupClassNames={chipGroupClassNames}
+          chipStyle={chipStyle}
+          chips={chips}
+          label={label} />
+      </BpkMobileScrollContainer>
+      <BpkBreakpoint query={BREAKPOINTS.ABOVE_TABLET}>
+        <Nudger ariaLabel={trailingNudgerLabel} chipStyle={chipStyle} scrollContainerRef={scrollContainerRef} />
```
olliecurtis, line 200: Same as above

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,254 @@
...
+    {ariaLabel && <legend className='visually-hidden'>{ariaLabel}</legend>}
+    {label && <BpkText textStyle={TEXT_STYLES.footnote} aria-hidden>{label}</BpkText>}
+    {chips.map((chip, index) => <Chip chipItem={chip} chipStyle={chipStyle} ariaMultiselectable={ariaMultiselectable} chipIndex={index} />)}
+  </fieldset>
+);
+
+const RailChipGroup = ({
+  ariaLabel,
+  ariaMultiselectable = true,
+  chipGroupClassNames,
```
olliecurtis, line 155: What is the usecase for having a prop to provide additional classes? This would indicate that folks have to do customisation away from the design/its not fit for purpose that we want folks to change things

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
olliecurtis: I wonder if there is something more clever we can do here instead of exposing two different types of chips just a difference of having state.

What is the difference of needing to have a stateless and stateful chip? I think this should either be put to the consumer (like all our other components as Backpack shouldn't know about state) or built in mechanisms to detect to use state

packages/bpk-component-chip-group/src/BpkChipGroupSingleSelect.tsx
```diff
@@ -0,0 +1,66 @@
...
+type CommonSingleSelectProps = {
+  chips: SingleSelectChipItem[];
+  onItemClick?: (item: SingleSelectChipItem, selected: boolean, index: number) => void,
+} & ChipGroupProps;
+
+export type SingleSelectProps = {
+  selectedIndex?: number;
+} & CommonSingleSelectProps;
+
+const BpkChipGroupSingleSelect = ({ chips, onItemClick, selectedIndex, ...rest }: SingleSelectProps) => {
```
olliecurtis, line 32: Can we rename this component to `BpkSingleSelectChipGroup` to align with apps?

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -0,0 +1,254 @@
...
+  label,
+}: WrapChipGroupProps & InternalProps) =>
+  <ChipGroupContent ariaLabel={ariaLabel}
+    ariaMultiselectable={ariaMultiselectable}
+    chipGroupClassNames={chipGroupClassNames}
+    chipStyle={chipStyle}
+    chips={chips}
+    label={label} />;
+
+const BpkChipGroup = (props: ChipGroupProps) => {
```
olliecurtis, line 221: Can we rename this component to `BpkMultiSelectChipGroup` to align with apps?

packages/bpk-component-chip-group/src/BpkChipGroupSingleSelect.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <BpkChipGroup chips={chipsWithSelection} ariaMultiselectable={false} {...rest} />
+  );
+};
+
+export type SingleSelectStateProps = {
+  initiallySelectedIndex?: number;
+} & CommonSingleSelectProps
+
+export const BpkChipGroupSingleSelectState = ({ initiallySelectedIndex = -1, onItemClick, ...rest }: SingleSelectStateProps) => {
```
olliecurtis, line 52: Same as previous comment

> I wonder if there is something more clever we can do here instead of exposing two different types of chips just a difference of having state.
>
> What is the difference of needing to have a stateless and stateful chip? I think this should either be put to the consumer (like all our other components as Backpack shouldn't know about state) or built in mechanisms to detect to use state

packages/bpk-component-chip-group/src/Nudger.tsx
```diff
@@ -0,0 +1,113 @@
...
+  [CHIP_TYPES.onDark]: BUTTON_TYPES.secondaryOnDark,
+  [CHIP_TYPES.onImage]: BUTTON_TYPES.primaryOnDark,
+}
+
+
+type Props = {
+  ariaLabel: string;
+  chipStyle?: ChipStyleType;
+  scrollContainerRef: MutableRefObject<HTMLElement | null>;
+  leading?: boolean;
```
olliecurtis, line 46: I would say rather than having this as a bool. I would use an enum that has leading or trailing in it.

Something like 
```  
  
const POSITION = {
   leading: 'leading',
   trailing: 'trailing'
}

position: Position;
```

packages/bpk-component-chip-group/src/Nudger.tsx
```diff
@@ -0,0 +1,113 @@
...
+    <div className={classNames}>
+      <BpkButtonV2
+        title={ariaLabel}
+        type={CHIP_STYLE_TO_BUTTON_STYLE[chipStyle]}
+        iconOnly
+        disabled={!show || !enabled}
+        onClick={() => {
+          if (scrollContainerRef.current) {
+            scrollContainerRef.current.scrollBy({
+              left: isLeft ? -SCROLL_DISTANCE : SCROLL_DISTANCE,
```
olliecurtis, line 101: Is this for RTL? Can we use that instead?

packages/bpk-component-chip-group/index.d.ts
olliecurtis, line 1: This file can be deleted as its now autogenerated and build time

packages/bpk-component-chip-group/src/BpkChipGroup.d.ts
olliecurtis, line 1: This file can be deleted as its now autogenerated and build time

packages/bpk-component-chip-group/src/BpkChipGroupSingleSelect.d.ts
olliecurtis, line 1: This file can be deleted as its now autogenerated and build time

packages/bpk-component-chip-group/src/Nudger.d.ts
olliecurtis, line 1: This file can be deleted as its now autogenerated and build time

olliecurtis (changes requested):

packages/bpk-component-chip-group/src/Nudger.tsx
Iain530: This is a little complex. `isLeft` is calculated as a combination of `isRtl` and `isLeading`. We need to know if it's left or right to choose the correct icon and to also scroll in the right direction when it's pressed.

So I think we are already doing it as simply as we can unfortuantely

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
Iain530: Good catch, this originally optional but should now be required so i've removed the default

packages/bpk-component-chip-group/src/BpkChipGroupSingleSelect.tsx
Iain530: It seemed like a nice utility to provide, especially since we had to implement it to show examples anyway. Alternatively this could be a HOC like some other components do. e.g. `withMultiSelectState(BpkMultiSelectChipGroup)` however I originally decided against this as I think we would still need 2 separate functions for MultiSelect and SingleSelect - so it only adds complexity to the consumer doing the importing and the setup before using the component

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
Iain530: I think we didn't want to duplicate this in both `RailChipGroup` and `WrapChipGroup`. Though I've removed the prop for now as I think it's less complex with the duiplication actually
```ts
  const chipGroupClassNames = getClassName(
    'bpk-chip-group',
    `bpk-chip-group--{type}`,
  );
```

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
Iain530: Decided to move the stateful ones to examples only as this aligns with the rest of backpack

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
```diff
@@ -22,30 +22,32 @@
...
+  margin: 0 (-4 * tokens.$bpk-one-pixel-rem);
   align-items: center;
   white-space: nowrap;
 }
 
 .bpk-chip-group {
   display: flex;
   margin: 0;
-  padding: 0;
+  padding: 4 * tokens.$bpk-one-pixel-rem;
```
olliecurtis, line 33: Can we use `bpk-spacing-sm()`which is 4px

```suggestion
  padding: tokens.bpk-spacing-sm();
```

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
```diff
@@ -22,30 +22,32 @@
 
 .bpk-chip-group-container {
   display: flex;
+  margin: 0 (-4 * tokens.$bpk-one-pixel-rem);
```
olliecurtis, line 25: Same here

```suggestion
  margin: 0 -(tokens.bpk-spacing-sm());
```

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
```diff
@@ -22,30 +22,32 @@
...
-
   &--wrap {
+    padding-top: 0;
+    padding-bottom: 0;
     flex-wrap: wrap;
   }
 }
 
+.bpk-sticky-chip-container:first-child {
+  margin-inline-start: 4 * tokens.$bpk-one-pixel-rem;
```
olliecurtis, line 46: Same here

```suggestion
  margin-inline-start: tokens.bpk-spacing-sm();
```

packages/bpk-component-chip-group/src/Nudger.module.scss
```diff
@@ -21,13 +21,11 @@
 .bpk-chip-group-nudger {
   &--leading {
     margin-inline-end: tokens.bpk-spacing-md();
+    margin-inline-start: 4 * tokens.$bpk-one-pixel-rem;
```
olliecurtis, line 24:
```suggestion
    margin-inline-start: tokens.bpk-spacing-sm();
```

packages/bpk-component-chip-group/src/Nudger.module.scss
```diff
@@ -21,13 +21,11 @@
...
   }
 
   &--trailing {
+    margin-inline-end: 4 * tokens.$bpk-one-pixel-rem;
```
olliecurtis, line 28:
```suggestion
    margin-inline-end: tokens.bpk-spacing-sm();
```

olliecurtis (changes requested):

olliecurtis (approved):

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
```diff
@@ -30,6 +30,7 @@
   display: flex;
   margin: 0;
   padding: 0;
+  align-items: baseline;
```
bjoern-wahle, line 34: @Iain530 @olliecurtis Do you maybe remember why this `baseline` here was necessary? It seems it is breaking alignment when setting `leadingAccessoryView` for any of the chips: 
image

Removing the `align-items` here fixes the issue, but I am afraid I might break something else...

PR #3209, opened 2024-02-01
[NO-JIRA] Make BpkModalV2 the default

ByAnthony:

- [ ] Component `README.md`
- [ ] Tests


olliecurtis (changes requested):
Are we going to look to introduce Polyfils for the HTML Dialog element as its not fully supported by our current list of browsers.

Specifically Safari 14 - 15.3: 
https://caniuse.com/dialog 
https://github.com/GoogleChrome/dialog-polyfill

Which we recently had an issue and incident because we were using tech that is not supported by our current list

ByAnthony:
> Are we going to look to introduce Polyfils for the HTML Dialog element as its not fully supported by our current list of browsers.
> 
> Specifically Safari 14 - 15.3: https://caniuse.com/dialog https://github.com/GoogleChrome/dialog-polyfill
> 
> Which we recently had an issue and incident because we were using tech that is not supported by our current list

There is already a polyfill for dialog which was tested in the V2

PR #3235, opened 2024-02-15
Experiment reduce mixins single stylesheet

mungodewar:
This is an experimentation on how we can still write CSS modules alongside our components, however, they are consumed and used at a global level using a single stylesheet. 

The idea here is that the CSS Module class hash will always resolve to the same value in different build systems if the same parameters are used. 

Thus we can generate classNames and attach them to DOM but *not* generate css chunks at the same moment. Instead a "human in the loop" will craft the single stylesheet using `import` statements. The single stylesheet in this example is `base.css` generated in `bpk-stylesheet`. 

This doesn't work seamlessly within storybook, there are some specific classes/styles that are now *not* included on the page & I've not bothered to work out a way of including them (should be simple based on resource path). Every component bar `badge v1` is broken in this way & some examples from `badge` look a little funny as their contextual environment (dark background for outline type) has been taken away.


.storybook/preview.tsx
```diff
@@ -22,6 +22,7 @@ import type { Preview } from '@storybook/react';
 
 import '../packages/bpk-stylesheets';
 import '../packages/bpk-stylesheets/font';
+import '../packages/bpk-stylesheets/base.css';
```
mungodewar, line 25: Add the base.css that is generated at build time in bpk-stylesheets

.storybook/webpack.config.js
```diff
@@ -30,6 +30,12 @@ const devMode = process.env.NODE_ENV !== "production";
 
 module.exports = ({ config }) => {
   config.plugins.push(new MiniCssExtractPlugin());
+
+  const i = config.module.rules.findIndex(
+    rule => rule.test.toString() === '/\\.css$/',
+  );
+  config.module.rules.splice(i, 1);
```
mungodewar, line 37: Remove defaults for css rules from storybook config

.storybook/webpack.config.js
```diff
@@ -89,15 +90,24 @@ module.exports = ({ config }) => {
...
+            localIdentName: '[hash:base64]',
+            localIdentHashFunction: 'md4',
+            localIdentContext: path.resolve(__dirname, "../packages/bpk-stylesheet"),
+            mode: (resourcePath, resourceQuery, resourceFragment) => {
+              if (/global.scss$/i.test(resourcePath)) {
+                return "global";
+              }
+
+              return "local";
+            },
```
mungodewar, line 110: This is setup specifically to generate the classes the exact same way that `bpk-stylesheet` does, it relies on the hashing function being totally deterministic for both of the separate `build`s to lineup. 

I don't think we need the `mode` function here, that is just left over from experimentation.

packages/bpk-stylesheets/index.js
```diff
@@ -17,6 +17,7 @@
  */
 
 import './index.scss';
+import '../bpk-component-badge/src/BpkBadge.module.scss';
```
mungodewar, line 20: Here a human gets to order the component scss, this will define the order in the produced stylesheet.

packages/bpk-stylesheets/webpack.config.babel.js
```diff
@@ -73,7 +73,17 @@ module.exports = {
...
+                localIdentName: '[hash:base64]',
+                localIdentHashFunction: 'md4',
+                mode: (resourcePath, resourceQuery, resourceFragment) => {
+                  if (/global.scss$/i.test(resourcePath)) {
+                    return "global";
+                  }
+    
+                  return "local";
+                },
+              },
```
mungodewar, line 86: As before, the same config so that the className hashes line up.

packages/bpk-component-badge/src/BpkBadge.module.scss
```diff
@@ -18,50 +18,108 @@
...
+}
 
-  &--warning {
-    @include bpk-badge--warning;
-  }
+$types: 'warning', 'normal', 'success', 'critical', 'inverse', 'outline', 'strong', 'brand';
 
-  &--success {
-    @include bpk-badge--success;
+@each $type in $types {
```
runmoore, line 82: could you explain this please?

mungodewar: This is sass basically what it is doing is using some loops to avoid repetitive typing. As we now know, CSS modules demands a single class per element. That makes variants a little tricky, in addition to having `.warning` we'd need `warning-centered` `warning-right` and `warning-left`. We need these for _every_ "type" of badge. That would be a lot of typing out the same thing. So instead we can write a loop. 

`$types: 'warning', 'normal', 'success', 'critical', 'inverse', 'outline', 'strong', 'brand';`

is equivalent to `const types = ['warning', 'normal', 'success', 'critical', 'inverse', 'outline', 'strong', 'brand']`

`@each $type in $types`

is equivalent to `for type in types`

Now we have a loop with all our types & it's time to consume that information: 

```
  .#{$type}-centered {
    composes: #{$type};
    vertical-align: text-bottom;
  }
```

`#{$type}` is the syntax for string interpolation, it's equivalent to: `'${type}'`. We are creating a new class, when `type=warning` we get: `warning-centered` & we are doing it for each "type"!.

This way, we can apply a _single_ style to each element that entirely defines it's behaviour. 

This notably doesn't allow for _multiple_ variants to be applied at once. Eg you can't be `right` and `centered`. We could support that, I've not done that in this PR though.

PR #3243, opened 2024-02-19
[THES-184] BpkBottomSheetV2 - use BpkDialogWrapper to fix focus

JToporkiewicz:
### **Migration guide**

To prevent TS errors, remove the properties `getApplicationElement` and `renderTarget` from use cases of the BpkBottomSheet component.

------

Using dialog element fixes all of our focus issues - relevant to multiple reported accessibility concerns. The dialog element has been made into a reusable wrapper because:
* it has more use cases than just the BpkBottomSheet (for example BpkModalV2)
* it requires a handling of some special cases - where a browser doesn't support dialog out of the box
* it can be difficult to figure out the handling of close and cancel (close by ESC key)

This PR changes the BpkBottomSheet component to use the new wrapper but **it does not update any other components that may find the wrapper relevant** due to the scope of work that would be required, and the priorities of the developers team.

[Example](https://github.com/Skyscanner/backpack/pull/3247/files#diff-7a8b63e05cbc892661ca35e029f843cacbb6da5a470c52a62e2b68055286da30R110) of use of BpkDialogWrapper in BpkModalV2 - to show the reusability of the wrapper

- [x] Component `README.md`
- [x] Tests


packages/bpk-dialog-utils/BpkDialogWrapper.tsx
```diff
@@ -0,0 +1,153 @@
...
+        />
+      )}
+      <CSSTransition
+        classNames={transitionClassNames}
+        in={!exiting}
+        appear={!exiting}
+        exit={exiting}
+        timeout={timeout}
+      >
+        <dialog
```
olliecurtis, line 135: Sharing this comment here as I think this will be important: https://github.com/Skyscanner/backpack/pull/3209#pullrequestreview-1890544627

Would be worth us running a test in Safari 14 to see how this works and make sure it is safe and doesn't break for folks

ByAnthony: If the code was copied form bpkmodalv2, there is a polyfill for old browser that don't support the dialog element.

JToporkiewicz: Yes, BpkDialogWrapper is essentially a copy of the BpkModalV2 dialog, with optional animations and disabling ESC or background click close.

packages/bpk-dialog-utils/src/BpkDialogWrapper.tsx
```diff
@@ -0,0 +1,170 @@
...
+      source: 'ESCAPE' | 'DOCUMENT_CLICK';
+    },
+) => void | null;
+  exiting?: boolean;
+  transitionClassNames?: {
+    appear?: string,
+    appearActive?: string,
+    exit?: string
+  };
+  timeout?: {appear?: number, exit?: number};
```
ByAnthony, line 49: like those improvements 👍

packages/bpk-dialog-utils/src/BpkDialogWrapper.tsx
```diff
@@ -0,0 +1,170 @@
...
+};
+
+type DialogProps = {
+  isDialogOpen: boolean;
+};
+
+// TODO: this check if the browser support the HTML dialog element. We can remove it once we drop support as a business for Safari 14
+const dialogSupported = typeof HTMLDialogElement === 'function';
```
ByAnthony, line 57: check if dialog is supported or not; and then use the polyfill if not supported 💪

packages/bpk-dialog-utils/src/BpkDialogWrapper.tsx
```diff
@@ -0,0 +1,170 @@
...
+      setPageProperties({ isDialogOpen: false })
+      window.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [id, isOpen, onClose, closeOnEscPressed, closeOnScrimClick]);
+
+  return isOpen ? (
+    <div
+      className={getClassName(
+        'bpk-dialog-wrapper',
+        dialogSupported ? '' : 'bpk-dialog-wrapper--polyfill',
```
ByAnthony, line 135: Polyfill if dialog not supported 💪

packages/bpk-dialog-utils/src/BpkDialogWrapper.tsx
```diff
@@ -0,0 +1,170 @@
...
+        />
+      )}
+      <CSSTransition
+        classNames={transitionClassNames}
+        in={!exiting}
+        appear={!exiting}
+        exit={exiting}
+        timeout={timeout}
+      >
+        <dialog
```
ByAnthony, line 152: 🚀

ByAnthony:
Great work @JToporkiewicz 🥇

README.md
```diff
@@ -85,6 +85,7 @@ To contribute please see [contributing.md](CONTRIBUTING.md).
 [`bpk-theming`](/packages/bpk-theming)
 [`bpk-component-ticket`](/packages/bpk-component-ticket)
 [`bpk-component-tooltip`](/packages/bpk-component-tooltip)
+[`bpk-dialog-utils`](/packages/bpk-dialog-utils/)
```
olliecurtis, line 88: Are these expected to be used outside of Backpack or is this new introduction purely for sharing code between components in Backpack?

If its the latter then we could bundle these easily into `bpk-react-utils` as this package is a suite for storing utilities that are shared between components. E.g. Scrim

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -15,86 +15,145 @@
...
   onClose: (
-    arg0?: TouchEvent | MouseEvent | KeyboardEvent,
+    arg0?: Event | KeyboardEvent | MouseEvent | PointerEvent | TouchEvent,
     arg1?: {
       source: 'ESCAPE' | 'DOCUMENT_CLICK';
     },
   ) => void;
   title?: string;
   wide?: boolean;
-  getApplicationElement: () => HTMLElement | null;
```
anambl, line 47: wouldn't removing these props make this into a breaking change? 🤔

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -15,86 +15,145 @@
...
   children: ReactNode;
   closeLabel?: string;
   closeOnEscPressed?: boolean;
   closeOnScrimClick?: boolean;
   id: string;
-  isOpen: boolean;
   onAction?: () => void;
   onClose: (
-    arg0?: TouchEvent | MouseEvent | KeyboardEvent,
+    arg0?: Event | KeyboardEvent | MouseEvent | PointerEvent | TouchEvent,
```
anambl, line 43: Does the generic `Event` type not include all the other event types?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -15,86 +15,145 @@
...
-) => {
-    setExitting(true)
+  ) => {
+    setExiting(true)
     setTimeout(() => {
       onClose(arg0, arg1)
-      setExitting(false)
-    }, 240)
+      setExiting(false)
+    }, timeout)
```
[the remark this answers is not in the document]
JToporkiewicz: This needs to be passed in from parent because when the component is on desktop (a different animation) this gets set to 0, as seen on [lines 102](https://github.com/Skyscanner/backpack/pull/3243/files#diff-74fb49523990b42587c46681a2766595137981a62c73ad7f2cc909d181a40e5eR102) and 134

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
JToporkiewicz: You're right. This throws a ts error when having the property in use but not expected. I'll change the label on this PR to major.

Surprisingly though running the storybook with that extra prop doesn't cause any breakages or even console.logs. Still not good to keep it in, but something interesting to know.

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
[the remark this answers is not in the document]
anambl: Ohh, you're right, I hadn't realised it's a parameter of `handleClose`. Makes sense then!

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
anambl: But doesn't `Event` include all the others, so we could just remove the others then? 🤔
What's the reason we need to call `onClose` with `e.nativeEvent` rather than just the event?

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
```diff
@@ -0,0 +1,181 @@
...
+      if (closeOnEscPressed
+        && event.key === 'Escape'
+        && (!dialogWithPolyfill || event.target === dialogWithPolyfill)) {
+        onClose(event, {source: "ESCAPE"});
+      }
+      event.stopPropagation();
+    };
+
+    if (isOpen) {
+      // There is a bug on older versions of browser using chromium (chrome, firefox, edge >114) where the dialog got an open attribute even before it is opened.
```
olliecurtis, line 113: Do we know which versions have this bug and are they still in the range of our supported browser list?

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
```diff
@@ -0,0 +1,181 @@
...
+      setPageProperties({ isDialogOpen: false })
+      window.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [id, isOpen, onClose, closeOnEscPressed, closeOnScrimClick]);
+
+  return isOpen ? (
+    <div
+      className={getClassName(
+        'bpk-dialog-wrapper',
+        dialogSupported ? '' : 'bpk-dialog-wrapper--polyfill',
```
olliecurtis, line 141: I believe this could be simplified to 
```suggestion
        !dialogSupported && 'bpk-dialog-wrapper--polyfill',
```

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -15,86 +15,145 @@
...
   children: ReactNode;
   closeLabel?: string;
   closeOnEscPressed?: boolean;
   closeOnScrimClick?: boolean;
   id: string;
-  isOpen: boolean;
   onAction?: () => void;
   onClose: (
-    arg0?: TouchEvent | MouseEvent | KeyboardEvent,
+    arg0?: Event,
```
olliecurtis, line 43: Should we be using generic types here which could introduce issues with types down the line?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -15,86 +15,145 @@
...
+    return <BpkBreakpoint query={BREAKPOINTS.ABOVE_MOBILE}>
+      {(isAboveMobile: boolean) =>
+      <BpkDialogWrapper
+      ariaLabelledby={ariaLabelledby}
+      dialogClassName={dialogClassName}
+      id={id}
       isOpen={isOpen}
-      onClose={handleClose}
+      onClose={(
+        arg0?: Event,
```
olliecurtis, line 99: Same here

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -15,86 +15,145 @@
...
+              >
+                {title}
+              </h2>
+            }
+            leadingButton={
+              <BpkCloseButton
+                className={getClassName('bpk-bottom-sheet--close-button')}
+                label={closeLabel}
+                onClick={(
+                  arg0?: Event,
```
olliecurtis, line 131: Same here

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -17,16 +17,156 @@
...
+    border-radius: tokens.$bpk-border-radius-lg tokens.$bpk-border-radius-lg 0 0;
+    overflow-x: hidden;
+  }
+
+  &::-webkit-scrollbar {
+    display: none;
+  }
+
+  &--appear {
+    animation-duration: 0.24s;
```
olliecurtis, line 69: Can we/should we be using Backpack standard animation values here? 

https://www.skyscanner.design/latest/foundations/animation/overview-DTJrpm76

Otherwise it calls into question what are these 'random' numbers mean

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -17,16 +17,156 @@
...
+      transform: scale(1);
+      opacity: tokens.$bpk-modal-opacity;
+    }
+  }
+
+  &--exit {
+    animation-direction: reverse;
+    animation-duration: 0.24s;
+    animation-name: slide-up;
+    animation-timing-function: ease-in-out;
```
olliecurtis, line 91: As this is common as above could we extract this to a separate local mixin to use across the scss in this file, to make maintenance easier?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -17,16 +17,156 @@
...
+
+    @include utils.bpk-rtl {
+      right: auto;
+      margin-right: calc(-#{tokens.bpk-spacing-sm()} / 2);
+      margin-left: tokens.bpk-spacing-lg();
+    }
+  }
+
+  &--action-button {
+    position: relative; // Override position: absolute coming from BpkNavigationBar
```
olliecurtis, line 149: Waht would be the reason for overriding BpkNavigationBar?

This could indicate something is wrong in the core component

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -17,16 +17,156 @@
...
+    right: auto;
+    width: calc(#{$close-button-width} - (#{tokens.bpk-spacing-sm()}));
+    margin-right: calc(#{tokens.bpk-spacing-sm()} / 2);
+    margin-left: calc(#{tokens.bpk-spacing-sm()});
+    word-break: break-all;
```
olliecurtis, line 154: We might want to check if we want to expect this behaviour!

olliecurtis (changes requested):
Some comments below.

Also could we understand a bit on the overriding of styles in `BpkNavigationBar` components and sub components? As this could be an issue with the component that we need to update :)

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
pangszenga: Can you elaborate? Are you talking about type issues or issues with the events?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
pangszenga: The X button and action button overlays each other as out of the box, the `BpkNavigationBar` has a styling for the leading item to have absolute positioning (see the screenshot) which causes the buttons to overlap.
image

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
pangszenga: Yes this is confirmed with design 👍

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
pangszenga: Need empty string here, otherwise the `className` computed is incorrect

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
olliecurtis: Is this a symptom of other styles we are also supplying to parts of the Nav bar as out of the box this should be supporting and aligning all the parts equally to make this possible without the need for overridding as otherwise we are introducing a risk of CSS non determinism and this problem coming up anway even with the overrides

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
pangszenga: Oops noob move on my part 😅

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
pangszenga: We initially used `TouchEvent | MouseEvent | KeyboardEvent` which resulted in a type error in the outer `BpkDialogueWrapper`, as we pass it an `onClose` function that it uses to handles the backdrop click and on escape key down. 

Without a generic `Event` type, we were running into a wall with the escape event throwing a type error relating to the mismatch type use of `SyntheticEvent<HTMLDialogElement, Event>` to the `TouchEvent | MouseEvent | KeyboardEvent` type casted as above. Using the generic `Event` type allows us to satisfy the `SyntheticEvent<HTMLDialogElement, Event>` type. 

Using the `SyntheticEvent<HTMLDialogElement, Event>` type to type check these event does not satisfy type checking as the other `onClose` calls in the `BpkDialogueWrapper` are using the native `addEventListener` which does not accept a `SyntheticEvent`, only accepting a listener on `MouseEvent` + 1 overload (which in our case, we are using a `PointerEvent`).

Do you have any suggestions on how to overcome this?

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
pangszenga: Im not sure what backpack's supported browser list is, but here is a list of versions that is supported https://caniuse.com/mdn-api_htmldialogelement_open

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
pangszenga: It doesn't appear so, as removing all styling relating to components at or within `BpkNavigationBar` is still resulting in the same effect

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
pangszenga: We have ended up using union events `TouchEvent | MouseEvent | KeyboardEvent | SyntheticEvent<HTMLDialogElement, Event>`

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -15,86 +15,145 @@
...
+                <BpkButtonLink
+                  className={getClassName('bpk-bottom-sheet--action-button')}
+                  onClick={onAction}
+                >
+                  {actionText}
+                </BpkButtonLink>
+              ) :
+              <div
+                className={getClassName('bpk-bottom-sheet--action-button')}
+              />
```
olliecurtis, line 148: Can we change this to
```suggestion
              null
```

As having this empty div causes some weird alignment behaviour when its just placed

Screenshot 2024-03-05 at 12 08 14

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -15,86 +15,145 @@
...
+          <BpkNavigationBar
+            id={headingId}
+            className={getClassName('bpk-bottom-sheet--navigation')}
+            title={title &&
+              <h2
+                id={headingId}
+                className={getClassName('bpk-bottom-sheet--heading')}
+              >
+                {title}
+              </h2>
```
olliecurtis, line 124: This could be used to be a `BpkText`?

```suggestion
              <BpkText id={headingId} textStyle={TEXT_STYLES.label1} tagName="h2">{title}</BpkText>
```

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -17,16 +17,159 @@
...
+
+    @include borders.bpk-border-bottom-sm(tokens.$bpk-line-day);
+  }
+
+  &--navigation {
     display: flex;
-    padding: tokens.bpk-spacing-lg();
+    justify-content: space-between;
+    background-color: tokens.$bpk-modal-background-color;
+  }
```
olliecurtis, line 125: No longer needed?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -17,16 +17,159 @@
...
+    width: calc(
+      100% - ((#{$close-button-width}) * 2 + #{tokens.bpk-spacing-sm()} / 2)
+    ); // Treat title as if it was always surrounded by close and action to center align it
+
+    margin-right: auto;
+    text-align: center;
+
+    @include typography.bpk-text;
+    @include typography.bpk-label-1;
+  }
```
olliecurtis, line 137: Same here?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -17,16 +17,159 @@
...
+    left: auto;
+    margin-right: tokens.bpk-spacing-lg();
+
+    @include utils.bpk-rtl {
+      right: auto;
+      margin-right: calc(-#{tokens.bpk-spacing-sm()} / 2);
+      margin-left: tokens.bpk-spacing-lg();
+    }
+  }
+
```
olliecurtis, line 150: Same here?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -17,16 +17,159 @@
...
+    margin-right: calc(#{tokens.bpk-spacing-sm()} / 2);
+    margin-left: calc(#{tokens.bpk-spacing-sm()});
+    word-break: break-all;
+
+    @include utils.bpk-rtl {
+      left: auto;
+      margin-right: calc(#{tokens.bpk-spacing-sm()});
+      margin-left: calc(#{tokens.bpk-spacing-sm()} / 2);
+    }
+  }
```
olliecurtis, line 164: Same here?

olliecurtis (approved):

PR #3249, opened 2024-02-23
[ARGG-1019]: Investigate adding new linter plugins for testing before global config

olliecurtis:
This is a PoC PR to use for sharing to internal docs to show possibility of using in our global config


olliecurtis:
Closing as this is a PoC and sharable link

PR #3277, opened 2024-03-12
[DTO-5126][bpk-component-chip] Create BpkIconChip

Iain530:
The `BpkIconChip` is a requirement for the new BpkChipGroup component (#3198)

- [See Figma](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?type=design&node-id=30119-37736&mode=design&t=tpAPzT03Uveajq2R-0)

image

---

- [x] Component `README.md`
- [x] Tests
- [x] Type declaration (`.d.ts`) files updated


examples/bpk-component-chip/examples.tsx
```diff
@@ -0,0 +1,309 @@
+/*
```
Iain530, line 1: This file changes:
- migrated to TS
- migrated class components to functional components
- added BpkIconChip examples

packages/bpk-component-chip/src/BpkIconChip.tsx
```diff
@@ -0,0 +1,61 @@
...
+import type { CommonProps as Props } from './commonTypes';
+import STYLES from './BpkSelectableChip.module.scss';
+
+const getClassName = cssModules(STYLES);
+
+export type BpkIconChipProps = {
+  leadingAccessoryView: ReactNode;
+} & Omit<Props, 'trailingAccessoryView' | 'leadingAccessoryView' | 'children'>;
+
+const BpkIconChip = ({
```
xalechez, line 39: Is something wrong with formatting here?

packages/bpk-component-chip/src/BpkSelectableChip.tsx
```diff
@@ -64,7 +65,7 @@ const BpkSelectableChip = ({
       {...rest}
     >
       {leadingAccessoryView && (
-        <span className={getClassName('bpk-chip__leading-accessory-view')}>
+        <span className={getClassName('bpk-chip__leading-accessory-view', leadingAccessoryClassName)}>
```
xalechez, line 68: I'm thinking whether we should introduce a new className prop instead of modifying component own styles? Looks like a source of potential non-determinism problems

packages/bpk-component-chip/src/BpkIconChip.tsx
Iain530: Yeah it looks inconsistent, will fix 👀

packages/bpk-component-chip/src/BpkSelectableChip.tsx
Iain530: How would you go about doing this? The styles we need to add override the default chip behaviour - this should work for the BpkIconChip afaik but does open up the potential for non-determinism problems if consumers start using this new prop

Iain530: I've removed the leadingAccessoryClassname prop and just optionally add the icon-only if the children are not set - this seemed slightly better to me than having a flag that should only be used by `BpkIconChip`, what do you think?

examples/bpk-component-chip/stories.tsx
```diff
@@ -20,13 +20,14 @@
...
 import BpkDropdownChip from '../../packages/bpk-component-chip/src/BpkDropdownChip';
+import BpkIconChip from '../../packages/bpk-component-chip/src/BpkIconChip';
 
 import {
   AllTypesExample,
   AllSelectableChipsExample,
   AllDropdownChipsExample,
   AllDismissibleChipsExample,
-  RadioGroupChipsExample,
+  RadioGroupChipsExample, AllIconChipsExample,
```
olliecurtis, line 30: nit: Can you push this to a new line?

packages/bpk-component-chip/src/BpkIconChip.tsx
```diff
@@ -0,0 +1,56 @@
...
+ className,
+ ...rest
+}: BpkIconChipProps) => {
+  const classNames = getClassName(`bpk-chip--icon-only`, className);
```
olliecurtis, line 43: This is going to be interesting one and will need some thinking, as we ideally don't want to introduce a className prop that we then send to override another Backpack component as it introduces a non-determinisim risk by doing so

olliecurtis (changes requested):

packages/bpk-component-chip/src/BpkIconChip.tsx
olliecurtis: Having a bit of think about this, I am wondering if we actually need a separate component here and instead to the existing `BpkSelectableChip` we just simply add a prop `iconOnly` similar to the `BpkButton` which will just apply the icon only CSS

Iain530: I've updated to use `!children` implies icon-only in BpkSelectableChip. This avoids className issues but might make it harder to extend BpkSelectableChip in future, what do you think? `iconOnly` may be a better long term solution?

I see the other chip types just hardcode certain props so it seemed consistent to have a new component. Happy to remove though if that's preferred.

olliecurtis (changes requested):

olliecurtis (approved):

PR #3281, opened 2024-03-14
[BpkMobileScrollContainer] Add  aria label

R1CC1M:

- Add aria-label to BpkMobileScrollContainer

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-horizontal-nav/src/__snapshots__/BpkHorizontalNav-test.js.snap
+++ b/packages/bpk-component-horizontal-nav/src/__snapshots__/BpkHorizontalNav-test.js.snap
@@ -10,7 +10,6 @@ exports[`BpkHorizontalNav should render correctly 1`] = `
       class="bpk-mobile-scroll-container__scroller"
     >
       <nav
-        aria-label=""
         class="bpk-mobile-scroll-container__inner"
       >
         <div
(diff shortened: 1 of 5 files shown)


packages/bpk-component-mobile-scroll-container/src/BpkMobileScrollContainer.js
```diff
@@ -97,17 +97,19 @@ type State = {
...
   className: PropTypes.string,
   leadingIndicatorClassName: PropTypes.string,
   trailingIndicatorClassName: PropTypes.string,
-  style: PropTypes.object,  
+  style: PropTypes.object,
   showScrollbar: PropTypes.bool,
 };
 
 const defaultProps = {
+  ariaLabel: '',
```
Sybsw, line 112: I don't think this is the correct approach because it modifies behaviour of consumers that don't use the prop unnecessarily.
The value should be able to be null if not supplied, in which case it should not apply the arialabel tag to the element.
This way it would provide the flexibility for consumers to give "" as the string if they want to but also maintain the old behaviour of not adding it to consumers that don't use it.

packages/bpk-component-horizontal-nav/src/__snapshots__/BpkHorizontalNav-test.js.snap
```diff
@@ -176,6 +182,7 @@ exports[`BpkHorizontalNav should render correctly with custom "trailingScrollInd
       class="bpk-mobile-scroll-container__scroller"
     >
       <nav
+        aria-label=""
```
Sybsw, line 185: In reference to the other comment, there should be a tests with both it adding and not adding the aria-label tag

Sybsw (changes requested):

Sybsw (approved):

PR #3280, opened 2024-03-13
Usage of className in BpkAccordionItem replaced with wrapper spans

Sybsw:
Placing usage of className prop in two internal components with spans set to the same style classes

Remember to include the following changes:

- [x] Ensure the PR title includes the name of the component you are changing so it's clear in the release notes for consumers of the changes in the version e.g `[KOA-123][BpkButton] Updating the colour`
- [ ] `README.md` (If you have created a new component)
- [ ] Component `README.md`
- [x] Tests
- [ ] Storybook examples created/updated
- [ ] Type declaration (`.d.ts`) files updated
- [ ] For breaking changes or deprecating components/properties, migration guides added to the description of the PR. If the guide has large changes, consider creating a new Markdown page inside the component's docs folder and link it here

PR #3283, opened 2024-03-14
LOOM-1275 BkpCloseButton - Move Icon style override into span

jronald01:
**CSS non-determinism / Class 2 - BackPack overrides**:
- Moving the icon styling into parent `<span>`
- Needed to change `vertical-align` from `-1` to `-1.75` to fix alignment after adding the `<span>`
  - The `<span>` appears to add additional vertical pixels, e.g. with zoom changes from 32x32 to 32x38.5 and without zoom changes from 16x16 to 16x19, which can be fixed via the `vertical-align`.

- [ ] Component `README.md`
- [ ] Tests


Sybsw (dismissed):

jronald01:
> From the Percy changes it looks like in the new changes it has shifted the margins up a little bit so the icon is shifted up in alignment which we need to check into

Ah, I see it now - I couldn't see the vertical shift at first as I had the side-by-side view selected in Percy rather than overlay mode.

mungodewar:
> > From the Percy changes it looks like in the new changes it has shifted the margins up a little bit so the icon is shifted up in alignment which we need to check into
> 
> Ah, I see it now - I couldn't see the vertical shift at first as I had the side-by-side view selected in Percy rather than overlay mode.

Further decreasing the vertical-alignment, seems to do the trick. `vertical-align: -2 * $button-padding;` although I'm not sure I understand _why_.

Sybsw:
Almost identical effect happening in the other PR: https://github.com/Skyscanner/backpack/pull/3280
https://percy.io/096a694a/backpack/builds/33186525/changed/1819236201?browser=chrome&browser_ids=56&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=original&width=1280&widths=375%2C1280

olliecurtis:
@Sybsw @jimronald - I am guessing the reason we are seeing this issue in this PR and the one Simon mentioned is probably due to the new `span` and `div` being used across both PRs are probably introducing this as I think Normalize CSS probably adds some change to the raw html of these so that it impacts these.

olliecurtis (approved):

PR #3297, opened 2024-03-20
LOOM-1261 BpkAccordionItem - Moved className props out into existing and new spans

Sybsw:
Replacement of https://github.com/Skyscanner/backpack/pull/3280

Moving classnames into existing outer spans
Put a span around the upper BpkText and modified the existing span around BpkIcon to have `display:inherit` because previous value of `display: inline-block` was invalid in the context

The parent element relies on flex so did not want to re-engineer that part unless necessary.

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-accordion/src/BpkAccordionItem.module.scss
```diff
@@ -57,7 +57,7 @@
   }
 
   &__title-text {
-    display: inline-block;
+    display: inherit;
```
Sybsw, line 60: `display: inline-block` doesn't make sense anyway as it is within the class above which has `display: inline-flex`

mungodewar: What `display` value is `inherit`ed?

olliecurtis (approved):

packages/bpk-component-accordion/src/BpkAccordionItem.module.scss
Sybsw: `display: inline-flex;` [from this line](https://github.com/Skyscanner/backpack/pull/3297/files#diff-5b1965da520f4d42beb4451d6dabd6724b1b130710c78ca5ced991e1709ce67aR52)

Sybsw: From Mungo's advice, changed to `display:block` with change to tagName to h3

packages/bpk-component-accordion/src/BpkAccordionItem.module.scss
```diff
@@ -57,7 +57,7 @@
   }
 
   &__title-text {
-    display: inline-block;
+    display: block;
```
mungodewar, line 60: We _could_ leave this out as title-text is a child of a flex container it will default to `block`. Maybe it's good signposting? 

Although, what is the point of setting a `span` as a `block` level element? A span with `display: block` might as well be a `div`:

```
<div className={titleTextClassNames.join(' ')}>
```

If this change is made, we can remove this display value.

mungodewar (approved):
LGTM, one suggestion/optimisation to be able to remove the `display`

mungodewar (approved):

PR #3298, opened 2024-03-21
BpKContentCards LOOM-1276

jronald01:
CSS non-determinism / Class 2 - BackPack overrides:

Moving the margin into new parent <div>
Also converted conditional breakpoint styling from scss into a regular breakpoint.

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-content-cards/src/BpkContentCards.module.scss
+++ b/packages/bpk-component-content-cards/src/BpkContentCards.module.scss
@@ -24,14 +24,8 @@
 @use '../../unstable__bpk-mixins/typography';
 
 .bpk-content-cards {
-  &--header-text {
+  &--header {
     margin-bottom: tokens.bpk-spacing-base();
-
-    @include typography.bpk-heading-2;
-
-    @include breakpoints.bpk-breakpoint-mobile {
-      @include typography.bpk-heading-3;
-    }
   }
 
   &--layout {
--- a/packages/bpk-component-content-cards/src/BpkContentCards.js
+++ b/packages/bpk-component-content-cards/src/BpkContentCards.js
@@ -18,9 +18,12 @@
 /* eslint react/no-array-index-key: 0 */
 
 // @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
-import BpkText from '../../bpk-component-text';
+import BpkText, { TEXT_STYLES } from '../../bpk-component-text';
 // @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
 import { cssModules } from '../../bpk-react-utils';
+import BpkBreakpoint, {
+  BREAKPOINTS,
+} from "../../bpk-component-breakpoint";
 
 import BpkContentCard from './BpkContentCard';
 import STYLES from './BpkContentCards.module.scss';
@@ -47,12 +50,25 @@ const BpkContentCards = ({ cards, heading }: Props) => {
 
   return (
     <div>
-      <BpkText
-        tagName="h2"
-        className={getClassName('bpk-content-cards--header-text')}
-      >
-        {heading}
-      </BpkText>
+      <div className={getClassName('bpk-content-cards--header')}>
+        <BpkBreakpoint query={BREAKPOINTS.MOBILE}>
+            {isActive => (isActive ?
+              <BpkText
+                tagName="h3"
+                textStyle={TEXT_STYLES.heading3}
+              >
+                {heading}
+              </BpkText>
+            :
+              <BpkText
+                tagName="h2"
+                textStyle={TEXT_STYLES.heading2}
+              >
+                {heading}
+              </BpkText>
+            )}
+        </BpkBreakpoint>
+      </div>
       <div role="list" className={getClassName('bpk-content-cards--layout')}>
         {cards.map((card, index) => (
           <div role="listitem" key={index}>


packages/bpk-component-content-cards/src/BpkContentCards.module.scss
```diff
@@ -23,17 +23,18 @@
 @use '../../unstable__bpk-mixins/breakpoints';
 @use '../../unstable__bpk-mixins/typography';
 
-.bpk-content-cards {
-  &--header-text {
-    margin-bottom: tokens.bpk-spacing-base();
+.headerText {
+  margin-top: inherit;
```
mungodewar, line 27: Is this new?

jronald01: Ah yeah, this is needed to keep the alignment:
Screenshot 2024-03-21 at 15 00 53

jronald01: Otherwise it slips down:

Screenshot 2024-03-21 at 15 01 48

mungodewar: Should we just set it to a defined value rather than `inherit` or is there a reason `inherit` is needed here?

jronald01: So I think without it, we are getting the browser default for <h2>, but yeah we could use a define value.

With:
Screenshot 2024-03-21 at 15 07 57

Without:
Screenshot 2024-03-21 at 15 07 40

jronald01: I can set it to 0.

jronald01: Browser default: https://www.w3schools.com/cssref/css_default_values.php
`<h2>` margin top is 0.83rem = 13.28px
We are seeing exactly double that above 26.56px which cannot be coincidental.
Anyway, setting it to 0 reinstates it to what it should be.

jronald01: Aha, nice, I've made the change and pushed it 👍

jronald01: Also now fixed stylelint issues

packages/bpk-component-content-cards/src/__snapshots__/BpkContentCards-test.js.snap
```diff
@@ -3,9 +3,7 @@
 exports[`Content Card widget renders correctly with 1 card 1`] = `
 <DocumentFragment>
   <div>
-    <h2
-      class="bpk-text bpk-text--body-default bpk-content-cards--header-text"
-    >
+    <h2>
```
mungodewar, line 6: It's odd that these snaps don't have any `className` included, do you know why?

packages/bpk-component-content-cards/src/BpkContentCards.module.scss
```diff
@@ -23,17 +23,18 @@
 @use '../../unstable__bpk-mixins/breakpoints';
 @use '../../unstable__bpk-mixins/typography';
 
-.bpk-content-cards {
-  &--header-text {
-    margin-bottom: tokens.bpk-spacing-base();
+.headerText {
+  margin-bottom: tokens.bpk-spacing-base();
```
mungodewar, line 27: Does this `margin-bottom` not get overridden by `@include typography.bpk-text;`?

jronald01: Ah, without it, the image and content are no longer aligned as they were previously, which means the Percy snapshot match will fail:

With:
Screenshot 2024-03-25 at 09 36 55

Without:
Screenshot 2024-03-25 at 09 40 10

mungodewar: Doesn't override, cool 👍

packages/bpk-component-content-cards/src/__snapshots__/BpkContentCards-test.js.snap
jronald01: Ah Mungo, sorry, still catching up with onboarding to Backpack, what do you mean by **odd**?

I cannot see any reference to `className` in the snap on `main` either, i.e. there are no class names in the existing snap.

For reference, there are 167 `.snap` files and only 84 have class names in them.

mungodewar: Just meaning that we are rendering: 

`<h2 className={STYLES.headerText}>{heading}</h2>`

but within the snapshot (above) we see no `class` on the `h2`.

I _think_ its cause we aren't now using `getClassName`?

packages/bpk-component-content-cards/src/BpkContentCards.module.scss
jronald01: Ah, I need to look back into this some user error on my side with branches - I need to check the above again.

packages/bpk-component-content-cards/src/__snapshots__/BpkContentCards-test.js.snap
jronald01: Reverted to using `getClassName` - DONE 👍

packages/bpk-component-content-cards/src/BpkContentCards.module.scss
jronald01: Cut a long story short:

jronald01: `bpk-paragraph` - https://github.com/Skyscanner/backpack/blob/main/packages/bpk-mixins/_typography.scss#L663
```
@mixin bpk-paragraph {
  margin-top: $bpk-spacing-none;
  margin-bottom: bpk-spacing-base();
}
```

**versus**

`bpk-text` https://github.com/Skyscanner/backpack/blob/main/packages/bpk-mixins/_typography.scss#L64
```
@mixin bpk-text {
  margin: 0;
}
```

jronald01: `margin: 0` isn't enough, we need that `margin-bottom` to be set to `bpk-spacing-base()` to match the alignment on `main`.

**I've also added some Visual Test stories**, we can reduce these though.

mungodewar: Thanks Jim, LGTM. On the VT, my take is that we don't need all of them. My bias would be to the more complex example 3 cards & have that be the only VT for the component. Just thinking purely of extra cost on Percy tbh.

jronald01: Ah cool, I'll leave the three card Visual Test and remove the others 👍

mungodewar (approved):

PR #3299, opened 2024-03-21
BpkBreakpoint: Fix Server Side Rendering

robaw:
This change simplifies some of the logic inside the BpkBreakpoint, moving the matching logic entirely within useMediaQuery. This also should fix Server Side Rendering, as previously this was giving the error that window was not defined, as we were doing a bad job at detecting whether the `window` and `matchMedia` function  was present (see [this](https://skyscanner.slack.com/archives/C0JHPDSSU/p1711001772312839) slack thread).

## Migration guide

Any component being rendered in tests that rely on the BpkBreakpoint will now require a mock to explicitely define what the expected behaviour is. See the [README](https://github.com/Skyscanner/backpack/blob/b4fd8e3900bbd4a18e9426b96f143ecbcbeadd58/packages/bpk-component-breakpoint/README.md) for more details and examples.

- [ ] Component `README.md`
- [x] Tests


packages/bpk-component-breakpoint/src/BpkBreakpoint-test.tsx
```diff
@@ -22,99 +22,111 @@ import ReactDOMServer from 'react-dom/server';
...
-          matches ? <div>matches</div> : <div>does not match</div>
-        }
-      </BpkBreakpoint>,
-    );
-    expect(asFragment()).toMatchSnapshot();
-  });
-
-  describe('SSR mode', () => {
-    beforeEach(() => {
+  describe('children as component', () => {
```
robaw, line 25: BpkBreakpoint responsibilities have changed. It now only takes the match response from `useMediaQuery()` and works out whether to:

- render the children
- call the children function

packages/bpk-component-breakpoint/src/useMediaQuery.ts
```diff
@@ -16,15 +16,17 @@
...
 
-const useMediaQuery = (query: string): boolean => {
   const [matches, setMatches] = useState(
-    window.matchMedia ? window.matchMedia(query).matches : false,
+    isClient ? window.matchMedia(query).matches : matchSSR,
   );
 
   useEffect(() => {
-    if (window.matchMedia) {
+    if (isClient) {
```
anambl, line 29: the `useEffect` only runs on client, do we need this check? Unless it's possible that the `window` is defined, but `matchMedia` isn't? 🤔

robaw: That is true, however, there are unit tests in backpack that are not mocking BpkBreakpoint, and so the useEffect runs and fails.

anambl: why do they fail? I'm not sure I understand

robaw: You can see from [this previous run](https://github.com/Skyscanner/backpack/actions/runs/8298787263/job/22713096089) that unit tests for components that relied on BpkBreakpoint were failing because:

- Unit tests do execute the `useEffect` unlike when SSR'ing
- Unit tests do not provide the full window object

anambl: Ahh the GH actions logs are a mess 😅 I can only see
```
TypeError: window.matchMedia is not a function

      2 |
      3 | const useMediaQuery = (query: string) => {
    > 4 |   const [matches, setMatches] = useState(window.matchMedia(query).matches);
```
I would think that mocking the `matchMedia` on the window object would solve the issue?
Doesn't feel right to add this extra check just for unit tests to pass 🤔

robaw: Added a mock to all of the tests that were failing due to this and removed the check within the useEffect

packages/bpk-component-breakpoint/src/useMediaQuery.ts
```diff
@@ -16,25 +16,24 @@
...
-import { useState, useEffect } from 'react';
+import { useEffect, useState } from 'react';
+
+const useMediaQuery = (query: string, matchSSR = false): boolean => {
+  const isClient = typeof window !== 'undefined' && !!window.matchMedia;
 
-const useMediaQuery = (query: string): boolean => {
   const [matches, setMatches] = useState(
-    window.matchMedia ? window.matchMedia(query).matches : false,
+    isClient ? window.matchMedia(query).matches : matchSSR,
```
runmoore, line 25: For my own benefit, what does "matches" mean in this context? Particularly in the context of `matchSSR`?

The docs mention:

> You can pass matchSSR which will instruct the breakpoint to match any time it is rendered on the server.

But this isn't clear to me. Maybe it's worth improving our guidance around this as I think our guidance is pretty light here

robaw: > Does it mean if matchSSR is true this component will always render on the server?

Yes, exactly

> Do we attempt to guess the screen size on the server at all?

Nope, this isn't something this component has done, which is why it has the "what do you want to match on the server" approach.

> But this isn't clear to me. Maybe it's worth improving our guidance around this as I think our guidance is pretty light here

I can have a think about re-phrasing this, feel free to suggest something that is clearer to someone that doesn't have the internal workings in their head

runmoore: I was just having a play around and in Skyscanner via request-context we have access to the booleans `isMobile`, `isTablet` and `isBrowser`(?). 

My first option would be to change the API to `alwaysRenderWhenSSR`, as this is unambiguous what will happen and the word `match` on it's own isn't clear in this context. If we don't want to change the API, expanding our docs would be the next best option. 

Something like the below with an explanation would help convey that any consumer of this API would want to make at least an informed guess on the device size for SSR, otherwise if you have multiple `BpkBreakpioint` components, chances are the majority of users will see a flash of one component in the wrong device setting, following by the correct component once the component has mounted and executed the client side only code

```jsx
// `matchSSR` - Always render this component on the server. You will likely want to make an informed guess from the request object what size the users device is.
<BpkBreakpoint query={BREAKPOINTS.TABLET} matchSSR={isMobilePhone}>
  {mobileContent}
<BplBrealpoint/>
```

robaw: Yup I think that's an improvement in the guidance, I like the push towards using other data to estimate what breakpoint is likely to match on the client-side. Pushed a change that is similar to this, let me know what you think.

packages/bpk-component-breakpoint/README.md
```diff
@@ -24,17 +24,21 @@ export default () => (
 
 ### Server Side Render (SSR) Support
 
-You can pass `matchSSR` which will instruct the breakpoint to match any time it is rendered on the server.
+You can pass `matchSSR` which will instruct the breakpoint to match any time it is rendered on the server. You can use
+this to estimate what breakpoint is likely to match on the client-side.
 
 ```js
-<BpkBreakpoint query={BREAKPOINTS.TABLET} matchSSR>
+<BpkBreakpoint query={BREAKPOINTS.TABLET} matchSSR={isTablet}>
```
anambl, line 31: I really like this, I didn't know we have this info from the request-context. Maybe we can add an import as well to clarify where the `isTablet` boolean would come from? Maybe it's not just the request-context and can be something else, but just to clarify that it's not part of the API of `BpkBreakpoint`

robaw: Did think about that, but request-context is something specific to Skyscanner and not open-source, do we lean towards providing Skyscanner-specific docs in backpack?

packages/bpk-component-breakpoint/src/useMediaQuery.ts
robaw: I think we'll have to give some guidance to people, as any consumers of BpkBreakpoint, or any component that uses BpkBreakpoint will have to do mocks in their tests too.

packages/bpk-component-breakpoint/README.md
anambl: > do we lean towards providing Skyscanner-specific docs in backpack?

No, we don't. I think it's fine to just add any dummy import just to clarify it's not part of the API 🤔

robaw: added dummy import

packages/bpk-component-breakpoint/src/useMediaQuery.ts
```diff
@@ -16,25 +16,24 @@
...
-      };
-      media.addEventListener('change', listener);
-      return () => media.removeEventListener('change', listener);
-    }
-    return undefined;
-  }, [query]);
+    };
+    media.addEventListener('change', listener);
+    return () => media.removeEventListener('change', listener);
+  }, [query, isClient]);
```
anambl, line 36: we don't have `isClient` as a dependency anymore so we can remove this?

robaw: Removed

anambl:
@robaw just one more quick thing - could you add a few words in the PR about what is the improvement for consumers to adopt this?

packages/bpk-component-breakpoint/README.md
```diff
@@ -24,17 +24,57 @@ export default () => (
...
 </BpkBreakpoint>
 ```
 
+If you match to a different breakpoint when rendering on the server, than what is matched to in the traveller's browser,
+then React will print a warning saying there is a mismatch.
+
+### Testing
+
+When writing tests for any components that use BpkBreakpoint, you will have to mock either the `BpkBreakpoint` component
+or the underlying `useMediaQuery`. This is because the `window.matchMedia` function that we rely on does not exist in
```
runmoore, line 47: Is it worth linking to the official jest docs on this? https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom

Do you think Backpack could provide a sample function that might make this easier for consumers? e.g. allow consumers to add a line in `jest.setup`

```
import '@skyscanner/backpack-web/bpk-breakpoint/matchMedia.mock'; 
```

robaw: It depends. If we recommend baking a mock into the jest config, we'd have to pick if we wanted the default to be to match everything or match nothing. Which might not be that useful.

The way we are recommending here is to be explicit about which viewport you are emulating and that seems safer (albeit more faff).

anambl (approved):

robaw:
@runmoore are you good with how this PR looks to approve?

LouiseReid:
Have we tested this in a microsite that uses this component?

robaw:
> Have we tested this in a microsite that uses this component?

Tested by copying/pasting into legal-pages, which server-side renders its components.

PR #3320, opened 2024-03-26
LOOM-1274 BpkBreadcrumb Class 2 Overrides

jronald01:
**Add existing Bpk child component overrides to wrappers.**

Had to set a specific line height after adding the wrapper around the 16x16 or 32x32(zoomed) icon to retain the correct alignment, i.e. adding the wrapper increases the height of the right arrow icon which then affects the alignment of the right arrow within the breadcrumb.

**Without** the line height fix, the right arrows are vertically shifted up:
<kbd> Screenshot 2024-03-28 at 11 37 19 </kbd>

**With** the line height fix, the right arrows are back to where they should be centrally aligned:
<kbd> Screenshot 2024-03-28 at 11 39 05 </kbd>

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-breadcrumb/src/BpkBreadcrumbItem.module.scss
+++ b/packages/bpk-component-breadcrumb/src/BpkBreadcrumbItem.module.scss
@@ -1,20 +1,20 @@
 /*
-* Backpack - Skyscanner's Design System
-*
-* Copyright 2016 Skyscanner Ltd
-*
-* Licensed under the Apache License, Version 2.0 (the "License");
-* you may not use this file except in compliance with the License.
-* You may obtain a copy of the License at
-*
-*   http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing, software
-* distributed under the License is distributed on an "AS IS" BASIS,
-* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-* See the License for the specific language governing permissions and
-* limitations under the License.
-*/
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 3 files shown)


packages/bpk-component-breadcrumb/src/BpkBreadcrumbItem.module.scss
```diff
@@ -19,6 +19,11 @@
 @use '../../unstable__bpk-mixins/tokens';
 @use '../../unstable__bpk-mixins/typography';
 
+/* After adding the wrapper around the icon, need to set the line
+    height to stop the 16x16 or 32x32(zoomed) icon from expanding
+    verically past it's 16 or 32 pixels height */
+$icon-height: tokens.$bpk-line-height-xs / 2;
```
olliecurtis, line 25: Given this is only used in one place does this need a whole dedicated variable just for one line further down?

olliecurtis (changes requested):

packages/bpk-component-breadcrumb/src/BpkBreadcrumbItem.module.scss
jronald01: Ah, this is because of Styelint, i.e.:
Screenshot 2024-03-28 at 12 02 28

It's fine if I am just doing:
```
  &__arrow {
    margin: 0 tokens.bpk-spacing-sm();
    line-height: tokens.$bpk-line-height-xs;    <<--------------
    fill: tokens.$bpk-text-disabled-day;
  }
```

but when I try and divide it by 2, stylelint wants a separate variable.

jronald01: The other option is I can set it to a specific pixel value of no greater than 12 pixels before the alignment starts to break again for the regular and zoomed in icon, e.g.:
`line-height: 12px;`

Or the other option is:
`line-height: 0;` which forces the element to only use the vertical space it needs.

But both will need a stylelint disable.

Or `line-height: 75%`, `line-height: 0.75`, etc.

But again, all will need stylelint disable:

```
42:5   ✖  Expected variable, function or keyword for "0.75" of "line-height"           scale-unlimited/declaration-strict-value
```

jronald01: Screenshot 2024-03-28 at 12 18 10

jronald01: Ah yeah, I actually first attached a screenshot with brackets and then I replaced it for one without the brackets:
Screenshot 2024-03-28 at 14 11 24

jronald01: So with or without brackets produces the same result basically.

olliecurtis: No worries in which case approved!

olliecurtis (approved):

PR #3335, opened 2024-04-04
[LOOM-1291][AnimateAndFade] className moved up one level

metalix2:
`AnimateAndFade` which wraps the `BpkAnimateHeight`, passes a `classname` prop to be consumed by the `BpkAnimateHeight`, The `BpkAnimateHeight` doesn't set any of it's own `classname` props so there is currently no risk of non-determinism style behaviour.

This change moves the class name to a parent element to eliminate any future risks where `BpkAnimateHeight` may have a classname in the future.

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-banner-alert/src/AnimateAndFade.tsx
+++ b/packages/bpk-component-banner-alert/src/AnimateAndFade.tsx
@@ -40,7 +40,7 @@ type Props = {
   animateOnLeave: boolean;
   children: ReactNode | string;
   show: boolean;
-  className?: string | null;
+  className?: string | undefined;
 };
 
 type State = {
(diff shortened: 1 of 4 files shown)


packages/bpk-component-banner-alert/src/AnimateAndFade.tsx
```diff
@@ -57,7 +57,6 @@ class AnimateAndFade extends Component<Props, State> {
   static defaultProps = {
     animateOnEnter: false,
     animateOnLeave: false,
-    className: null,
```
metalix2, line 60: We shouldn't be setting a null value as a className on a react element. It should be defined or undefined.

mungodewar (approved):
LGTM,  could be a potential breaking change given its user supplied className and the location of it's application has changed. However, risk is low.

PR #3337, opened 2024-04-07
[MAT-1205] implement skeleton

felix-luo-sc:

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/examples/bpk-component-skeleton/examples.tsx
+++ b/examples/bpk-component-skeleton/examples.tsx
@@ -1,0 +1,92 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import BpkBaseSkeleton from '../../packages/bpk-component-skeleton/src/BpkBaseSkeleton';
+import{ 
+  BpkImageSkeleton, 
+  BpkBodyTextSkeleton, 
+  BpkCircleSkeleton,
+  BpkHeadlineSkeleton,
+  SIZE_TYPES, 
+  IMAGE_SKELETON_STYLE 
+} from '../../packages/bpk-component-skeleton';
+
+/* eslint-disable backpack/use-tokens */
+const BaseSkeletonExample = () => <BpkBaseSkeleton className='' />;
+
+const ImageDefaultSizeWithDefaultStyleExample = () => <BpkImageSkeleton ariaLabel='loading' />
+const ImageDefaultSizeWithRoundedStyleExample = () => <BpkImageSkeleton ariaLabel='loading' style={IMAGE_SKELETON_STYLE.rounded} />
+const ImageCustomNumberSizeWithDefaultStyleExample = () => <BpkImageSkeleton size={{width: 7, height: 6}} ariaLabel='loading' />
+const ImageCustomStringSizeWithRoundedStyleExample = () => <BpkImageSkeleton size={{width: '7rem', height: '6rem'}} ariaLabel='loading' style={IMAGE_SKELETON_STYLE.rounded} />
+
+const BodyTextSmallSizeExample = () => <BpkBodyTextSkeleton size={SIZE_TYPES.small} ariaLabel='loading' />
+const BodyTextDefaultSizeExample = () => <BpkBodyTextSkeleton size={SIZE_TYPES.default} ariaLabel='loading' />
+const BodyTextLargeSizeExample = () => <BpkBodyTextSkeleton size={SIZE_TYPES.large} ariaLabel='loading' />
+const BodyTextCustomSizeExample = () => <BpkBodyTextSkeleton size={{width: 16, height: 1}} ariaLabel='loading' />
+
+const CircleSmallSizeExample = () => <BpkCircleSkeleton size={SIZE_TYPES.small} ariaLabel='loading'/>;
+const CircleDefaultSizeExample = () => <BpkCircleSkeleton ariaLabel='loading' />;
+const CircleCustomSizeExample = () => <BpkCircleSkeleton size={{width: 5, height: 5}} ariaLabel='loading' />;
+
+
+const HeadlineSmallSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.small} ariaLabel='loading' />
+const HeadlineDefaultSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.default} ariaLabel='loading' />
+const HeadlineLargeSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.large} ariaLabel='loading' />
+const HeadlineCustomSizeExample = () => <BpkHeadlineSkeleton size={{width: 18, height: 1.5}} ariaLabel='loading' />
+
+const VisualTestExample = () => 
+  <>
+    <ImageDefaultSizeWithDefaultStyleExample /><br />
+    <ImageDefaultSizeWithRoundedStyleExample /><br />
+    <ImageCustomNumberSizeWithDefaultStyleExample /><br />
+    <ImageCustomStringSizeWithRoundedStyleExample /><br />
+
+    <BodyTextSmallSizeExample /><br />
+    <BodyTextDefaultSizeExample /><br />
+    <BodyTextLargeSizeExample /><br />
+    <BodyTextCustomSizeExample /><br />
+
+    <CircleSmallSizeExample /><br />
+    <CircleDefaultSizeExample /><br />
+    <CircleCustomSizeExample /><br />
+
+    <HeadlineSmallSizeExample /><br />
+    <HeadlineDefaultSizeExample /><br />
+    <HeadlineLargeSizeExample /><br />
+    <HeadlineCustomSizeExample /><br />
+  </>
+/* eslint-enable backpack/use-tokens */
+export {
+  BaseSkeletonExample,
+  ImageDefaultSizeWithDefaultStyleExample,
+  ImageDefaultSizeWithRoundedStyleExample,
+  ImageCustomNumberSizeWithDefaultStyleExample,
+  ImageCustomStringSizeWithRoundedStyleExample,
+  BodyTextSmallSizeExample,
+  BodyTextDefaultSizeExample,
+  BodyTextLargeSizeExample,
+  BodyTextCustomSizeExample,
+  CircleSmallSizeExample,
+  CircleDefaultSizeExample,
+  CircleCustomSizeExample,
+  HeadlineSmallSizeExample,
+  HeadlineDefaultSizeExample,
+  HeadlineLargeSizeExample,
+  HeadlineCustomSizeExample,
+  VisualTestExample,
+};
--- a/examples/bpk-component-skeleton/stories.ts
+++ b/examples/bpk-component-skeleton/stories.ts
@@ -1,0 +1,70 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import BpkBaseSkeleton from '../../packages/bpk-component-skeleton/src/BpkBaseSkeleton';
+
+import {
+  BaseSkeletonExample,
+  ImageDefaultSizeWithDefaultStyleExample,
+  ImageDefaultSizeWithRoundedStyleExample,
+  ImageCustomNumberSizeWithDefaultStyleExample,
+  ImageCustomStringSizeWithRoundedStyleExample,
+  BodyTextSmallSizeExample,
+  BodyTextDefaultSizeExample,
+  BodyTextLargeSizeExample,
+  BodyTextCustomSizeExample,
+  CircleSmallSizeExample,
+  CircleDefaultSizeExample,
+  CircleCustomSizeExample,
+  HeadlineSmallSizeExample,
+  HeadlineDefaultSizeExample,
+  HeadlineLargeSizeExample,
+  HeadlineCustomSizeExample,
+  VisualTestExample,
+} from './examples';
+
+export default {
+  title: 'bpk-component-skeleton',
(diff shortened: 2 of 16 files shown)


examples/bpk-component-skeleton/stories.ts
```diff
@@ -0,0 +1,81 @@
...
+export const HeadlineSmallSize = HeadlineSmallSizeExample;
+export const HeadlineDefaultSize = HeadlineDefaultSizeExample;
+export const HeadlineLargeSize = HeadlineLargeSizeExample;
+export const HeadlineCustomSize =HeadlineCustomSizeExample;
+
+export const VisualTest = VisualTestExample;
+export const VisualTestWithZoom = VisualTest.bind({});
+VisualTestWithZoom.args = {
+  zoomEnabled: true
+};
```
olliecurtis, line 81: For this component I think a visual test is not going to be great as the snapshots will be inconsistent between runs and constantly cause Percy CI to fail for PRs.

So my suggestion here would be to remove these stories and we will just need to use manual checks to verify all is well

packages/bpk-component-skeleton/src/BpkBaseSkeleton.module.scss
```diff
@@ -0,0 +1,49 @@
...
+  background-repeat: no-repeat;
+  animation: shimmer-anim tokens.$bpk-duration-base * 4 infinite linear;
+  animation-direction: normal;
+
+  @include utils.bpk-rtl {
+    animation-direction: reverse;
+  }
+}
+
+@keyframes shimmer-anim {
```
lynn-sc, line 41: Is it better to write its name clearly here, such as `shimmer-animation`?

packages/bpk-component-skeleton/src/BpkBaseSkeleton.tsx
```diff
@@ -0,0 +1,49 @@
...
+const BpkBaseSkeleton = ({
+  ariaLabel,
+  className,
+  styleObj
+ }: Props) => {
+
+  const classNames = [getClassName('bpk-skeleton')];
+  if (className) {
+    classNames.push(className);
+  }
```
lynn-sc, line 43: Could this be simplified to `const classNames = getClassName('bpk-skeleton', className)`? What do you think?

packages/bpk-component-skeleton/src/BpkBaseSkeleton.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <div role="status" aria-label={ariaLabel} className={classNames.join(' ')} style={styleObj}  />
```
lynn-sc, line 46: So I think `className={classNames.join(' ')}` can be `className={classNames}`

packages/bpk-component-skeleton/src/BpkBodyTextSkeleton.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+type Props = {
+  size?: BodyTextSizeType,
+  className?: string;
+  ariaLabel?: string;
+};
+
+const BpkBodyTextSkeleton = ({
+  ariaLabel,
+  className,
```
lynn-sc, line 46: Same here

packages/bpk-component-skeleton/src/BpkCircleSkeleton.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+type Props = {
+  size?: CircleSizeType,
+  className?: string;
+  ariaLabel?: string;
+};
+
+const BpkCircleSkeleton = ({
+  ariaLabel,
+  className,
```
lynn-sc, line 45: Same here

packages/bpk-component-skeleton/src/BpkHeadlineSkeleton.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+type Props = {
+  size?: HeadlineSizeType,
+  className?: string;
+  ariaLabel?: string;
+};
+
+const BpkHeadlineSkeleton = ({
+  ariaLabel,
+  className,
```
lynn-sc, line 46: Same above

examples/bpk-component-skeleton/stories.ts
felix-luo-sc: That makes sense. The visual test has been removed.

packages/bpk-component-skeleton/README.md
```diff
@@ -0,0 +1,46 @@
...
+    size={SIZE_TYPES.small}
+    ariaLabel="loading"
+    className="custom-class"
+  />
+
+   <BpkHeadlineSkeleton
+    size={SIZE_TYPES.small}
+    ariaLabel="loading"
+    className="custom-class"
+  />
```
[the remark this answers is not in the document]
felix-luo-sc: Yes, agree with you. Updated.

packages/bpk-component-skeleton/src/BpkBaseSkeleton.module.scss
felix-luo-sc: Updated.

packages/bpk-component-skeleton/src/BpkBaseSkeleton.tsx
felix-luo-sc: Updated.

packages/bpk-component-skeleton/src/BpkBaseSkeleton.tsx
felix-luo-sc: Updated.

packages/bpk-component-skeleton/src/BpkBaseSkeleton.tsx
```diff
@@ -0,0 +1,49 @@
...
+    width: string;
+    height: string;
+  };
+};
+
+const BpkBaseSkeleton = ({
+  ariaLabel,
+  className,
+  styleObj
+ }: Props) => {
```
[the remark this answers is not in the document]
felix-luo-sc: Updated.

lynn-sc (approved):
LGTM

examples/bpk-component-skeleton/examples.tsx
```diff
@@ -0,0 +1,93 @@
...
+const CircleDefaultSizeExample = () => <BpkCircleSkeleton ariaLabel='loading' />;
+const CircleCustomSizeExample = () => <BpkCircleSkeleton size={{width: 5, height: 5}} ariaLabel='loading' />;
+
+
+const HeadlineSmallSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.small} ariaLabel='loading' />
+const HeadlineDefaultSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.default} ariaLabel='loading' />
+const HeadlineLargeSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.large} ariaLabel='loading' />
+const HeadlineCustomSizeExample = () => <BpkHeadlineSkeleton size={{width: 18, height: 1.5}} ariaLabel='loading' />
+
+const CombinedComponentExample = () =>
```
felix-luo-sc, line 52: Add a CombinedComponentExample for better manual test.

packages/bpk-component-skeleton/src/BpkBaseSkeleton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  className: string;
+  ariaLabel?: string;
+  styleObj?: {
+    width: string;
+    height: string;
+  };
+};
+
+const BpkBaseSkeleton = ({
+  ariaLabel = '',
```
olliecurtis, line 35: We should probably check this is ok to have a default for this vs making this a required field from an a11y perspective

packages/bpk-component-skeleton/src/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { CUSTOM_SIZE_TYPE } from "./common-types"
+
+const getCustomStyles = ({ height, width } : CUSTOM_SIZE_TYPE) => {
```
olliecurtis, line 21: I wonder if we can maybe avoid having to do this function by maybe making it clear in the API that this should be a size string of what you want. `6rem`, `4px`, etc that way it moves away from needing to have a custom function and this prop being used differently across the product so keeps it consistant for all

olliecurtis (changes requested):
A few changes/questions below.

I am also checking with the team relating to the use of className as we have recently introduced changes that means components shouldn't expose and use classNames for setting styles but instead doing it via props.

So it could be that the way we apply the styles to each type of Skeleton, is by passing a prop down from the outer component to the base.

E.g. type={SKELETON.circle}

So that all the styles are applied at one layer but not exposed to the consumer to avoid CSS non-determinism issues

examples/bpk-component-skeleton/examples.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+
+const HeadlineSmallSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.small} ariaLabel='loading' />
+const HeadlineDefaultSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.default} ariaLabel='loading' />
+const HeadlineLargeSizeExample = () => <BpkHeadlineSkeleton size={SIZE_TYPES.large} ariaLabel='loading' />
+const HeadlineCustomSizeExample = () => <BpkHeadlineSkeleton size={{width: 18, height: 1.5}} ariaLabel='loading' />
+
+const CombinedComponentExample = () => 
+  <>
+    <ImageDefaultSizeWithDefaultStyleExample /><br />
```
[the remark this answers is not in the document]
felix-luo-sc: Will change.

packages/bpk-component-skeleton/src/BpkBaseSkeleton.tsx
felix-luo-sc: Considering the translation issues, I think it would be better to make this a required field.

packages/bpk-component-skeleton/src/utils.ts
felix-luo-sc: Yes, only keeps strings is better.

packages/bpk-component-skeleton/src/BpkSkeleton.tsx
```diff
@@ -0,0 +1,97 @@
...
+  [SKELETON_TYPES.image]: (typeof SIZE_TYPES)['default'];
+  [SKELETON_TYPES.bodyText]: (typeof SIZE_TYPES)[keyof typeof SIZE_TYPES];
+  [SKELETON_TYPES.circle]: Exclude<(typeof SIZE_TYPES)[keyof typeof SIZE_TYPES], 'large'>
+  [SKELETON_TYPES.headline]: (typeof SIZE_TYPES)[keyof typeof SIZE_TYPES];
+};
+
+type ComponentProps = {
+  type: Extract<SkeletonType, 'image'>
+  size?: SizeMap['image'] | CUSTOM_SIZE_TYPE;
+  ariaLabel: string;
```
olliecurtis, line 62: Just speaking with A11y folks on this one we will not want to have an ariaLabel at all on this Skeleton component and actually the skeleton should be `aria-hidden` until it loads. 

I have updated the API in the Figma design to reflect this

packages/bpk-component-skeleton/src/BpkSkeleton.tsx
```diff
@@ -0,0 +1,97 @@
...
+    styleObj = size;
+  } else {
+    classNames.push(getClassName(`bpk-skeleton__${type}--${size}`))
+  }
+
+  if(type === SKELETON_TYPES.image) {
+    if(props.style === IMAGE_SKELETON_STYLE.rounded) {
+      classNames.push(getClassName('bpk-skeleton__image--rounded'));
+    }
+  }
```
olliecurtis, line 85: This can be simplified to be

```suggestion
  const isImageRounded = type === SKELETON_TYPES.image && props.style === IMAGE_SKELETON_STYLE.rounded
  const classNames:string = getClassName(
      `bpk-skeleton__${type}`,
      typeof size != 'object' && `bpk-skeleton__${type}--${size}`,
      isImageRounded && 'bpk-skeleton__image--rounded'
   );
  const styleObj = typeof size === 'object' ? size : null;
```

packages/bpk-component-skeleton/src/BpkSkeleton.tsx
```diff
@@ -0,0 +1,97 @@
...
+  default: 'default',
+} as const;
+export type ImageSkeletonStyle = (typeof IMAGE_SKELETON_STYLE)[keyof typeof IMAGE_SKELETON_STYLE];
+
+type SizeMap = {
+  [SKELETON_TYPES.image]: (typeof SIZE_TYPES)['default'];
+  [SKELETON_TYPES.bodyText]: (typeof SIZE_TYPES)[keyof typeof SIZE_TYPES];
+  [SKELETON_TYPES.circle]: Exclude<(typeof SIZE_TYPES)[keyof typeof SIZE_TYPES], 'large'>
+  [SKELETON_TYPES.headline]: (typeof SIZE_TYPES)[keyof typeof SIZE_TYPES];
+};
```
olliecurtis, line 57: Thinking of these to make this file a bit more readable to place these inside the `common-types.tsx` file

olliecurtis (changes requested):

packages/bpk-component-skeleton/src/BpkSkeleton.tsx
felix-luo-sc: I see. Thank you for confirming. Let me update the pr.

packages/bpk-component-skeleton/src/BpkSkeleton.tsx
felix-luo-sc: Updated.

packages/bpk-component-skeleton/src/BpkSkeleton.tsx
felix-luo-sc: Yes, just like what `BpkButtonV2` does. Updated.

packages/bpk-component-skeleton/src/BpkSkeleton.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <BpkBaseSkeleton skeletonStyle={classNames} styleObj={styleObj} />
+  )
+};
+
+BpkSkeleton.defaultProps = {
+  size: SIZE_TYPES.default,
+  style: IMAGE_SKELETON_STYLE.default,
+}
```
olliecurtis, line 50: In TS we wouldn't declare default props like this but more up in the props definition on L31

felix-luo-sc: The reason why I don't deconstruct at line L31 is that it will cause ts error. (The style property only exist when type is 'image')
image

olliecurtis: Here what we can do is what you had before before

`const BpkSkeleton = ({type, size = SiZE_TYPES.default, style = IMAGE_SKELETON_STYLE.default})` 

Then you can remove the const {} = props declaration and just use the props as normal :) 

Then

packages/bpk-component-skeleton/src/BpkSkeleton.tsx
```diff
@@ -29,7 +29,7 @@ import STYLES from './BpkSkeleton.module.scss';
 const getClassName = cssModules(STYLES);
 
 const BpkSkeleton = (props: ComponentProps) => {
-  const { size, type } = props;
+  const { size = SIZE_TYPES.default, type } = props;
```
olliecurtis, line 32: Here you can do

```suggestion
const BpkSkeleton = ({
    type,
    size = SIZE_TYPES.default,
    style = IMAGE_SKELETON_STYLE.default
}: Props)
```

olliecurtis (approved):

PR #3339, opened 2024-04-08
LOOM-1292 BpkInfoBannerInner - Class 2 Overrides 

jronald01:
**Class 2 Overrides:**

1. BpkIcon - Styles added to a wrapper and Down Chevron icon added to avoid misalignment that comes from adding a wrapper which essentially takes the height from the parent and causes a misalignment in one direction when you apply `scaleY(-1)`.
2. BpkCloseButton - Duplicated styles removed
3. BpkLink - Style is redundant

- [ ] Component `README.md`
- [ ] Tests


jronald01:
After playing around and trying to fix the alignment, I’ve decided it’s simpler just to pull in the down chevron icon and make the icon conditional.

The problem was that once we introduced the new wrapper and flipped the icon `scaleY (-1)`, we are now flipping an area that is the same height as the `button` rather than the `svg`, hence one of the directions up or down would no longer be aligned due to flipping a different sized area, for example, at 4x zoom, we are now flipping a 39.06x56.15 rather than the original 39.06x46.39 which causes a misalignment.

packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
```diff
@@ -56,6 +58,7 @@ import STYLES from './BpkInfoBanner.module.scss';
 const getClassName = cssModules(STYLES);
 
 const ExpandIcon = withButtonAlignment(ChevronDownIcon);
+const ContractIcon = withButtonAlignment(ChevronUpIcon);
```
olliecurtis, line 61: It took me a while to understand what `Contract` meant here, maybe better and what we have used in other components would be `Collapse` so keeps it consistent across Backpack, so would be 

```suggestion
const CollapseIcon = withButtonAlignment(ChevronUpIcon);
```

packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
```diff
@@ -107,7 +107,9 @@ const ToggleButton = (props: ToggleButtonProps) => {
       aria-expanded={props.expanded}
       title={props.label}
     >
-      <ExpandIcon className={classNames} />
+      <div className={classNames}>
+        {props.expanded ? <ExpandIcon/> : <ContractIcon/>}
```
olliecurtis, line 111: With the above suggestion
```suggestion
        {props.expanded ? <ExpandIcon/> : <CollapseIcon/>}
```

olliecurtis (changes requested):

packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
jronald01: Ah, yeah, I was thinking expansion and contraction, but yeah, **collapse** makes way more sense 👍

jronald01:
Reversed chevron direction by a mistake when converting to a conditional, fixed it now:
`{props.expanded ? <CollapseIcon/> : <ExpandIcon/> }`

i.e. when it's expanded, we want to show the **Collapse Icon** and when it's not expanded, we want to show the **Expand Icon**.

olliecurtis (approved):

PR #3343, opened 2024-04-09
[LOOM-1341]: update eslint-config-skyscanner@18 & include forbid rule

mungodewar:
Tasks:
- add forbid-props rule and inline ignore `className` usage with TODO statement to be removed in future work

Changes:
- Refactored `commonButtonTests` into an actual test that imports different buttons, rather than exporting a test helper that was used in each button test suite. This was because we have a `*test*` regex and `commonButtonTests` wasn't matching this. It was an unusual pattern of testing and I thought it cleaner to remove it and have a single place for testing the common tests for Button instead. 

- [ ] Component `README.md`
- [ ] Tests


decisions/className.md
```diff
@@ -0,0 +1,96 @@
+### Avoid className props
```
mungodewar, line 1: Docs basically ripped from banana

metalix2 (approved):
Looks fine to me 👍

PR #3347, opened 2024-04-10
[LOOM-1341]: update & --fix to @skyscanner/eslint-config-skyscanner@17.4.0

mungodewar:
Mostly extracted from: https://github.com/Skyscanner/backpack/pull/3343

- Removal of backpack/use-tokens is because of: https://github.com/Skyscanner/eslint-plugin-backpack/pull/85
- Large number of changes related to https://github.com/Skyscanner/eslint-config-skyscanner/pull/623
- Calendar change required due to overriding CSS/conflict: see: https://percy.io/096a694a/backpack/builds/33616452?utm_campaign=096a694a&utm_content=backpack&utm_source=github_status_public

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,6 +123,7 @@
     "@percy/storybook": "^5.0.1",
     "@skyscanner/bpk-foundations-web": "^17.9.0",
     "@skyscanner/bpk-svgs": "^19.1.0",
+    "@skyscanner/eslint-config-skyscanner": "^17.4.0",
     "@skyscanner/stylelint-config-skyscanner": "^10.0.0",
     "@storybook/addon-a11y": "^7.6.17",
     "@storybook/addon-actions": "^7.6.17",
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "@skyscanner/bpk-foundations-web": "^17.9.0",
     "@skyscanner/bpk-svgs": "^19.1.0",
     "@skyscanner/eslint-config-skyscanner": "^17.4.0",
-    "@skyscanner/stylelint-config-skyscanner": "^10.0.0",
+    "@skyscanner/stylelint-config-skyscanner": "^11.1.0",
     "@storybook/addon-a11y": "^7.6.17",
     "@storybook/addon-actions": "^7.6.17",
     "@storybook/addon-docs": "7.6.17",
--- a/packages/bpk-component-calendar/src/BpkCalendarGrid.module.scss
+++ b/packages/bpk-component-calendar/src/BpkCalendarGrid.module.scss
@@ -20,7 +20,6 @@
 @use '../../unstable__bpk-mixins/utils';
 
 .bpk-calendar-grid {
-  width: 100%;
   border-collapse: separate;
   border-spacing: 0;
 
--- a/packages/bpk-component-table/src/BpkTableRow-test.js
+++ b/packages/bpk-component-table/src/BpkTableRow-test.js
@@ -28,7 +28,7 @@ describe('BpkTableRow', () => {
       <table>
         <tbody>
           <BpkTableRow>
-            <td />
+            <td aria-label="test" />
           </BpkTableRow>
         </tbody>
       </table>,
@@ -41,7 +41,7 @@ describe('BpkTableRow', () => {
       <table>
         <tbody>
           <BpkTableRow className="my-custom-class">
-            <td />
+            <td aria-label="test" />
           </BpkTableRow>
         </tbody>
       </table>,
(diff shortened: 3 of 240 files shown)


package.json
```diff
@@ -123,6 +123,7 @@
     "@percy/storybook": "^5.0.1",
     "@skyscanner/bpk-foundations-web": "^17.9.0",
     "@skyscanner/bpk-svgs": "^19.1.0",
+    "@skyscanner/eslint-config-skyscanner": "^17.4.0",
```
olliecurtis, line 126: Ah this is actually also pulled in via `stylelint-config-skyscanner` so it might be worth us bumping in there and then bumping the below dep up a version to avoid multiple versions of e-c-s being installed and conflicting

packages/bpk-component-calendar/src/BpkCalendarGrid.module.scss
```diff
@@ -20,7 +20,6 @@
 @use '../../unstable__bpk-mixins/utils';
 
 .bpk-calendar-grid {
-  width: 100%;
```
olliecurtis, line 23: Be careful with removing this and making sure to test this as this can cause alignment issues across browser and also with the `fixed-size` styles too, especially with the generation of the scrollable calendar and datepicker

packages/bpk-component-table/src/BpkTableRow-test.js
olliecurtis, line 1: How come these styles have changed for a linting update?

olliecurtis (changes requested):

packages/bpk-component-table/src/BpkTableRow-test.js
mungodewar: New linting rule that requires `td` elements to have `label` https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/control-has-associated-label.md

packages/bpk-component-calendar/src/BpkCalendarGrid.module.scss
mungodewar: Is there an exhaustive list that I can look at for manual testing? It seemed that `width: 100` previously was _always_ being overriden. This is applied to a `div` which also has `width: 100%` as default anyway, so I think it's safe, but I'll carry out any testing that you can think of to make sure this is safe. I understand it's a really complex component.

package.json
mungodewar: Yeah, good point. I think everything will still shake out to the same version due to the `^`. In `main` it's the same version, do you know why it's a dependency and not a dev-dep? https://github.com/Skyscanner/stylelint-config-skyscanner/blob/76c753b382d9a608a569f8efde0f75873197dab5/package.json#L40

mungodewar:
```
└─┬ @skyscanner/stylelint-config-skyscanner@10.2.0
  └── @skyscanner/eslint-config-skyscanner@16.3.0
```
Oh yeah, it's actually quite out of date.

mungodewar: Ah, found this reasoning: https://github.com/Skyscanner/stylelint-config-skyscanner/blob/76c753b382d9a608a569f8efde0f75873197dab5/docs/eslint-as-dependency.md?plain=1#L3

mungodewar: I've updated stylelint to latest, 

This results in this tree (no dupe)
```
├── @skyscanner/eslint-config-skyscanner@17.4.0
└─┬ @skyscanner/stylelint-config-skyscanner@11.1.0
  └── @skyscanner/eslint-config-skyscanner@17.4.0 deduped
```

packages/bpk-component-calendar/src/BpkCalendarGrid.module.scss
olliecurtis: Sadly no list but I know for the calendar yes this was always being overriden as I believe it sets the width to a fixed one with a class, but if you change it here the knockon affect for this to a consumer can be a broken calendar, so maybe with this component we need to do a test in a consumer before we merge this just for sanity sake as this is the monster component

package.json
mungodewar: I interpreted the doc as the reason that stylelint consumes eslint-config, not that we shouldn't install the eslint-config if we consume it. 

> ...allowing consumers to update to the latest of each and have a tree that is in sync.

This reads, to me, like we _should_ consume both. It's, IMO, not right to rely on a transitive dependency. Many (all the ones I've checked) of our consumers use both stylelint-config and eslint-config alongside one another.

olliecurtis (approved):

packages/bpk-component-calendar/src/BpkCalendarGrid.module.scss
mungodewar: Note: tested on product and posted results offline 👍

PR #3349, opened 2024-04-11
LOOM-1304 BpkNudger Class 2 Overrides

jronald01:
**Class 2 Overrides**:

1. Moved all `classname` overrides in `BpkNudger` into wrappers and re-jigged styling to compensate.

This ended-up being more involved as no matter what I did, I couldn't get the **zoomed in mobile view** to match the Percy snapshot.

So I moved the **title** and **subtitle** from within the `BpkLabel`, which is used for the person icon and moved them to the same level as the `BpkLabel` rather than being children.

This was required because when we move the **classname override to a wrapper**, we _lose_ the **flex** arrangement, as `BpkLabel` uses `display: block;` **_after_ the wrapper**, which breaks the arrangement:
```
@mixin bpk-label {
  display: block;
```

There are no doubt multiple ways to fix this, but the current version fixes for all visual tests, zoomed or not zoomed and mobile and desktop whereas the previous fixes I experimented with always had at least one snapshot not matching.

Due to the linting rules, I also had to remove the `classname` from the component and removed the now redundant test that checked for a `classname`. 

-- -- 

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-nudger/src/BpkNudger.tsx
```diff
@@ -59,24 +59,21 @@ const BpkNudger = ({
...
       )}
       <BpkConfigurableNudger
         inputClassName={getClassName('bpk-nudger__input--numeric')}
         compareValues={compareValues}
         incrementValue={incrementValue}
         decrementValue={decrementValue}
         formatValue={formatValue}
-        // TODO: className to be removed
-        // eslint-disable-next-line @skyscanner/rules/forbid-component-props
-        className={className}
```
olliecurtis, line 79: If we are removing this here should we also remove it from the API and also this PR would then become a major as we are removing functionality that folks could be using?

olliecurtis (changes requested):

packages/bpk-component-nudger/src/BpkNudger.tsx
mungodewar: Agreed. I don't think we can remove these "consumer pass-through" `className` cases. Tbh quite unsure on how to tackle these long term. As you say Ollie they can only be removed through a breaking change.

jronald01: Ah sorry, my bad, I got confused about this failure: https://github.com/Skyscanner/backpack/actions/runs/8649230361/job/23715083356#step:8:134
```
/home/runner/work/backpack/backpack/packages/bpk-component-nudger/src/BpkNudger.tsx
Error:   77:9  error  Prop className is forbidden. See ./decisions/className.md  @skyscanner/rules/forbid-component-props
```

jronald01: Do I just need to disable the linter for the line? @olliecurtis @mungodewar

jronald01: Or do I add it to `.eslintrc.js` in an `allowedFor` rules?

mungodewar: Let's leave it inline till we work out what is to be done with these cases.

olliecurtis (approved):

PR #3354, opened 2024-04-12
LOOM-1314 BpkSectionListItem Class 2 Overrides

jronald01:
### Class 2 Overrides:

1. Added `className` override to wrapper for both the `href` and `onClick`.
2. Needed to use `display: contents` on the wrapped style after adding the wrapper to retain the icon alignment.

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-section-list/src/BpkSectionListItem.module.scss
+++ b/packages/bpk-component-section-list/src/BpkSectionListItem.module.scss
@@ -40,6 +40,7 @@
   }
 
   &__chevron {
+    display: contents;
     fill: tokens.$bpk-text-secondary-day;
 
     #{$root}:hover & {
--- a/packages/bpk-component-section-list/src/BpkSectionListItem.js
+++ b/packages/bpk-component-section-list/src/BpkSectionListItem.js
@@ -63,11 +63,9 @@ const BpkSectionListItem = (props: Props) => {
         {...rest}
       >
         {children}
-        <BpkLargeChevronRightIconWithRtlSupport
-          // TODO: className to be removed
-          // eslint-disable-next-line @skyscanner/rules/forbid-component-props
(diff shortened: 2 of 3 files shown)


olliecurtis (approved):

packages/bpk-component-section-list/src/BpkSectionListItem.module.scss
```diff
@@ -40,6 +40,7 @@
   }
 
   &__chevron {
+    display: contents;
```
mungodewar, line 43: Screenshot 2024-04-12 at 15 46 29

I think previously the parent of the svg would have been that `button` & that has `display: flex` testing quickly locally I think that gives the same result. 

Tbh I'm not quite sure what `display: contents` does... never come across it before!

jronald01: Ah, here's a blurb on the use of `contents`:

> display: contents causes an element's children to appear as if they were direct children of the element's parent, ignoring the element itself. This can be useful when a wrapper element should be ignored when using CSS grid or similar layout techniques.

mungodewar: Interesting, so it's like it ignores the `span` for the wrapping. Nice. Seems appropriate!

jronald01: Bizarrely, when I tried `flex`, it wasn't aligned as per the snapshot on `main` (zoomed mobile):
Screenshot 2024-04-12 at 16 02 57

When I changed it to `contents` it matches the original snapshot on `main` (zoomed mobile):
Screenshot 2024-04-12 at 16 00 30

To me, `flex` is the right one to use, but it doesn't match the snapshot on `main`.

PR #3368, opened 2024-04-16
Luna 1326: Changes BpkSlider base library for mobile accessibility

LouiseReid:
Base library updated to [radux-ui/react-slider](https://www.radix-ui.com/primitives/docs/components/slider) which allows for screen reader swipe interaction on mWeb.

All props of previous `react-slider` library have been changed to ensure 99% backwards compatibility however `ariaLabel` is now a required prop

**The Problem we had - swiping does not move thumb**

https://github.com/Skyscanner/backpack/assets/28352715/bc461f79-5f5a-4cf8-8de6-2dfa8e10f9af

**With new base library - note this is using the dist build from this PR in banana**

https://github.com/Skyscanner/backpack/assets/28352715/5e46ba44-16bd-427f-bb1f-8d0277fffa80

**Breaking change**

- `ariaLabel` is now a required prop
- `className` prop removed

- [ ] Component `README.md`
- [x] Tests


examples/bpk-component-slider/stories.js
```diff
@@ -40,11 +40,13 @@ export default {
...
-              **Note:** When you're representing non-integer values (eg time, dates) please ensure you use \`ariaLabel\` and \`ariaValuetext\` to ensure that assistive technologies will be able to understand the value better.
-              `
-            }
+            {`**Note**: The aria props are the values that will be passed to the thumb of the slider. If your slider is for times for instance you would likely pass something like the following to ensure the value of the thumb is read out in a formatted state rather than just the value of where the thumb is on the track. If no \`ariaValuetext\` is passed to the component the screen reader will read just the value of the thumb
+
+            ariaLabels={['From', 'To']}
+    ariaValuetext={[getSliderTime(finalSliderStart), getSliderTime(finalSliderEnd)]}
+
+          `}
+
```
LouiseReid, line 49: This looks nice when rendered
Screenshot 2024-04-16 at 16 10 50

packages/bpk-component-slider/src/BpkSlider-test.tsx
```diff
@@ -30,35 +30,37 @@ window.ResizeObserver =
...
+  }
   it('should render correctly', () => {
-    const { asFragment } = render(<BpkSlider min={0} max={100} value={25} />);
-    expect(asFragment()).toMatchSnapshot();
-  });
-
-  it('should render correctly with a "className" attribute', () => {
-    const { asFragment } = render(
-      <BpkSlider min={0} max={100} value={25} className="my-slider" />,
-    );
```
LouiseReid, line 41: className prop removed

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @skyscanner/rules/forbid-component-props */
```
LouiseReid, line 1: This is needed here as a way to set styles on the external component

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -16,60 +17,74 @@
...
+  };
+
+  const handleOnChange = (sliderValues: number[]) => {
+    processSliderValues(sliderValues, onChange);
+  };
 
-  const isRange = (rest.value || rest.defaultValue || []).length > 1;
+  const handleOnAfterChange = (sliderValues: number[]) => {
+    processSliderValues(sliderValues, onAfterChange);
+  };
```
LouiseReid, line 58: This looks a bit wild but its to handle differences between the base libraries. This new library always returns the value in an array, so if there is just one thumb it is passed back like `[value]` where as for a range its the expected `[value, value]`

The previous library returned a one thumb slider value back as just a number rather than a number in an array

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -16,60 +17,74 @@
...
-    classNames.push(getClassName('bpk-slider--range'));
-  }
-  if (className) {
-    classNames.push(getClassName(className));
-  }
 
   return (
-    <Slider
+    <Slider.Root
+      className={getClassName('bpk-slider')}
```
Jennahowieson, line 63: Why is it that we remove className throughout, when we still need it here for the root styling?

packages/bpk-component-slider/src/accessibility-test.tsx
```diff
@@ -30,44 +30,52 @@ window.ResizeObserver =
...
-          min={0}
-          value={[20, 80]}
+          value={[0, 1290]}
+          ariaLabel={['from', 'to']}
+          ariaValuetext={['00:00', '21:30']}
         />
       </div>,
     );
+
+    debug()
```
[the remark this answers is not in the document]
LouiseReid: Woops, thanks 🫡

packages/bpk-component-slider/src/BpkSlider.tsx
LouiseReid: We still need to give the slider skyscanner the styles but removing the className prop from BpkSlider prevents the likes of Banana adding their own styles on top of that

Jennahowieson (approved):

HelloMaggieP (approved):

packages/bpk-component-slider/src/BpkSlider.tsx
LouiseReid: Yea, that is from the internal workings of the library. Somewhere in here, its state is an array

https://github.com/radix-ui/primitives/blob/main/packages/react/slider/src/Slider.tsx#L99

olliecurtis:
> > Thanks @LouiseReid for the PR this is interesting this issue came up as we recently completed an audit for this component based on the old library and there were no screenreader issues coming. So this PR comes as a shock as its not working when it was passed in audits
> 
> Its strange as the keyboard arrow keys for navigation work on desktop which is what I believe should also power the mobile screen reader swipe but they don't

Gotcha I see, now trying again on some different devices!

Could you also check the storybook for this? https://backpack.github.io/storybook-prs/3368/?path=/story/bpk-component-slider--visual-test

I just tried on Chrome on Android and its not working as expected, its the same behaviour of not slidiing but this time no highlighting of the slider controls

olliecurtis (changes requested):
Just adding a merge prevention here whilst we are figuring out some A11y weirdness with Android

LouiseReid:
@olliecurtis I've hit a bit of a wall with trying to get this working with Android talkback. I propose we merge as is at the next appropriate time given it enhances iOS and does nothing detrimental to android

packages/bpk-component-slider/src/BpkSlider.tsx
olliecurtis: Here we can add an exclusion to our linter so its centralised in one place, see here for examples: https://github.com/Skyscanner/backpack/blob/main/.eslintrc#L139

olliecurtis (commented):
One small comment see below. Also to check do we need to have a migration guide for folks to adopt the new version of the Slider to help manage this change?

olliecurtis:
Adding DON'T MERGE label so we don't merge until the point of next major and batch together

olliecurtis (approved):

PR #3369, opened 2024-04-16
ChipGroup - Add support for visual label inc. examples

steviehailey-skyscanner:
WIP - Addition of label

### DIFF
--- a/examples/bpk-component-chip-group/examples.tsx
+++ b/examples/bpk-component-chip-group/examples.tsx
@@ -189,6 +189,17 @@ export const OnImageChipGroup = () => {
   );
 };
 
+export const BpkChipGroupWithLabel = () => (
+     <div className={getClassName('bpk-chip-group-examples__with-label')}>
+       <BpkChipGroupState
+         type={CHIP_GROUP_TYPES.rail}
+         chips={chips}
+         chipStyle={CHIP_TYPES.default}
+         label='Filter'
+         ariaLabel="Select cities to filter by"
+       />
+     </div>
+   );
 
 export const AllChipTypesGroup = () => {
   const [dismissed, setDismissed] = useState(false);
@@ -286,6 +297,10 @@ export const MixedExample = () => (
       On Image
     </BpkText>
     <OnImageChipGroup />
+    <BpkText textStyle={TEXT_STYLES.heading3} tagName="h2">
+      With Label
+    </BpkText>
+    <BpkChipGroupWithLabel />
     <BpkText textStyle={TEXT_STYLES.heading3} tagName="h2">
       Wrapped
     </BpkText>
--- a/packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
+++ b/packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
@@ -30,6 +30,7 @@
   display: flex;
   margin: 0;
   padding: 0;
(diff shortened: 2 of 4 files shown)


packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
```diff
@@ -30,6 +30,7 @@
   display: flex;
   margin: 0;
   padding: 0;
+  align-items: baseline;
```
steviehailey-skyscanner, line 33: The label doesn't fill the height (like chips do) so needs alignment to be properly positioned

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -178,6 +181,7 @@ const BpkChipGroup = ({
           role={ariaMultiselectable ? 'group' : 'radiogroup'}
         >
           {ariaLabel && <legend className='visually-hidden'>{ariaLabel}</legend>}
+          {label && <BpkText textStyle={TEXT_STYLES.footnote} aria-hidden>{label}</BpkText>}
```
steviehailey-skyscanner, line 184: Conditionally show the label
- can be aria-hidden because we use the legend for the non-visual representation
- defaulting to `footnote` size here - we could make it overridable but erring on keeping simple given we have no use case for that yet

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
Iain530: Can you add the built css file too?

examples/bpk-component-chip-group/examples.tsx
```diff
@@ -189,6 +189,17 @@ export const OnImageChipGroup = () => {
...
+     <div className={getClassName('bpk-chip-group-examples__with-label')}>
+       <BpkChipGroupState
+         type={CHIP_GROUP_TYPES.rail}
+         chips={chips}
+         chipStyle={CHIP_TYPES.default}
+         label='Filter'
+         ariaLabel="Select cities to filter by"
+       />
+     </div>
+   );
```
Iain530, line 202: Is it worth having one with the sticky chip on also?

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
```diff
@@ -42,6 +42,10 @@
   &--wrap {
     flex-wrap: wrap;
   }
+
+  &--withLabel {
```
Iain530, line 46: I think this should be `--with-label` according to BEM naming ([docs](https://github.com/Skyscanner/backpack/blob/0d75ec32c3b0fafce594f898f1bcad0b7c4d73b0/CONTRIBUTING.md#css))

examples/bpk-component-chip-group/examples.tsx
steviehailey-skyscanner: They are meant to be mutually exclusive. First we could tackle making that so and then add an example as apt depending on the expected behaviour.

packages/bpk-component-chip-group/src/BpkChipGroup.module.scss
steviehailey-skyscanner: Added 👍

PR #3370, opened 2024-04-18
Split definitions for group types

steviehailey-skyscanner:
Splitting up definitions of the group types to reduce logic

### DIFF
--- a/packages/bpk-component-chip-group/src/BpkChipGroup.tsx
+++ b/packages/bpk-component-chip-group/src/BpkChipGroup.tsx
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import type { ReactElement, ReactNode} from 'react';
+import type { ReactNode } from 'react';
 import { useRef, useState } from 'react';
 
 import BpkBreakpoint, { BREAKPOINTS } from '../../bpk-component-breakpoint';
@@ -70,60 +70,31 @@ export type ChipItem = {
   hidden?: boolean;
 } & SingleSelectChipItem;
 
+export type InternalProps = {
+  chipGroupClassNames: string
+};
+
 export type CommonProps = {
   label?: string;
   ariaLabel?: string;
-  type?: ChipGroupType;
   chipStyle?: ChipStyleType;
-  // only required when type = rail
-  leadingNudgerLabel?: string;
-  trailingNudgerLabel?: string;
   chips: ChipItem[];
-  stickyChip?: ChipItem;
   ariaMultiselectable?: boolean;
 };
 
 export type RailChipGroupProps = {
-  type: "rail"
+  stickyChip?: ChipItem;
   leadingNudgerLabel?: string;
   trailingNudgerLabel?: string;
-};
+} & CommonProps;
 
 export type WrapChipGroupProps = {
-  type: "wrap"
-};
+} & CommonProps;
 
-export type ChipGroupProps = (RailChipGroupProps | WrapChipGroupProps) & CommonProps;
+export type ChipGroupProps = (RailChipGroupProps & { type: typeof CHIP_GROUP_TYPES.rail } | WrapChipGroupProps & { type: typeof CHIP_GROUP_TYPES.wrap });
 
-const BpkChipGroup = ({
-  ariaLabel,
-  ariaMultiselectable = true,
-  chipStyle = CHIP_TYPES.default,
-  chips,
-  label,
-  leadingNudgerLabel = '',
-  stickyChip,
-  trailingNudgerLabel = '',
-  type = CHIP_GROUP_TYPES.rail,
-}: ChipGroupProps) => {
-  const scrollContainerRef = useRef<HTMLElement | null>(null);
-
-  const containerClassnames = getClassName('bpk-chip-group-container')
-
-  const chipGroupClassNames = getClassName(
-    'bpk-chip-group',
-    `bpk-chip-group--${type}`,
-    label && 'bpk-chip-group--with-label'
-  );
-
-  const stickyChipContainerClassnames = getClassName(
-    'bpk-sticky-chip-container',
-    `bpk-sticky-chip-container--${chipStyle}`,
-  );
-
-  const scrollContainerIndicator = getClassName('bpk-chip-group-scroller-indicator');
-
-  const renderChipItem = ({
+const Chip = ({ ariaMultiselectable, chipIndex, chipItem, chipStyle }: { chipIndex: number, chipItem: ChipItem, chipStyle: ChipStyleType, ariaMultiselectable: boolean }, index: number) => {
+  const {
     accessibilityLabel,
     component = CHIP_COMPONENT.selectable,
     hidden = false,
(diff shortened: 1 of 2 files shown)


packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
   hidden?: boolean;
 } & SingleSelectChipItem;
 
+export type InternalProps = {
```
steviehailey-skyscanner, line 73: Props common to both components which are not exposed to external consumers

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
+  chipGroupClassNames: string
+};
+
 export type CommonProps = {
   label?: string;
   ariaLabel?: string;
-  type?: ChipGroupType;
```
steviehailey-skyscanner, line 76: This is now only used where we define the union of component types

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
-export type ChipGroupProps = {
-  chips: ChipItem[];
+export type RailChipGroupProps = {
   stickyChip?: ChipItem;
-  ariaMultiselectable?: boolean;
+  leadingNudgerLabel?: string;
+  trailingNudgerLabel?: string;
+} & CommonProps;
+
+export type WrapChipGroupProps = {
```
steviehailey-skyscanner, line 91: There isn't anything specific in here but open question whether we like having a marker type to describe the scenario

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
 } & CommonProps;
 
+export type ChipGroupProps = (RailChipGroupProps & { type: typeof CHIP_GROUP_TYPES.rail } | WrapChipGroupProps & { type: typeof CHIP_GROUP_TYPES.wrap });
```
steviehailey-skyscanner, line 94: We could move the type field into the respective sub-types but it isn't used by any direct consumers so eslint isn't happy with that - defining it with the union here covers what we need without exposing it to places that don't use it

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
 
-const BpkChipGroup = ({
+const Chip = ({ ariaMultiselectable, chipIndex, chipItem, chipStyle }: { chipIndex: number, chipItem: ChipItem, chipStyle: ChipStyleType, ariaMultiselectable: boolean }, index: number) => {
```
steviehailey-skyscanner, line 96: Extracted from the inline func to build a chip

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
+      role={ariaMultiselectable ? 'checkbox' : 'radio'}
+      leadingAccessoryView={leadingAccessoryView}
+      {...rest}
+    >
+      {text}
+    </Component>
+  );
+}
+
+const ChipGroupContent = (
```
steviehailey-skyscanner, line 128: Extracted from the common core legend component of each variety of chipgroup

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
+    className={chipGroupClassNames}
+    role={ariaMultiselectable ? 'group' : 'radiogroup'}
+  >
+    {ariaLabel && <legend className='visually-hidden'>{ariaLabel}</legend>}
+    {label && <BpkText textStyle={TEXT_STYLES.footnote} aria-hidden>{label}</BpkText>}
+    {chips.map((chip, index) => <Chip chipItem={chip} chipStyle={chipStyle} ariaMultiselectable={ariaMultiselectable} chipIndex={index} />)}
+  </fieldset>
+);
+
+const RailChipGroup = ({
```
steviehailey-skyscanner, line 153: Less conditionals now we know this is a rail group

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
+          label={label} />
+      </BpkMobileScrollContainer>
+      <BpkBreakpoint query={BREAKPOINTS.ABOVE_TABLET}>
+        <Nudger ariaLabel={trailingNudgerLabel} chipStyle={chipStyle} scrollContainerRef={scrollContainerRef} />
+      </BpkBreakpoint>
+    </>
   );
 };
 
+const WrapChipGroup = ({
```
steviehailey-skyscanner, line 211: Simpler logic now we know doesn't support rail logic

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
+  label,
+}: WrapChipGroupProps & InternalProps) =>
+  <ChipGroupContent ariaLabel={ariaLabel}
+    ariaMultiselectable={ariaMultiselectable}
+    chipGroupClassNames={chipGroupClassNames}
+    chipStyle={chipStyle}
+    chips={chips}
+    label={label} />;
+
+const BpkChipGroup = ({ label, type, ...rest }: ChipGroupProps) => {
```
steviehailey-skyscanner, line 226: External interface which now defers to the respective sub-types

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
Iain530: Reading this I'm expecting `typeof CHIP_GROUP_TYPES.rail` to be `string`, instead of `'rail'` which is what I think we want?

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
Iain530: Can we remove the `index` param now we have `chipIndex`?

Iain530 (commented):
Nice one 🚀 couple of Qs

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
steviehailey-skyscanner: No, thanks to the `as const` on the CHIP_GROUP_TYPES (line 39) the type is specifically `rail` instead of string 

Here is a preview of the evaluated type:
image

packages/bpk-component-chip-group/src/BpkChipGroupSingleSelect.tsx
```diff
@@ -18,14 +18,13 @@
...
-import BpkChipGroup, { type ChipItem, type SingleSelectChipItem, type CommonProps } from './BpkChipGroup';
+import BpkChipGroup, { type ChipItem, type SingleSelectChipItem, type ChipGroupProps } from './BpkChipGroup';
 
 export type SingleSelectProps = {
   chips: SingleSelectChipItem[];
   selectedIndex?: number;
   onItemClick?: (item: SingleSelectChipItem, selected: boolean, index: number) => void,
-} & CommonProps;
-
+} & ChipGroupProps;
```
steviehailey-skyscanner, line 27: Want to use the entire ChipGroupProps because consumer can choose `type` and then may need to provided specific args based on selection

Iain530: I think this might have cause the `chips` prop to now include the `ChipItem` as well as `SingleSelectChipItem`
image

packages/bpk-component-chip-group/src/BpkChipGroup.tsx
```diff
@@ -70,131 +70,172 @@ export type ChipItem = {
...
+  ariaMultiselectable?: boolean;
 };
 
-export type ChipGroupProps = {
-  chips: ChipItem[];
+export type RailChipGroupProps = {
   stickyChip?: ChipItem;
-  ariaMultiselectable?: boolean;
+  leadingNudgerLabel?: string;
+  trailingNudgerLabel?: string;
```
Iain530, line 88: These two labels should now be required since they were previously required for rail but not needed for wrap. Now this type is rail only they can be required and the default of `""` removed

PR #3386, opened 2024-04-25
LOOM-1307 BpkPhoneInput Class 2 Overrides

jronald01:
**Class 2 Overrides**:

1. `BpkInput` and `BpkSelect` classname overrides are duplicated and therefore redundant.  Can still see classnames in each component after the local override removed.
2. `<div>` classname override added as a parent wrapper to each `BpkLabel`.

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -149,23 +149,18 @@ const BpkPhoneInput = (props: Props) => {
...
+          </BpkLabel>
+        </div>
         {/* $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'. */}
         <BpkSelect
           {...commonProps}
           {...dialingCodeProps}
-          // TODO: className to be removed
-          // eslint-disable-next-line @skyscanner/rules/forbid-component-props
-          className={getClassName(dialingCodeProps.className)}
-          wrapperClassName={getClassName(dialingCodeProps.wrapperClassName)}
```
metalix2, line 168: So on closer inspection the `dialingCodeProps` set the className and wrapperClassName which means these ones here are overrides for the default props. But also we don't detect that were setting className here as it's being set from the spread from `dialingCodeProps` above. There may be more cases of this which might have been missed.

mungodewar: spread props should be caught by the _parent_ component so we are covered there (and if it's available on the public interface of the component there is nothing we can do either). 

"named" className props are potentially still problematic, however, there are far fewer of them and so I'm far less concerned.

jronald01: Ah, I reverted the `BpkInput` and `BpkSelect` changes and just left the `BpkLabel` ones in there as per the original Backpack audit doc.

Are we saying we are fine with the `BpkInput` and `BpkSelect` changes being included here?

mungodewar: Lets leave the PR as is, is there another ticket for the BpkInput? If not, let's create one to see if anything can be done about the remaining `className` there or we move it to allow.

jronald01: I'll create another ticket for it, cannot see anything currently that matches 👍

jronald01: New ticket created for `BpkInput`.

mungodewar (approved):

packages/bpk-component-phone-input/src/BpkPhoneInput.js
```diff
@@ -178,28 +177,27 @@ const BpkPhoneInput = (props: Props) => {
...
+            {...rest}
+            id={id}
+            name={name}
+            value={displayValue}
+            type={INPUT_TYPES.tel}
+            onChange={handleChange}
+            // TODO: className to be removed
+            // eslint-disable-next-line @skyscanner/rules/forbid-component-props
+            className={getClassName(className)}
+          />
```
metalix2, line 200: This seems to be indented by 1 too many.

jronald01: Oh yeah, thanks for spotting Matt, fixed now.

PR #3390, opened 2024-04-26
[ARGG-1063][BpkPopover]: Migrate to floating UI library

olliecurtis:
Migrating the underlying library of the BpkPopover to use the maintainers replacement library `floating-ui`, a much more lightweight library and has a lot of functionality built in that means we can remove our use of custom built react Portals!

This is one PR of a few to update components away from `popperjs` to `floating-ui` so can be used as a base!

Changes here are as follows:
- Removed PopoverPortal, as we no longer need to use our custom React Portal implementation as this is handled via `floating-ui`
- Removed the `renderTarget` prop as this is longer needed to apply where in the DOM the popover should be and is located by `floating-ui` relative to the `target`
- Added the ability to turn the arrow pointer off via the `showArrow` prop
- Changed the type of the `target` prop to only now expect a `ReactElement` this is required as we need to add the refs that `floating-ui` generate as part of the creation of the popover, which cannot be applied to any function elements or inputs for the target.
- Made `isOpen` optional, for if consumers want to have the popover open on first render, as the state and appearance in the DOM is now controlled through state with `floating-ui` instead so no longer relies on consumers to manage this state object
- Kept `onClose` property as consumers may wish to do other actions onClose or log events, but is not required to manage state via this now
- Converted component to Typescript and generated types
- Removed snapshot testing as this is unreliable and not best practice

- [x] Component `README.md`
- [x] Tests
- [x] Type declaration (`.d.ts`) files updated


packages/bpk-component-datepicker/src/BpkDatepicker.tsx
```diff
@@ -381,12 +380,10 @@ class BpkDatepicker extends Component<Props, State> {
             <BpkPopover
               id={`${id}-popover`}
               target={input}
-              renderTarget={renderTarget}
```
anambl, line 384: I guess this prop would have to be deprecated from `BpkDatepicker` once we migrate the Modal as well? 🤔

packages/bpk-component-popover/src/BpkPopover-test.tsx
```diff
@@ -16,99 +16,139 @@
...
         My popover content
       </BpkPopover>,
     );
-    expect(asFragment()).toMatchSnapshot();
+    expect(screen.getByText('My popover content')).toBeVisible();
   });
 
-  it('should render correctly with "padded" attribute equal to false', () => {
-    const { asFragment } = render(
+  // it('should render to the left', () => {
```
anambl, line 67: Should we delete this or uncomment it?

packages/bpk-component-datepicker/src/BpkDatepicker.tsx
```diff
@@ -381,12 +380,10 @@ class BpkDatepicker extends Component<Props, State> {
...
               onClose={this.props.onClose || this.onClose}
               isOpen={this.state.isOpen}
               label={title}
               closeButtonText={closeButtonText}
-              tabIndex={0}
```
anambl, line 389: How come we don't need this anymore?

packages/bpk-component-popover/src/BpkPopover.tsx
```diff
@@ -0,0 +1,240 @@
...
+    </div>
+  );
+  const classNames = getClassName('bpk-popover', className);
+  const bodyClassNames = getClassName(padded && 'bpk-popover__body--padded');
+
+  const labelId = `bpk-popover-label-${id}`;
+
+  return (
+    <>
+      {targetElement}
```
anambl, line 146: Why do _we_ need to render the targetElement rather than letting the consumer do it as it was before?

anambl: And related to this - I think it'd be good to create a short migration guide for consumers?

packages/bpk-component-popover/src/BpkPopover.tsx
```diff
@@ -0,0 +1,240 @@
...
+                {showArrow && (
+                  <FloatingArrow
+                    ref={arrowRef}
+                    context={context}
+                    id={ARROW_ID}
+                    // eslint-disable-next-line @skyscanner/rules/forbid-component-props
+                    className={getClassName('bpk-popover__arrow')}
+                    role="presentation"
+                    stroke={surfaceHighlightDay}
+                    strokeWidth={0.0625}
```
anambl, line 176: Hmm could be use a token here or declare it as a constant at the top?

packages/bpk-component-popover/src/BpkPopover-test.tsx
olliecurtis: Ah yeah this is one I need to action to make it work as it would be good to test it but it was giving weird results!

packages/bpk-component-datepicker/src/BpkDatepicker.tsx
olliecurtis: This is not used by the popover anymore as the floating-ui library automatically handles it :)

packages/bpk-component-popover/src/BpkPopover.tsx
olliecurtis: > think it'd be good to create a short migration guide for consumers?

Yep on the todo!

> Why do we need to render the targetElement rather than letting the consumer do it as it was before?

This is because we need to attach the generated refs from the floating UI library to it for all its magic and handling of it. So if we don't provide it there is no where for the refs to be provided and so the popover won't attach and operate.

Doing digging through our code folks are only ever passing the target in anyway instead of rendering separately so this is no change to what we were doing before and what was required of the consumer :) You can see this in the now deleted PopoverPortal here: https://github.com/Skyscanner/backpack/pull/3390/files#diff-03fea3ca881f4c6ce5e50658360c673e40526260b7fb538ffdd5ee36c51d3af9L197

packages/bpk-component-popover/src/BpkPopover.tsx
olliecurtis: I can move it to a const at the top. We can't use a token here sadly as Backpack Foundations only provide them with a string and include the word `rem` in them whereas floating UI expects a number

packages/bpk-component-datepicker/src/BpkDatepicker.tsx
olliecurtis: Good question, will take a look at this!

packages/bpk-component-popover/src/BpkPopover.tsx
anambl: Awesome, makes sense!

packages/bpk-component-datepicker/src/BpkDatepicker.tsx
olliecurtis: Yes you are right this prop will begin to go away when we pick up the modal ticket!

anambl:
LGTM - the only thing left is adding a migration guide for consumers seeing as we're releasing as major?

olliecurtis:
@anambl have added a migration guide here: https://github.com/Skyscanner/backpack/pull/3390/commits/24b413039eba0903e8346f995fb4149a6bdb5bd7

Let me know your thoughts :)

anambl:
> @anambl have added a migration guide here: [24b4130](https://github.com/Skyscanner/backpack/commit/24b413039eba0903e8346f995fb4149a6bdb5bd7)
> 
> Let me know your thoughts :)

Looks great!

Only thing is I've noticed in Storybook at the base of the arrow the line of the popover is still there 🤔 Is that expected from a design perspective? As I noticed that in Figma it's not the same

anambl (approved):
LGTM! 🎉

PR #3393, opened 2024-04-29
[IRN-5018][BpkNavigationBar][BpkBottomSheet] Add handling for long title text

steviehailey-skyscanner:
It was discovered that long titles overlap the buttons on the BpkBottomSheet component. The issue arises from the underlying BpkNavigationBar component lacking handling for this scenario.
Fix improves layout based on design requirement to clamp the text with ellipsis when there is not enough space to fit without encroaching on buttons or wrapping to an additional line.

# Screenshots

## Issue of long heading overlapping buttons
image

## Design guidance on expected layout
image

## Examples after fix
* Navigation Bar
* BottomSheet

| Tablet  | Large Mobile | Small Mobile |
| ------------- | ------------- |------------- |
|  |  |   |
|   |  |   |

## Other selected cases - Small Mobile
| Action Button  | No Title | Graphic Logo Header |
| ------------- | ------------- |------------- |
|  |  |   |

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/examples/bpk-component-bottom-sheet/examples.tsx
+++ b/examples/bpk-component-bottom-sheet/examples.tsx
@@ -184,6 +184,26 @@ const OverflowingExample = () => (
   </BottomSheetContainer>
 );
 
+const WrapHeaderTextExample = () => (
+  <BottomSheetContainer
+    title="Long Bottom Sheet title which wraps"
+    closeLabel="Close Bottom Sheet"
+  >
+    This is a default bottom sheet. You can put anything you want in here.
+  </BottomSheetContainer>
+);
+
+const WrapHeaderTextWithActionButtonExample = () => (
+  <BottomSheetContainer
+    title="Long Bottom Sheet title which wraps"
+    closeLabel="Close Bottom Sheet"
+    actionText="Action"
+    onAction={action('Action clicked')}
+  >
+    This is a default bottom sheet. You can put anything you want in here.
+  </BottomSheetContainer>
+);
+
 const NoHeaderExample = () => (
   <BottomSheetContainer
     closeLabel="Close Bottom Sheet"
@@ -219,8 +239,8 @@ const NestedExample = () => (
     wide
     closeOnEscPressed
     closeOnScrimClick
-    >
-      Outer Bottom Sheet
+  >
+    Outer Bottom Sheet
     <BottomSheetContainer
       title="Inner Bottom Sheet"
       closeLabel="Close Inner Bottom Sheet"
@@ -274,6 +294,8 @@ export {
   BackdropClickCloseExample,
   EscapeCloseExample,
   OverflowingExample,
+  WrapHeaderTextExample,
+  WrapHeaderTextWithActionButtonExample,
   NoHeaderExample,
   NoHeaderWithActionButtonExample,
   ActionButtonExample,
--- a/examples/bpk-component-bottom-sheet/examples.tsx
+++ b/examples/bpk-component-bottom-sheet/examples.tsx
@@ -184,18 +184,18 @@ const OverflowingExample = () => (
   </BottomSheetContainer>
 );
 
-const WrapHeaderTextExample = () => (
+const LongHeaderTextExample = () => (
   <BottomSheetContainer
-    title="Long Bottom Sheet title which wraps"
+    title="Bottom Sheet title which is long"
     closeLabel="Close Bottom Sheet"
   >
     This is a default bottom sheet. You can put anything you want in here.
   </BottomSheetContainer>
 );
 
-const WrapHeaderTextWithActionButtonExample = () => (
+const LongHeaderTextWithActionButtonExample = () => (
   <BottomSheetContainer
-    title="Long Bottom Sheet title which wraps"
+    title="Bottom Sheet title which is long"
     closeLabel="Close Bottom Sheet"
     actionText="Action"
     onAction={action('Action clicked')}
@@ -294,8 +294,8 @@ export {
   BackdropClickCloseExample,
   EscapeCloseExample,
   OverflowingExample,
-  WrapHeaderTextExample,
-  WrapHeaderTextWithActionButtonExample,
+  LongHeaderTextExample,
+  LongHeaderTextWithActionButtonExample,
   NoHeaderExample,
   NoHeaderWithActionButtonExample,
   ActionButtonExample,
--- a/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
+++ b/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
@@ -30,6 +30,7 @@
   animation-timing-function: ease-in-out;
 }
 
+// Q: can this be removed?
 $close-button-width: tokens.bpk-spacing-lg() * 2;
 
 .bpk-bottom-sheet {
@@ -96,6 +97,13 @@ $close-button-width: tokens.bpk-spacing-lg() * 2;
     @include breakpoints.bpk-breakpoint-above-mobile {
       animation: none;
     }
+  }
+
+  &--title {
+    text-align: center;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
   }
 
   &--content {
--- a/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
+++ b/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
@@ -30,9 +30,6 @@
   animation-timing-function: ease-in-out;
 }
 
-// Q: can this be removed?
-$close-button-width: tokens.bpk-spacing-lg() * 2;
-
 .bpk-bottom-sheet {
   z-index: tokens.$bpk-zindex-modal;
   width: 100%;
@@ -97,10 +94,6 @@ $close-button-width: tokens.bpk-spacing-lg() * 2;
     @include breakpoints.bpk-breakpoint-above-mobile {
       animation: none;
     }
-  }
-
-  &--action {
-    padding-inline-start: tokens.bpk-spacing-lg();
   }
 
   &--content {
--- a/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
+++ b/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
@@ -99,11 +99,8 @@ $close-button-width: tokens.bpk-spacing-lg() * 2;
     }
   }
 
-  &--title {
-    text-align: center;
-    text-overflow: ellipsis;
-    white-space: nowrap;
-    overflow: hidden;
+  &--action {
+    padding-inline-start: tokens.bpk-spacing-lg();
   }
 
   &--content {
--- a/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
+++ b/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
@@ -112,8 +112,13 @@
     position: sticky;
     top: 0;
     z-index: tokens.$bpk-zindex-tooltip - 1;
+    padding: tokens.bpk-spacing-sm() 0;
 
-    @include borders.bpk-border-bottom-sm(tokens.$bpk-line-day);
+    @include utils.bpk-themeable-property(
+      background-color,
+      --bpk-navigation-bar-background-color,
+      tokens.$bpk-surface-default-day
+    );
   }
 
   @keyframes slide-up {
--- a/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
+++ b/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
@@ -117,7 +117,7 @@ const BpkBottomSheet = ({
               id={headingId}
               title={title}
               titleTextStyle={TEXT_STYLES.label1}
-              titleTagName={title ? "h2" : undefined}
+              titleTagName={title ? "h2" : "span"}
               leadingButton={
                 <BpkCloseButton
                   label={closeLabel}
--- a/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
+++ b/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
@@ -24,7 +24,7 @@ import BpkCloseButton from '../../bpk-component-close-button';
 // @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
 import { BpkButtonLink } from '../../bpk-component-link';
 import BpkNavigationBar from "../../bpk-component-navigation-bar";
-import BpkText, { TEXT_STYLES } from "../../bpk-component-text/src/BpkText";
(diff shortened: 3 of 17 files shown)


examples/bpk-component-bottom-sheet/examples.tsx
```diff
@@ -178,6 +178,26 @@ const OverflowingExample = () => (
   </BottomSheetContainer>
 );
 
+const WrapHeaderTextExample = () => (
```
steviehailey-skyscanner, line 181: Demo the issue
image

examples/bpk-component-bottom-sheet/examples.tsx
```diff
@@ -178,6 +178,26 @@ const OverflowingExample = () => (
...
+  <BottomSheetContainer
+    title="Long Bottom Sheet title which wraps"
+    closeLabel="Close Bottom Sheet"
+  >
+    This is a default bottom sheet. You can put anything you want in here.
+  </BottomSheetContainer>
+);
+
+const WrapHeaderTextWithActionButtonExample = () => (
```
steviehailey-skyscanner, line 190: Demo issue with Action button
image

examples/bpk-component-bottom-sheet/examples.tsx
```diff
@@ -178,6 +178,26 @@ const OverflowingExample = () => (
   </BottomSheetContainer>
 );
 
+const LongHeaderTextExample = () => (
```
steviehailey-skyscanner, line 181: Demo of misbehaving header at Bottom Sheet level - now fixed

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -30,8 +30,6 @@
   animation-timing-function: ease-in-out;
 }
 
-$close-button-width: tokens.bpk-spacing-lg() * 2;
```
steviehailey-skyscanner, line 33: This was unused so removed

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -87,63 +87,63 @@ const BpkBottomSheet = ({
...
+        closeOnEscPressed={closeOnEscPressed}
+        closeOnScrimClick={closeOnScrimClick}
+        timeout={{ appear: animationTimeout, exit: isAboveMobile ? 0 : animationTimeout }}
       >
-      <>
-        <header className={getClassName('bpk-bottom-sheet--header')}>
-          <BpkNavigationBar
-            id={headingId}
-            title={title &&
-              <BpkText id={headingId} textStyle={TEXT_STYLES.label1} tagName="h2">{title}</BpkText>
```
steviehailey-skyscanner, line 119: We can now just pass the text and options for which element and style we want (which lets the component perform the text overflow for us)

packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
```diff
@@ -23,12 +23,19 @@
 
 .bpk-navigation-bar {
   position: relative;
-  display: flex;
+  display: grid;
```
steviehailey-skyscanner, line 26: Attempted to layout with flex after removing some absolute positioning on the buttons but the desired spacing and handing of buttons being optional for different scenarios goes beyond what we can easily do with Flexbox. Grid can easily achieve the different cases we need.

packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
```diff
@@ -23,12 +23,19 @@
...
   min-height: tokens.bpk-spacing-xl() + tokens.bpk-spacing-lg();
-  padding: tokens.bpk-spacing-base();
-  flex-direction: row;
-  justify-content: center;
+  padding: tokens.bpk-spacing-lg();
```
steviehailey-skyscanner, line 28: Margin adjustments based on new design requirements

packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
```diff
@@ -41,40 +48,45 @@
   }
 
   &__title {
+    grid-column: 2;
+    text-align: center;
+    text-overflow: ellipsis;
```
steviehailey-skyscanner, line 53: Clamp the text for small screen/long title scenarios

packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
```diff
@@ -41,40 +48,45 @@
...
+    white-space: nowrap;
+    overflow: hidden;
+
     @include utils.bpk-themeable-property(
       color,
       --bpk-navigation-bar-title-color,
       tokens.$bpk-text-primary-day
     );
 
+    // ensure we still apply ellipsis overflow when a custom title element is specified e.g. h2
```
steviehailey-skyscanner, line 63: BottomSheet is an example where we want an h2 element but it needs to be treated as inline to clamp the text

packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
```diff
@@ -41,40 +48,45 @@
...
     }
   }
 
+  &__title-container {
+    justify-self: center;
+  }
+
   &__leading-item,
   &__trailing-item {
-    position: absolute;
```
steviehailey-skyscanner, line 57: This positioning was causing overlapping text and buttons

packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
```diff
@@ -34,6 +36,8 @@ export type BarStyle = (typeof BAR_STYLES)[keyof typeof BAR_STYLES];
 export type Props = {
   id: string;
   title: ReactNode;
+  titleTextStyle?: TextStyle;
```
steviehailey-skyscanner, line 39: Make it easy for consumers to control the markup without needing to do their own title layout

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -114,8 +112,13 @@ $close-button-width: tokens.bpk-spacing-lg() * 2;
     position: sticky;
     top: 0;
     z-index: tokens.$bpk-zindex-tooltip - 1;
+    padding: tokens.bpk-spacing-sm() 0;
```
steviehailey-skyscanner, line 115: Adding additional padding to the heading based on BottomSheet design (avoiding adding to Nav Bar directly to avoid all other consumers gaining height)

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.module.scss
```diff
@@ -114,8 +112,13 @@ $close-button-width: tokens.bpk-spacing-lg() * 2;
...
 
-    @include borders.bpk-border-bottom-sm(tokens.$bpk-line-day);
+    @include utils.bpk-themeable-property(
```
steviehailey-skyscanner, line 117: Avoid seeing scrolling content "through" the padding around the header
image

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -87,63 +87,63 @@ const BpkBottomSheet = ({
...
-    </>
-    </BpkDialogWrapper>
-  }
+        <>
+          <header className={getClassName('bpk-bottom-sheet--header')}>
+            <BpkNavigationBar
+              id={headingId}
+              title={title}
+              titleTextStyle={TEXT_STYLES.label1}
+              titleTagName={title ? "h2" : undefined}
```
olliecurtis, line 120: Should we be passing in undefined or a better alternative value? Otherwise we could just do

```suggestion
              titleTagName={title && "h2"}
```

packages/bpk-component-bottom-sheet/src/__snapshots__/BpkBottomSheet-test.tsx.snap
```diff
@@ -52,7 +52,7 @@ exports[`BpkBottomSheet renders correctly with action props 1`] = `
               class="bpk-navigation-bar__title bpk-navigation-bar__title--default"
             >
               <span
-                class="bpk-text bpk-text--heading-5"
+                class="bpk-text bpk-text--label-1"
```
olliecurtis, line 55: Interesting these have now changed as this could lead me to believe these used to be wrong!

packages/bpk-component-modal/src/__snapshots__/BpkModalInner-test.tsx.snap
olliecurtis, line 1: As these have changed have we checked this is still ok on the BpkModal?

packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
```diff
@@ -50,6 +54,8 @@ const BpkNavigationBar = (props: Props) => {
     leadingButton,
     sticky = false,
     title,
+    titleTagName,
+    titleTextStyle,
```
olliecurtis, line 58: As you have made these optional, we will need to set these up as defaults otherwise it is going to cause issues when not provided. 

That way you don't need to have what you have below and can just pass the props through.

Currently though there isn't even a fallback for `titleTagName` so that might cause some havoc

I would say here it probably should be

```suggestion
    titleTagName = 'span',
    titleTextStyle = TEXT_STYLES.heading5,
```

packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
```diff
@@ -87,13 +93,14 @@ const BpkNavigationBar = (props: Props) => {
         )}>
           <BpkText
             id={titleId}
-            textStyle={TEXT_STYLES.heading5}
+            textStyle={titleTextStyle ?? TEXT_STYLES.heading5}
```
olliecurtis, line 96: With the above you can then just make this

```suggestion
            textStyle={titleTextStyle}
```

olliecurtis (changes requested):

packages/bpk-component-navigation-bar/src/BpkNavigationBarButtonLink.d.ts
```diff
@@ -0,0 +1,35 @@
+/*
```
steviehailey-skyscanner, line 1: Auto-generated missing typedefs...

packages/bpk-component-navigation-bar/index.d.ts
```diff
@@ -0,0 +1,41 @@
+/*
```
steviehailey-skyscanner, line 1: Auto-generated missing typedefs...

packages/bpk-component-navigation-bar/src/BpkNavigationBar.d.ts
```diff
@@ -0,0 +1,39 @@
+/*
```
steviehailey-skyscanner, line 1: Auto-generated missing typedefs...

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
steviehailey-skyscanner: Yeah, the title text can be an empty string, in which case we don't really want to output an empty `h2` in case causes screen-reader confusion when navigating by heading. So, `undefined` here is equivalent to not providing a customised tagName if we don't have a title (maintaining previous behaviour of an empty span being generated instead). If we `&&` we'll be asking for an "" element in that scenario which TS will object to.

packages/bpk-component-bottom-sheet/src/__snapshots__/BpkBottomSheet-test.tsx.snap
steviehailey-skyscanner: Indeed!

packages/bpk-component-modal/src/__snapshots__/BpkModalInner-test.tsx.snap
steviehailey-skyscanner: Yeah, the behaviour in terms of not clamping the text is consistent with before but it now respects the newer margins (and avoid overlapping the buttons) so long text will be slightly different than before

## New (example from: https://backpack.github.io/storybook-prs/3393/?path=/story/bpk-component-modal--long-title)
image

## Old (existing behaviour)
image

packages/bpk-component-bottom-sheet/src/__snapshots__/BpkBottomSheet-test.tsx.snap
steviehailey-skyscanner: Although since the span was empty it wouldn't have been a noticeable difference either way.

packages/bpk-component-modal/src/__snapshots__/BpkModalInner-test.tsx.snap
steviehailey-skyscanner: Note that this doesn't apply to the new Modal v2 which has it's own bespoke header
image

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
olliecurtis: I am a bit confused with the response, so apologies.

Looking at the code for the NavBar if the title is empty then it wouldn't even render anything other than a div so there would always be something rendered anyway.

I am just thinking here it would be better if we passed something down even if a span to avoid using undefined as even with undefined BpkText will have a fallback anyway but it just might make sense to have somehting here so that we don't have to drill into BpkText to find out what happens if there is undefined passed down to it

olliecurtis: https://github.com/Skyscanner/backpack/pull/3393/files#diff-8307bb32853b57180e184be5c1921bca36b55cb56431c7012848f328736f674cR89-R103

packages/bpk-component-modal/src/__snapshots__/BpkModalInner-test.tsx.snap
olliecurtis: We might wanna check with design if we are ok with the modal doing this in the case of these changes

Screenshot 2024-05-01 at 15 50 07

Taken from (accessory-modal)[https://backpack.github.io/storybook-prs/3393/?path=/story/bpk-component-modal--with-accessory-view]

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
steviehailey-skyscanner: Sure, happy to be explicit about a span be the appropriate fallback here 👍

packages/bpk-component-modal/src/__snapshots__/BpkModalInner-test.tsx.snap
steviehailey-skyscanner: Spoke to design and they think this scenario is unrealistic and the handling is reasonable (certainly better than the existing behaviour) so no need to adjust any further
image

olliecurtis (approved):

PR #3402, opened 2024-05-01
[ZOOT-1102][BpkDialogInner]: Adds optional aria-modal to the dialog

michmaml:
## Background:
It was discovered that it is possible to escape the modal with the screen reader. This can cause confusion if the content under the modal is unclear, and leaving the dialog may not be obvious, which can make travelers trapped behind an undismissable modal.

## Action
In order to address the issue, it was decided to set an extra property to the modal that would prevent travelers from escaping the modal. To avoid breaking any other pages that use the modal, the property will be set as optional.

## Changes
Fairly simple:
1. Added aria-modal={ariaModal} to the modal, which has the ariaModal value set to false by default.
2. Set the type of the variable as a boolean to prevent teams from setting the value to incorrect values.

https://github.com/Skyscanner/backpack/assets/55076036/a2e15fb8-038e-4ea7-8055-42362d7e1188

---

- [ ] Component `README.md`
- [x] Tests


packages/bpk-component-dialog/src/common-types.ts
```diff
@@ -26,11 +26,12 @@ export const HEADER_ICON_TYPES = {
 
 export type DialogInnerProps = {
   ariaLabel: string;
+  ariaModal: boolean;
```
olliecurtis, line 29: You will need to make sure you add this to the `common-types.d.ts` file so that it will come through in the TS types otherwise it will cause issues

packages/bpk-component-dialog/src/BpkDialogInner.tsx
```diff
@@ -30,6 +30,7 @@ const getClassName = cssModules(STYLES);
 const BpkDialogInner = (props: Props) => {
   const {
     ariaLabel,
+    ariaModal = true,
```
olliecurtis, line 33: You will need to make sure you add this to the `BpkDialogInner.d.ts` file so that it will come through in the TS types otherwise it will cause issues

olliecurtis (commented):
You will also need to mak

packages/bpk-component-dialog/src/BpkDialogInner.tsx
michmaml: Where exactly @olliecurtis? I am not sure where to add `ariaModal` to the aforementioned file 😕.

olliecurtis (approved):

PR #3408, opened 2024-05-02
[LOOM-1296][withOpenEvents]: Place `openable` styling within BpkInput

mungodewar:
Place `openable` styling within BpkInput.

- Using `data-` attribute field we can signal to BpkInput styling that it should be rendered with a `cursor: pointer`.

Why not use a regular prop? 

- I didn't want to add to the public api of BpkInput (it's really only valid with the `withOpenEvents` HOC)
- I didn't want to make a breaking change (to bring the logic of `withOpenEvents` inside BpkInput) and control that behaviour through props.

An alternative could be: 
- Bring most of the logic inside BpkInput
- Provide a `openable` prop to BpkINput
- `withOpenEvents` controls just the `openable` prop
- BpkInput then controls it's own `cursor` styling

The above seems like a much larger change, so thought before I'd embark on that to get some feedback on the above first. 

- [ ] Component `README.md`
- [ ] Tests


packages/bpk-component-input/src/withOpenEvents.tsx
```diff
@@ -24,11 +24,7 @@ import type {
   ComponentProps,
 } from 'react';
 
-import { cssModules, wrapDisplayName } from '../../bpk-react-utils';
-
-import STYLES from './BpkInput.module.scss';
-
-const getClassName = cssModules(STYLES);
+import { wrapDisplayName } from '../../bpk-react-utils';
```
metalix2, line 27: What's this being used for?

metalix2:
I think it's going to be a challenge for HOC's where they're purpose is to wrap elements, which can then change anything about them. I'm happy with this solution. The `data-` attribute is an interesting solution. And it does avoid changing the API.

packages/bpk-component-input/src/withOpenEvents.tsx
mungodewar: The `wrapDisplayName` basically will render something like: `withOpenEvents(MyCustomComponent)` when viewed in the react debug tools. 

It's used when defining the `displayName`.Without it we'd loose the knowledge that the component had been wrapped when debugging. 

Screenshot 2024-05-02 at 10 54 14

metalix2: Nice I see! That's very cool!

metalix2 (approved):
Ship it!

olliecurtis (approved):

PR #3420, opened 2024-05-03
[NoJIra][BpkModal] Update modal stories to tsx and add visual tests

metalix2:

- [ ] Component `README.md`
- [ ] Tests


examples/bpk-component-modal/examples.tsx
```diff
@@ -18,9 +18,9 @@
 
 /* @flow strict */
```
mungodewar, line 19: flow comment no longer required

examples/bpk-component-modal/examples.tsx
```diff
@@ -108,147 +111,113 @@ const content = [
...
-            {children}
-          </BpkModal>
-        </div>
+  return (
+    <div id="modal-container" {...wrapperProps}>
+      <div id="pagewrap">
+        <BpkButton onClick={() => setIsOpen(true)}>
+          {buttonLabel || 'Open modal'}
+        </BpkButton>
+        {/* $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'. */}
```
mungodewar, line 133: flow comment not required.

examples/bpk-component-modal/stories.tsx
```diff
@@ -59,3 +59,24 @@ export const NoPadding = NoPaddingExample;
 
 export const WithAccessoryView = WithAccessoryViewExample;
 export const Contrast = ContrastExample;
+export const VisualTestDefault = () => Default(true);
+export const VisualTestDefaultWithZoom = {
+  render: () => Default(true),
```
mungodewar, line 64: I think we can do this: 
```suggestion
  render: () => <><Default isOpen/><Contrast isOpen/><WithAccessoryView isOpen/>
```

Slims down the amount of snaps we need. 

You'll have to turn the examples to read props: 

```
const NoPaddingExample = ({isOpen}) => (
  <ModalContainer title="Modal title" closeLabel="Close modal" padded={false} isOpen={isOpen}>
    This is a default modal. You can put anything you want in here.
  </ModalContainer>
);
```

metalix2: I was curious how this might appear on mobile view perhaps they are just one after the other and not stacked?

examples/bpk-component-modal/stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+export const Contrast = ContrastExample;
+
+// Due to how iframes work we can pass a local url to load the stories above.
+// Attempted to use a Custom Iframe component with a react portal and ref to
+// render components but it didn't have the desired effect.
+const visualWrapper = (id: string, zoomEnabled: boolean = false) => (
+  <div style={{ height: '640px', width: '100%' }}>
+    <iframe
+      title={`Embedded Storybook ${id}`}
+      src={`/iframe.html?id=${id}&viewMode=story&args=zoomEnabled:${zoomEnabled}`}
```
olliecurtis, line 69: For deployed storybook we will have to have a think how we can make this work. As locally its fine because the address is

`localhost:9001/${src}` but on the deployed version the path is `backpack.github.io/storybook` so by doing `/iframe` it looks for just `backpack.github.io` which produces the following:

Screenshot 2024-05-06 at 20 15 54

olliecurtis: Unless we don't mind for this test because its purely for visual tests? If so we should add a banner or text above the first iFrame to make it clear?

metalix2: Hmm that is is a good point. 
The other option was to try use the react portal again in the iframe.. I suppose we might not care for visual test scenarios in the deployed storybook? 🤔

metalix2: I don't think we can manage to load the modal in the iframe with a portal. 🤔 Is there a flag that could be passed in that adds the subpath to storybook so that the url can generate with/without or we just update the tests to match the release version?

olliecurtis (approved):

PR #3427, opened 2024-05-08
[IRN-5039] BottomSheet supports aria-label

steviehailey-skyscanner:
# Description

Previously BottomSheet consumers had to provide an `aria-labelledBy` to be attached to the `dialog` element but it is rare that they know the relevant title element id to provide.

Now BottomSheet consumers can provide either:
a. `aria-labelledBy` - if they are not providing a string title and want to reference an element within the content as a label
b. `aria-label` - if they would like to provide an a11y string (e.g. same value as title or a more accessible equivalent)  

## Screenshot - BottomSheet consumer can now control aria-label on dialog element
image

- [ ] Component `README.md`
- [ ] Tests


examples/bpk-component-bottom-sheet/examples.tsx
```diff
@@ -146,7 +146,7 @@ BottomSheetContainerProps,
             Open bottom sheet
           </BpkButtonV2>
           <BpkBottomSheet
-            ariaLabelledby='test'
```
steviehailey-skyscanner, line 149: Make the default demo more of a valid example - there was no `#test` element. Now screen reader reads the a11y title.

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -31,9 +31,8 @@ import STYLES from './BpkBottomSheet.module.scss';
 
 const getClassName = cssModules(STYLES);
 
-export type Props = {
+interface CommonProps {
```
steviehailey-skyscanner, line 34: Extract out the non-a11y props so we can conditionally allow either label or labelledBy...

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -92,7 +93,7 @@ const BpkBottomSheet = ({
   return <BpkBreakpoint query={BREAKPOINTS.ABOVE_MOBILE}>
     {(isAboveMobile: boolean) =>
       <BpkDialogWrapper
-        ariaLabelledby={ariaLabelledby}
+        {...props}
```
steviehailey-skyscanner, line 96: Pass down the a11y props (we don't know which one we'll have but TS assures we'll have one)

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
```diff
@@ -40,16 +39,18 @@ export interface Props {
...
   transitionClassNames?: {
     appear?: string,
     appearActive?: string,
     exit?: string
   };
-  timeout?: {appear?: number, exit?: number};
+  timeout?: { appear?: number, exit?: number };
 };
 
+export type Props = CommonProps & ({ ariaLabelledby: string } | { ariaLabel: string; });;
```
steviehailey-skyscanner, line 52: Same pattern as above

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
```diff
@@ -135,6 +135,11 @@ export const BpkDialogWrapper = (props: Props) => {
     };
   }, [id, isOpen, onClose, closeOnEscPressed, closeOnScrimClick]);
 
+  const ariaProps = {
```
steviehailey-skyscanner, line 138: Extract whichever aria prop has been provided to add the the dialog element

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -51,20 +50,22 @@ export type Props = {
...
-  id,
-  isOpen,
-  onAction = () => null,
-  onClose,
-  title = '',
-  wide = false
-}: Props) => {
+export type Props = CommonProps & ({ ariaLabelledby: string } | { ariaLabel: string; });
+
+const BpkBottomSheet = (props: Props) => {
```
anambl, line 55: why not keep it as it was previously and destructure it here? 🤔

steviehailey-skyscanner: This is a limitation of the mutually exclusive aria props unfortunately. TS won't allow destructuring of those fields because it knows one will not be there so by keeping the `props` intact we can safely reference or pass them on (subject to performing checks for existence to keep TS happy).

In this component we just pass whichever we have by spreading the `props`

In BpkDialog wrapper we have an example of checking and using a specific one here:  https://github.com/Skyscanner/backpack/pull/3427/files#diff-8ae3b880e3772a84f85517514640da0930f295cffbdfd531fcf063a29ba84606R139

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper-test.tsx
```diff
@@ -100,6 +101,30 @@ describe('BpkDialogWrapper', () => {
...
+      const { container } = render(
+        <BpkDialogWrapper
+          {...props}
+          ariaLabelledby='my-title-id'
+        >
+          Dialog content
+        </BpkDialogWrapper>
+      );
+
+      expect(container.querySelector('dialog[aria-labelledby="my-title-id"]')).toBeInTheDocument();
```
steviehailey-skyscanner, line 114: Check aria correctly applied

packages/bpk-react-utils/src/BpkDialogWrapper/accessibility-test.tsx
```diff
@@ -22,18 +22,35 @@ import { axe } from 'jest-axe';
...
-        timeout={{appear: 0, exit: 0}}
+      >
+        Dialog content
+      </BpkDialogWrapper>,
+    );
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+
+  it('should not have programmatically-detectable accessibility issues using airaLabel', async () => {
```
steviehailey-skyscanner, line 49: Additional axe check for alternative aria

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
anambl: I see, but we could also destructure the props here and spread the aria props rather than explicitly destructure them? Something like
```
const BpkBottomSheet = ({actionText = '', children, etc, ...ariaProps}) => {...}
```
?
In which case we'd also pass whichever we have

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
anambl, line 52: I think there's an extra `;`?

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
```diff
@@ -177,6 +182,6 @@ export const BpkDialogWrapper = (props: Props) => {
           </div>
         </dialog>
       </CSSTransition>
-    </div>
+    </div >
```
anambl, line 185: Is this an automatic linting change?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -51,20 +50,22 @@ export type Props = {
...
-  closeOnEscPressed = false,
-  closeOnScrimClick = false,
-  id,
-  isOpen,
-  onAction = () => null,
-  onClose,
-  title = '',
-  wide = false
-}: Props) => {
+export type Props = CommonProps & ({ ariaLabelledby: string } | { ariaLabel: string; });
```
anambl, line 53: Interesting, I thought interfaces could only be extended and `&` wouldn't work on them

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
steviehailey-skyscanner: Ah, yeah we could use that syntax to keep the destructure - will push an update 🚀

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
steviehailey-skyscanner: Yeah there is some subtlety there, you can mention interfaces when using intersection to define a `type` (you just can define an `interface` using intersection).
- https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx
steviehailey-skyscanner: Not sure what happened there - the linter is happy with or without the space so will remove 😆

packages/bpk-component-bottom-sheet/src/BpkBottomSheet-test.tsx
```diff
@@ -85,17 +91,39 @@ describe('BpkBottomSheet', () => {
...
+      </BpkBottomSheet>
+    );
+
+    expect(container.querySelector('dialog[aria-labelledby="my-bottomsheet-title-id"]')).toBeInTheDocument();
+  });
+  it('renders correctly with ariaLabel prop', () => {
+    const { container } = render(
+      <BpkBottomSheet
+        {...props}
+        ariaLabel='my a11y title'
```
anambl, line 120: I'm guessing this will take precedence since it's added after `ariaLabelledBy`? 🤔

packages/bpk-component-bottom-sheet/src/BpkBottomSheet-test.tsx
```diff
@@ -85,17 +91,39 @@ describe('BpkBottomSheet', () => {
...
         Bottom Sheet content
       </BpkBottomSheet>
     );
     expect(asFragment()).toMatchSnapshot();
   });
+  it('renders correctly with ariaLabelledBy prop', () => {
+    const { container } = render(
+      <BpkBottomSheet
+        {...props}
+        ariaLabelledby='my-bottomsheet-title-id'
```
anambl, line 108: it's already part of `props` so I don't think we need it here?

packages/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper-test.tsx
```diff
@@ -100,6 +101,30 @@ describe('BpkDialogWrapper', () => {
       );
       expect(asFragment()).toMatchSnapshot();
     });
+    it('renders correctly with ariaLabelledBy prop', () => {
+      const { container } = render(
+        <BpkDialogWrapper
+          {...props}
+          ariaLabelledby='my-title-id'
```
anambl, line 108: Same here, it's already declared in `props`

anambl:
It all looks great to me!
What happens if you don't pass neither or if you pass both aria props? For the first, I assume a TS error, but I don't think it's possible to unit test that 😅 And for the 2nd, the latter takes precedence? 🤔

packages/bpk-component-bottom-sheet/src/BpkBottomSheet-test.tsx
steviehailey-skyscanner: Yeah I we could avoid it but I thought it was nice to be explicit when we are asserting based on that property?

packages/bpk-component-bottom-sheet/src/BpkBottomSheet-test.tsx
steviehailey-skyscanner: I have done some testing in Safari with VoiceOver and if you give it a valid `ariaLabelledBy` then it will use that. If the labelledBy was missing or invalid then it falls back to the `aria-label` instead.

steviehailey-skyscanner:
> It all looks great to me! What happens if you don't pass neither or if you pass both aria props? For the first, I assume a TS error, but I don't think it's possible to unit test that 😅 And for the 2nd, the latter takes precedence? 🤔

Indeed TS will prevent the former but if someone ignored the TS error then things function but you get a degraded screenreader experience because it often describes the "Close button" before describing what the dialog is for. For the later case, it seems that `labelledBy` is used in favour of `label` if someone did provide both which seems reasonable.

anambl (approved):

PR #3463, opened 2024-05-20
[NOJIRA][BpkBreakpoint]: Bug fix initial incorrect value of isClient on client rendered applications

mungodewar:
Initially incorrect value (`false`) was used when using BpkBreakpoint on a client rendered application. This was due to the forced re-render functionality also doubling as client environment detection. In this PR they have been separated and thus the issue is removed.

`useEffect and useState` is used to solve forced re-rendering and `typeof window` statement is used to detect the initial `isClient` value. 

Change(s):
- use typeof check rather than boolean to determine if environment is client. Continue to support forced re-renders

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-breakpoint/src/BpkBreakpoint.tsx
+++ b/packages/bpk-component-breakpoint/src/BpkBreakpoint.tsx
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import type { ReactElement, ReactNode } from 'react';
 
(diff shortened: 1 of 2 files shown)


packages/bpk-component-breakpoint/src/BpkBreakpoint.tsx
```diff
@@ -45,6 +45,18 @@ type Props = {
   matchSSR?: boolean;
 };
 
+const useLegacyWarning = (query: string, legacy: boolean, isClient: boolean) =>
+  useMemo(() => {
```
mungodewar, line 49: Ensures that the `console.log` only occurs once per query/instance to not duplicate the console output (as we are forcing a re-render)

olliecurtis (approved):

PR #3467, opened 2024-05-22
[ARGG-1143]: Add changed files check for running Percy

olliecurtis:
This PR updates the Percy running logic for CI to only run if core component files have changed that can have an impact on visual tests.

It looks to see if any files have changed in: 

- `examples` folder
- `packages` folder on the following files
   - All non test `.js`, `.ts` and `.tsx` files
   - All `.scss` files 

If these files change then an output variable will be recorded to say that tests are required to pass which is then used by the `Percy` job to determin to run.

If no files matching the criteria change, then the output variable will be recorded to say `Percy` job can be skipped and a separate status event is sent to the PR to inform that the steps can be skipped

Screenshot 2024-05-22 at 10 37 49

For more examples of the testing you can see under this test repo

Example of Percy running: https://github.com/olliecurtis/backpack/pull/1
Example of Percy skip: https://github.com/olliecurtis/backpack/pull/2


.github/workflows/_build.yml
```diff
@@ -103,8 +142,9 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   PercyTests:
+    if: (needs.CheckVisualTests.outputs.requires-visual == 'true')
```
metalix2, line 145: This will run all percy snapshots if any components change correct?

metalix2 (approved):
Looks good 👍

.github/workflows/_build.yml
olliecurtis: Yes, as there are going to be some components that use other components, so we might need to test those too without crazy custom logic to check which are connected or not.

This is how we have done this in other projects in Skyscanner too, so its more just don't run if we change non component code!

PR #3490, opened 2024-06-07
[NO-JIRA]: Correct logic for when to skip Percy

olliecurtis:
Previously without the checking logic when PRs were merged to main it would attempt to skip Percy tests but fail because there is no `pr_sha` to be able to update the reference for.

However we always want to run visual tests on main and not skip them at all, so we need to add the logic for only running on PRs and when no visual affecting files have been changed.


metalix2 (approved):
Thanks @olliecurtis

arnaugm (approved):

PR #3495, opened 2024-06-12
[NOJIRA][BpkPopover] Fix Popover issue that floating-ui's useClick does not support stopPropagation

yifeng-ruan:
After upgrading the Backpack version, the functionality that allowed the popover to support triggering the target's events stopped working. 

The issue seems to be that floating-ui's useClick does not support stopPropagation, which causes the target's click event to fail when stopPropagation is needed. 

For more details: https://github.com/floating-ui/floating-ui/issues/1800

- [ ] Component `README.md`
- [x] Tests

### DIFF
--- a/packages/bpk-component-popover/src/BpkPopover-test.tsx
+++ b/packages/bpk-component-popover/src/BpkPopover-test.tsx
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { render, screen, fireEvent } from '@testing-library/react';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
 import BpkPopover from './BpkPopover';
@@ -46,7 +46,7 @@ describe('BpkPopover', () => {
     expect(screen.getByText('My popover content')).toBeVisible();
   });
 
-  it('should rerender correctly with "isOpen" provided', () => {
+  it('should rerender correctly with "isOpen" provided', async () => {
     const props = {
       id: 'my-popover',
       onClose: () => null,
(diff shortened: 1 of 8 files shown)


packages/bpk-component-popover/src/BpkPopover.tsx
```diff
@@ -154,9 +154,20 @@ const BpkPopover = ({
...
+      ...getReferenceProps(
+        {
+          onClick: (event) => {
+            if (targetProps.onClick) {
+              event.stopPropagation();
+              targetProps.onClick(event);
+            }
+          },
+        }
+      ),
```
olliecurtis, line 170: Can we make sure we add this to line 175 too?

packages/bpk-component-popover/src/BpkPopover.tsx
```diff
@@ -154,9 +154,20 @@ const BpkPopover = ({
     dismiss,
   ]);
 
+  const targetProps = target.props;
```
olliecurtis, line 157: I'd say I wouldn't extract all the props to a variable here as it will mean we aren't spreading any of the others anywhere else, so just either set this to be

`const targetClick = target.props.onClick` or on line 164 just simply call it as `if (target.props.onClick) {`

olliecurtis (changes requested):

packages/bpk-component-popover/src/BpkPopover.tsx
yifeng-ruan: Thank you very much for your suggestion. 
The target is defined as `ReactElement<any>`, and with the any type, directly using `target.props.onClick` would result in a TypeScript type error. 

However, defining `const targetClick = target.props.onClick` works fine. 

The code has been updated. Thanks.

olliecurtis (approved):

packages/bpk-component-popover/src/BpkPopover-test.tsx
```diff
@@ -191,4 +191,41 @@ describe('BpkPopover', () => {
...
+    const event = new MouseEvent('click', { bubbles: true });
+    jest.spyOn(event, 'stopPropagation');
+
+    act(() => {
+      fireEvent(button, event);
+    });
+
+    expect(handleClick).toHaveBeenCalled();
+    expect(event.stopPropagation).toHaveBeenCalled();
+    expect(handleClick.mock.calls[0][0].afterStopPropagation).toBe(true);
```
olliecurtis, line 228: To avoid the error, change this to similar as above to be

```suggestion
    await waitFor(async () => {
      await fireEvent(button, event);
      expect(handleClick).toHaveBeenCalled();
      expect(event.stopPropagation).toHaveBeenCalled();
      expect(handleClick.mock.calls[0][0].afterStopPropagation).toBe(true);
    });
```

olliecurtis (approved):

PR #3497, opened 2024-06-14
[NO JIRA][BpkPopover]: Update popover to attach to body outside of target elements

olliecurtis:
Wraps the popover in `FloatingPortal` this is part of the floating ui library that allows you to determine when in the DOM you render the popover to be displayed.

Previously the popover would have been attached to the node directly adjacent to the target that triggers it, this creates an issue whereby when rendered in a small window space would cause the popover to render inside/behind content even with a `z-index` applied.

Screenshot 2024-06-14 at 15 22 58

With adding this new portal to the popover it will now attach the popover to the main `body` in HTML so means it will always display over content regardless of where its contained, fixing the above issue

Screenshot 2024-06-14 at 15 23 59

- [ ] Component `README.md`
- [ ] Tests

### DIFF
--- a/packages/bpk-component-popover/src/BpkPopover-test.tsx
+++ b/packages/bpk-component-popover/src/BpkPopover-test.tsx
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { render, screen, fireEvent, waitFor, act } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


mungodewar (approved):
LGTM

PR #3508, opened 2024-06-21
[ARGG-1158][BpkDialog]: Migrate Dialog to not use Portal

olliecurtis:
Migrates the BpkDialog to not use the custom portal implementation from Backpack.

All functionality has been retained, the only main change here is remove the use of BpkPortal and replace the state management to be local to the component.

Alongside this I have added a visual test to help us keep track of the component changes

- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests


examples/bpk-component-dialog/examples.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const onClose = () => {
+    setIsOpen(false);
+  };
+
+  return (
+    <div id="dialog-container">
+      <div id="pagewrap">
+        <BpkButton onClick={onOpen}>Open dialog</BpkButton>
+      </div>
+      {/* $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'. */}
```
anambl, line 67: We won't need this anymore, right?

packages/bpk-component-dialog/src/BpkDialog.tsx
```diff
@@ -53,31 +55,44 @@ const BpkDialog = ({
...
-          </span>
-        )}
-        {children}
-      </BpkDialogInner>
-    </Portal>
+    <>
+      {isOpen && (
+        <BpkDialogInner
+          onClose={onClose}
+          closeOnScrimClick={dismissible}
```
anambl, line 79: I'm a little confused because actually I can't find this prop in `BpkDialogInner` 🤔 I see we were already passing it, but I wonder if it actually has an effect?

packages/bpk-component-dialog/src/BpkDialog.tsx
```diff
@@ -53,31 +55,44 @@ const BpkDialog = ({
...
+      window.addEventListener('keydown', handleKeyDown);
+    }
+
+    return () => {
+      window.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [isOpen, onClose, dismissible]);
+
   return (
-    <Portal
```
anambl, line 57: Have we validated that indeed there is no use case/ we do not need to render the dialog in a different part of the DOM tree?
For example, one use case for which we needed portals in the past was to place an element outside the main `pagewrap` element which all microsites render and to which our scrim component applies `aria-hidden` property so it results in the component being hidden from AT.

anambl:
I am very confused because I can see in Storybook the `scrim` component being rendered in the tree, however, I cannot tell where it's coming from 😮

examples/bpk-component-dialog/examples.tsx
olliecurtis: Good catch!

packages/bpk-component-dialog/src/BpkDialog.tsx
olliecurtis: This goes through to the `withScrim` wrapper that is around the Inner, this was part of the original API and not been touched so would remain as it was :) 

https://github.com/Skyscanner/backpack/blob/main/packages/bpk-scrim-utils/src/withScrim.tsx#L66

All that has happened here is the `isOpen` wrapper :)

olliecurtis:
> I am very confused because I can see in Storybook the `scrim` component being rendered in the tree, however, I cannot tell where it's coming from 😮

This comes from the `withScrim` HOC that we have in the scrim component: https://github.com/Skyscanner/backpack/blob/main/packages/bpk-scrim-utils/src/withScrim.tsx

packages/bpk-component-dialog/src/BpkDialog.tsx
olliecurtis: We can setup a test or storybook example to test this! This works similar to the new modal v2, but you might be right now there is no `renderTarget` but I think the `withScrim` `getApplicationElement` property might be the main handler here, but you are right will test this!

olliecurtis: I did a check here and the property that does handle this with the `pagewrap` is the `getApplicationElement` property as I suspected so this will still work for consumers as this is what controlled where it goes and where the `aria-hidden` is applied to

olliecurtis: To confirm the line foudn it here whilst testing an example :) https://github.com/Skyscanner/backpack/pull/3508/files#diff-268f629c295e29c7e8875462a2fd7ef18a97eb0058f784296bddae08e1375635R74

anambl:
> > I am very confused because I can see in Storybook the `scrim` component being rendered in the tree, however, I cannot tell where it's coming from 😮
> 
> This comes from the `withScrim` HOC that we have in the scrim component: https://github.com/Skyscanner/backpack/blob/main/packages/bpk-scrim-utils/src/withScrim.tsx
> 
> Which we wrap the DialogInner around: https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-dialog/src/BpkDialogInner.tsx#L75

OHH I looked at the code 100 times and did not see that export 😂 thanks

packages/bpk-component-dialog/src/BpkDialog.tsx
anambl: Makes sense, yeah, somehow my eyes totally skipped over the export so I just didn't see the scrim 😂

packages/bpk-component-dialog/src/BpkDialog.tsx
anambl: Yes, it's the scrim that applies the `aria-hidden` to the `getApplicationElement`, but it was the portal that was rendering it outside it.
Because looking at the [example](https://github.com/Skyscanner/backpack/pull/3508/files#diff-268f629c295e29c7e8875462a2fd7ef18a97eb0058f784296bddae08e1375635R63-R67), I can see that we explicitly render the dialog outside the `pagewrap` element so the examples would look good, but microsites would normally render it within the `pagewrap` element which sits at the very top for example here in [profile](https://github.skyscannertools.net/identity/profile/blob/e503cea49ba9a9c2b9209317936ffb6319464d15/app/routes/priceAlerts/index.html#L30)
Let me know if you'd like to zoom to chat through this 😄

olliecurtis: Taking a look at the code linked and live running of the website it seems that what we have in the examples and what we render is the exact same, whereby the dialog is rendered outside even when the pagewrap is the top element :) 

Screenshot 2024-06-27 at 17 02 53

examples/bpk-component-dialog/examples.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const onClose = () => {
+    setIsOpen(false);
+  };
+
+  return (
+    <div id="dialog-container">
+      <div id="pagewrap">
+        <BpkButton onClick={onOpen}>Open dialog</BpkButton>
+      </div>
+      <BpkDialog
```
anambl, line 67: Can we add some examples (or change the existing ones) to render the dialog within the `pagewrap` element to reproduce how it would actually be like in production and more easily test that it's accessible?

anambl:
There seem to be some issues with the visual tests 🤔
The [storybook](https://backpack.github.io/storybook-prs/3508/?path=/story/bpk-component-dialog--visual-test-default) link shows the supernova page, and [percy](https://percy.io/096a694a/backpack/builds/35311641/unchanged/1929221020?browser=chrome&browser_ids=59&category=changed&device_name=&extra_category=missing-snapshots&group_snapshots_by=&os=&other-browser=&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280) shows a blank page probably because it takes a while to load. Any clues why that is?

anambl (approved):

PR #3518, opened 2024-07-01
[QUAR-425] [BpkInsetBanner] Add component

kirstybryce:
### JIRA: QUAR-425

We want to use the inset banner component in our ads in web like we already [do for our ads in apps](https://www.skyscanner.design/latest/components/inset-banner/compose-SIxM16J2). In this PR we're adding the inset banner component, based on the guidance set out by the design team (see the [inset banner variants designs](https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?node-id=38191-6271&t=aoKCa5ErjKIKuN2W-4) and [props](https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?node-id=38191-5613&t=aoKCa5ErjKIKuN2W-4)). 

The only thing missing in the component from the designs is the ad info button which when clicked opens a popover window with legal information about the ad. The implementation of this is complex and requires further investigation into how we want to do it, so it has been postponed for a future iteration of the component (see slack thread for more details). The call to action section in the banner currently only accepts the `text` prop, to display `Sponsored` in our ads, but in the future we will add the ad info functionality props to the `callToAction` object prop.

## Checklist

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests


Homeralf (approved):
LGTM good job 🚀

LexSwed (approved):
Looking goood 🚀

examples/bpk-component-inset-banner/examples.js
olliecurtis, line 1: Can you write this in TS?

examples/bpk-component-inset-banner/stories.js
olliecurtis, line 1: Can you write this in TS?

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -0,0 +1,114 @@
...
+import STYLES from './BpkInsetBanner.module.scss';
+
+const getClassName = cssModules(STYLES);
+
+type Props = {
+  title?: string;
+  subheadline?: string;
+  logo?: string;
+  backgroundColor: string;
+  variant: 'onLight' | 'onDark';
```
olliecurtis, line 29: Can this be changed to have an exported object of `VARIANTS`

So something like:

```tsx
export const VARIANTS = {
  onLight: 'on-light',
  onDark: 'on-dark'
} as const;
```

Which will allow consumers to supply variant of `variant={VARIANT.onLight}`

Which can then mean this type can be 

```suggestion
  variant: (typeof VARIANT)[keyof typeof VARIANT];
```

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -0,0 +1,114 @@
...
+};
+
+const BpkInsetBanner = ({
+  backgroundColor,
+  body,
+  callToAction,
+  logo,
+  subheadline,
+  title,
+  variant,
```
olliecurtis, line 48: Should any of these have defaults?

E.g. variant could have a default of onLight

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const classNames = [];
+  if (variant === 'onLight') {
+    classNames.push('bpk-inset-banner--on-light');
+  }
+  if (variant === 'onDark') {
+    classNames.push('bpk-inset-banner--on-dark');
+  }
+  if (body) {
+    classNames.push('bpk-inset-banner--with-body');
+  }
```
olliecurtis, line 59: With the above suggestion for Variants, this could then become
 
```suggestion
  const classNames = getClassName(
      'bpk-inset-banner',
      variant && `bpk-inset-banner--${variant}`
      body && 'bpk-inset-banner--with-body'
  );
```

Then on L64 you can just do

`className={classNames}`

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -0,0 +1,114 @@
...
+        style={{
+          backgroundColor,
+        }}
+      >
+        <div className={getClassName('bpk-inset-banner--content-container')}>
+          {logo ? (
+            <img
+              className={getClassName('bpk-inset-banner--image')}
+              src={logo}
+              alt=""
```
olliecurtis, line 74: I think logos would need an alt looking at the figma designs, which would be a partner name

So it could be that we make logo an ReactElement and all this is passed into instead of building ourselves?

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -0,0 +1,114 @@
...
+              aria-hidden
+            />
+          ) : null}
+          <div className={getClassName('bpk-inset-banner--text-container')}>
+            <div className={getClassName('bpk-inset-banner--title')}>
+              {title}
+            </div>
```
olliecurtis, line 81: Looking at the CSS for this, we could make this a `BpkText`? And then just use the props to set the size to be label2?

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -0,0 +1,114 @@
...
+            <div className={getClassName('bpk-inset-banner--subheadline')}>
+              {subheadline}
+            </div>
```
olliecurtis, line 84: Similar to above but using caption

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -0,0 +1,114 @@
...
+          </div>
+        </div>
+        <div className={getClassName('bpk-inset-banner--cta-container')}>
+          {callToAction?.text ? (
+            <div className={getClassName('bpk-inset-banner--cta-text')}>
+              {callToAction.text}
+            </div>
+          ) : null}
```
olliecurtis, line 92: For this right now I think you can do, then avoiding needing to return back a null

```suggestion
          {callToAction?.text && (
            <div className={getClassName('bpk-inset-banner--cta-text')}>
              {callToAction.text}
            </div>
          )}
```

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -0,0 +1,114 @@
...
+              href={body.link}
+              className={getClassName(
+                'bpk-inset-banner-body-container--link-text',
+              )}
+            >
+              {body.linkText}
+            </a>
+          ) : null}
+        </div>
+      ) : null}
```
olliecurtis, line 109: Similar to above

olliecurtis (changes requested):

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
kirstybryce: If I add default values for backgroundColor and variant, should these then become non-required props?

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
kirstybryce: The current usage in the ads is that the partner logo is hidden in the voice over and the partner info is passed to voice over via an aria-label as `Advertisement by {partner}` that is read out at the start of the advert. In this case we can pass some a11y text as a prop that can be added to the content container div?

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
olliecurtis: Yeah I would say have some defaults and make them not required so that it can be the most simple form to make the component work.

Which looks like this has been updated now!

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
olliecurtis: Hmm good question maybe this is something we need to confirm up with design and make sure that is reflected in Figma to be surrounding rather than just the logo.

I haven't tried it yet but what would the impact be for a screenreader if the whole thing is surrounded with the aria label and then there is more things to track inside it

kirstybryce: I just checked with Eileen and she confirmed we want to use the current method used in our ads which is to have the `Advertisement by {partner}` aria-label added to the container of the card, so that all the elements are contained in a group and can be skipped if the user doesn't want to listen to the content of the card.

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
kirstybryce: I made the props not required here :) https://github.com/Skyscanner/backpack/pull/3518/commits/53fc0926c8412f54c64cd8ec14838654255b9445

olliecurtis (approved):
LGTM! Thank you for the contribution!

examples/bpk-component-inset-banner/stories.ts
```diff
@@ -0,0 +1,38 @@
...
+export default {
+  title: 'bpk-component-inset-banner',
+  component: BpkInsetBanner,
+};
+
+export const DefaultWithTitle = DefaultExampleTitleOnly;
+export const DefaultWithTitleAndSubheadline = DefaultExampleTitleAndSubheadline;
+export const WithLogoAndCtaTextLight = WithLogoAndCtaTextExampleLight;
+export const WithBodyTextLight = WithBodyTextExampleLight;
+export const WithBodyTextAndLinkDark = WithBodyTextAndLinkExampleDark;
```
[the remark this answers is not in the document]
kirstybryce: Added it here https://github.com/Skyscanner/backpack/pull/3518/commits/3cf329ef1d63290cdb6d39cf999f0bb6fdf829a8 and approved build https://percy.io/096a694a/backpack/builds/35278409/changed/1927945039?browser=chrome&browser_ids=59&group_snapshots_by=similar_diff&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280 :)

PR #3522, opened 2024-07-03
[ARGG-1159][BpkModal]: Remove portal from modal

olliecurtis:
Similar to #3508 

Migrates the BpkModal to not use the custom portal implementation from Backpack.

All functionality has been retained, the only main change here is remove the use of BpkPortal and replace the state management to be local to the component.

- [ ] Component `README.md`
- [x] Tests
- [x] Accessibility tests


packages/bpk-component-modal/src/BpkModal.tsx
```diff
@@ -71,12 +74,28 @@ const BpkModal = ({
   modalStyle = MODAL_STYLING.default,
   onClose = () => null,
   padded = true,
-  renderTarget = null,
+  renderTarget = () => null,
```
anambl, line 77: looking at the TS error in CI, it seems like we'll have to do the same for the datepicker

packages/bpk-component-modal/src/BpkModal.tsx
```diff
@@ -37,7 +40,7 @@ export type Props = Partial<ModalDialogProps> & {
   isOpen: boolean;
   closeOnScrimClick?: boolean;
   closeOnEscPressed?: boolean;
-  renderTarget?: null | HTMLElement | (() => null | HTMLElement);
```
anambl, line 40: Interesting that `() => null` default value satisfies the `() => HTMLElement` type and this is not needed 🤔

olliecurtis: This I had followed from our other components that seems to work. I guess its satisfied under the `null` part of the type

anambl (approved):

PR #3525, opened 2024-07-08
[Luna-1372][BpkAccordion] Remove description filter

juanmataruiz:
### Description
🎫 [Luna-1371](https://skyscanner.atlassian.net/browse/LUNA-1371) for reference.

There is an issue on iOS where VoiceOver reads the word "description" as it goes through the flight's filter Accordion.

We have identified that this is due to the usage of `dt` `dl` and `dd` in the component as VioceOver will read the element attached to it.

### Checklist

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests

### DIFF
--- a/packages/bpk-component-accordion/src/BpkAccordion.js
+++ b/packages/bpk-component-accordion/src/BpkAccordion.js
@@ -35,9 +35,9 @@ const BpkAccordion = (props: Props) => {
 
   return (
     // $FlowFixMe[cannot-spread-inexact] - inexact rest. See decisions/flowfixme.md
-    <dl className={classNames} {...rest}>
+    <div className={classNames} {...rest}>
       {children}
-    </dl>
+    </div>
   );
 };
 
--- a/packages/bpk-component-accordion/src/BpkAccordion.js
+++ b/packages/bpk-component-accordion/src/BpkAccordion.js
@@ -19,8 +19,9 @@
 /* @flow strict */
 
 import PropTypes from 'prop-types';
-import React, { type Node } from 'react';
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -180,5 +180,8 @@
     "webpack": "^4.30.0",
     "webpack-cli": "^3.3.2",
     "webpack-dev-server": "^3.3.1"
+  },
+  "dependencies": {
+    "@skyscanner/backpack-web": "^34.10.1"
```
olliecurtis, line 185: This isn't to have been installed can you remove?

packages/bpk-component-accordion/src/BpkAccordion.js
```diff
@@ -35,9 +35,9 @@ const BpkAccordion = (props: Props) => {
 
   return (
     // $FlowFixMe[cannot-spread-inexact] - inexact rest. See decisions/flowfixme.md
-    <dl className={classNames} {...rest}>
+    <div className={classNames} {...rest}>
```
olliecurtis, line 38: Let me do some checking here for this change, as I think some of these were made for SEO purposes so we may need to do some logic to support both

package.json
juanmataruiz: Oh, I missed this. Done now thanks!

packages/bpk-component-accordion/src/BpkAccordion.js
juanmataruiz: Thanks for doing this. 

If the description elements can't be removed, I would need to chat with the backpack team about the best approach to fix this bug.

PR #3527, opened 2024-07-10
[NO JIRA][BpkPopover]: Reinstate renderTarget prop for rendering in DOM at specific location

olliecurtis:
We had an issue where when a popover is used within a modal that meant that the popover would render outside of the modal and to the main page below, meaning it would be hidden from the view and rendered behind.

This is due to the default behvaiour without specification to render to the body. So to solve this we return a previously removed property in v34 `renderTarget`.

This property is used to specify to the `FloatingPortal`, where to attach the popover giving consumers the previous ability to specify where in the DOM to place the component when its time to render meaning that it can then be displayed within the modal and not to content behind

Screenshot 2024-07-10 at 09 38 26

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests

### DIFF
--- a/examples/bpk-component-popover/examples.js
+++ b/examples/bpk-component-popover/examples.js
@@ -144,6 +144,13 @@ const DefaultExample = () => (
   </Spacer>
 );
 
+const WithCustomRenderTargetExample = () => (
+  <Spacer>
+    <div id="my-target" />
+    <PopoverContainer id="my-popover-1" renderTarget={() => document.getElementById('my-target')} />
+  </Spacer>
+);
+
 const WithoutArrowExample = () => (
   <Spacer>
     <PopoverContainer id="my-popover-2" displayArrow={false} />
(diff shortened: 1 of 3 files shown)


olliecurtis:
> Why was the prop removed before?

It had been previously removed as it wasn't required for the new library implementation, but we hadn't factored in that people may popover in a modal or dialog so its default rendering would be outside of the regular DOM

metalix2:
For cases that don't use BpkPopover in a modal they can continue to not specify a `renderTarget` Prop?

mungodewar:
> > Why was the prop removed before?
> 
> It had been previously removed as it wasn't required for the new library implementation, but we hadn't factored in that people may popover in a modal or dialog so its default rendering would be outside of the regular DOM

It's tough that we can't solve this within the library, cause I think the desired behaviour (modals over content, popups over modal) is clear. Can it's rendering not be fixed with an appropriate `z-index` value?

PR #3540, opened 2024-07-16
[LOOM-1525] - Refactor Nudger remove readonly attribute from input and Emit native events from buttons

metalix2:
## Description

The Backpack Nudger can be part of a form but it doesn't behave like a form element. The readonly attribute removes any change event that would be associated with the input. The name attribute is also missing but would be required for submission with FormData. 

To get around this we have:
 - Removed `readonly` attribute from input element
 - replaced input element's `value` with `defaultValue`
 - onClick functions now fire additional Native 'change' Event that would be expected for interacting with a form element.
 - Added `name` attribute which allows the nudger to be part of a form and consumed with FormData.
 
- [ ] Component `README.md`
- [ ] Tests
- [x] Accessibility tests

        https://github.com/user-attachments/assets/88844995-318a-460f-a00c-c79823a09887


packages/bpk-component-nudger/src/BpkConfigurableNudger.tsx
```diff
@@ -82,32 +85,47 @@ const BpkConfigurableNudger = ({
...
       <input
         type="text"
         aria-live="polite"
-        readOnly
-        value={formatValue(value)}
+        defaultValue={formatValue(value)}
         id={id}
+        ref={inputRef}
+        name={name || id}
+        onChange={(event) => onChange(parseInt(event?.target.value, 10))}
```
metalix2, line 115: This ensures that we're able to parse the user input as a number. Otherwise if user enters "1" and we increment the value we get "11" 🎖️ Javascript at it's best.

metalix2: Perhaps it has to happen further up?

olliecurtis (approved):

PR #3542, opened 2024-07-17
[NO-JIRA] [BpkInsetBanner] Add classname and rest props

kirstybryce:
Added the `className` and `rest` props to BpkInsetBanner as we need them for the usage in the sponsored booking panel ad.

- `className` prop:
   - While adding the `BpkInsetBanner` component in the sponsored booking ad to replace the custom made banner, I noticed that the padding for the custom banner is different than the backpack component. 
   - Since the banner UI has to match the one of the current custom banner in the sponsored booking panel ad, I added the `className` prop so that a custom class can be passed to the banner container. 
   - With this solution we can pass a className to change the padding of the inset banner in the sponsored booking panel ad, instead of changing the padding directly in the backpack component as it would unalign the padding with the [BpkInsetBanner designs](https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?node-id=38191-5527&t=McTTZOu8DoQfEzg9-4).
- `rest` props:
   - Added the `rest` prop to allow any additional props that might be needed to be passed.
   - For the sponsored booking panel ad we need to add an `onClick` prop to the banner, so adding `rest` props allows us to do this without setting up a dedicated `onClick` prop and click handling in the backpack component (since the whole banner isn't intended to be clicked in most cases).

- [x] Component `README.md`
- [x] Tests
- [ ] Accessibility tests

### DIFF
--- a/examples/bpk-component-inset-banner/examples.module.scss
+++ b/examples/bpk-component-inset-banner/examples.module.scss
@@ -1,0 +1,23 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


kirstybryce:
After discussing with design we want to use the current padding in the booking panel ad instead of the padding defined in the original designs, so we no longer need the custom class name. We also decided since the click functionality on the whole banner is only needed for the booking panel ad that we will add this in an enclosing div instead of inside the backpack component.

PR #3531, opened 2024-07-15
[CYB-2251][Snippet] Add new Snippet component

taofang2021:
Add the Snippet component for web, which align with the [app](https://www.skyscanner.design/latest/components/snippet/compose-CaNywPG3)
image
image

Remember to include the following changes:

- [x] Ensure the PR title includes the name of the component you are changing so it's clear in the release notes for consumers of the changes in the version e.g `[KOA-123][BpkButton] Updating the colour`
- [x] `README.md` (If you have created a new component)
- [x] Component `README.md`
- [x] Tests
- [ ] Accessibility tests
    - The following checks were performed:
        - [ ] Ability to navigate using a [keyboard only](https://webaim.org/techniques/keyboard/)
        - [ ] Zoom functionality ([Deque University explanation](https://dequeuniversity.com/checklists/web/text)):
            - [ ] The page SHOULD be functional AND readable when only the text is magnified to 200% of its initial size
            - [ ] Pages must reflow as zoom increases up to 400% so that content continues to be presented in only one column i.e. Content MUST NOT require scrolling in two directions (both vertically and horizontally)
        - [ ] Ability to navigate using a [screen reader only](https://webaim.org/articles/screenreader_testing/)
- [x] Storybook examples created/updated
- [ ] For breaking changes or deprecating components/properties, migration guides added to the description of the PR. If the guide has large changes, consider creating a new Markdown page inside the component's docs folder and link it here

PR #3550, opened 2024-07-22
[NO JIRA][BpkImage]: Update image radius to 12px

olliecurtis:
There is a design requirement to move the BpkImage border radius to be 12px instead of 8px. This change got raised as part of the #3531 where it was found that the current BpkImage design is not aligned to the Figma component version.

So this PR changes this to align to Figma and be able to support the BpkSnippet contribution. This has been signed off by design as needing to be done!

- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-image/src/BpkImage.module.scss
+++ b/packages/bpk-component-image/src/BpkImage.module.scss
@@ -37,6 +37,6 @@
   }
 
   &--border-radius-sm {
-    @include radii.bpk-border-radius-sm;
+    @include radii.bpk-border-radius-md;
   }
 }


jronald01 (approved):

packages/bpk-component-image/src/BpkImage.module.scss
```diff
@@ -37,6 +37,6 @@
   }
 
   &--border-radius-sm {
-    @include radii.bpk-border-radius-sm;
+    @include radii.bpk-border-radius-md;
   }
```
taofang2021, line 41: Hi, is it better to keep the sytle name consistent with the size? so that user can better to understand the actual radius size they are using.

PR #3564, opened 2024-07-29
LOOM-1521: add name props for input

Parisistan:
Loom is implementing name props in splitInput and its parent inputField so that it can be submitted in a form.

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests


packages/bpk-component-split-input/src/BpkInputField.js
```diff
@@ -62,6 +63,7 @@ BpkInputField.propTypes = {
   value: PropTypes.string,
   focus: PropTypes.bool.isRequired,
   index: PropTypes.number.isRequired,
+  name: PropTypes.string.isRequired,
```
metalix2, line 66: We probably don't want to make it a required field.. as it would be a breaking change and a major release. The component can still work without a name attribute but we want to allow consumers to define if needed.

Parisistan: Makes sense!

packages/bpk-component-split-input/src/BpkInputField.js
```diff
@@ -62,6 +63,7 @@ BpkInputField.propTypes = {
   value: PropTypes.string,
   focus: PropTypes.bool.isRequired,
   index: PropTypes.number.isRequired,
+  name: PropTypes.string,
```
olliecurtis, line 66: As this is not marked as required it will need a default value otherwise the value when rendered will be blank and/or crrash

olliecurtis (changes requested):

packages/bpk-component-split-input/src/BpkInputField.js
Parisistan: Thanks for pointing that out!

olliecurtis (approved):

PR #3574, opened 2024-08-05
NO-JIRA: Added submit to loading button props

amburrrshand:
Minor change to add the submit prop to the loading button props. Due to confusion in the past, it's probably best to make the submit prop type for the loading button explicit for consumers of Backpack as in the past, some users have set the type to be submit without knowing the submit prop exists and inherits the props from BpkButton via ...rest

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/examples/bpk-component-button/examples.tsx
+++ b/examples/bpk-component-button/examples.tsx
@@ -28,6 +28,7 @@ import {
 } from '../../packages/bpk-component-icon';
 import LargeLongArrowRightIcon from '../../packages/bpk-component-icon/lg/long-arrow-right';
 import SmallLongArrowRightIcon from '../../packages/bpk-component-icon/sm/long-arrow-right';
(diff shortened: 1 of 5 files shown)


examples/bpk-component-button/examples.tsx
```diff
@@ -176,6 +177,18 @@ const FullWidthExample = (props: any) => (
   </BpkButtonV2>
 );
 
+const SubmitButtonExample = (props: any) => (
+  <BpkButtonV2 submit {...props}>
+    Submit Button
+  </BpkButtonV2>
+);
+
+const SubmitLoadingButtonExample = (props: any) => (
```
anambl, line 186: ohh this one should probably sit in the bpk-component-loading-button example file 😄

amburrrshand:
> Awesome! I think it'd be a `patch` rather than `minor` though? Wdyt? 🤔

Yeah I agree, thanks!

PR #3576, opened 2024-08-06
[BpkNudger] Simplify nudger to make it support only number types

metalix2:
Updates to the Nudger to remove the ConfigurableNudger Component. To keep aligned with the designs the nudger is meant to only support number values. 

Nudger Input can now be directly interacted with providing a way to set the desired value directly.

https://github.com/user-attachments/assets/2dd98201-3067-4244-8494-01e040823ee5

- [ ] Component `README.md`
- [x] Tests
- [x] Accessibility tests


packages/bpk-react-utils/src/nativeEventHandler.ts
```diff
@@ -20,17 +20,20 @@
...
     inputProto,
     'value',
   ) as PropertyDescriptor;
   const setValue = descriptor.set;
+  
   if (setValue) {
-    const event = new Event('change', { bubbles: true });
     setValue.call(element, value);
-    element.dispatchEvent(event);
+    if (shouldDispatchChange) {
```
metalix2, line 33: Added this as the nudger may set the value on the input and we set the state but we don't want to prematurely send the onChange before it does it on it's own.

packages/bpk-component-nudger/src/BpkNudger.module.scss
```diff
@@ -23,7 +23,6 @@
 .bpk-nudger {
   @include utils.bpk-rtl {
     display: flex;
-    flex-direction: row-reverse;
```
metalix2, line 26: RTL fix?

packages/bpk-component-nudger/src/BpkNudger.module.scss
```diff
@@ -37,6 +36,16 @@
...
     @include typography.bpk-heading-5;
 
+    &::-webkit-outer-spin-button,
+    &::-webkit-inner-spin-button {
+      margin: 0;
+      appearance: none;
+    }
+
+    &[type='number'] {
+      appearance: textfield;
```
metalix2, line 46: Hides the little number widgets 
Screenshot 2024-08-06 at 15 00 14

packages/bpk-component-nudger/src/BpkNudger.module.scss
metalix2: https://github.com/user-attachments/assets/f4e89874-09c3-4ded-9e0e-3523ae1e9cf8

examples/bpk-component-nudger/examples.js
```diff
@@ -71,7 +69,7 @@ class NudgerContainer extends Component<
         <BpkNudger
           aria-labelledby="passenger-label"
           id={id}
-          min={1}
+          min={-10}
```
olliecurtis, line 72: I wonder if we should be allowing values to have a minimum less than zero 🤔 can't see the usecase for it

packages/bpk-component-nudger/src/BpkNudger.tsx
```diff
@@ -16,41 +16,79 @@
...
 const BpkNudger = ({
   buttonType = 'secondary',
   className = null,
+  decreaseButtonLabel,
   icon,
   id,
+  increaseButtonLabel,
+  inputClassName,
+  max,
+  min,
```
olliecurtis, line 46: Wonder if this should be clamped down in the code and be set as a default to 0, so it can never drop below zero as minus numbers never come up for nudgers

examples/bpk-component-nudger/examples.js
metalix2: Possibly.. but it's up to the consumers what values they want to set.  this is just an example anyway.

packages/bpk-component-nudger/src/BpkNudger.tsx
metalix2: Could that be to restrictive? I think consumers will most likely not go below 0. We can set a default min of 0?

examples/bpk-component-nudger/examples.js
metalix2: Can remove this and set it back to 1

packages/bpk-component-nudger/src/BpkNudger.tsx
olliecurtis: I think it wouldn't be too restrictive, as we would be setting a correct expectation of what the system should be and avoid people possibly putting bad data into our systems and causing errors, so if you try to go beyond 0 it would bring it back in and not accidently get set below

metalix2: Is this what you had in mind - https://github.com/Skyscanner/backpack/pull/3576/commits/954cbd823df21c986270fe831e19ee6d558cb27f ?

packages/bpk-component-nudger/src/BpkNudger.tsx
```diff
@@ -16,41 +16,81 @@
...
+  name,
+  onValueChange,
+  step = 1,
   subtitle,
   title,
+  value,
   ...rest
 }: CommonProps) => {
+  // Ensure that the minimum value is 0 or greater it's not designed to handle negative values.
+  const minimum = min < 0 ? 0 : min;
```
olliecurtis, line 56: With this setup it seems that once you go passed zero upwards you can never get back to zero according to this: https://backpack.github.io/storybook-prs/3576/?path=/story/bpk-component-nudger--stateful

metalix2: Hmmm that is strange I think i need to make it `<=`

metalix2: Ahh that example is set to 1 for minium. with a defualt value of 0 😂

olliecurtis (approved):

PR #3579, opened 2024-08-07
[Loom-1535] - add form-test to bpkSlider

Parisistan:
Adding a form-test to our slider component to check if it renders correctly. 

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-slider/src/form-test.tsx
+++ b/packages/bpk-component-slider/src/form-test.tsx
@@ -1,0 +1,103 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2016 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useState } from 'react';
+
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import BpkSlider from './BpkSlider';
+
+window.ResizeObserver =
+  window.ResizeObserver ||
+  jest.fn().mockImplementation(() => ({
+    disconnect: jest.fn(),
+    observe: jest.fn(),
+    unobserve: jest.fn(),
+  }));
+
+describe('BpkSlider form test', () => {
+  it('should work as a form component in a form', async () => {
+    const Wrap = () => {
+      const [sliderValue, setSliderValue] = useState<number[]>([50]);
+      return (
+        <form data-testid="form">
+          <BpkSlider
+            id="slider"
+            min={0}
+            max={100}
+            value={sliderValue}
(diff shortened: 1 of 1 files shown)


packages/bpk-component-slider/src/form-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+          <BpkSlider
+            id="slider"
+            min={0}
+            max={100}
+            value={sliderValue}
+            name="slider"
+            ariaLabel={['slider']}
+            ariaValuetext={['min', 'max']}
+            data-testid="mySlider"
+            onChange={(value) => setSliderValue(Array.isArray(value) ? value : [value])}
```
mungodewar, line 49: Why is the value sometimes within an array and sometimes not?

packages/bpk-component-slider/src/form-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+            id="slider"
+            min={0}
+            max={100}
+            name="slider"
+            ariaLabel={['slider']}
+            ariaValuetext={['min', 'max']}
+            value={sliderValue}
+            onChange={(value) => {
+              handleChange(value);
+              setSliderValue(Array.isArray(value) ? value : [value]);
```
mungodewar, line 85: Within this test, if we are not interested in the value, we can just remove the state that does the setting:

```
onChange={handleChange}
```

Although, another option is to have this `handleChange` assertion in "should work as a form component in a form" test?

packages/bpk-component-slider/src/form-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    const sliderThumb = screen.getByLabelText('slider');
+    expect(sliderThumb).toBeInTheDocument();
+
+    await sliderThumb.focus();
+    await userEvent.keyboard('{ArrowRight}');
+    expect(handleChange).toHaveBeenCalledTimes(1);
+  });
+
+  it('should emit change event when both sides of slider value are changed', async () => {
```
mungodewar, line 102: Nice, I think this test has overlap with "should emit change event when slider value is changed" & thus I think you only need one of them & I think this test is better as it tests both slider value changes.

Parisistan: Yeah fair!

packages/bpk-component-slider/src/form-test.tsx
Parisistan: Because sometimes there aren't a min and max value that could both change and only one side of the slider is changeable. So I wanted one of the tests to check for when both `min, max` can change to cover various types. Does that make sense?

packages/bpk-component-slider/src/form-test.tsx
Parisistan: Eliminated this test as it was covered in the following.

packages/bpk-component-slider/src/form-test.tsx
Parisistan: image

mungodewar: I think I see, so if the `sliderValue={[20,80]}` then the bottom "two thumb" slider will be shown? 

I think that if we want to test both cases (single slider & multi slider) we should create two test cases. Or we choose the most simple case only to test the form interaction.

Parisistan: > so if the sliderValue={[20,80]} then the bottom "two thumb" slider will be shown?
yes that is right!
And okay, I can make it be clear in the description too so there is no confusion for users. :)

mungodewar (approved):
LGTM - nice 🎉

PR #3582, opened 2024-08-12
[NO JIRA][BpkSwitch]: Remove label and just ship switch

olliecurtis:
Removes the label property from BpkSwitch so the component aligns with App to just provide the switch itself, and so the label can then be provided by the consumer and placed either side of the component.

It also adds a required `ariaLabel` so that if the switch it being used Assistive Technologies can read out the context of the component.

The component has also been converted to Typescript.

To update the component to this updated API you will now be required to provide your own text and layout when using the component. E.g.

```tsx
// Before

<BpkSwitch label="My switch" />

// After

<div>
   <BpkText> My switch </BpkText>
   <BpkSwitch ariaLabel="Toggle my switch" />
</div>
```

- [ ] Component `README.md`
- [x] Tests
- [x] Accessibility tests


packages/bpk-component-switch/src/BpkSwitch.tsx
```diff
@@ -27,44 +23,34 @@ import STYLES from './BpkSwitch.module.scss';
...
+  small = false,
+  ...rest
+}: Props) => {
   const switchClassNames = getClassName(
     'bpk-switch__switch',
     small && 'bpk-switch__switch--small',
   );
 
   return (
     <label className={getClassName('bpk-switch', className)}>
```
LouiseReid, line 44: does this still need to be wrapped in a label?

olliecurtis: Taking a look yes it still needs to be a label so that the component is still fully selectable and operational as a switch with a Assistive Tech, otherwise looks like this

Screenshot 2024-08-12 at 13 16 45

metalix2 (approved):
This looks great!

PR #3585, opened 2024-08-13
[LOOM-1609]: React 18 migration

olliecurtis:

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests


dangerfile.ts
```diff
@@ -149,10 +149,11 @@ const reactRecogniseProp = ["React does not recognize"]
 const invalidTags = ["The tag <rect>", "The tag <g>", "The tag <text>"]
 const passingTests = ["✓"]
 const unknownEventHandler = ["Unknown event handler"]
-const propType = ["Failed prop type"]
+const propType = ["Failed prop type", "Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.", 'A props object containing a "key" prop is being spread into JSX:']
```
olliecurtis, line 152: The first new addition has been added because we will need to ignore this until we have full conversion of components to Typescript. 

For the second it shall be addressed in this PR or a separate one as this is new

dangerfile.ts
```diff
@@ -149,10 +149,11 @@ const reactRecogniseProp = ["React does not recognize"]
...
 const componentWillReceiveProps = ["componentWillReceiveProps"]
 const invalidCSSProperties = ["is an invalid value for the .* css style property."]
 const invalidProps = ["for a non-boolean attribute", "Invalid ARIA attribute"]
+const actTests = ["inside a test was not wrapped in act(...)"]
```
olliecurtis, line 156: This is a newly introduced warning, tests still pass and there is a few conversations open for this

`https://github.com/reactwg/react-18/discussions/102`

As its not a blocker adding it here now and we should revisit in the future

dangerfile.ts
```diff
@@ -149,10 +149,14 @@ const reactRecogniseProp = ["React does not recognize"]
...
 const unknownEventHandler = ["Unknown event handler"]
-const propType = ["Failed prop type"]
+const propType = ["Failed prop type", "Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.", 'A props object containing a "key" prop is being spread into JSX:']
 const componentWillReceiveProps = ["componentWillReceiveProps"]
 const invalidCSSProperties = ["is an invalid value for the .* css style property."]
 const invalidProps = ["for a non-boolean attribute", "Invalid ARIA attribute"]
+// TODO: Address tests being wrapped in act
+const actTests = ["inside a test was not wrapped in act(...)"]
+// TODO: Convert components that use CSSTransition to functional components and allow for using refs
+const findDOMNode = ["findDOMNode is deprecated and will be removed in the next major release."];
```
olliecurtis, line 159: This is coming from the underlying `react-transition-group` library

packages/bpk-component-breakpoint/src/useMediaQuery-test.tsx
```diff
@@ -66,10 +66,8 @@ describe('useMediaQuery', () => {
     }));
 
     const view = renderHook(() => useMediaQuery('(min-width: 768px)'));
-
-    expect(view.result.all.length).toBe(1);
-
-    expect(view.result.all[0]).toBe(true);
+    
+    expect(view.result.current).toBe(true);
```
olliecurtis, line 70: The structure of `renderHook` returns a different object where the value is located

packages/bpk-component-carousel/src/utils.test.tsx
olliecurtis, line 1: Need to check where or what this has come from 🤔 

I think it might have been an error thrown that there was no tests for this

packages/bpk-theming/src/BpkThemeProvider.js
```diff
@@ -56,36 +55,26 @@ const createStyle = (theme, themeAttributes) => {
...
-  render() {
-    const {
-      children,
-      component: WrapperComponent,
-      style: userStyle,
-      theme,
-      themeAttributes,
-      ...rest
-    } = this.props;
+const BpkThemeProvider = (props) => {
```
olliecurtis, line 58: We change this to be a functional component as the `getChildContext` is deprecated and will be removed in a future version.

packages/package.json
```diff
@@ -49,8 +49,8 @@
   },
   "peerDependencies": {
     "node-sass": ">= 7",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": ">=17.0.2",
+    "react-dom": ">=17.0.2",
```
olliecurtis, line 53: We don't want to block folks from upgrading Backpack in parallel to adopting React

scripts/jest/setup.js
```diff
@@ -19,6 +19,8 @@
 import 'jest-axe/extend-expect';
 import '@testing-library/jest-dom';
 import 'raf/polyfill';
+import { TextEncoder } from 'util'
```
olliecurtis, line 22: This mock is required as part of the upgrade as in a newer version of jsdom this is not exposed/available to the environment so causes tests to fail due to underlying global not being available

packages/package.json
xalechez: Should we limit it to `17.0.2 || ^18.3.1`? I don't think we're ready to tell that we support 19 as well, whenever it's released

olliecurtis: Ah yes good point! Completely forgot about that will update that now

PR #3586, opened 2024-08-14
[NO JIRA][BpkSpinner]: Updating TS types and removing propTypes ahead of React 18

olliecurtis:
As part of the migration to React 18 when running tests we found errors that were being through due to the mix of TS types and and propTypes which are deprecated and discouraged since React 17

```
console.error
    Warning: BpkExtraLargeSpinner: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
        at className (/Users/olivercurtis/Documents/Backpack/backpack/packages/bpk-component-spinner/src/BpkExtraLargeSpinner.tsx:39:11)
```

- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests


packages/bpk-component-spinner/src/BpkExtraLargeSpinner.tsx
```diff
@@ -31,12 +29,16 @@ import STYLES from './BpkSpinner.module.scss';
 const getClassName = cssModules(STYLES);
 
 type Props = {
-  type: SpinnerTypes,
-  className?: string,
+  type?: SpinnerTypes;
+  className?: string | null;
```
xalechez, line 33: I'm wondering if we need semantic difference between null and undefined here?

olliecurtis: Possibly? I was jsut setting it here based on what we had as the previous right now to avoid any unexpected change or issues

xalechez (approved):

PR #3608, opened 2024-09-16
LOOM-1668 backpack date picker fires change event on value change

amburrrshand:

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-datepicker/src/BpkDatepicker.tsx
+++ b/packages/bpk-component-datepicker/src/BpkDatepicker.tsx
@@ -278,8 +278,9 @@ class BpkDatepicker extends Component<Props, State> {
           DateUtils.isSameDay(newEndDate, selectionConfiguration.startDate))
       ) {
         onDateSelect(selectionConfiguration.startDate, newEndDate);
+        this.myRef && setNativeValue(this.myRef, this.props.formatDate(newEndDate));
       } else {
-        this.myRef && setNativeValue(this.myRef, this.props.formatDate(newStartDate))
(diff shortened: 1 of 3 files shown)


packages/bpk-component-datepicker/src/BpkDatepicker.tsx
```diff
@@ -107,7 +109,11 @@ type State = {
 };
 
 class BpkDatepicker extends Component<Props, State> {
-  inputRef: React.RefObject<HTMLInputElement>;
+  inputRef: (ref:HTMLInputElement) => void;
+
+  myRef?: HTMLInputElement;
```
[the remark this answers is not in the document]
amburrrshand: Good shout! I've changed that now

packages/bpk-component-datepicker/src/BpkDatepicker.tsx
```diff
@@ -107,7 +109,11 @@ type State = {
 };
 
 class BpkDatepicker extends Component<Props, State> {
-  inputRef: React.RefObject<HTMLInputElement>;
+  inputRef: (ref:HTMLInputElement) => void;
```
metalix2, line 112: part of me wants to make the change in BpkInput to support forwardRef... but that's a task for another day. So we're stuck with inputRef prop for now. 
I think we should rename `divRef` to `focusRef` as that's it's purpose. Then I think that should be ok 👍

amburrrshand: 👍

metalix2 (approved):
I'm happy with this 👍

PR #3645, opened 2024-10-17
[NoJira][BpkSlider] - Fix native event handling for the Slider

metalix2:
### Description

Resolve native event bug where dragging over many slider steps result in multiple change events. 

This change fixes that once a users `mousedown` event is lifted the `"click"` will then register a final `change` event.

Works for keyboard as well the `"keyup"` event where a user can hold down the arrow key and the change sent at that moment.

Tested across Chrome, Firefox and  Safari.

https://github.com/user-attachments/assets/b87232a8-7633-4944-aca5-ee74f84dc3ca

https://github.com/user-attachments/assets/a0dbeffb-2f24-43a6-8ca4-c49cd7cba904

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests


packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -117,20 +126,67 @@ const BpkSlider = ({
...
+              setNativeValue(input, newValue);
+            }
+          }
+        };
+
+        const focusInHandler = () => {
+          // Controller needed as we may click more than once when in focus. On focusout we can then abort the event listeners
+          const controller = new AbortController();
+          thumb.addEventListener('focusout', () => controller.abort(), {
+            once: true,
```
metalix2, line 157: I attempted to use the removeEventListener but this approach worked a lot better.

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -117,20 +126,67 @@ const BpkSlider = ({
...
-        setNativeValue(input, `${value}`);
+      const thumb = thumbRef.current;
+      const input = ref.current;
+      if (thumb) {
+        const interactionEndHandler = (prevValue: number) => {
+          if (input) {
+            const newValue = parseFloat(input.getAttribute('value') as string);
+            if (prevValue !== newValue) {
+              setNativeValue(input, newValue);
+            }
```
mungodewar, line 149: Why do we parse things as a float? 
Why do we need to check the previous value vs the new?

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -117,20 +126,67 @@ const BpkSlider = ({
...
+  ) => {
+    const { thumbRef, value, ...inputProps } = props;
     const ref = useRef<HTMLInputElement>();
     const composedRefs = useComposedRefs(forwardedRef, ref);
 
-    const prevValue = usePrevious(value);
-
-    // Bubble value change to parents (e.g form change event)
+    // Provide native behaviour that the input range type provides.
     useEffect(() => {
```
mungodewar, line 140: I think generally this useEffect could use heavy documentation referencing the browser spec/MDN etc.

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -117,20 +126,67 @@ const BpkSlider = ({
...
+            const newValue = parseFloat(input.getAttribute('value') as string);
+            if (prevValue !== newValue) {
+              setNativeValue(input, newValue);
+            }
+          }
+        };
+
+        const focusInHandler = () => {
+          // Controller needed as we may click more than once when in focus. On focusout we can then abort the event listeners
+          const controller = new AbortController();
```
mungodewar, line 155: We good for browser support (AFAIK) https://caniuse.com/?search=abort%20controller

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -117,20 +126,67 @@ const BpkSlider = ({
...
+          thumb.addEventListener('focusout', () => controller.abort(), {
+            once: true,
+          });
+          // needed on document as users can drag the thumb while out of the thumb elements mouse area
+          document.addEventListener(
+            'click',
+            () => interactionEndHandler(value as number),
```
mungodewar, line 162: Why do we need to cast the value? Shouldn't `interactionEndHandler` expect to handle all the diferent types of values that come from the event listener?

mungodewar: General comment on this area of the code is that the nested nature of the options/functions etc make it a little difficult to read, is there a way to make it easier? I think, if we are able to not have to pass the casting there is a small increase in readability, eg:

```
document.addEventListener('click', interactionEndHandler, options)
```

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -117,20 +126,67 @@ const BpkSlider = ({
...
+            {
+              signal: controller.signal,
+            },
+          );
+          thumb.addEventListener(
+            'keyup',
+            (e: KeyboardEvent) => {
+              ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'].includes(
+                e.key,
+              ) && interactionEndHandler(value as number);
```
mungodewar, line 172: Should this logic not be within the handler itself? 
Even if there are different types of handlers with shared common logic: 

```js
// common & shared
function interactionValueParsing(){}

function keyupInteractionHandler (e){
   if(!['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'].includes( e.key)) return;
   const value = interactionValueParsing(e)
   // stuff
}

File: packages/bpk-component-slider/src/form-test.tsx [hunk 7]
```diff
@@ -67,9 +67,12 @@ describe('BpkSlider form-test for single-thumb slider', () => {
...
 
   it('should emit change event when both sides of slider value are changed for two-thumb slider', async () => {
-    const handleChange = jest.fn();
+    const formValidation = jest.fn();
     const Wrap = () => {
-      const [sliderValue, setSliderValue] = useState<number[]>([0, 100]);  // two-thumb slider with min and max value
+      const [sliderValue, setSliderValue] = useState<number[]>([0, 100]); // two-thumb slider with min and max value
+      useEffect(() => {
+        document.addEventListener('change', formValidation);
+      }, []);
```
mungodewar, line 75: How come this is within the `Wrap` component? Is there a reason for that?

mungodewar (changes requested):

packages/bpk-component-slider/src/BpkSlider.tsx
metalix2: > Why do we parse things as a float?
The input `type` is a 'number' and though we could use `parseInt` the range can actually support a float value. 
When we parse a string "3" as float we still get a whole number 3 rather than 3.0. And as were not parsing hex strings. Float fits this purpose really nicely. Defaulting to an int when it's a whole number.

> Why do we need to check the previous value vs the new?
If a user interacts and doesn't move the value then there is no "change" so we don't fire an event.

packages/bpk-component-slider/src/BpkSlider.tsx
metalix2: > Why do we need to cast the value? Shouldn't interactionEndHandler expect to handle all the diferent types of values that come from the event listener?

I thought we had originally needed them but the value remains the same once instanciated.

metalix2: >General comment on this area of the code is that the nested nature of the options/functions etc make it a little difficult to read, is there a way to make it easier? I think, if we are able to not have to pass the casting there is a small increase in readability, eg:


I'll see If I can refactor it a little.

packages/bpk-component-slider/src/BpkSlider.tsx
metalix2: Well It would mean checking for key events when it was a click event. This just sort of wraps and takes care of the custom keyboard logic.

packages/bpk-component-slider/src/form-test.tsx
metalix2: It's to setup the eventListener on render so that formValidation can be mocked and valdiated against.

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -117,20 +126,74 @@ const BpkSlider = ({
...
+            // parseFloat works for both integers and floats and the Slider supports decimal stepping. e.g. 0 - 1
+            const newValue = parseFloat(input.getAttribute('value') as string);
+            if (value !== newValue) { // newValue is changed if the slider has moved, clicking away from the thumb will not fire the change event
+              setNativeValue(input, newValue);
+            }
+          }
+        };
+
+        // the focusInHandler is used to add event listeners to the document when the thumb is in focus
+        // Allows us to track at which moment the user
```
mungodewar, line 167: Comment left on a cliffhanger :D

packages/bpk-component-slider/src/form-test.tsx
mungodewar: Why does it need to be setup on render?

packages/bpk-component-slider/src/BpkSlider.tsx
mungodewar: > It would mean checking for key events when it was a click event

I don't understand. Why? Don't we attach the handler twice, once for click and once for keyup - these could just be different handlers, no?

packages/bpk-component-slider/src/BpkSlider.tsx
mungodewar: What if the slider is `strong`, `medium`, `light` or do sliders not work like that? https://www.skyscanner.design/latest/components/slider/web-aNXvlY7y#section-non-integer-values-54 I think it doesn't work like that. 

Always makes me a little nervous when JS used in this way to have happy coincidences. [Math.trunc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) might be an alternative. 

Why should we not care about the 10ths/100th etc within the float like value returned? Would it cause problems if we did `3.1 === 3.1`?

packages/bpk-component-slider/src/BpkSlider.tsx
```diff
@@ -117,20 +126,82 @@ const BpkSlider = ({
...
+          thumb.addEventListener('focusout', () => controller.abort(), {
+            once: true, // not necessary to fire more than once and will restart on the next focusin
+          });
+
+          // These three EventListeners signal the end of the interaction with the thumb
+          document.addEventListener('click', interactionEndHandler, {
+            // needed on document as users can drag the thumb while out of the thumb elements mouse area
+            signal: controller.signal,
+          });
+          thumb.addEventListener('touchend', interactionEndHandler, {
```
metalix2, line 185: touchend needed for mobile devices 👍 Tested in storybook

mungodewar (approved):

PR #3670, opened 2024-11-14
[COR-481][BpkDrawer] Removing header border

TristanGrayCC:
- Removes divider below the header
- Fixes additionalClass param
- Corrects imports

- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests


packages/bpk-component-drawer/src/BpkDrawerContent-test.tsx
```diff
@@ -50,7 +50,7 @@ describe('BpkDrawerContent', () => {
     const { asFragment } = render(
       <BpkDrawerContent
         id="my-drawer"
-        className="my-classname"
+        additionalClassName="my-classname"
```
amburrrshand, line 53: Would it be possible to keep it as `className`, otherwise it will be a breaking change for consumers

TristanGrayCC: I get an error saying that className is a banned
`Prop className is forbidden. See https://github.com/Skyscanner/eslint-plugin-rules?tab=readme-ov-file#forbid-component-props`

This shouldn't impact consumers as the value is passed from BpkDrawer, which hasn't changed its prop names?

amburrrshand: Interesting - thanks for sharing this!

LexSwed: `BpkDrawerContent` should also be an internal component, right? Not exposed to the consumers that are using just `BpkDrawer`?

amburrrshand (approved):
Looks good!

amburrrshand (changes requested):
I've spoken to the team about this, and we can't pass classNames in any into Bpk components so the styles can't be manipulated from the outside

amburrrshand (approved):
Thanks for making these changes

PR #3696, opened 2024-12-17
[NO-JIRA][BpkModal] Add custom padding element to scrim util and modal

ss-tombrown:
This PR adds support to specify a custom padding element to Backpack Modal. 

When a modal is opened, padding is added to the body to prevent the page from horizontally shifting. However, some microsites (like Banana) are setup in a way that has no effect, for example when using an absolutely positioned content.

This addition allows passing in a custom element instead, so that the padding can be applied to the correct element, instead of the body.

- [x] Component `README.md`
- [x] Tests
- [ ] Accessibility tests


amburrrshand (approved):
Looks good to me!

PR #3076, opened 2023-11-08
BpkCardList

FireRedNinja:
This PR add the new component BpkCardList.

### Rail
image

### Grid
image

### Stack

Remember to include the following changes:

- [x] `README.md` (If you have created a new component)
- [x] Component `README.md`
- [x] Tests
- [x] Storybook examples created/updated
- [ ] For breaking changes or deprecating components/properties, migration guides added to the description of the PR. If the guide has large changes, consider creating a new Markdown page inside the component's docs folder and link it here

PR #3697, opened 2024-12-19
[BDES-96] [BpkCardList] Grid and Stack Layouts

FireRedNinja:

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### Grid with Header Button
image

### Grid

### DIFF
--- a/examples/bpk-component-card-list/examples.tsx
+++ b/examples/bpk-component-card-list/examples.tsx
@@ -16,16 +16,104 @@
  * limitations under the License.
  */
 
+import { useState } from "react";
+
+import BpkCard from "../../packages/bpk-component-card";
 import BpkCardList from "../../packages/bpk-component-card-list";
+import BpkImage from "../../packages/bpk-component-image";
+import BpkText, { TEXT_STYLES } from "../../packages/bpk-component-text/src/BpkText";
+
+import STYLES from "./exampless.module.scss";
+
+const DestinationCard = (i: number) => (
+  <BpkCard href="/" padded={false}>
+    <div className={STYLES['bpkdocs-consumer-level']}>
+      <BpkImage
+        aspectRatio={3000 / 1800}
+        altText="card image"
+        src="https://content.skyscnr.com/7adba3a46af3ca29695f96937d19fcf1/GettyImages-149127892.jpg?crop=960px:640px&quality=100"
+      />
+
+      <div className={STYLES['bpk-bottom']}>
+        <div className={STYLES['bpk-column']}>
+          <BpkText textStyle={TEXT_STYLES.heading4}>
+            {`Destination Name ${i}`}
+          </BpkText>
+          <BpkText>Country</BpkText>
+        </div>
+
+        <div className={STYLES['bpk-column']}>
+          <BpkText>Direct</BpkText>
+          <BpkText textStyle={TEXT_STYLES.heading4}>£100</BpkText>
+        </div>
+      </div>
+    </div>
+  </BpkCard>
+);
+
+type ExampleCard = typeof DestinationCard;
+
+const cards = (cardType: ExampleCard) => {
+  const cardList = [];
+  for (let i = 0; i < 14; i += 1) {
+    cardList.push(cardType(i));
+  }
+  return cardList;
+};
 
 const BasicExample = () => (
   <BpkCardList
     title="Must-visit spots"
     description="Check out these world-famous destinations perfect for visiting in spring."
     buttonText="Explore More"
-    buttonHref="https://www.skyscanner.net/"
-    onButtonClick={() => {}}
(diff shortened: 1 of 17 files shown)


examples/bpk-component-card-list/examples.tsx
```diff
@@ -16,16 +16,97 @@
...
   <BpkCardList
     title="Must-visit spots"
     description="Check out these world-famous destinations perfect for visiting in spring."
     buttonText="Explore More"
-    buttonHref="https://www.skyscanner.net/"
-    onButtonClick={() => {}}
+    buttonHref=""
+    onButtonClick={() => null}
+    cardList={cards(DestinationCard)}
+    layoutDesktop="grid"
```
anambl, line 74: I'd create some constants and export them for all these to avoid consumers having to hardcode these

examples/bpk-component-card-list/exampless.module.scss
anambl, line 1: small typo in the file name?

packages/bpk-component-card-list/src/BpkCardList.tsx
```diff
@@ -16,31 +16,45 @@
...
+    initiallyShownCards = DEFAULT_ITEMS,
+    layoutDesktop,
+    layoutMobile,
     onButtonClick,
     title,
   } = props;
 
-  const button = buttonText && (
-    <BpkButtonV2 onClick={onButtonClick} href={buttonHref}>{buttonText}</BpkButtonV2>
+  const button = buttonText && accessory !== 'button' && (
```
anambl, line 50: likewise with this - a constant would be useful to avoid consumers misusing the component due to a typo

packages/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.tsx
```diff
@@ -0,0 +1,103 @@
...
+    buttonText,
+    children,
+    expandText,
+    initiallyShownCards,
+    layout,
+    onButtonClick,
+  } = props;
+
+  let defaultInitiallyShownCards: number;
+  if (accessory === ACCESSORY_TYPES.Expand) {
```
anambl, line 42: I don't think this depends on the accessory. From the API table:
> 3 cards displayed by default, but you can change it.

packages/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.tsx
```diff
@@ -0,0 +1,103 @@
...
+  let accessoryContent;
+  if (accessory === ACCESSORY_TYPES.Expand) {
+    accessoryContent = (
+      <BpkExpand
+        showContent={showContent}
+        hideContent={hideContent}
+        collapsed={collapsed}
+        setCollapsed={setCollapsed}
+      >
+        {expandText || ''}
```
anambl, line 74: I wonder if there's a way in TS to ensure expandText is passed when accessory is expand 🤔 likewise for button and button text

packages/bpk-component-card-list/src/BpkExpand/BpkExpand.tsx
```diff
@@ -0,0 +1,67 @@
...
+}: ExpandProps) => {
+  const buttonIcon = collapsed ? (
+    <AlignedChevronDownIcon />
+  ) : (
+    <AlignedChevronUpIcon />
+  );
+  const buttonOnClick = () => {
+    if (collapsed) {
+      showContent();
+      setCollapsed(false);
```
anambl, line 47: I think it's better not to directly change the state here and instead pass a callback from the parent for the onClick functionality. In fact, I think `showContent` handles that already

packages/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.tsx
FireRedNinja: This is for the case when more than 3 cards are given but the user don't want to use an expand accessory. It would look confusing for the user as it would simply look like the component is broken with only 3 cards showing

packages/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.tsx
FireRedNinja: There is, but it is hard to understand and maintain ([my last attempt](https://github.com/Skyscanner/backpack/pull/3076/files#diff-903ad3bf2d2c9c01e2bd93ed01293027a4cff06bc81985d89cc29fe4a5aa281cR96-R100)). I thought it would be simpler. If its preferred to be more strict, then I can fall back to my last attempt.

PR #3699, opened 2025-01-02
[BpkCalendar] Accessibility roles

gert-janvercauteren:
# BpkCalendar

Add the appropriate 'roles' to different components to make the calendar more accessible and easier to navigate for people using screen readers

You can play test the screen reader and keyboard interaction in [the preview](https://backpack.github.io/storybook-prs/3699)

* Enabled grid navigation (Screen reader will announce that)
* Removed aria-hidden of disabled date cells, it would mess up the grid structure.

- [ ] Component `README.md`
- [x] Tests
- [x] Accessibility tests


packages/bpk-component-calendar/src/BpkCalendarDate.tsx
```diff
@@ -193,7 +193,6 @@ class BpkCalendarDate extends PureComponent<Props> {
         type="button"
         style={style}
         className={classNames.join(' ')}
-        aria-hidden={isBlocked}
```
anambl, line 196: I think this might be causing the blocked dates in the scrollable calendar to be read by the screen reader which doesn't seem to be happening on the main branch 🤔

anambl (approved):

packages/bpk-component-calendar/src/BpkCalendarDate.tsx
olliecurtis: Agree with this point, currently the calendar wouldn't allow a AT user to go to dates that are in the past and make them inaccessible so with this removal this now allows them to select previous dates

olliecurtis: I guess are we happy with users being able to get to that date and with it mentioning 'dimmed' this is ok?

anambl: Had a chat with Gert about this - it's not possible to keep this with the role of grid as it completely messes it up. Adding grid though brings a lot of benefits though like navigating up and down so even though you can now navigate through the blocked dates, it will be at most 30 days blocked, but you can just move down through the calendar so it's quick

gert-janvercauteren: Yes, I purposefully unblocked the dimmed dates as otherwise, Grid navigation wouldn't work correctly :(

olliecurtis (approved):

PR #3717, opened 2025-01-17
[GRAV-1370] [BpkOverlay] Add video overlay style

kirstybryce:
JIRA: https://skyscanner.atlassian.net/browse/GRAV-1370

The new homepage hero video ad POC requires a [new video overlay](https://www.figma.com/design/o7vYhyVecpNyC92vTDPNTf/Video-Moments?node-id=2742-4578&m=dev). This PR adds that new overlay type to `BpkOverlay` as `videoTop` and `videoBottom`. 

Link to Figma video overlay design: https://www.figma.com/design/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=27362-1451&m=dev 

| No overlay | Video Overlay |
| --- | --- |
|  | Screenshot 2025-01-21 at 15 17 57 |

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-overlay/src/BpkOverlay.tsx
+++ b/packages/bpk-component-overlay/src/BpkOverlay.tsx
@@ -41,6 +41,7 @@ export const OVERLAY_TYPES = {
   rightMedium: 'rightMedium',
   rightHigh: 'rightHigh',
   vignette: 'vignette',
+  videoOverlay: 'videoOverlay',
   off: 'off',
 } as const;
 
@@ -61,6 +62,7 @@ const overlayTypeClassSuffixes = {
   [OVERLAY_TYPES.rightMedium]: 'right-medium',
   [OVERLAY_TYPES.rightHigh]: 'right-high',
   [OVERLAY_TYPES.vignette]: 'vignette',
+  [OVERLAY_TYPES.videoOverlay]: 'video-overlay',
   [OVERLAY_TYPES.off]: 'off',
 } as const;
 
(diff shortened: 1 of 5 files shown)


packages/bpk-component-overlay/src/BpkOverlay.tsx
```diff
@@ -61,6 +62,7 @@ const overlayTypeClassSuffixes = {
   [OVERLAY_TYPES.rightMedium]: 'right-medium',
   [OVERLAY_TYPES.rightHigh]: 'right-high',
   [OVERLAY_TYPES.vignette]: 'vignette',
+  [OVERLAY_TYPES.videoOverlay]: 'video-overlay',
```
olliecurtis, line 65: We will need to have a `videoTop` and `videoBottom` overlay types looking at the figma design :)

kirstybryce: Hi, the reason I made it as one overlay is for 2 reasons:
- simplicity of having one overlay used for the homepage video component instead of 2
- since the top overlay is only taking up 40% of the height at the top of the overlay, and the bottom overlay is taking up 60% of the height at the bottom of the overlay, it seemed better to have a full overlay with those value since the current top/bottom overlay are taking up 100% of the height of the overlay. If I created a separate video top/bottom overlay they wouldn't be consistent with the other top/bottom overlays so it seemed better to have a separate video overlay like the vignette.
Do you think it would be better as top/bottom? Happy to change it if so :)

olliecurtis: Thanks for the response!

Checking the Backpack designs and with the designers, it would be expected to be two separate types of overlay for videos :) so shouldn't ever have both the top and bottom turned on at the same time.

If we would expect that to happen, then we should talk with design to get this changed at its design level to then reflect that into the code

kirstybryce: ok makes sense, will update :)

olliecurtis (approved):

Sybsw (approved):

PR #3725, opened 2025-01-21
[ST-6228] [BpkSlider] Pin react-slider to version that still supports React 17

dominicfraser:
`react-slider` currently errors when being used alongside Webpack and React 17.

```
Module not found: Error: Can't resolve 'react/jsx-runtime' in '.../banana/packages/webapp/node_modules/@radix-ui/react-slider/dist'
Did you mean 'jsx-runtime.js'?
BREAKING CHANGE: The request 'react/jsx-runtime' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
```

This is not fixed until React 18. At this point we still need to support React 17.

https://github.com/facebook/react/issues/20235#issuecomment-1095340542

This PR reverts to an exact version that still supports React 17. 

It also pins to exactly that version, via `npm i --registry=https://registry.npmjs.org/ @radix-ui/react-slider@1.1.2 --save-exact` preventing NPM silently updating the minor version when consumers do an install.

---

- [x] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/package-lock.json
+++ b/packages/package-lock.json
@@ -12,7 +12,7 @@
         "@floating-ui/react": "^0.26.12",
         "@popperjs/core": "^2.11.8",
         "@radix-ui/react-compose-refs": "^1.1.1",
-        "@radix-ui/react-slider": "^1.2.2",
+        "@radix-ui/react-slider": "1.1.2",
         "@react-google-maps/api": "^2.19.3",
         "@skyscanner/bpk-foundations-web": "^19.1.0",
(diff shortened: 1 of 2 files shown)


metalix2 (approved):
Looks good!

PR #3730, opened 2025-01-29
[QUAR-642] [BpkCardWrapper] Add more info button and body section to card wrapper

kirstybryce:
https://skyscanner.atlassian.net/browse/QUAR-642

Figma designs: https://www.figma.com/design/22unPHkDCzqJkd9rKzTiGY/T%26Cs-Ads?node-id=79-11694&t=MTrz7gFIDvIgP4k2-4

For the inline ads, we need to have a dropdown box for T&C in the ad card wrapper. We're currently use a [custom component](http://github.skyscannertools.net/skyscanner/banana/blob/main/packages/webapp-acorn/src/components/adverts/FlightsAdvert/ItineraryInlinePlus/ItineraryInlinePlusWrapper/ItineraryInlinePlusWrapper.tsx) to do this for desktop ads, but we need to add it to mobile ads as well. Instead of duplicating the custom wrapper into the mobile web ad components, we want to use the `BpkCardWrapper` component instead. To use it though we need to add the `more info` button and dropdown body element.

## `More Info`
Before Screenshot

## `Less Info`
After Screenshot

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-card/src/BpkCardWrapper-test.tsx
+++ b/packages/bpk-component-card/src/BpkCardWrapper-test.tsx
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { render,screen, fireEvent } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
 
 import { coreAccentDay } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
 
@@ -73,7 +73,7 @@ describe('BpkCardWrapper', () => {
           openBtnLabel: 'More Info',
           closeBtnLabel: 'Less Info',
         }}
-      />
+      />,
     );
 
     const button = screen.getByRole('button', { name: /more info/i });
@@ -81,5 +81,33 @@ describe('BpkCardWrapper', () => {
 
     expect(screen.getByText('More Info Content')).toBeInTheDocument();
   });
-});
 
+  it('hides card content when Less Info button is clicked', () => {
+    render(
+      <BpkCardWrapper
+        header={headerContent}
+        card={cardContent}
+        backgroundColor={coreAccentDay}
+        body={{
+          text: 'More Info Content',
+          openBtnLabel: 'More Info',
+          closeBtnLabel: 'Less Info',
+        }}
+      />,
+    );
+
+    const openbutton = screen.getByRole('button', { name: /more info/i });
+    fireEvent.click(openbutton);
+
+    expect(screen.getByText('More Info Content')).toBeInTheDocument();
+
+    const closeButton = screen.getByRole('button', { name: /less info/i });
+    fireEvent.click(closeButton);
+
+    // find element that contains the text "More Info Content" and get its closest ancestor with the class "bpk-card-wrapper--body"
+    const contentWrapper = screen.getByText('More Info Content').closest('.bpk-card-wrapper--body');
+
+    // Assert that the found element does NOT have the class "expanded", the content should be collapsed
+    expect(contentWrapper).not.toHaveClass('expanded');
+  });
+});
--- a/packages/bpk-component-card/src/BpkCardWrapper-test.tsx
+++ b/packages/bpk-component-card/src/BpkCardWrapper-test.tsx
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { render } from '@testing-library/react';
+import { render,screen, fireEvent } from '@testing-library/react';
 
 import { coreAccentDay } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
 
@@ -61,4 +61,25 @@ describe('BpkCardWrapper', () => {
     );
     expect(asFragment()).toMatchSnapshot();
   });
+
+  it('shows card content when More Info button is clicked', () => {
+    render(
+      <BpkCardWrapper
+        header={headerContent}
+        card={cardContent}
+        backgroundColor={coreAccentDay}
+        body={{
+          text: 'More Info Content',
+          openBtnLabel: 'More Info',
+          closeBtnLabel: 'Less Info',
+        }}
+      />
+    );
+
+    const button = screen.getByRole('button', { name: /more info/i });
+    fireEvent.click(button);
+
+    expect(screen.getByText('More Info Content')).toBeInTheDocument();
+  });
 });
+
--- a/packages/bpk-component-card/src/BpkCardWrapper.module.scss
+++ b/packages/bpk-component-card/src/BpkCardWrapper.module.scss
@@ -97,7 +97,6 @@
       align-items: flex-end;
 
       &--button {
-        cursor: pointer;
         all: unset;
       }
 
@@ -105,6 +104,7 @@
         display: flex;
         margin: tokens.bpk-spacing-md() tokens.bpk-spacing-base();
         align-items: center;
+        cursor: pointer;
 
         &--text {
           margin-inline-end: tokens.bpk-spacing-md();
@@ -113,6 +113,7 @@
     }
 
     &--link-text {
+      width: max-content;
       color: tokens.$bpk-private-button-link-normal-foreground-day;
       text-decoration: none;
 
--- a/packages/bpk-component-card/src/BpkCardWrapper.module.scss
+++ b/packages/bpk-component-card/src/BpkCardWrapper.module.scss
@@ -36,15 +36,15 @@
 
   &--content {
     position: relative;
+    transition: border-radius 0.25s ease-in-out;
     border-radius: tokens.$bpk-border-radius-md;
     background-color: tokens.$bpk-private-info-banner-default-day;
-    transition: border-radius 0.25s ease-in-out;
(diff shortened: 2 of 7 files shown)


hir06:
Can you add screenshots for the visual changes in PR description?

packages/bpk-component-card/src/BpkCardWrapper-test.tsx
```diff
@@ -61,4 +61,25 @@ describe('BpkCardWrapper', () => {
...
+          openBtnLabel: 'More Info',
+          closeBtnLabel: 'Less Info',
+        }}
+      />
+    );
+
+    const button = screen.getByRole('button', { name: /more info/i });
+    fireEvent.click(button);
+
+    expect(screen.getByText('More Info Content')).toBeInTheDocument();
```
hir06, line 82: can add another test case where the content is hidden when the card is collapsed!

packages/bpk-component-card/src/BpkCardWrapper.module.scss
```diff
@@ -20,8 +20,11 @@
 @use '../../unstable__bpk-mixins/borders';
 @use '../../unstable__bpk-mixins/cards';
 @use '../../unstable__bpk-mixins/radii';
+@use '../../unstable__bpk-mixins/utils';
 
 .bpk-card-wrapper {
+  padding: 0.1rem;
```
hir06, line 26: why give fixed number here? will it not impact the visual test cases of other consumers already using `bpk-card-wrapper`?

packages/bpk-component-card/src/BpkCardWrapper.tsx
```diff
@@ -31,27 +37,133 @@ type Props = {
...
   const classNames = getClassName('bpk-card-wrapper', className);
 
+  const [isBodyOpen, setIsBodyOpen] = useState(false);
+  const bodyRef = useRef<HTMLDivElement>(null);
+
+  const toggleExpand = () => {
+    setIsBodyOpen(!isBodyOpen);
+  };
+
+  const toggleLabel = isBodyOpen ? body?.closeBtnLabel : body?.openBtnLabel;
```
hir06, line 66: You will not need `?.` here with `body?.closeBtnLabel` and `body?.openBtnLabel` as you have marked them mandatory above. if you want to make them optional define them optional and give some default text.

packages/bpk-component-card/src/BpkCardWrapper.tsx
```diff
@@ -31,27 +37,133 @@ type Props = {
   className?: string | null;
   backgroundColor: string;
   header: ReactNode;
+  body?: {
+    text: string;
+    openBtnLabel: string;
```
hir06, line 42: are these optional or mandatory params?

packages/bpk-component-card/src/BpkCardWrapper-test.tsx
sharada1022: Add the test for the same. Please check.

packages/bpk-component-card/src/BpkCardWrapper.module.scss
sharada1022: yes, you are right. I’ve updated the padding to use the predefined spacing token.

packages/bpk-component-card/src/BpkCardWrapper.tsx
sharada1022: Since `body` is optional, it is not required when using the component. However, if body is provided, its properties—`text`, `openBtnLabel`, and `closeBtnLabel`—are mandatory. On the other hand, `link`, `linkText`, and `moreInfoBtnColor` are optional.

packages/bpk-component-card/src/BpkCardWrapper.tsx
sharada1022: The `?` ensures extra security in case `body` is not defined. It will only attempt to return `closeBtnLabel` or `openBtnLabel` if the `body` prop is provided. Removing it results in a type error.

hir06: Ah yep I misread it to be `body?.oprnBtnLabel?.` ignore it

hir06 (approved):
LGTM but please check for build failures and warnings and visual tests for consumer of the component

hir06 (changes requested):
I see some design conflict with expand version of the wrapper for eg. rounded corner for the body details are not present on desktop or mobile.

sharada1022:
> LGTM but please check for build failures and warnings and visual tests for consumer of the component

All the checks has been passed and visual test is picked up by Percy to run on CI.

sharada1022:
> I see some design conflict with expand version of the wrapper for eg. rounded corner for the body details are not present on desktop or mobile.

I checked with the design team, and they confirmed that we should maintain the full card look with rounded corners as long as the transition is smooth.

packages/bpk-component-card/src/BpkCardWrapper.module.scss
```diff
@@ -20,10 +20,12 @@
 @use '../../unstable__bpk-mixins/borders';
 @use '../../unstable__bpk-mixins/cards';
 @use '../../unstable__bpk-mixins/radii';
+@use '../../unstable__bpk-mixins/utils';
 
 .bpk-card-wrapper {
+  padding: tokens.bpk-spacing-sm() * 0.5;
+
   @include cards.bpk-card;
```
kirstybryce, line 28:
```suggestion
  @include cards.bpk-card;
  border-radius: tokens.$bpk-one-pixel-rem * 14;
```
Because you added `0.125rem` of padding to the card wrapper, we need to add that also to the border radius so that it's event around the inner card.


| Border radius calculation | Before & After |
| --- | --- |
|  <br/> In our case: <br/> `R1 = .75rem` from the inner card border radius. <br/> `D = .125rem` from the wrapper padding. <br/> So the wrapper border radius is `R2 = .875rem` | **Before (border-radius: .75rem)** <br/> Screenshot 2025-02-21 at 12 26 12 <br/> **After (border-radius: .875rem)** <br/> Screenshot 2025-02-21 at 12 34 45 |

kirstybryce (commented):
Just a small UI change needed, otherwise LGTM :)

hir06 (approved):
LGTM please check the pipeline failures.

packages/bpk-component-card/src/BpkCardWrapper.tsx
```diff
@@ -31,27 +37,127 @@ type Props = {
   className?: string | null;
   backgroundColor: string;
   header: ReactNode;
+  body?: {
```
olliecurtis, line 40: body as a name I think might be confusing to consumers, maybe we could call this `expandContent` so its clear where this would be.

wdyt?

packages/bpk-component-card/src/BpkCardWrapper.tsx
```diff
@@ -31,27 +37,127 @@ type Props = {
...
+  body,
   card,
   className = null,
   header,
 }: Props) => {
   const classNames = getClassName('bpk-card-wrapper', className);
 
-  return (
-    <CardContext.Provider value={{ elevated: false }}>
+  const [isBodyOpen, setIsBodyOpen] = useState(false);
```
olliecurtis, line 59: For this I would call this

```suggestion
  const [isExpanded, setIsExpanded] = useState(false);
```

packages/bpk-component-card/src/BpkCardWrapper.tsx
```diff
@@ -31,27 +37,127 @@ type Props = {
...
+  const bodyRef = useRef<HTMLDivElement>(null);
```
olliecurtis, line 60:
```suggestion
  const expandRef = useRef<HTMLDivElement>(null);
```

packages/bpk-component-card/src/BpkCardWrapper.tsx
```diff
@@ -31,27 +37,127 @@ type Props = {
...
       >
-        <div className={getClassName('bpk-card-wrapper--header')}>{header}</div>
-        <div className={getClassName('bpk-card-wrapper--content')}>{card}</div>
+        <BpkText textStyle={TEXT_STYLES.caption}>{toggleLabel}</BpkText>
+      </div>
+      {isBodyOpen ? <BpkIconChevronUp /> : <BpkIconChevronDown />}
+    </div>
+  );
+
+  const headerWithBodyDiv = body && (
```
olliecurtis, line 84: I would call this `expandableHeader`

packages/bpk-component-card/src/BpkCardWrapper.module.scss
```diff
@@ -69,4 +79,56 @@
       opacity: 1;
     }
   }
+
+  &--body {
```
olliecurtis, line 83:
```suggestion
  &--expand {
```

floracheng292:
After discussing with quar-gf, we decided to close this PR.

PR #3731, opened 2025-01-30
SDPLAT-26630: use GH_APP in workflows

olliecurtis:
Applying the same changes as https://github.com/Skyscanner/backpack-android/pull/2195 provided by Catalyst to use the Github App for CI


olliecurtis:
As the change needed to be merged to main for the label check to work, we overrided the settings so we can merge the change to unblock the CI from failing

arnaugm (approved):

PR #3732, opened 2025-01-31
[QUAR-497] [BpkInsetBanner] Add CTA button & popover

kirstybryce:
JIRA: https://skyscanner.atlassian.net/browse/QUAR-497

When the `BpkInsetBanner` component was created last year we added the CTA object prop accepting a text element. Now we need to add the CTA button and popover functionality so that we can add required legal info in the adverts using the inset banner.

Props added to `CallToAction` object prop:
- `popoverMessage`: message that will be displayed inside the popover dialog when the CTA button is clicked.
- `popoverPlacement`: where the popover dialog should appear when CTA button is clicked (ex. 'top', 'bottom', etc).
- `buttonCloseLabel`: text to display in the close button of the popover dialog.
- `buttonA11yLabel`: accessibility aria label for the CTA button.
- `popverLabel`: label of the popover dialog, and if `labelTitle` is true, title to display at the top of the popover dialog.
- `popoverId`: id of the popover dialog.
- `labelTitle`: if true, displays `popoverLabel` as popover title.
- `closeBtnIcon`: if true, displays close icon instead of `buttonCloseLabel` in popover close button.

| Without CTA Popover | With CTA Popover (`onLight` variant) | With CTA Popover (`onDark` variant) | 
| --- | --- | --- |
| Screenshot 2025-02-03 at 16 03 56 | Screenshot 2025-02-03 at 16 03 46 | Screenshot 2025-02-05 at 18 12 11

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests

### DIFF
--- a/packages/bpk-component-inset-banner/src/BpkInsetBanner-test.tsx
+++ b/packages/bpk-component-inset-banner/src/BpkInsetBanner-test.tsx
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { render } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
 
 import '@testing-library/jest-dom';
 
@@ -50,6 +50,28 @@ describe('BpkInsetBanner', () => {
     );
 
     expect(getByText('Sponsored')).toBeInTheDocument();
+  });
+
+  it('should render call to action popover text if provided', () => {
+    render(
+      <BpkInsetBanner
+        title="Lorem ipsum"
+        subheadline="Lorem ipsum dolor sit amet"
+        logo="https://content.skyscnr.com/m/7950ed6f30581485/Medium-Skyscanner-Vertical-White.png"
+        backgroundColor="#F55D42"
+        callToAction={{
+          text: 'Sponsored',
+          popoverMessage: 'Popover message',
+        }}
+        variant={VARIANT.onDark}
+      />,
+    );
+
+    const ctaButton = screen.getByTestId('adInfoBtn');
+
+    fireEvent.click(ctaButton);
+
+    expect(screen.getByText('Popover message')).toBeInTheDocument();
   });
 
   it('should render body if provided', () => {
--- a/packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
+++ b/packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
@@ -17,8 +17,12 @@
  */
 import { surfaceHighlightDay } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
 
+import InfoIcon from '../../bpk-component-icon/sm/information-circle';
+import BpkPopover from '../../bpk-component-popover/src/BpkPopover';
 import BpkText, { TEXT_STYLES } from '../../bpk-component-text/src/BpkText';
 import { cssModules } from '../../bpk-react-utils';
+
+import type { Placement } from '@floating-ui/react';
 
 import STYLES from './BpkInsetBanner.module.scss';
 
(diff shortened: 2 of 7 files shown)


packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -37,9 +41,16 @@ export type Props = {
...
     linkText?: string;
   };
-  // [TO-DO] In future iteration we will add a button to CTA section with popover functionality
   callToAction?: {
     text?: string;
+    popoverMessage?: string;
+    popoverPlacement?: Placement;
+    buttonCloseLabel?: string;
+    buttonA11yLabel?: string;
+    popverLabel?: string;
```
hir06, line 50: would it be nice to have popoverA11yLabel too?

hir06: I mean for popover

packages/bpk-component-inset-banner/src/BpkInsetBanner-test.tsx
```diff
@@ -52,6 +52,28 @@ describe('BpkInsetBanner', () => {
     expect(getByText('Sponsored')).toBeInTheDocument();
   });
 
+  it('should render call to action popover text if provided', () => {
```
MaxHun, line 55: should we also test that the popover can be closed properly?

MaxHun (approved):
left 1 comment, rest lgtm

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
kirstybryce: `buttonA11yLabel` is used to pass a value to `aria-label` on the button so I think that's sufficient, as `aria-labelledby` requires an id to be passed to refer to an element to be used as the label of the object. But with the way backpack components work in the consumer, the backpack component is abstracted and we don't see the IDs inside the backpack component. In this case I think having `buttonA11yLabel` is preferable :)

packages/bpk-component-inset-banner/src/BpkInsetBanner-test.tsx
kirstybryce: The popover close functionality is [tested in the popover component](https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-popover/src/BpkPopover-test.tsx#L172), so I think there's enough coverage for that there, and in inset banner we just want to test the popover message prop is passed correctly and shown in the popover dialog :)

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
hir06: I see so here the button would be info icon and the label typically could be description of popover?

kirstybryce: Button would be info icon, and if CTA text provided then text + info icon is in the button, so what would be read out by the screen reader is either `"aria-label text", button` or `"aria-label text", button, "cta text"`

hir06 (approved):
LGTM

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -86,13 +97,70 @@ const BpkInsetBanner = ({
...
+              label={callToAction?.popverLabel || ''}
+              placement={callToAction?.popoverPlacement || 'bottom'}
+              onClose={(e: {
+                stopPropagation: () => void;
+                preventDefault: () => void;
+              }) => {
+                e.stopPropagation();
+                e.preventDefault();
+              }}
+              aria-label={callToAction?.buttonA11yLabel}
```
olliecurtis, line 121: Will need to check this providing this alongside the label, as looking at the BpkPopover this might compete with `label` which also sets an A11y label for the popover based on the prop

olliecurtis: Just testing this and this property is having no effect to a screenreader as the `aria-labelledby` inside the popover is being prioritsed over it and using the `label` (which is the popover title) to be the aria for the user :) so this can be omitted

kirstybryce: Ok I'll remove it :)

olliecurtis (approved):

PR #3740, opened 2025-02-13
[LOOM-2050]: Updating docs for different build tools usage

olliecurtis:
This PR updates the docs for use of bpk-stylesheets based on the build tools that we are using in Skyscanner.

By default if you use the default JS export when using say vite it will not transpile the Sass and imports so intead you need to use the pre-transpiled code which exists under the `base.js` and `base.css` files, this is to ensure that in the output bundles the compiled css files have the removed `:global` declarations and operate at a base style


metalix2 (approved):
Looks good

PR #3742, opened 2025-02-14
[BD-9345][BpkPrice] Add a new icon property to BpkPrice

wuyq0808:
## Changes
1. Adds a new `icon` property to BpkPrice.
2. Includes small improvements to the HTML structure and CSS properties.

## Design

[Figma link](https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?m=auto&node-id=16468-17004&t=jCtrbc7zmVUe19YT-1)

image

## Screenshots

| Large | Small | Right aligned |
| -- | -- | -- |
| image | image | image |

- [x] Component `README.md`
- [x] Tests
- [ ] Accessibility tests


packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -111,34 +132,38 @@ const BpkPrice = (props: Props) => {
...
+              'bpk-price__icon',
+              !isAlignRight && 'bpk-price__spacing',
+            )}
+          >
+            {icon}
           </span>
         )}
+        {!isAlignRight && getTrailingTextNode()}
       </div>
+      {isAlignRight && getTrailingTextNode()}
```
wuyq0808, line 166: Reorganize the elements for better clarity: When aligned right, the trailing text appears on a new line and is wrapped in a `<div>`. Otherwise, it remains inline and is wrapped in a `<span>`.

packages/bpk-component-price/src/BpkPrice.module.scss
```diff
@@ -17,19 +17,14 @@
...
   flex-direction: column;
   color: tokens.$bpk-text-secondary-day;
 
   &--right {
-    text-align: right;
-
-    @include utils.bpk-rtl {
-      text-align: left;
-    }
+    align-items: flex-end;
```
wuyq0808, line 27: Replace `text-align` with `align-items` to simplify the code.

packages/bpk-component-price/src/BpkPrice.module.scss
```diff
@@ -53,9 +48,12 @@
     margin: 0 tokens.bpk-spacing-sm();
   }
 
-  &__column-container {
-    display: flex;
-    flex-direction: column;
+  &__main {
```
wuyq0808, line 51: Rename the css class name to better reflect the main part of the component.

packages/bpk-component-price/src/BpkPrice.module.scss
```diff
@@ -66,12 +64,11 @@
 
   &__spacing::after {
     content: '';
-    margin-right: tokens.bpk-spacing-sm();
+    margin-inline-end: tokens.bpk-spacing-sm();
```
wuyq0808, line 67: Use `margin-inline-end` for simplification.

packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -39,6 +40,7 @@ type Props = {
   leadingClassName: ?string,
   trailingText: ?string,
   previousPrice: ?string,
+  icon?: Node,
```
wuyq0808, line 43: If `?` is placed before the type `Node`, the IDE shows a warning about a missing required property in the example code.
I looked it up and found that placing `?` after the property name is the correct syntax in Flow. Is there any context on why we place the ? before the type?

anambl: I think it might be that when we first started using flow `?` was only supported before the type because looking at other flow'ed components, they all follow the same syntax. You're right though that the docs show it should be place after the property name.

packages/bpk-component-price/src/BpkPrice.js
anambl: why does it need to be moved outside the span if it's aligned right?

wuyq0808: If aligned right, the trailing text appears on a new row, so it should be placed in a separate `div` within the flex container instead of a `span`, which is more suited for inline text.

wuyq0808: What we had if aligned right:
```
<div>{leadingText}</div>
<div>
  <span>{price}</span>
  <span>{trailingText}</span>   // 2 span displayed in 2 rows
</div>
```

After change:
```
<div>{leadingText}</div>
<div>
  <span>{price}</span>
  <span>{icon}</span>  // 2 span displayed in the same row
</div>
<div>
  <span>{trailingText}</span>
</div>
```

Does this make it clearer?

anambl (approved):

PR #3751, opened 2025-02-19
[BD-9484][BpkInfoBannerExpandable] Fix an a11y issue

wuyq0808:
### Context
Fix a11y issue - element with button role can not be operable via keyboard

We identified this issue in hotels day-view:

### Changes
1. Change the toggle button from `<button>` to `<div>` and not clickable.
2. Change the expendable header to a `<button>`.
3. Style adjustments to maintain the original appearance.

- [ ] Component `README.md`
- [x] Tests
- [x] Accessibility tests

### DIFF
--- a/packages/bpk-component-info-banner/src/BpkInfoBanner-test.tsx
+++ b/packages/bpk-component-info-banner/src/BpkInfoBanner-test.tsx
@@ -55,6 +55,8 @@ describe('BpkInfoBanner', () => {
         message={message}
       />,
     );
-    expect(screen.getByRole('presentation')).toHaveClass('custom-banner-class-name');
+    expect(screen.getByText(message).parentElement?.parentElement).toHaveClass(
+      'custom-banner-class-name',
+    );
   });
 });
--- a/packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
+++ b/packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
@@ -92,13 +92,11 @@ type ToggleButtonProps = {
 };
 
 const ToggleButton = (props: ToggleButtonProps) => (
-    <div
-      className={getClassName('bpk-info-banner__toggle-button')}
-      title={props.label}
-    >
-      {props.expanded ? <CollapseIcon /> : <ExpandIcon />}
-    </div>
-  );
+  <div className={getClassName('bpk-info-banner__toggle-button')}>
+    {props.expanded ? <CollapseIcon /> : <ExpandIcon />}
+    <span className="visually-hidden">{props.label}</span>
+  </div>
+);
 
 type Props = CommonProps & {
   action?: ExpandableBannerAction;
@@ -150,7 +148,7 @@ const BpkInfoBannerInner = ({
   const dismissable = configuration === CONFIGURATION.DISMISSABLE;
   const showChildren = isExpandable && expanded;
 
-  const sectionClassNames = getClassName(
+  const classNames = getClassName(
     'bpk-info-banner',
    `bpk-info-banner--${type}`,
    `bpk-info-banner--style-${style}`,
@@ -168,8 +166,8 @@ const BpkInfoBannerInner = ({
 
   const BannerHeader = isExpandable ? 'button' : 'div';
 
-  // Disabling 'click-events-have-key-events and interactive-supports-focus' because header element is not focusable.
-  // ToggleButton is focusable and works for this.
+  const childrenContainerId = 'children-container';
+
   return (
     <AnimateAndFade
       animateOnEnter={animateOnEnter}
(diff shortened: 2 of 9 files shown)


packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
```diff
@@ -189,8 +178,12 @@ const BpkInfoBannerInner = ({
       {...rest}
     >
       <section className={sectionClassNames} role="presentation">
-        <div
+        <BannerHeader
           role={isExpandable ? 'button' : undefined}
```
gert-janvercauteren, line 182: role button is not needed on a <button>, you might want to set type="button" when expandable :D

packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
```diff
@@ -91,23 +91,14 @@ type ToggleButtonProps = {
...
-
-  return (
-    <button
-      type="button"
+const ToggleButton = (props: ToggleButtonProps) => (
+    <div
       className={getClassName('bpk-info-banner__toggle-button')}
-      aria-label={props.label}
-      aria-expanded={props.expanded}
       title={props.label}
```
gert-janvercauteren, line 97: We should remove the 'title' now that it's no longer a button

packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
```diff
@@ -91,23 +91,14 @@ type ToggleButtonProps = {
...
     >
-      <div className={classNames}>
-        {props.expanded ? <CollapseIcon/> : <ExpandIcon/> }
-      </div>
-    </button>
+      {props.expanded ? <CollapseIcon /> : <ExpandIcon />}
```
gert-janvercauteren, line 99: We can add a <span className="visually-hidden">Icon text</span> or add altText to the icon to have 'view more' added.

packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
```diff
@@ -189,8 +178,11 @@ const BpkInfoBannerInner = ({
       {...rest}
     >
       <section className={sectionClassNames} role="presentation">
-        <div
-          role={isExpandable ? 'button' : undefined}
+        <BannerHeader
+          aria-label={isExpandable ? toggleButtonLabel : undefined}
```
gert-janvercauteren, line 182: We shouldn't set the aria-label here, as it will override the visible text. (Look at the prev. comment)

Ideally the button screen reader should read:
`{visible text}, {toggleButtonLabel}, <expanded|collapsed>, button` 

We also need to review the 'aria-controls' for this expand/collapse behaviour:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

gert-janvercauteren (commented):
Reviewed with keyboard and screen reader. Let me know if I can support on testing

packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
```diff
@@ -189,8 +178,11 @@ const BpkInfoBannerInner = ({
       {...rest}
     >
       <section className={sectionClassNames} role="presentation">
```
gert-janvercauteren, line 180: nit: This probably can be a div, and remove the role :D

wuyq0808: Yes, removed it.

packages/bpk-component-info-banner/src/BpkInfoBannerInner.tsx
wuyq0808: Removed.

packages/bpk-component-info-banner/src/withBannerAlertState-test.tsx
```diff
@@ -134,9 +134,7 @@ describe('withBannerAlertState(BpkInfoBannerExpandable)', () => {
       </BpkInfoBannerExpandableState>,
     );
 
-    const expandButton = screen.getByRole('button', { name: 'View more' });
-
-    await userEvent.click(expandButton);
```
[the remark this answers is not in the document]
wuyq0808: Thanks for reviewing! Yes you're right. There is still a button with the role, but 'View more' no longer matches. It'd work with just removing `{ name: 'View more' }` but it didn't occur to me :)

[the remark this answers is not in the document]
mungodewar: The `getByRole` might be slightly more [robust to future changes](https://testing-library.com/docs/queries/about#priority)

packages/bpk-component-info-banner/src/BpkInfoBanner-test.tsx
```diff
@@ -55,6 +55,8 @@ describe('BpkInfoBanner', () => {
         message={message}
       />,
     );
-    expect(screen.getByRole('presentation')).toHaveClass('custom-banner-class-name');
+    expect(screen.getByText(message).parentElement?.parentElement).toHaveClass(
+      'custom-banner-class-name',
+    );
```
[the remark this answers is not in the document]
wuyq0808: That's true 👍 I considered querying by class and asserting its existence, but it didn’t account for the DOM structure. 
I think this way may fail more easily than before, but it doesn’t offer less protection.

packages/bpk-component-info-banner/src/withBannerAlertState-test.tsx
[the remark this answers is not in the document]
wuyq0808: Oh thanks! I'll push a new commit.

mungodewar (approved):
LGTM

PR #3770, opened 2025-03-17
[GOOSE-148][BpkSegmentedControl] Refactor for use by Flights Search Controls

ss-tombrown:
[JIRA](https://skyscanner.atlassian.net/browse/GOOSE-148)

### What's changed

This PR contains changes required for the Segmented Control component to be used by the Flights Search Controls date switcher.

Specifically, this PR...

- Switches CSS Flex with CSS Grid so that the tab items maintain equal widths when used and centred within a flex container.
- Removes style `text-overflow` from control item, as it was agreed that consumers should be in charge of how the content is displayed within each item, and not enforce via this component.
- A11y fixes including ([context](https://skyscanner.slack.com/archives/C0JHPDSSU/p1742283184632619?thread_ts=1742227344.368179&cid=C0JHPDSSU)):
  - Removal of `radiogroup` role. This is because the component does not behave as would be expected for radio group so removing for now until it's decided on a better approach
  - Fix to show focus indicator when navigating the component via keyboard

# Screenshots / Recordings

#### Custom button story

N/A

#### Switch to grid

Screenshot 2025-03-20 at 14 52 36

#### Custom button story (FSC use case)

https://github.com/user-attachments/assets/dcc7bf3b-eaab-4f2e-b104-30e8b9ee1452

#### Switch to grid

Screenshot 2025-03-20 at 14 53 13

#### Focus indicator fix

https://github.com/user-attachments/assets/a1d17150-ecea-47c5-aec5-8c1fb7a585fa

- [ ] Component `README.md`
- [ ] Tests
- [x] Accessibility tests

### DIFF
--- a/packages/bpk-component-segmented-control/src/BpkSegmentedControl.module.scss
+++ b/packages/bpk-component-segmented-control/src/BpkSegmentedControl.module.scss
@@ -38,8 +38,6 @@
   min-height: tokens.bpk-spacing-xl();
   padding: tokens.bpk-spacing-md() tokens.bpk-spacing-base();
   border: none;
-  text-overflow: ellipsis;
-  white-space: nowrap;
   cursor: pointer;
   overflow: hidden;
 
--- a/packages/bpk-component-segmented-control/src/BpkSegmentedControl.module.scss
+++ b/packages/bpk-component-segmented-control/src/BpkSegmentedControl.module.scss
@@ -39,7 +39,6 @@
   min-height: tokens.bpk-spacing-xl();
   padding: tokens.bpk-spacing-md() tokens.bpk-spacing-base();
   border: none;
-  text-align: center;
   text-overflow: ellipsis;
   white-space: nowrap;
   cursor: pointer;
--- a/packages/bpk-component-segmented-control/src/BpkSegmentedControl.module.scss
+++ b/packages/bpk-component-segmented-control/src/BpkSegmentedControl.module.scss
@@ -23,8 +23,9 @@
 @use '../../unstable__bpk-mixins/shadows';
 
 .bpk-segmented-control-group {
-  display: flex;
-  flex-wrap: nowrap;
+  display: grid;
+  grid-auto-columns: 1fr;
+  grid-auto-flow: column;
   overflow: hidden;
 
   @include radii.bpk-border-radius-sm;
@@ -37,9 +38,10 @@
 .bpk-segmented-control {
   min-height: tokens.bpk-spacing-xl();
   padding: tokens.bpk-spacing-md() tokens.bpk-spacing-base();
-  flex: 1;
   border: none;
+  text-align: center;
   text-overflow: ellipsis;
+  white-space: nowrap;
   cursor: pointer;
   overflow: hidden;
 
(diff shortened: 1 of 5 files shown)


packages/bpk-component-segmented-control/src/BpkSegmentedControl.module.scss
```diff
@@ -37,9 +38,10 @@
 .bpk-segmented-control {
   min-height: tokens.bpk-spacing-xl();
   padding: tokens.bpk-spacing-md() tokens.bpk-spacing-base();
-  flex: 1;
   border: none;
+  text-align: center;
```
Supremeyh, line 42: it seems user agent stylesheet has default set text-align to center, do we need rewrite here?

image

Faye-Xiao:
NIt:  Is it possible to add a new property to maintain these two text display modes? 🤔  I feel like, for FSC, the refactored version is what it needed, while the previous use case remains useful.

ss-tombrown:
> NIt: Is it possible to add a new property to maintain these two text display modes? 🤔 I feel like, for FSC, the refactored version is what it needed, while the previous use case remains useful.

@Faye-Xiao

After speaking with design, my understanding was that the current text-wrapping behaviour is incorrect for all existing use cases. The switch to grid is non-changing, and ensures the items remain equal.

Therefore, in line with the YAGNI principle, I recommend not placing this behind a prop, as there is no known use case for the current behavior.

FireRedNinja (approved):
LGTM from Goose 🪿

Faye-Xiao:
> @Faye-Xiao
> 
> After speaking with design, my understanding was that the current text-wrapping behaviour is incorrect for all existing use cases. The switch to grid is non-changing, and ensures the items remain equal.
> 
> Therefore, in line with the YAGNI principle, I recommend not placing this behind a prop, as there is no known use case for the current behavior.

Hi @ss-tombrown thank you for the message.
The reason I raised that question is I noticed a use case for `BpkSegmentedControl` on the Acorn day view page. You can find it here: https://github.skyscannertools.net/skyscanner/banana/blob/1fbe342366f6f6e231e2d74dafa4f8b18b3d3579/packages/webapp/src/components/FqsTabs/FqsTabs.tsx#L112. 

I suppose this is another use case that needs to display all relevant information, as both the price and flight duration are essential details for travellers. I guess it might result in a visual change in this market when using this refactor. Just double-check if the refactor is suitable for this use case. 
- mWeb 
Screenshot 2025-03-19 at 11 37 41

- edge case
Screenshot 2025-03-19 at 11 38 19

ss-tombrown:
@Faye-Xiao 

After discussion with design, they agreed it's probably too much of an edge case to be a major concern, however, we also agreed it's best to let the consumer decide on wrapping behaviour like happens in banana [here](https://github.skyscannertools.net/skyscanner/banana/blob/55de13cdbc73187e6551f90189ec93d68712f5ae/packages/webapp/src/components/FqsTabs/FqsTabs.scss#L30C47-L30C77).

Given that it's possible to pass in react components with their own styles, this means removing text wrapping-related styles from this component, so I've removed both the `text-overflow` and `white-space`styles from the control.

Faye-Xiao (approved):
LGTM

Supremeyh:
Hi Tom, basically lgtm, it seems there are some changes since last review, could you also update PR description and snapshots after the change, or we might consider add override example in [stories](https://github.com/Skyscanner/backpack/blob/main/examples/bpk-component-segmented-control/stories.tsx) as well/

Supremeyh (approved):
lgtm

PR #3771, opened 2025-03-20
[MAT-2306][BpkDrawer] Update BpkDrawer by allowing dynamic width and animation

Evelyn-Xiang:
[[MAT-2306]](https://skyscanner.atlassian.net/browse/MAT-2306)
[BpkDrawer](https://www.skyscanner.design/latest/components/drawer/web-QAxL5e0N) needs a code contribution:
- Any width can be applied by consumers. This essentially means removing the 400px max-width and allowing any width.
- Update transition into duration-base for both entering and exiting

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

[MAT-2306]: https://skyscanner.atlassian.net/browse/MAT-2306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/bpk-component-drawer/src/BpkDrawer.tsx
+++ b/packages/bpk-component-drawer/src/BpkDrawer.tsx
@@ -21,15 +21,12 @@
 import type { ReactNode } from 'react';
 import { useState, useEffect } from 'react';
 
-import type { animations } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
-
 import { Portal, isDeviceIpad, isDeviceIphone } from '../../bpk-react-utils';
 import { withScrim } from '../../bpk-scrim-utils';
 
 import BpkDrawerContent from './BpkDrawerContent';
 
 const BpkScrimDrawerContent = withScrim(BpkDrawerContent);
-type ValueOf<T> = T[keyof T];
 
 export type Props = {
   id: string,
@@ -44,7 +41,6 @@ export type Props = {
   title: string,
   getApplicationElement: () => HTMLElement | null,
   width?: string,
-  transitionDuration?: ValueOf<typeof animations>,
   renderTarget?: null | HTMLElement | (() => null | HTMLElement),
   dialogRef?: (ref: HTMLElement | null | undefined) => void,
   className?: string,
@@ -76,7 +72,6 @@ const BpkDrawer = ({
   padded = true,
   renderTarget = null,
   title,
-  transitionDuration,
   width,
 }: Props) =>  {
 
@@ -106,7 +101,6 @@ const BpkDrawer = ({
           closeLabel={closeLabel || ""}
           closeText={closeText}
           width={width}
-          transitionDuration={transitionDuration}
           // eslint-disable-next-line @skyscanner/rules/forbid-component-props
           className={className}
           contentClassName={contentClassName}
--- a/packages/bpk-component-drawer/src/BpkDrawer.tsx
+++ b/packages/bpk-component-drawer/src/BpkDrawer.tsx
@@ -21,12 +21,15 @@
 import type { ReactNode } from 'react';
 import { useState, useEffect } from 'react';
 
+import type { animations } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
+
 import { Portal, isDeviceIpad, isDeviceIphone } from '../../bpk-react-utils';
 import { withScrim } from '../../bpk-scrim-utils';
 
 import BpkDrawerContent from './BpkDrawerContent';
 
 const BpkScrimDrawerContent = withScrim(BpkDrawerContent);
+type ValueOf<T> = T[keyof T];
 
 export type Props = {
   id: string,
@@ -40,6 +43,8 @@ export type Props = {
   ) => void;
   title: string,
   getApplicationElement: () => HTMLElement | null,
+  width?: string,
+  transitionDuration?: ValueOf<typeof animations>,
   renderTarget?: null | HTMLElement | (() => null | HTMLElement),
   dialogRef?: (ref: HTMLElement | null | undefined) => void,
   className?: string,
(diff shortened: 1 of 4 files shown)


packages/bpk-component-drawer/src/BpkDrawerContent.tsx
```diff
@@ -114,6 +119,11 @@ const BpkDrawerContent = ({
           role="dialog"
           key="dialog"
           aria-labelledby={headingId}
+          style={{
+            // @ts-expect-error TS is reporting this incorrectly as styles are valid
+            '--dynamic-width': width,
+            '--duration': transitionDuration,
```
lynn-sc, line 125: Could you take a screenshot of what it looks like when neither width nor transitionDuration is passed in? I'm wondering if we need to give it a default value.

Evelyn-Xiang: > Could you take a screenshot of what it looks like when neither width nor transitionDuration is passed in? I'm wondering if we need to give it a default value.

Hi Lynn, it looks like this. Seems no need to add a default value for width but not sure if a default transitionDuration is necessary.

https://github.com/user-attachments/assets/1fc17234-063c-491b-abcd-50c9eaead33a

Evelyn-Xiang: > Could you take a screenshot of what it looks like when neither width nor transitionDuration is passed in? I'm wondering if we need to give it a default value.

If there is no content, it will look like this screenshot. But if it has content, the content will stretch the width.
Screenshot 2025-03-20 at 16 01 24

https://github.com/user-attachments/assets/994a736f-8517-4db8-9cad-baf8174f8390

lynn-sc: Because your video and screenshots do not look the same as the screenshots below, so I need you to confirm whether the drawer after not passing these two values ​​is the same as the original one?
image

Evelyn-Xiang: @lynn-sc , no they are not the same as the original one if not passing default values. Because the max-width is deleted as required, it looks a bit wider.

lynn-sc (approved):
LGTM

felix-luo-sc (approved):
LGTM

packages/bpk-component-drawer/src/BpkDrawer.tsx
```diff
@@ -40,6 +40,7 @@ export type Props = {
   ) => void;
   title: string,
   getApplicationElement: () => HTMLElement | null,
+  width?: string,
```
Supremeyh, line 43: since width is optional, do we need a default value in case there are no width passed through?

Evelyn-Xiang: > since width is optional, do we need a default value in case there are no width passed through?

Yes, it is added here. 
Screenshot 2025-03-21 at 11 11 13

packages/bpk-component-drawer/src/BpkDrawer.tsx
```diff
@@ -71,6 +72,7 @@ const BpkDrawer = ({
   padded = true,
   renderTarget = null,
   title,
+  width,
```
Faye-Xiao, line 75: Hi @Evelyn-Xiang I suppose the default value @Supremeyh [mentioned](https://github.com/Skyscanner/backpack/pull/3771/files#r2006774110) is supposed to be here.  
According to the code, `BpkDrawer` has an optional `width` without a default value, and passes it to the `BpkScrimDrawerContent ` while the `width` in `BpkScrimDrawerContent` is also optional and has no default value. 
Therefore, for the edge case, `width` is not given, maybe `undefined`, looks like we would have the below values
```
 style={{
            // @ts-expect-error TS is reporting this incorrectly as styles are valid
            '--dynamic-width': undefined,
 }}
```
I'm curious about what the CSS below is in production. 🤔 Correct me if I missed something. 

```
width: var(--dynamic-width, 90%);  -> ?
```

Faye-Xiao:
**Not blocker**: 
Just to clarify, there is no visual test case for BpkDrawer in the Storybook. I'm not sure who the main contributor to this component is. According to the [backpack contribution](https://github.com/Skyscanner/backpack/blob/main/CONTRIBUTING.md) guide, every component should include a visual test story case for regression testing.

packages/bpk-component-drawer/src/BpkDrawerContent.tsx
```diff
@@ -114,6 +116,10 @@ const BpkDrawerContent = ({
           role="dialog"
           key="dialog"
           aria-labelledby={headingId}
+          style={{
+            // @ts-expect-error TS is reporting this incorrectly as styles are valid
+            '--dynamic-width': width,
```
Supremeyh, line 121: maybe it can explicitly define `CSSProperties` for style, rather than suppress the typescript complaint, eg

```
import type { CSSProperties } from 'react';

 style={
          {
              '--dynamic-width': width,
            } as CSSProperties
          }
```

packages/bpk-component-drawer/src/BpkDrawer.tsx
Evelyn-Xiang: Thanks for clarifying!
`width: var(--dynamic-width, 90%);  -> ?` means if --dynamic-width is undefined, width will use the default value 90%.

But I have updated the default value to the position you mentioned. For better consistency and be visible in the documentation

packages/bpk-component-drawer/src/BpkDrawerContent.tsx
Evelyn-Xiang: > maybe it can explicitly define `CSSProperties` for style, rather than suppress the typescript complaint, eg
> 
> ```
> import type { CSSProperties } from 'react';
> 
>  style={
>           {
>               '--dynamic-width': width,
>             } as CSSProperties
>           }
> ```


Thanks! have updated

Faye-Xiao (approved):
LGTM.

Supremeyh (approved):
lgtm

PR #3772, opened 2025-03-20
[PANGOO-2946][BpkPopover]Fix BpkDatepicker onClose

senzg:
This ticket is originally raised for fixing BpkDatepicker issues while eventually found it is the problem of BpkPopover. See more on:  [JIRA Link] https://skyscanner.atlassian.net/browse/PANGOO-2946?atlOrigin=eyJpIjoiNzJkYmM2Y2U3MDU0NGEwN2EwZjEzODQ5MTllMDgwYWMiLCJwIjoiaiJ9

In this ticket we fix the bug that bpkPopover state isOpen left true on close

- onClose function is not called when clicking region outside of the popover page

- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-drawer/src/BpkDrawer.tsx
+++ b/packages/bpk-component-drawer/src/BpkDrawer.tsx
@@ -40,6 +40,7 @@ export type Props = {
   ) => void;
   title: string,
   getApplicationElement: () => HTMLElement | null,
+  width?: string,
   renderTarget?: null | HTMLElement | (() => null | HTMLElement),
   dialogRef?: (ref: HTMLElement | null | undefined) => void,
   className?: string,
@@ -71,6 +72,7 @@ const BpkDrawer = ({
   padded = true,
   renderTarget = null,
   title,
+  width = '90%',
 }: Props) =>  {
 
   const [isDrawerShown, setIsDrawerShown] = useState(true);
@@ -98,6 +100,7 @@ const BpkDrawer = ({
           dialogRef={dialogRef}
           closeLabel={closeLabel || ""}
           closeText={closeText}
+          width={width}
           // eslint-disable-next-line @skyscanner/rules/forbid-component-props
           className={className}
           contentClassName={contentClassName}
--- a/packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
+++ b/packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
@@ -51,6 +51,7 @@ export type Props = {
     popoverId?: string;
     labelTitle?: boolean;
     closeBtnIcon?: boolean;
+    zIndexCustom?: number;
   };
   logo?: string;
   subheadline?: string;
@@ -121,11 +122,12 @@ const BpkInsetBanner = ({
               closeButtonText={callToAction?.buttonCloseLabel}
               closeButtonIcon={callToAction?.closeBtnIcon}
               labelAsTitle={callToAction?.labelTitle}
+              zIndexValue={callToAction?.zIndexCustom}
               target={
                 <button
                   aria-label={callToAction?.buttonA11yLabel}
                   className={getClassName('bpk-inset-banner--cta-button')}
-                  data-testid="adInfoBtn"
+                  data-testid="ctaBtn"
                   aria-hidden="false"
                   type="button"
                 >
--- a/packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
+++ b/packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
@@ -115,9 +115,11 @@ const BpkInsetBanner = ({
               onClose={(e: {
                 stopPropagation: () => void;
                 preventDefault: () => void;
-              }) => {
-                e.stopPropagation();
-                e.preventDefault();
+              } | null) => {
+                if (e) {
+                  e.stopPropagation();
+                  e.preventDefault();
+                }
               }}
               closeButtonText={callToAction?.buttonCloseLabel}
               closeButtonIcon={callToAction?.closeBtnIcon}
--- a/packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
+++ b/packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
@@ -115,11 +115,9 @@ const BpkInsetBanner = ({
               onClose={(e: {
                 stopPropagation: () => void;
                 preventDefault: () => void;
-              } | null) => {
-                if (e) {
-                  e.stopPropagation();
-                  e.preventDefault();
-                }
+              } | null) => {         
+                  e?.stopPropagation();
+                  e?.preventDefault();
               }}
               closeButtonText={callToAction?.buttonCloseLabel}
               closeButtonIcon={callToAction?.closeBtnIcon}
--- a/packages/bpk-component-popover/src/BpkPopover.tsx
+++ b/packages/bpk-component-popover/src/BpkPopover.tsx
@@ -162,6 +162,13 @@ const BpkPopover = ({
 
   const click = useClick(context);
   const dismiss = useDismiss(context);
+
+  useEffect(() => {
+    if (!isOpenState && onClose) {
+      onClose(null as unknown as SyntheticEvent<HTMLButtonElement>, { source: EVENT_SOURCES.CLOSE_BUTTON });
+    }
+  }, [isOpenState, onClose]);
+
   const hover = useHover(context, {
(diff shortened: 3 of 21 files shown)


packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -115,9 +115,11 @@ const BpkInsetBanner = ({
...
                 stopPropagation: () => void;
                 preventDefault: () => void;
-              }) => {
-                e.stopPropagation();
-                e.preventDefault();
+              } | null) => {
+                if (e) {
+                  e.stopPropagation();
+                  e.preventDefault();
+                }
```
Supremeyh, line 122: it could simplify to `e?.stopPropagation(); e?.preventDefault()`

packages/bpk-component-popover/src/BpkPopover.tsx
```diff
@@ -170,6 +170,12 @@ const BpkPopover = ({
     }),
   });
 
+  useEffect(() => {
+    if (!isOpenState && onClose) {
+      onClose(null, { source: EVENT_SOURCES.CLOSE_BUTTON });
```
Supremeyh, line 175: seems the source could be `CLOSE_BUTTON` or `CLOSE_LINK`, is there reason only set it to `CLOSE_BUTTON`?

Faye-Xiao: I'm not quite sure if this can resolve the issue with component state mismatches in a real production scenario. 

I encountered this problem previously with the CE budget filter feature. I discovered that when we use `BpkPopover` in a controlled component, the state of the parent component cannot be updated due to the unmounting issue of `BpkPopover`. This issue arises from the rendering behaviour of its wrapper, `FloatingPortal`.

If we look at the actual DOM changes, we see that a DOM element with a dynamic ID is created when the popover is opened. However, when the close button is clicked, the wrapper DOM is removed immediately, preventing `BpkPopover` from executing the `onClose` state-changing function.

Screenshot 2025-03-21 at 14 56 56

You can add some ·console.log· in each place to test the code execute steps.

Faye-Xiao: See related notes from https://skyscanner.atlassian.net/wiki/spaces/PANGOO/pages/1275366197/Backpack+Component+Pain+Point+Feedback

senzg: So the origin purpose is to fix the bug in BpkDatepicker, and eventually I found it is because onClose is not called if we don't close it through the close button or the close link. This PR is the synchronize the state through activating the onClose function once the popover's own state changes, I test it in BpkDatepicker which shows it works

packages/bpk-component-drawer/src/BpkDrawer.tsx
```diff
@@ -72,7 +71,6 @@ const BpkDrawer = ({
   padded = true,
   renderTarget = null,
   title,
-  width = '90%',
```
Supremeyh, line 75: it seems the changes has been reverted after merge main branch?
https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-drawer/src/BpkDrawer.tsx#L75

packages/bpk-component-popover/src/BpkPopover.tsx
senzg: > seems the source could be `CLOSE_BUTTON` or `CLOSE_LINK`, is there reason only set it to `CLOSE_BUTTON`?

Hi Ezreal, thanks for your reminding! After a further look at it, I realized there is indeed some issues relates to it, and decided to change the way to implement the logic. That is to remove the `useEffect` bloc, change the function inside `useFloating` hook and add a new Event Name. In this way, we can correctly call onClose and specify the source of calling the function as `Click_outside`, separating from button click and link click

senzg: > I'm not quite sure if this can resolve the issue with component state mismatches in a real production scenario.
> 
> I encountered this problem previously with the CE budget filter feature. I discovered that when we use `BpkPopover` in a controlled component, the state of the parent component cannot be updated due to the unmounting issue of `BpkPopover`. This issue arises from the rendering behaviour of its wrapper, `FloatingPortal`.
> 
> If we look at the actual DOM changes, we see that a DOM element with a dynamic ID is created when the popover is opened. However, when the close button is clicked, the wrapper DOM is removed immediately, preventing `BpkPopover` from executing the `onClose` state-changing function.
> 
> Screenshot 2025-03-21 at 14 56 56
> 
> You can add some ·console.log· in each place to test the code execute steps.

May I ask it is clicking the close button that prevents `BpkPopover` or it is clicking outside the popover box? Or both? 

Currently there is indeed some issues with outside clicking, but if clicking the button also cause the problem, that's another story.

Faye-Xiao: > So the origin purpose is to fix the bug in BpkDatepicker, and eventually I found it is the problem of BpkPopover, and it is because in Popover, onClose is not called if we don't close it through the close button or the close link. This PR is to synchronize the states of BpkPopover and its parent through activating the onClose function once the popover's own state changes (and only if it is going to close for sure), I test it in BpkDatepicker which shows it works

OK. Let's fix this case first. For the case I mentioned, I'll create a ticket for that later.

senzg: > > So the origin purpose is to fix the bug in BpkDatepicker, and eventually I found it is the problem of BpkPopover, and it is because in Popover, onClose is not called if we don't close it through the close button or the close link. This PR is to synchronize the states of BpkPopover and its parent through activating the onClose function once the popover's own state changes (and only if it is going to close for sure), I test it in BpkDatepicker which shows it works
> 
> OK. Let's fix this case first. For the case I mentioned, I'll create a ticket for that later.

Maybe I can help with this issue since I have had the context relates to it😄

packages/bpk-component-drawer/src/BpkDrawer.tsx
senzg: Can I just delete the change from github(right here)

senzg: Hi, the clash is resolved

packages/bpk-component-popover/src/BpkPopover.tsx
```diff
@@ -143,14 +144,22 @@ const BpkPopover = ({
...
+      onClose(null, { source: EVENT_SOURCES.CLOSE_OUTSIDE });
     }
   }, [isOpen]);
 
   const arrowRef = useRef(null);
 
+  const onFloatingChange = (isFloatingOpen: boolean) => {
+    setIsOpenState(isFloatingOpen);
+    if (!isFloatingOpen && onClose) {
+      onClose(null, { source: EVENT_SOURCES.CLOSE_OUTSIDE});
```
Supremeyh, line 156: it seems when isOpenState changed to false, onClose would called twice, is that expected?

senzg: Since I've delete the useEffect block I wrote, it won't call the onClose twice I think. May I ask where will call it for the second time?

senzg: > it seems when isOpenState changed to false, onClose would called twice, is that expected?

I see. It is very wired to call useEffect after setIsOpenState(false), which calls setIsOpenState(false) again. Somehow the state pass to the parent then pass down to BpkPopover again, not a good movement

senzg: > it seems when isOpenState changed to false, onClose would called twice, is that expected?

I see. I am going to delete the onClose call from the useEffect block, in this way, I will assume the parent who use BpkPopover will call the onClose function itself, and solely changing the isOpen prop without calling onClose from the parent is not allow, which will cause a bug. But that make more sense since it is out of BpkPopover's jurisdiction. In another word, it makes sense not to care about the careless user (parent)

senzg: > it seems when isOpenState changed to false, onClose would called twice, is that expected?

For example, in `BpkDatepicker`, `onClose` and `onOpen` will be call if `isOpen` state is changed. Any parent use this `BpkPopover` should follow this rule. Do you think it is a better choice?
image

senzg: onClose is deleted from the useEffect block

Supremeyh (approved):

PR #3804, opened 2025-04-21
[BD-9736] [BpkPrice] Append data attribute to Price Element

LuoTuxiu:
[BD-9736](https://skyscanner.atlassian.net/browse/BD-9736)

In some user case, some crawlers need to inspect the real price(not the previous price). So it will be convenient that we provide a data attribute to the price element(not the whole BpkPrice Component). This pr will add a prop: `dataPriceYourPick`. if set it true, then it will append a data attribute: `data-price-your-pick="true"` to the price element.

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

[BD-9736]: https://skyscanner.atlassian.net/browse/BD-9736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/bpk-component-price/src/BpkPrice-test.js
+++ b/packages/bpk-component-price/src/BpkPrice-test.js
@@ -97,7 +97,7 @@ describe.each([
         dataPriceYourPick
       />,
     );
-    expect(screen.getbyText(price)).toHaveAttribute(
+    expect(screen.getByText(price)).toHaveAttribute(
       'data-price-your-pick',
       'true',
     );
--- a/packages/bpk-component-price/src/BpkPrice-test.js
+++ b/packages/bpk-component-price/src/BpkPrice-test.js
@@ -94,7 +94,7 @@ describe.each([
         previousPrice={previousPrice}
         leadingText={leadingText}
         trailingText={trailingText}
-        dataAttributes={{ 'data-price-your-pick': true }}
+        dataAttributes={{ 'data-price-your-pick': 'true' }}
       />,
     );
     expect(screen.getByText(price)).toHaveAttribute(
--- a/packages/bpk-component-price/src/BpkPrice-test.js
+++ b/packages/bpk-component-price/src/BpkPrice-test.js
@@ -87,14 +87,14 @@ describe.each([
     expect(asFragment()).toMatchSnapshot();
   });
 
-  it('should append data-price-your-pick to price element when dataPriceYourPick is true', () => {
+  it('should append data-price-your-pick to price element when pass dataAttributes', () => {
     render(
       <BpkPrice
         {...props}
         previousPrice={previousPrice}
         leadingText={leadingText}
         trailingText={trailingText}
-        dataPriceYourPick
+        dataAttributes={{ 'data-price-your-pick': true }}
       />,
     );
     expect(screen.getByText(price)).toHaveAttribute(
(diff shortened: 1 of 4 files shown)


packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -177,6 +183,7 @@ BpkPrice.propTypes = {
   trailingText: PropTypes.string,
   previousPrice: PropTypes.string,
   leadingClassName: PropTypes.string,
+  dataPriceYourPick: PropTypes.bool,
```
wuyq0808, line 186: Not sure if we should introduce business idea (`YourPick`) into Backpack components. Should we use something like `priceProps` with a structure? Like this `inputProps` https://github.com/Skyscanner/backpack/blob/7d9573d4db382428afa018e7ecefeb8e67cfee75/examples/bpk-component-slider/examples.js#L72

packages/bpk-component-price/src/BpkPrice.js
```diff
@@ -147,7 +149,11 @@ const BpkPrice = (props: Props) => {
             !isAlignRight && 'bpk-price__spacing',
           )}
         >
-          <BpkText textStyle={priceTextStyle} tagName="span">
+          <BpkText
+            textStyle={priceTextStyle}
+            tagName="span"
+            data-price-your-pick={dataPriceYourPick ? 'true' : undefined}
```
jasonhuang-sky, line 155:
```suggestion
            {...dataAttributes}
```

The dataAttributes could be defined as

```
dataAttributes: Record<string, string>
```

when consume this component

```
<BpkPrice
        {...props}
        previousPrice={previousPrice}
        leadingText={leadingText}
        trailingText={trailingText}
       dataAttributes={ 'data-price-your-pick': 'true' }
      />
```

Has this property name been discussed and decided? If not, then I suggest that it should be more general, rather than designed for a specific need. For example, this actually adds [data attributes](https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Solve_HTML_problems/Use_data_attributes) to this html element, so we can consider allowing custom data attributes, maybe they will be used in other scenarios in the future

LuoTuxiu: Yes, it would be better for general, I will update it.

packages/bpk-component-price/src/BpkPrice.js
LuoTuxiu: Very thanks for review. Yes, we should be more general and should not limit YourPick logic in Backpack.
I will udpate it.

packages/bpk-component-price/src/BpkPrice.js
LuoTuxiu: Have confirmed with backpack team we can make it.

packages/bpk-component-price/src/BpkPrice-test.js
```diff
@@ -88,6 +87,22 @@ describe.each([
...
   });
 
+  it('should append data-price-your-pick to price element when pass dataAttributes', () => {
+    render(
+      <BpkPrice
+        {...props}
+        previousPrice={previousPrice}
+        leadingText={leadingText}
+        trailingText={trailingText}
+        dataAttributes={{ 'data-price-your-pick': true }}
```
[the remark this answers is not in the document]
LuoTuxiu: Have updated. Thanks.

wuyq0808 (approved):
LGTM

jasonhuang-sky (approved):

PR #3806, opened 2025-04-22
[NO-JIRA][BpkInsetBanner] Fix UI issues

kirstybryce:
### Context
We noticed that in our sponsored booking panel ad that uses the bpk inset banner, on iPhones the popover is cutoff at the end due to the width of the popover being too large. There's also an issue with the width of certain partner logos being too big and pushing content out of the banner.
| Popover UI issue on iPhone |
| --- |
| Screenshot 2025-04-22 at 13 10 58 |

### Suggested solution
- Added a `popoverWidth` prop to pass a string width value (so that either pixel or rem values can be passed) to the popover
- Added a max width of 120px for logos in mobile screen sizes
- Added 1rem margin to the right of the popover so it isn't stuck to the right side of the screen like it currently is.

### Screenshots

| Before | After |
| --- | --- |
| Screenshot 2025-04-22 at 16 47 01 Screenshot 2025-04-22 at 16 47 05 | Screenshot 2025-04-22 at 16 48 35 Screenshot 2025-04-22 at 16 48 52 |

- [x] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/examples/bpk-component-inset-banner/examples.tsx
+++ b/examples/bpk-component-inset-banner/examples.tsx
@@ -92,13 +92,37 @@ const WithCtaTextAndPopoverExampleLight = () => (
     logo="https://content.skyscnr.com/m/49503c4388cb05ab/original/Skyland_Black_172x96.png"
     callToAction={{
       text: 'Sponsored',
-      popoverMessage: 'This is a popover message',
+      popoverMessage:
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
+      closeBtnIcon: true,
+      labelTitle: true,
+      popverLabel: 'Info',
+      buttonCloseLabel: 'Close',
+      buttonA11yLabel: 'More info',
+    }}
+    backgroundColor="#FFE300"
+    variant={VARIANT.onLight}
+    accessibilityLabel="Sponsored by Skyscanner"
+  />
+);
+
+const WithCustomPopoverWidthAndMarginsExample = () => (
+  <BpkInsetBanner
+    title="Lorem ipsum"
+    subheadline="Lorem ipsum dolor sit amet"
+    logo="https://content.skyscnr.com/m/49503c4388cb05ab/original/Skyland_Black_172x96.png"
+    callToAction={{
+      text: 'Sponsored',
+      popoverMessage:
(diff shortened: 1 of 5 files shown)


examples/bpk-component-inset-banner/examples.tsx
```diff
@@ -98,6 +98,7 @@ const WithCtaTextAndPopoverExampleLight = () => (
       popverLabel: 'Info',
       buttonCloseLabel: 'Close',
       buttonA11yLabel: 'More info',
+      popoverWidth: '15rem',
```
Supremeyh, line 101: it could be better add another example with `popoverWidth`

packages/bpk-component-inset-banner/src/BpkInsetBanner.module.scss
```diff
@@ -51,6 +51,7 @@
     padding-inline-end: tokens.bpk-spacing-base();
 
     @include breakpoints.bpk-breakpoint-mobile {
+      max-width: 120 * tokens.$bpk-one-pixel-rem;
```
Supremeyh, line 54: `3 * tokens.bpk-spacing-xxl()`

packages/bpk-component-inset-banner/src/BpkInsetBanner.tsx
```diff
@@ -109,6 +112,7 @@ const BpkInsetBanner = ({
             }}
           >
             <BpkPopover
+              style={{ width: popoverWidth, marginInlineEnd: '1rem' }}
```
Supremeyh, line 115: we should get rid of hardcode `marginInlineEnd`, if necessary, it should pass like popoverWidth

senzg: Would it be better if we can add a non-required prop `popoverWidth ` in BpkPopover to allow consumers controlling it? cc @Ezreal-Yang

kirstybryce: I can add it as a prop :)

kirstybryce: Added margin props:
https://github.com/Skyscanner/backpack/pull/3806/commits/d0b2ecb7c209c5e80fd61846f0728bbe9658c0a1#diff-ffe744a7ae973ded3495ae16e5128ba06fb5df4cf3e573c0c96245418fc24fa8R53-R54

Supremeyh (approved):

PR #3808, opened 2025-04-28
[LUNA-2516][BPKSegmentedControls] Updates the implemenation to support arrow navigation

michmaml:
# BPKSegmentedControls

As part of the web audit, we identifed that the Segmented Controls component does not allow arrow navigation. This was the case because all the control panels were buttons. To allow the panels to be navigable using arrow keys, we need to replace the buttons with input filed of type `radio`.

The solution preserves the previous styles and behaviour. The only change is the aforementioned keyboard navigation.

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests


gert-janvercauteren:
Closing this to ensure we follow a different approach.

PR #3816, opened 2025-05-02
[CAPY-1169][BpkChipGroup]: Set Chip Group Nudgers edge threshold

lucascomp:
There is an intermittent edge case where the right hand nudger is never disabled due to `1px` calculation error, which may occurs depending on a combination of parents width and screen width. Unfortunately, I was not able to find the pattern that causes the issue, but looking for some external references the suggestion is to introduce a threshold of 2 pixels before evaluating nudgers conditions.

| Before | After |
| - | - |
| <video src="https://github.com/user-attachments/assets/d6576307-632e-4a56-a54b-019d981a9bac" /> | <video src="https://github.com/user-attachments/assets/aa44c51c-dcb1-41db-8d06-bd0481bd9956" /> |

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests

### DIFF
--- a/packages/bpk-component-chip-group/src/Nudger.tsx
+++ b/packages/bpk-component-chip-group/src/Nudger.tsx
@@ -56,6 +56,9 @@ const AlignedRightArrowIcon = withButtonAlignment(ArrowRight);
 // Chosen based on feeling good with the example stories
 const SCROLL_DISTANCE = 150;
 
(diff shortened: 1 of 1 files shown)


Iain530 (approved):
Nice find 👍

GCpigsic (approved):

PR #3824, opened 2025-05-12
[CLOV-230][BpkTooltip]Fix arrow alignment

GCpigsic:
Fix the arrow alignment of tooltip

## screenshot

|before|after|
|---|---|
|  | |

- [x] Component `README.md`
- [x] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-tooltip/src/BpkTooltip.tsx
+++ b/packages/bpk-component-tooltip/src/BpkTooltip.tsx
@@ -37,7 +37,7 @@ import {
   useRole,
 } from '@floating-ui/react';
 
-import { surfaceHighlightDay } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
+import { surfaceHighlightDay, onePixelRem } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
 
 import { TransitionInitialMount, cssModules } from '../../bpk-react-utils';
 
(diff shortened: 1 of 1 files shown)


packages/bpk-component-tooltip/src/BpkTooltip.tsx
```diff
@@ -77,10 +77,10 @@ export type Props = {
 // so we need to compensate slightly to make it look as one.
 const getArrowAlignment = (placement: Placement) => {
   if (placement.includes('bottom')) {
-    return { bottom: '98%' };
+    return { bottom: `calc(100% - ${onePixelRem})` };
```
GCpigsic, line 80: onePixelRem here is the border of the tooltip https://github.com/Skyscanner/backpack/blob/edfcbbae1c82d99868fccc18392bd3a5104073dc/packages/bpk-component-tooltip/src/BpkTooltip.module.scss#L36 


https://github.com/Skyscanner/backpack/blob/af182b97055fe627a06427ab11958be3149cca4f/packages/bpk-mixins/_layers.scss#L41
The reason why we subtract the border width is to make sure the position of the arrow is aligned with the content area of the tooltip, not with the outer edge of the border.
image

Before, we made it as a fixed value '98%', it will make the position of the arrow not correct when the tooltips is high, like ⬇️
image

Supremeyh (approved):

PR #3835, opened 2025-05-20
[BpkSegmentedControl] adjust role and add label prop

gert-janvercauteren:
CLOV-252, adjust segmented control

To ensure we make the segmented control more accessible we're setting a group role and adding an aria-label that is settable using the label prop.

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-segmented-control/src/BpkSegmentedControl.tsx
+++ b/packages/bpk-component-segmented-control/src/BpkSegmentedControl.tsx
@@ -72,7 +72,7 @@ const BpkSegmentedControl = ({
     <div
       className={containerStyling}
       role="group"
-      aria-label={label}
+      {...(label ? { 'aria-label': label } : {})}
     >
       {buttonContents.map((content, index) => {
         const isSelected = index === selectedButton;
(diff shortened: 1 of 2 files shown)


packages/bpk-component-segmented-control/src/BpkSegmentedControl.tsx
```diff
@@ -46,6 +50,7 @@ export type Props = {
 
 const BpkSegmentedControl = ({
   buttonContents,
+  label,
```
Supremeyh, line 53: since it's optional, it would be undefined, so need default string, eg `label=''`

gert-janvercauteren: I would like the label to be undefined (go away) if not provided. that would work right?

kerrie625: It could read 'undefined'. I encountered this issue when trying to add labels to the NavTabs.

gert-janvercauteren: Addressed this in https://github.com/Skyscanner/backpack/pull/3835/commits/da75b5cdcc6e958108e981019ae558fc45b2670a also added a test :D

kerrie625 (approved):
LGTM

PR #3846, opened 2025-05-30
[BpkHorizontalNav] fix focus outline

gert-janvercauteren:
# BpkHorizontalNav

Adjust the focus outline to be inside so it doesn't get clipped by the surrounding container

| Before | After | 
| --- | --- |
| image | 
image |

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-horizontal-nav/src/BpkHorizontalNavItem.module.scss
+++ b/packages/bpk-component-horizontal-nav/src/BpkHorizontalNavItem.module.scss
@@ -43,6 +43,11 @@
     cursor: pointer;
     fill: currentcolor;
 
+    &:focus-visible {
+      @include utils.bpk-focus-indicator;
+      outline-offset: -(tokens.$bpk-one-pixel-rem * 2);
+    }
+
     &--default {
       @include utils.bpk-themeable-property(
         color,


packages/bpk-component-horizontal-nav/src/BpkHorizontalNavItem.module.scss
```diff
@@ -43,6 +43,10 @@
     cursor: pointer;
     fill: currentcolor;
 
+    &:focus-visible {
+      outline-offset: -(tokens.$bpk-one-pixel-rem * 2);
```
gert-janvercauteren, line 47: This is the value from the original mix-in, but negative.

gert-janvercauteren: Previously, it showed just 2 lines, which is undesired. 
I agree that the colour contrast is not great, but I have just fixed the appearance so the focus indicator is around the entire button. 

We can discuss with design if we want to adjust how the focus indicator looks, but that is out of scope for this PR.

senzg (commented):
LGTM.
Just a heads up that in Firefox, the tab key seems lack of contrast. But I understand it is out of scope of this PR
image

senzg (approved):
LGTM

packages/bpk-component-horizontal-nav/src/BpkHorizontalNavItem.module.scss
gert-janvercauteren: Discussed this with the design team, and they are okay with allowing this modification on the old component.

PR #3867, opened 2025-07-01
Revert "Bump @floating-ui/react from 0.26.12 to 0.27.8 in /packages"

Faye-Xiao:
Reverts Skyscanner/backpack#3812 for a quick fix. We found an adoption issue in v37.9.0 . The consumer microsite, built with React 17.x, encounters dependency errors when attempting to adopt Backpack v37.9.0 due to the 'react/jsx-runtime' alias.
To remove the blocker of new version adoption for some cases, we decided to revert @floating-ui/react  0.26.12  for a while and **will use a new PR to solve the related issues**.

### DIFF
--- a/packages/package-lock.json
+++ b/packages/package-lock.json
@@ -9,7 +9,7 @@
       "version": "21.0.1",
       "license": "Apache-2.0",
       "dependencies": {
-        "@floating-ui/react": "^0.27.8",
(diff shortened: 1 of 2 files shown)


Faye-Xiao:
The known issue from https://github.com/floating-ui/floating-ui/issues/3240

PR #3876, opened 2025-07-10
[maintainence] upgrade node to 22.x

Faye-Xiao:

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,1 +1,1 @@
-lts/hydrogen
+22.14.0
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "license": "Apache-2.0",
   "private": true,
   "engines": {
-    "node": ">=18.16.0",
-    "npm": ">=9.5.1"
+    "node": ">=18.20.4",
+    "npm": ">=10.7.0"
   },
   "repository": {
     "type": "git",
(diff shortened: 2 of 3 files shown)


Faye-Xiao:
Percy failed due to the flaky tests.

package.json
```diff
@@ -4,8 +4,8 @@
   "license": "Apache-2.0",
   "private": true,
   "engines": {
-    "node": ">=18.16.0",
-    "npm": ">=9.5.1"
+    "node": ">=18.20.4",
+    "npm": ">=10.7.0"
```
GCpigsic, line 8: Why do we need to change it?

GCpigsic (approved):

package.json
Faye-Xiao: To make it consistent with the Global component and the internal consumer microsite.

.nvmrc
```diff
@@ -1 +1 @@
-lts/hydrogen
+22.14.0
```
gert-janvercauteren, line 1: Should we use lts/jod for this?

Faye-Xiao: No strong need for this. Let's make it consistent with the global component first.

PR #3878, opened 2025-07-11
[Clover-402][BpkCardList] Update BpkCardList scroll behaviour

xiaogliu:
This PR mainly focuses on changes in the "row" layout scenario on desktop:

1. Click and slide behavior: Originally, clicking the next page would jump to one card. After the modification, it will jump to a whole set of cards (the specific number depends on initiallyShownCards).
2. Card rendering: Before the modification, mobile would render all cards at once, while desktop only rendered the cards in the current view. After the modification, the rendering behavior of mobile and desktop is consistent: only render the current view + RENDER_BUFFER + already rendered cards.
3. The style has been fine-tuned according to the design, and the specific modifications have been pasted into the PR code.

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests


examples/bpk-component-card-list/examples.tsx
```diff
@@ -152,7 +151,6 @@ const snippetProps = {
 };
 
 const DestinationCard = (i: number) => (
-  // Usage Suggestion: define minWidth from consumer side by using className
```
xiaogliu, line 155: The Card width is determined by the number of cards set to be displayed ([desktop](https://github.com/Skyscanner/backpack/pull/3878/files#diff-49825811eb055632d14400d08a33814bc582358ef6b4b5f9ed8ec089c1ca131aL41), [mobile](https://github.com/Skyscanner/backpack/pull/3878/files#diff-49825811eb055632d14400d08a33814bc582358ef6b4b5f9ed8ec089c1ca131aR73)), so the minWidth cannot be specified, otherwise it will [conflict](https://github.com/Skyscanner/backpack/pull/3878/files#diff-6a0e148934bce32a7b29700751ad5e7d3eabc660d11743cf2229efcd522e159eR51) with `initiallyShownCardsDesktop`/`initiallyShownCardsMobile`.

packages/bpk-component-card-list/src/BpkCardList.module.scss
```diff
@@ -17,12 +17,17 @@
...
 @use '../../unstable__bpk-mixins/tokens';
+@use '../../unstable__bpk-mixins/breakpoints';
 
 .bpk-card-list {
   display: flex;
   flex-direction: column;
   gap: tokens.bpk-spacing-lg();
 
+  @include breakpoints.bpk-breakpoint-mobile {
+    gap: tokens.bpk-spacing-base();
```
xiaogliu, line 28: Based on design, if mobile gap change to `tokens.bpk-spacing-base()`

packages/bpk-component-card-list/src/BpkCardList.tsx
```diff
@@ -34,7 +34,8 @@ import STYLES from './BpkCardList.module.scss';
 
 const getClassName = cssModules(STYLES);
 
-const DEFAULT_ITEMS = 3;
+const DEFAULT_ITEMS_DESKTOP = 3;
+const DEFAULT_ITEMS_MOBILE = 2;
```
xiaogliu, line 38: Based on design `DEFAULT_ITEMS_MOBILE` be `2` makes more sense.

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.module.scss
```diff
@@ -17,27 +17,33 @@
...
+@use '../../../unstable__bpk-mixins/utils';
 
 .bpk-card-list-row-rail {
   &__row,
   &__rail {
     display: flex;
     padding-top: tokens.bpk-spacing-sm();
     padding-bottom: tokens.bpk-spacing-sm();
-    overflow-x: scroll;
+    overflow-x: hidden;
```
xiaogliu, line 29: In desktop, scroll card by wheel/touchpad will change [currentIndex](https://github.com/Skyscanner/backpack/pull/3878/files#diff-b442a953b34523f87814a56f06504909e6ea34b995d5e1965d4e3359c92417cdL83), And pagination will also change the currentIndex. It is rather difficult to keep the two in perfect synchronization, and the functionality is redundant with that of pagination so disable this feature.

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.module.scss
```diff
@@ -17,27 +17,33 @@
...
+    }
+
     &::-webkit-scrollbar {
       display: none;
     }
 
     &__card {
       position: relative;
-      padding: tokens.bpk-spacing-md();
+      padding: 0 tokens.bpk-spacing-md();
```
xiaogliu, line 46: It's parent element already set padding top/bottom, if add padding top/bottom here, will cause spacing much big than design.
image

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.module.scss
```diff
@@ -52,6 +58,31 @@
       overflow: visible;
       box-sizing: border-box;
       scroll-snap-align: start;
+
+      @include breakpoints.bpk-breakpoint-mobile {
+        &:first-child {
+          padding-left: tokens.bpk-spacing-sm();
```
xiaogliu, line 64: On mobile, the first card should be as close as possible to monetization. Considering the shadow, it cannot be set to 0 completely.
image

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.module.scss
```diff
@@ -52,6 +58,31 @@
...
+        flex: 0 0
+          calc(
+            (
+                100% -
+                  (
+                    tokens.bpk-spacing-md() *
+                      (var(--initially-shown-cards, 3) - 1)
+                  )
+              ) /
+              (var(--initially-shown-cards, 3) - 0.5)
```
xiaogliu, line 83: 0.5 means that when on mobile, half of the last card in the visible area will be hidden.

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
```diff
@@ -69,10 +73,44 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
...
+    index <= lastVisibleIndex + RENDER_BUFFER_SIZE
+      ? 1
+      : 0,
+  );
+
+  const cardRefCallback = (index: number) => (el: HTMLDivElement | null) => {
+    cardRefs.current[index] = el;
+    observerVisibility(el, index);
+    setA11yTabIndex(el, index, visibilityList);
+    // record the first card's width and height when it becomes visible
```
xiaogliu, line 101: The width and height of all cards in the Cardlist are consistent, including the placeholders for unrendered cards.

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
```diff
@@ -127,30 +165,43 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
...
+        const cardDimensionStyle: CSSProperties = {};
+        if (firstCardWidthRef.current) {
+          cardDimensionStyle.width = `${firstCardWidthRef.current}px`;
+        }
+        if (firstCardHeightRef.current) {
+          cardDimensionStyle.height = `${firstCardHeightRef.current}px`;
+        }
+
+        // Only render cards that are within the renderList range or have been visible before
+        if (renderList[index] !== 1 && !hasBeenVisibleRef.current.has(index)) {
```
xiaogliu, line 177: If the current card has not entered the visible area, replace it with an empty element. The empty element should be consistent with the card width/height to avoid reflow.

packages/bpk-component-card-list/src/BpkCardListRowRail/utils.tsx
```diff
@@ -65,27 +65,6 @@ export const setA11yTabIndex = (
...
 
-export const useUpdateCurrentIndexByVisibility = (
-  isMobile: boolean,
-  visibilityList: number[],
-  setCurrentIndex: (index: number) => void,
-  stateScrollingLockRef: React.MutableRefObject<boolean>,
-  openSetStateLockTimeoutRef: React.MutableRefObject<NodeJS.Timeout | null>,
-) => {
-  useEffect(() => {
-    if (isMobile) return; // No pagination on mobile, so no need to update the current index
```
xiaogliu, line 76: https://github.com/Skyscanner/backpack/pull/3878/files#r2200174759

examples/bpk-component-card-list/examples.module.scss
```diff
@@ -18,10 +18,6 @@
 
 @use '../../packages/unstable__bpk-mixins/tokens';
 
-.bpk-card {
-  min-width: tokens.$bpk-one-pixel-rem * 250;
-}
-
```
Faye-Xiao, line 24: 👍

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
```diff
@@ -69,10 +73,44 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
   const openSetStateLockTimeoutRef = useRef<NodeJS.Timeout | null>(null);
 
   const observerVisibility = useIntersectionObserver(
```
Faye-Xiao, line 75: I took a quick look at the `useIntersectionObserver ` hook. It lacks the `disconnect` logic. 
Could you help add that for this hook?  🙇‍♀️ 

https://github.com/Skyscanner/backpack/blob/15273e9aedf930eefa106323b30df9377db6c22e/packages/bpk-component-card-list/src/BpkCardListRowRail/utils.tsx#L116-L168

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
```diff
@@ -69,10 +73,44 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
...
+    currentIndex * initiallyShownCards,
+    root,
+    cardRefs,
+    stateScrollingLockRef,
+  );
+
+  // Similar to Virtual Scrolling to improve performance
+  const firstVisibleIndex = Math.max(0, visibilityList.indexOf(1));
+  const lastVisibleIndex = firstVisibleIndex + initiallyShownCards - 1;
+  const renderList = visibilityList.map((_, index) =>
```
Faye-Xiao, line 90: 🤔  Question: Is there a maximum length limit for ·visibilityList·? Is it possible to have hundreds of items? 
Currently, the `map` is executed once per render and generates a new reference to `renderList`. Therefore, the `cardRefCallback` generates a closure in the map. 
Maybe we can use the `useMemo` to mitigate that.

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
```diff
@@ -127,30 +165,43 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
...
+          return (
+            <div
+              key={`placeholder-${index.toString()}`}
+              style={{
+                ...shownNumberStyle,
+                ...cardDimensionStyle,
+                flexShrink: 0,
+                visibility: 'hidden',
+              }}
+              aria-hidden="true"
```
Faye-Xiao, line 187: 👍

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
xiaogliu: > Is it possible to have hundreds of items?
👍 It's possible, will update

packages/bpk-component-card-list/src/common-types.ts
```diff
@@ -102,7 +102,6 @@ type CardListCarouselProps = {
   initiallyShownCards: number;
   layout: typeof LAYOUTS.row | typeof LAYOUTS.rail;
   currentIndex: number;
-  setCurrentIndex: Dispatch<SetStateAction<number>>;
```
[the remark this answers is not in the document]
xiaogliu: > Currently, there is no way to pass the visible page number back to the container between components. Is this the expected behavior? 

Based on the previous implementation, this is expected. Passing setCurrentIndex to BpkCardListCarousel was for [useUpdateCurrentIndexByVisibility](https://github.com/Skyscanner/backpack/pull/3878/files), but we no longer need this logic.

> I’m considering a couple of use cases: for instance, when opening a link, the page should display a specific index card, but the current component can't achieve that. Additionally, for the New Relic operational event, we need to track the index of the displayed card.

This is a valid concern. In fact, we've already received feedback to trigger events. But I think this is related to changing the API. However, before that, we should review the current cardlist again to see which callback functions we need to pass out. For example, besides `onIndexChange` and `onFilterChange`, should we also expose them? I suggest that we make unified modifications to the API after the discussion. This PR only fixes the current problems in CardList and does not change any API.

image

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
```diff
@@ -127,30 +165,43 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
...
-            style={cardStyle}
+            ref={cardRefCallback(index)}
+            style={{
+              ...shownNumberStyle,
+              ...cardDimensionStyle,
+            }}
             key={`carousel-card-${index.toString()}`}
             role="group"
             aria-label={slideLabel(index, childrenLength)}
-            aria-current={index === currentIndex ? 'true' : 'false'}
```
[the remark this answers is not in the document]
xiaogliu: Since multiple cards can be displayed on one screen currently and these cards have no primary or secondary distinctions (for example, three cards are displayed on one screen and all three are equally important), the current aria-current is not accurate: it only marks the first one, but there are three "current" ones. If aria-current is added to all three actually displayed pictures, multiple "current" designations will also cause confusion.

But love to discuss more.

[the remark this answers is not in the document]
gert-janvercauteren: Removing this is ok, aria-current="page" can be used on the pagnation dots, but not on a single slide :D

examples/bpk-component-card-list/examples.tsx
```diff
@@ -129,7 +129,6 @@ const commonProps = {
   ),
   buttonContent: 'See more',
   buttonHref: 'https://www.skyscanner.net/',
-  initiallyShownCardsMobile: 2,
```
Faye-Xiao, line 132: You removed this value, but I saw there is an example of giving `initiallyShownCardsMobile : 1`.
Forget to update other exmaples?

https://github.com/Skyscanner/backpack/blob/e55c824471a9cf971d47b58ae2cee5e34f4e216c/examples/bpk-component-card-list/examples.tsx#L334

xiaogliu:
> > Visit https://backpack.github.io/storybook-prs/3878 to see this build running in a browser.
> 
> Hi @xiaogliu, I compared this storybook with the [production storybook](https://backpack.github.io/storybook/?path=/story/bpk-component-card-list--row-to-rail&globals=viewport:mobile1) cases and found some visual differences, as the below screenshots show. Are these changes what we expected?
> 
> Screenshot 2025-07-17 at 18 39 29 Screenshot 2025-07-17 at 18 41 45 Screenshot 2025-07-17 at 18 41 56

This comment contains two issues:
1. The content displayed on the last card in the mobile visible area has become larger, which is as expected. Below is the design. 
Before the change, we could only be [hardcoded](https://github.com/Skyscanner/backpack/pull/3878/files#diff-318d1d29a60e09dab9927274e74c21248b0c37d2e0e4f1bb73b2107d75bad839L22), so you maybe see different width for last in different screen.  
After change, I calculated with a ratio of 100. Currently, I've set it to 0.5, but it would be more appropriate to change it to [0.8](https://github.com/Skyscanner/backpack/pull/3878/commits/9df07f2b85e62986246a642469597cf939cb3a16#diff-49825811eb055632d14400d08a33814bc582358ef6b4b5f9ed8ec089c1ca131aR72) (since the width of the last card shown is less than 50%, so it can only be estimated).
image
2. When there is only one mobile card, its width becomes larger. This was a bug and has been fixed. https://github.com/Skyscanner/backpack/pull/3878/commits/9df07f2b85e62986246a642469597cf939cb3a16#diff-49825811eb055632d14400d08a33814bc582358ef6b4b5f9ed8ec089c1ca131aR72

See this link for the lasted status. https://backpack.github.io/storybook-prs/3878/?path=/story/bpk-component-card-list--multi-components-scrolling-test

examples/bpk-component-card-list/examples.tsx
xiaogliu: I removed it here is because I change mobile default `initiallyShownCardsMobile` [be 2](https://github.com/Skyscanner/backpack/pull/3878/files#r2200154591), consistent with design, so need to pass 2.

But in https://github.com/Skyscanner/backpack/blob/e55c824471a9cf971d47b58ae2cee5e34f4e216c/examples/bpk-component-card-list/examples.tsx#L334 this case, we want to test different initial cards so still keep it.
image

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.module.scss
```diff
@@ -52,6 +58,29 @@
...
+          calc(
+            (
+                100% -
+                  (
+                    tokens.bpk-spacing-md() *
+                      (var(--initially-shown-cards, 3) - 1)
+                  )
+              ) /
+              max(1, (var(--initially-shown-cards, 3) - 0.8))
+          );
```
Faye-Xiao, line 73: Could you explain the key numbers in this code and provide a note to help others understand it for future iterations? 🙇‍♀️

Faye-Xiao:
Approved the Percy [failure](https://percy.io/09e69e44/web/backpack/builds/41647758?utm_campaign=09e69e44&utm_content=backpack&utm_source=github_status_public) as the cardlist case change is what we expected while other change are flaky tests.

Faye-Xiao (approved):
LGTM. 🚀 
During the review and discussion, we also identified some other improvements outside of this PR scope. Will improve that in other PRs.

PR #3931, opened 2025-09-08
[Clov-343][BpkAutosuggestV2]Fix issue in GC Carhire-Search-Control

kerrie-wu:
**Why This PR**
This PR addresses several integration and UX issues when using BpkAutosuggestV2 in GC carhire-search-control. The goal is to make the component more predictable, customizable, and aligned with real-world usage.

https://skyscanner.atlassian.net/wiki/spaces/~kerriewu/pages/1443364968/GC+car-hire-search-control+issue

**Fix: Input not clearing when using input clear button**

- Previously, clicking the clear button would not trigger suggestion fetching (onSuggestionsFetchRequested('')) because internal inputValue wasn't properly reset.

- Now we explicitly call setInputValue('') and invoke fetch on clear, even when input is empty.

- Result: popularSuggestions are shown again as expected after clearing.

**Behavior Change: Always fetch suggestions when input is empty**

- Instead of calling onSuggestionsClearRequested(), we now always trigger onSuggestionsFetchRequested('').

- Why: Better aligns with consumer expectations — clearing the input shouldn't hide the dropdown but repopulate it with fallback (popularSuggestions or recentSearches).

- Result: Desktop and mobile now behave consistently.

**Improve: Floating UI reference alignment**

- In modal views, the suggestion dropdown was incorrectly sized due to Floating UI binding to the input element instead of the full wrapper.

- We now bind refs.setReference() to the outer container via useRef, ensuring dropdown width matches design.

- Result: Dropdowns in mobile modals are now consistently positioned and styled.

**Fix: Enter key caused exception when suggestions were sections**

- Previously, pressing Enter when sections could cause onSuggestionSelected to be called with undefined or in invalid state.

- The offending onKeyDown logic has been removed — Downshift’s built-in onSelectedItemChange already handles Enter, click, and keyboard selection safely.

- Result: Pressing Enter with or without sections no longer throws runtime errors.
- 

- [ ] Component `README.md`
- [x] Tests
- [x] Accessibility tests

### DIFF
--- a/packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
+++ b/packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
@@ -512,7 +512,7 @@ const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
       });
 
       const setInputRef = (node: HTMLInputElement | null) => {
-        if (refs.reference.current === node) return;
+        if (refs.reference?.current === node) return;
 
         // convert input ref from Downshift
         if (typeof downshiftInputRef === 'function') {


packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
```diff
@@ -357,22 +355,6 @@ const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
       }
     };
 
-    const onKeyDown = (e: KeyboardEvent): void => {
-      if (e.key !== 'Enter') {
```
kerrie625, line 361: Removed custom Enter key handler to avoid double triggering onSuggestionSelected. Downshift already handles Enter selection internally — no need to override manually.Keeps selection behavior consistent across mouse and keyboard input

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
```diff
@@ -475,9 +461,11 @@ const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
             role="group"
             aria-labelledby={sectionId}
           >
-            <div id={sectionId} className={theme.sectionTitle}>
-              {sectionTitleElement}
-            </div>
+            {sectionTitleElement && (
+              <div id={sectionId} className={theme.sectionTitle}>
```
kerrie625, line 465: sectionTitleElement sometimes not inclue in section suggest

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
```diff
@@ -525,7 +512,7 @@ const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
       });
 
       const setInputRef = (node: HTMLInputElement | null) => {
-        refs.setReference(node);
+        if (refs.reference.current === node) return;
```
kerrie625, line 515: Optimization: Prevent unnecessary re-setting of input ref.Avoids triggering Floating UI re-renders when the input node remains the same. This prevents layout shift / dropdown flicker on every render.

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
```diff
@@ -568,24 +548,32 @@ const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
             </span>
             <div className={getClassName(theme.inputWrapper)}>
               <BpkInput
-                value={normalizedInputValue}
+                value={inputValue}
```
kerrie625, line 551: Use `inputValue` from useCombobox instead of `value` from getInputProps, because it's the actual live state. `value` might be stale or mismatched, especially during input clearing or keyboard selection.

IrinaWei:
Hi Kerrie, I have a questions maybe related to components accessibility. As you can see if the list is too long and need to scroll the view the rest, purely press key down won't scroll the list thus the selection is changed but customers won't see them unless they scroll the window manually.

https://github.com/user-attachments/assets/28bc3d94-efd6-483c-804a-9c08dc85443a

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
```diff
@@ -382,6 +364,10 @@ const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
     const clearSuggestions = (e?: SyntheticEvent<HTMLButtonElement>) => {
       e?.stopPropagation();
       setInputValue('');
+      if (alwaysRenderSuggestions) {
+        hasLoadedInitiallyRef.current = true;
+        onSuggestionsFetchRequested('');
```
IrinaWei, line 369: Hi would you consider make it more readable? 🤔 If I didn't get it wrong, `onSuggestionsFetchRequested('')` means show all the suggestions while `onSuggestionsFetchRequested()` means close suggestion.

kerrie625:
> Hi Kerrie, I have a questions maybe related to components accessibility. As you can see if the list is too long and need to scroll the view the rest, purely press key down won't scroll the list thus the selection is changed but customers won't see them unless they scroll the window manually.

Yes, you are right. Currently, Autosuggest does not support scrolling, just like V1 and the flight search control. We can consider this as an optimization for a future iteration, but for now the behavior should remain consistent with V1.
image

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
```diff
@@ -285,7 +281,7 @@ const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
       placement: 'bottom-start',
       middleware: isDesktop
         ? [
-            offset(17),
+            offset(4),
```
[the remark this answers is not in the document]
kerrie625: Keep the same with [V1 margin-top setting](https://github.com/Skyscanner/backpack/blob/0e4df68bed89c6653e9eb99d9d30d0b7a644ec92/packages/bpk-component-autosuggest/src/BpkAutosuggest.module.scss#L41), add extra top margin to visually separate the dropdown from the input. I found consumer aslo set this [margin top ](https://github.skyscannertools.net/skyscanner/global-components/blob/b11a3c11ab5551aeade51d2eef9f35a7ac8b2449/libs/car-hire-search-controls/components/LocationSelector/LocationSelector.module.scss#L57). So Offset(4) must remain unchanged in Floating UI, so margin is used here，to ensure the arrow does not overlap the input field.

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
IrinaWei, line 515: Will `refs.reference` return `null` or `undefine`? Maybe we can use `refs.reference?.current` here?

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
```diff
@@ -251,12 +247,12 @@ const BpkAutosuggest = forwardRef<HTMLInputElement, BpkAutoSuggestProps<any>>(
...
-        if (newInputValue?.length) {
+        if (newInputValue?.length > 0) {
           if (newIsOpen) {
             onSuggestionsFetchRequested(newInputValue);
           }
-        } else if (suggestionsCount) {
-          onSuggestionsClearRequested();
+        } else {
+          onSuggestionsFetchRequested('');
         }
```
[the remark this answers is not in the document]
IrinaWei: Hi when `alwaysRenderSuggestions` is false, empty input always trigger `onSuggestionsFetchRequested('')` and display all the suggestions. Is this aligns with what we expected?

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
kerrie625: Updated in [b03a922](https://github.com/Skyscanner/backpack/pull/3931/commits/b03a922ac9e971a167fd0431b7db2cd33e3cc39f)

packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.tsx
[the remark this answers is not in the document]
kerrie625: Yes, that’s expected — when alwaysRenderSuggestions is false, triggering onSuggestionsFetchRequested('') on empty input still allows the consumer to decide whether to show suggestions (like popular/recent). We're not enforcing any UI here, just notifying input value changes.

PR #3950, opened 2025-09-22
[STAR-161] Replace DOM-based virtualization with content-visibility

dyhagho:
## Context

Safari exhibits a UX issue where pagination buttons cause visual teleporting/jumping during smooth scrolling on `BpkCardListCarousel`. This happens when navigating rails with man cards (it was detected with 100 cards). The issue was reproduced on Safari 18.6 [(see example)](https://www.skyscanner.net/transport/flights-from/bcn/?adultsv2=1&cabinclass=economy&childrenv2=&ref=home&rtn=1&preferdirects=false&outboundaltsenabled=false&inboundaltsenabled=false&oym=2510&iym=2510&OverrideFeatureTest=ce_e2e_prototype_dev:B), though other versions may also be affected.

Root cause analysis showed that the carousel's virtualization logic swaps placeholder elements with real cards mid-animation. Combined with Safari's scroll-snap handling, this leads to recalculated landing points during scroll. The main detected factors are:

- Native smooth scrolling (`behavior: smooth`)
- CSS scroll-snap (`scroll-snap-type: x mandatory`)
- DOM mutations during scroll (from placeholder to real cards)

## Solution
We are replacing JS-based virtualization with CSS `content-visibility` to eliminate DOM mutations during scroll. Which ensures:

- Stability on DOM structure: All cards remain in the DOM throughout the scroll animation
- Browser-native virtualization: off-screen cards are efficiently managed by the browser

Additionally, CSS `containIntrinsicSize` is used to preserve dimensions for off-screen cards. Both properties are fully supported per Skyscanner's [browserlist](https://github.com/Skyscanner/browserslist-config-skyscanner/blob/main/index.js)

### What has changed?
- Remove programmatic scroll detection logic
- Remove JS virtual scroll logic
- Use `content-vivibility: auto` for off-screen cards
- Use `contain-intrinsic-size` to match actual card dimensions

### Visual changes

**Safari 18.6 with JS Virtual Scroll**

https://github.com/user-attachments/assets/b6a0abec-8609-46f2-a3cf-0d7682006d1c

**Safari 18.6 with CSS `content-visibility`**

https://github.com/user-attachments/assets/16988288-a824-4a5f-bb8f-f803db3f53e7

**Chrome with CSS `content-visibility`**

https://github.com/user-attachments/assets/ea0648fe-9427-4e52-a0bd-a472aad0e4b5

**Firefox with CSS `content-visibility`**

https://github.com/user-attachments/assets/462eba00-4ed5-4d4b-9a4b-ce273b370990

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
+++ b/packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
@@ -96,7 +96,6 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
 
   // Similar to Virtual Scrolling to improve performance
   const firstVisibleIndex = Math.max(0, visibilityList.indexOf(1));
-  const lastVisibleIndex = firstVisibleIndex + initiallyShownCards - 1;
 
   const dynamicRenderBufferSize = useMemo(() => {
     if (childrenLength === 0 || initiallyShownCards === 0 || isMobile) return RENDER_BUFFER_SIZE;
@@ -109,22 +108,6 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
       (shownIndicatorCount - 1) * initiallyShownCards,
     );
   }, [childrenLength, initiallyShownCards, isMobile]);
-
-  const renderList = useMemo(
-    () =>
-      visibilityList.map((_, index) =>
-        index >= firstVisibleIndex - dynamicRenderBufferSize &&
-        index <= lastVisibleIndex + dynamicRenderBufferSize
-          ? 1
-          : 0,
-      ),
-    [
-      visibilityList,
-      firstVisibleIndex,
-      lastVisibleIndex,
-      dynamicRenderBufferSize,
-    ],
-  );
 
   const cardRefFns = useMemo(
     () =>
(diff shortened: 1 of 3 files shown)


packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
```diff
@@ -232,20 +227,19 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
...
-            />
-          );
+        // Determine visibility based on virtualization logic
+        const shouldBeVisible = shouldCardBeVisible(index);
+
+        // Card virtualization optimization styles
+        const cardVirtualizationStyle: CSSProperties = {};
+        if (!shouldBeVisible) {
+          // Use content-visibility for browser-native virtualization
+          cardVirtualizationStyle.contentVisibility = 'auto';
```
bjoern-wahle, line 237: https://caniuse.com/?search=content-visibility
=> 90% , the main problematic part could be IE 11. 
Although the good thing is that even if the css property does not work, the functionality should work - potentially a bit laggy though?

dyhagho: Good point! I was just about to add some comments about this :)

Both `content-visibility` and `contain-intrinsic-size` are used for more modern browsers. I checked [Skyscanner's browserlist config](https://github.com/Skyscanner/browserslist-config-skyscanner/blob/main/index.js) and for that we have a full support.

Worth checking what Clover think about this

Faye-Xiao: No worries. Backpack currently doesn't support IE 11. 
https://github.com/Skyscanner/backpack/blob/f69464e9317e80d21035aaec81c0eb9dc2d7811d/package.json#L102-L108

packages/bpk-component-card-list/src/BpkCardListRowRail/BpkCardListCarousel.tsx
```diff
@@ -255,6 +249,7 @@ const BpkCardListCarousel = (props: CardListCarouselProps) => {
             style={{
               ...shownNumberStyle,
               ...cardDimensionStyle,
+              ...cardVirtualizationStyle,
```
Faye-Xiao, line 252: Nitpick: 
Is it possible to have a unit test for this change?

PR #3989, opened 2025-10-02
[IRN-6175] [BpkCardList] Make Title Optional

rabisse:
Making the title prop optional to prevent an unintended gap from being added. There have been uses where people do `title=""` to prevent a title from being added, and then just compensating for the gap that's added - but we want a way to not have this gap at all.

To prevent this from being a breaking change for those already setting `title=""` we're explicitly checking for `title !== undefined` so that the existing uses won't be effected.

## Migration from `title=""` to `title=undefined` - not needed but may allow removal of hacky styling
For uses of BpkCardList where `title=""` (or otherwise falsy):
1. Remove the prop entirely
2. Compensate for the loss of gap at the top of the component (check mobile and desktop)

## Screenshots
| title="" or other sting | title=undefined |
|---|---|
| Screenshot 2025-10-02 at 4 11 48 PM | Screenshot 2025-10-02 at 4 23 21 PM  |

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests

### DIFF
--- a/examples/bpk-component-page-indicator/examples.tsx
+++ b/examples/bpk-component-page-indicator/examples.tsx
@@ -18,6 +18,8 @@
 
 import { useState } from 'react';
 
+import { backgroundElevation03DarkColor } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
+
 import { BpkBackgroundImage } from '../../packages/bpk-component-image';
 import BpkPageIndicator, {
   VARIANT,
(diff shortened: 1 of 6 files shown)


packages/bpk-component-card-list/src/common-types.ts
```diff
@@ -70,7 +68,15 @@ type CardListBaseProps = {
...
-};
+} & TitleProps;
+
+type TitleProps = {
+  title: string;
+  description?: string;
+} | {
+  title?: never;
+  description?: never;
+}
```
rabisse, line 79: Clearer type that reflects the SectionHeader component
Both are optional, but the description is only allowed if there's also a title

PR #4004, opened 2025-10-08
[IRN-6177] [BpkCardList] Fix issue with carousel alignment

jimmycook:
This change is to address the issues with alignment when using row to rail. 

There are a few concerns:
- Drop shadows on cards getting cut off
- Ensuring the width of the cards is correct
- Where is the edge of the container

This change fixes is by when we are in the row state, we can have negative margin equal to the padding on the cards, and adjust the flex basis of the cards to fill the space correctly.

image

Because of the peaking on the cards, we don't apply this to rail mode, so it only fixes the issue on desktop.

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests


rabisse (dismissed):

Faye-Xiao (approved):
LGTM.

PR #4052, opened 2025-11-24
[GOOSE-559][BpkSwapButton] Override padding alignment set by Safari User Agent

FireRedNinja:
Ref: https://skyscanner.atlassian.net/browse/GOOSE-559

Safari ships its own user-agent styles, and our unstyled button ends up inheriting a few defaults we don’t actually want:

```css
padding-inline-start: 6px;
padding-block-start: 2px;
padding-block-end: 3px;
```

These extra paddings throw off the visual alignment compared to other browsers. This PR zeroes them out so the component isn’t relying on Safari’s quirks. Flexbox already handles centering, so resetting the padding brings consistent cross-browser rendering without changing layout logic.

### Before
image

### After (1px diff)
image

> No changes noticed in Firefox and Chrome

---

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-swap-button/src/BpkSwapButton.module.scss
+++ b/packages/bpk-component-swap-button/src/BpkSwapButton.module.scss
@@ -37,14 +37,13 @@
     display: flex;
     width: tokens.bpk-spacing-xxl();
     height: tokens.bpk-spacing-xxl();
+    padding: 0;
     justify-content: center;
     align-items: center;
     transition: transform tokens.$bpk-duration-base;
     border: tokens.$bpk-border-size-xl solid tokens.$bpk-core-primary-day;
     border-radius: 50%;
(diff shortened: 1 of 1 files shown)


jan-bock (dismissed):

packages/bpk-component-swap-button/src/BpkSwapButton.module.scss
```diff
@@ -43,6 +43,8 @@
     border: tokens.$bpk-border-size-xl solid tokens.$bpk-core-primary-day;
     border-radius: 50%;
     background-color: tokens.$bpk-surface-default-day;
+    padding-block: 0;
+    padding-inline: 0;
```
Supremeyh: it could simplify to `padding: 0;`

Supremeyh (approved):

PR #4175, opened 2026-01-29
[IRN-6372] [BpkBottomSheet] - aria title read on Android TalkBack

steviehailey-skyscanner:

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
+++ b/packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
@@ -24,6 +24,7 @@ import BpkCloseButton from '../../bpk-component-close-button';
 import BpkLink from '../../bpk-component-link';
 import BpkNavigationBar from '../../bpk-component-navigation-bar';
 import { TEXT_STYLES } from '../../bpk-component-text/src/BpkText';
+import BpkVisuallyHidden from '../../bpk-component-visually-hidden';
 import { BpkDialogWrapper, cssModules } from '../../bpk-react-utils';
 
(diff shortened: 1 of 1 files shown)


packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -155,6 +156,7 @@ const BpkBottomSheet = ({
   );
 
   const headingId = `bpk-bottom-sheet-heading-${id}`;
+  const headingIsHidden = !title && 'ariaLabel' in ariaProps;
```
steviehailey-skyscanner: Detect if we're in the situation where we only have a heading for ScreenReader purposes to describe the dialog (no visual title)

packages/bpk-component-bottom-sheet/src/BpkBottomSheet.tsx
```diff
@@ -185,7 +187,7 @@ const BpkBottomSheet = ({
       <>
         <header className={getClassName('bpk-bottom-sheet--header-wrapper')}>
           <BpkNavigationBar
-            id={headingId}
+            id={headingIsHidden ? `bpk-bottom-sheet-title-hidden-${id}` : headingId}
```
steviehailey-skyscanner: Since we adding a Screen Reader heading below using `headingId` we don't want to pass the same id to Nav Bar and risk a duplicated element id inside with the same id so use an alternative in this case.

steviehailey-skyscanner:
Experimental dup of https://github.com/Skyscanner/backpack/pull/4198

PR #4233, opened 2026-02-12
[INL-3675] feat: add dev release workflow for PR-based npm publishing

Jack-Waller:
## What

Adds a `workflow_dispatch` dev release workflow that publishes pre-release npm packages from open pull requests, so contributors can validate unreleased changes in downstream projects before merging.

Heavily inspired by [global-components' dev release workflow](https://github.com/Skyscanner/global-components/blob/main/.github/workflows/global-components-dev-release.yml) and [backpack's release workflow](https://github.com/Skyscanner/backpack/blob/main/.github/workflows/release.yml).

## Why

There is currently no way to test a Backpack change in a consuming application without merging and releasing. A dev release lets engineers publish a tagged pre-release (`x.y.z-dev-v<run>.<attempt>` under the `dev` npm dist-tag) from any open PR, significantly shortening the feedback loop.

Originally needed to safely roll out a Backpack fix (#4231) for [INL-3675](https://skyscanner.atlassian.net/browse/INL-3675).

## Files added

- **`.github/workflows/dev-release.yml`** — orchestrates security validation, build, npm publish, and PR comment reporting.
- **`.github/actions/pr-security-validation/action.yml`** — composite action that validates the PR is open, lives in the same repo (no forks), and the triggering user has write access.
- **`.github/actions/upsert-pr-comment/action.yml`** — composite action that creates or updates a single PR comment identified by an HTML marker, keeping the thread tidy.
- **`DEV_RELEASE.md`** — user-facing docs: when to use it, how to trigger from the UI or CLI, how to consume the published version, and the security model.

## Workflow flow

1. **SecurityCheck** — validates PR + caller, posts a "Dev Release Started" comment.
2. **DevRelease** — checkouts the validated SHA, runs `npm ci` and `npm run build`, computes the dev version. If `dry_run=false`, runs `cd packages/backpack-web/dist && npm version ... && npm publish --tag dev`.
3. **ReportResults** — updates the same PR comment with the final outcome and install command.

## Security considerations

- Fork PRs are rejected — the workflow only runs for branches in `Skyscanner/backpack`.
- Caller permission is verified via the GitHub API (must have write access).
- All `${{ }}` expressions are passed via `env` blocks to prevent script injection.
- Workflow permissions follow least-privilege (`contents: read`, `pull-requests: write`).
- GitHub API calls authenticate via the Backpack GitHub App because Skyscanner's organisation IP allow list rejects the default `GITHUB_TOKEN`.
- Real publishes (`dry_run=false`) are gated by the `Publishing` GitHub environment, which requires `backpack-web` team approval. Dry runs skip the environment so they don't need approval.
- npm publishes use `NPM_TOKEN` from the `Publishing` environment secrets (trusted publishing / OIDC is out of scope for this PR — see [#14](https://github.com/Skyscanner/backpack/security/code-scanning/14)).

## Changes since hand-over

The original workflow was authored by @Jack-Waller; @vincentliu took over and pushed five additional fix commits:

| Commit | Purpose |
|---|---|
| `bff9db98f` | `npm run transpile` → `npm run build` (the repo migrated to Nx, no `transpile` script exists), `cd dist` → `cd packages/backpack-web/dist`, bump action SHAs to match `release.yml`, add a "Dev Release Started" comment after the security check passes |
| `880033fd2` | Add GitHub App token to all jobs that hit the GitHub API (the org IP allow list rejects the default `GITHUB_TOKEN`) |
| `7380843cd` + `955efc2fa` | Conditional `environment` so dry runs skip the `Publishing` approval gate; real publishes still require approval |
| `e0bd11c6c` | Add `DEV_RELEASE.md` |

## Testing

End-to-end dry run on this PR:

- [Run #26558464180](https://github.com/Skyscanner/backpack/actions/runs/26558464180) — SecurityCheck 9s, Build 41s, Report 6s, total ~1 min
- "Dev Release Started" comment posted on this PR after security check
- Same comment upserted to "Dev Release Dry Run Successful" with version `41.13.0-dev-v26558464180.1` (base version `41.13.0` because the branch is behind `main`; will be `42.x` after rebase)
- Real publish (`dry_run=false`) deferred — dry run validates everything except the literal `npm publish` line

## Checklist

- [x] No component changes — CI, workflow, action and docs files only
- [x] `skip-changelog` label appropriate (no consumer-facing code changes)

[INL-3675]: https://skyscanner.atlassian.net/browse/INL-3675


.github/actions/pr-security-validation/action.yml
```diff
@@ -0,0 +1,148 @@
...
+
+        if [ -z "$BASE_REPO" ] || [ "$BASE_REPO" = "null" ]; then
+          echo "::error::Cannot determine base repository. Rejecting as a security precaution."
+          exit 1
+        fi
+
+        if [ "$HEAD_REPO" != "$BASE_REPO" ]; then
+          echo "::error::Dev releases are not permitted from forked repositories."
+          echo ""
+          echo "Head repo: $HEAD_REPO"
```
Jack-Waller: **Fork detection.** Comparing `head_repo` to `base_repo` ensures the PR branch lives in the same repository. Fork PRs are rejected because `workflow_dispatch` runs in the context of the base repo and would expose secrets to untrusted code from a fork.

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,205 @@
...
+    permissions:
+      contents: read
+    outputs:
+      dev_version: ${{ steps.version.outputs.dev_version }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ needs.SecurityCheck.outputs.pr_sha }}
+          persist-credentials: false
+          fetch-depth: 0
```
Jack-Waller: `fetch-depth: 0` is needed for `git describe` in the version step.

DEV_RELEASE.md
```diff
@@ -0,0 +1,150 @@
...
+After validating, switch back to a real version before merging your
+consumer change:
+
+```bash
+npm install @skyscanner/backpack-web@latest
+```
+
+---
+
+## Dry run vs real publish
```
RichardSyq, line 105: Is a dry run still necessary now?

RichardSyq (commented):
One security hardening suggestion for the dev release publish step.

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,235 @@
...
+          echo "Dev version: ${DEV_VERSION}"
+
+      - name: Build
+        run: npm run build
+
+      - name: Publish dev release to npm
+        if: ${{ inputs.dry_run != true }}
+        run: |
+          cd packages/backpack-web/dist
+          npm version $DEV_VERSION --no-git-tag-version
```
RichardSyq: Could we harden this publish step so PR-controlled npm lifecycle scripts never run with `NPM_TOKEN`? The current PR runs `npm version` and `npm publish` inside `packages/backpack-web/dist`, which is built from the PR head, while `NODE_AUTH_TOKEN` is in the environment. Since the PR can modify `packages/backpack-web/package.json`, it could add lifecycle scripts like `preversion`, `version`, `prepublishOnly`, `prepack`, or `prepare`; npm would execute those during `npm version` / `npm publish` after the `Publishing` environment is approved, and those scripts could read the token. The approval gate controls when the token is released, but not what PR-controlled package scripts do with it once released. Could we avoid `npm version` by updating `package.json` directly, then publish with scripts disabled, e.g. `npm publish --ignore-scripts --tag dev`? This preserves the dev release behavior while preventing unmerged PR code from running with the npm publish token.

xiaogliu: Thanks for catching this — agreed it's the right defence-in-depth posture even with the same-repo restriction in place.

Pushed `c2a6d7692`:
- Replaced `npm version` with a direct edit of `dist/package.json` via `node -e`, so `preversion` / `version` / `postversion` hooks can't run.
- Added `--ignore-scripts` to `npm publish`, so `prepack` / `prepare` / `prepublishOnly` / `postpack` / `postpublish` can't either.

I verified that `packages/backpack-web/package.json` has no `scripts` block today and the build only copies it verbatim (`scripts/transpilation/strip-exports.js` only deletes `exports`), so neither change can regress an existing legitimate hook.

Worth flagging that `release.yml` has the same shape (`npm version` + `npm publish` without `--ignore-scripts`), but the threat model is different there — real releases are triggered by a tag publish, never from a PR head — so I haven't touched it in this PR. Happy to open a separate one if you'd like the same hardening applied there.

DEV_RELEASE.md
xiaogliu: Fair question — the dry run isn't redundant with PR CI even though both run `npm run build`. PR CI doesn't exercise:

- the SecurityCheck job (PR validation, caller-permission check, GitHub App token / IP allow-list path),
- the dev-version computation (`git describe` + run id),
- the PR comment upsert flow,
- the version-stamping step that's now in the publish job (since #4233 it's no longer guarded by `dry_run`, so dry runs catch a malformed stamp before a real publish needs it).

The other reason it's useful is that contributors can self-test their workflow change without burning a `Publishing` approval. Real publishes are gated by `backpack-web` team review, dry runs aren't.

Happy to tighten the README wording to make this distinction clearer — let me know if you'd prefer the table to drop the "runs build" row (since that's the same in both columns) and lean on the new framing.

RichardSyq (approved):
LGTM

PR #4303, opened 2026-03-20
[CLOV-1589][BpkAutosuggest] Migrate JS to TypeScript

kerrie-wu:
- Convert all .js/.jsx source, test, and example files to .tsx/.ts
- Remove Flow type annotations and PropTypes; replace with TypeScript interfaces
- Use @ts-expect-error for untyped react-autosuggest import per decisions/imports-ts-suppressions.md
- Move defaultProps defaults to destructuring parameters
- Fix class component constructor to pass props per TypeScript strict mode
- Add missing onSuggestionsClearRequested method caught by TypeScript
- Remove now-unused @ts-expect-error directives from consumer examples/stories
- Delete all original .js files

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests


packages/bpk-component-autosuggest/src/BpkAutosuggest.tsx
```diff
@@ -51,26 +52,27 @@ Autosuggest.defaultProps.theme = {
...
         ref(element);
 
         if (typeof inputRef === 'function') {
           inputRef(element);
         }
       }}
-      autoComplete={autoComplete}
-      {...rest}
+      autoComplete={autoComplete as string}
+      {...(rest as any)}
```
IrinaWei: Could we delete `any` here?

packages/bpk-component-autosuggest/src/BpkAutosuggest.tsx
```diff
@@ -51,26 +52,27 @@ Autosuggest.defaultProps.theme = {
...
+      inputRef={(element: HTMLInputElement) => {
         ref(element);
 
         if (typeof inputRef === 'function') {
           inputRef(element);
         }
       }}
-      autoComplete={autoComplete}
-      {...rest}
+      autoComplete={autoComplete as string}
```
IrinaWei: Do we really need `as string` here?

packages/bpk-component-autosuggest/src/BpkAutosuggest.tsx
```diff
@@ -51,26 +52,27 @@ Autosuggest.defaultProps.theme = {
   sectionTitle: getClassName('bpk-autosuggest__section-title'),
 };
 
-type Props = {
-  ref: (?HTMLInputElement) => mixed,
-  inputRef: (?HTMLInputElement) => mixed,
-  autoComplete: string,
+type InputComponentProps = InputHTMLAttributes<HTMLInputElement> & {
+  ref: (element: HTMLInputElement | null) => void;
+  inputRef?: ((element: HTMLInputElement | null) => void) | null;
```
IrinaWei, line 60: `inputRef` is required before but now it's optional, is it intentional?

kerrie-wu: Yes, this is intentional. inputRef is supplied by the consumer via <Autosuggest inputProps={{ inputRef }} /> rather than by react-autosuggest itself, so it's genuinely optional at runtime — the existing typeof inputRef === 'function' guard a few lines below confirms this. The original Flow type marked it as required but the runtime check tells the real story; this TS migration corrects that. InputComponentProps is also a local (non-exported) type used only by the internal renderInputComponent default, so this has no impact on the public API of BpkAutosuggestLegacy.

packages/bpk-component-autosuggest/src/BpkAutosuggest.tsx
kerrie-wu: Updated in [0dbd905](https://github.com/Skyscanner/backpack/pull/4303/commits/0dbd905c247c5b11080dd2b7ee33def052ed6d77)

packages/bpk-component-autosuggest/src/BpkAutosuggest.tsx
kerrie-wu: Updated in [0dbd905](https://github.com/Skyscanner/backpack/pull/4303/commits/0dbd905c247c5b11080dd2b7ee33def052ed6d77)

IrinaWei (approved):
LGTM

PR #4342, opened 2026-04-01
[BpkHorizontalNav] Migrate to TypeScript

xiaogliu:
## Summary
- Migrate `bpk-component-horizontal-nav` from JavaScript (Flow) to TypeScript
- Convert class components (`BpkHorizontalNav`, `BpkHorizontalNavItem`) to functional components with hooks
- Replace Flow types and PropTypes with TypeScript types
- No public API changes — all exports remain identical

## Test plan
- [x] All 19 existing tests pass (3 suites)
- [x] ESLint: 0 errors, 0 warnings
- [x] TypeScript: 0 type errors from this package
- [x] Snapshots regenerated and verified


xiaogliu:
Addressed remaining Copilot comments:

**Fixed (commit c6c39335):**
- `scrollAdjustment` was computed as a delta between bounding-rect positions but passed as the absolute `left` argument to `Element.scroll({ left })` / `scrollLeft`. When the container was already scrolled, the selected item would land in the wrong position. Now uses `scrollRef.current.scrollLeft + scrollAdjustment` as the absolute target.

**Already addressed in earlier commit 7c97d4e3:**
- `ariaLabel` / className props typed as `string | null` (with `?? undefined` adapter for `BpkMobileScrollContainer`).
- `BpkHorizontalNavItem` uses `forwardRef` to preserve ref support.
- `useEffect` skips first render via `isMounted` ref to match the original `componentDidUpdate` behavior.
- Inexact rest comments added to both components.

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.tsx
```diff
@@ -0,0 +1,192 @@
...
+  useEffect(() => {
+    requestAnimationFrame(() => {
+      scrollSelectedIntoView(false);
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
+    scrollSelectedIntoView(true);
+  });
```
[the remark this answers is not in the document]
xiaogliu: Already addressed in commit 7c97d4e3 — the second `useEffect` uses an `isMounted` ref to skip the initial render, matching the original `componentDidUpdate` behaviour:

```tsx
useEffect(() => {
  if (isMounted.current) {
    scrollSelectedIntoView(true);
  } else {
    isMounted.current = true;
  }
});
```

The first `useEffect` (with `[]` deps) runs `scrollSelectedIntoView(false)` for the mount path (no smooth scroll), and the second one only triggers smooth scroll on subsequent updates. Not adding a dependency array is intentional: the original `componentDidUpdate` ran unconditionally on every update.

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.tsx
```diff
@@ -0,0 +1,197 @@
...
+        scrollRef.current.scroll &&
+        typeof scrollRef.current.scroll === 'function' &&
+        useSmoothScroll
+      ) {
+        scrollRef.current.scroll({
+          left: targetScrollLeft,
+          behavior: 'smooth',
+        });
+      } else {
+        scrollRef.current.scrollLeft = targetScrollLeft;
```
xiaogliu: Here, the old logic of `left: scrollAdjustment` / `scrollLeft = scrollAdjustment` has been changed to accumulate the offset based on the current `scrollLeft`. Even though this may correct the scrolling calculation, it will alter the behavior of `autoScrollToSelected` when there is an existing horizontal scroll position. The goal of this PR is only to migrate to TypeScript, and logical fixes should not be included. It is recommended to revert to the original `scrollAdjustment` logic; if you need to fix the scrolling bug, please create a separate PR.

xiaogliu: Reverted in commit e24efea4b — the `scrollAdjustment` logic now matches the original JS implementation byte-for-byte. The bug will be addressed in a separate PR.

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.js
```diff
@@ -1,223 +0,0 @@
...
-  /**
-   * Ensures that the selected item is within view when loaded on narrow-screened devices.
-   */
-  autoScrollToSelected: PropTypes.bool,
-  className: PropTypes.string,
-  leadingScrollIndicatorClassName: PropTypes.string,
-  /**
-   * When set to "false", the bottom border on the component isn't included. This refers
-   * to the underline on the whole "BpkHorizontalNav", not the line that appears under the selected item.
-   */
```
IrinaWei, line 206: The comment has been deleted in migration, would it be better to add them back?

packages/bpk-component-horizontal-nav/src/BpkHorizontalNavItem.tsx
```diff
@@ -93,36 +87,13 @@ class BpkHorizontalNavItem extends Component<Props> {
...
-  }
-}
-
-BpkHorizontalNavItem.propTypes = {
-  children: PropTypes.node.isRequired,
-  className: PropTypes.string,
-  /**
-   * **Note:**
-   * "disabled" and "selected" are mutually exclusive for true values, i.e. only one of them can be true for a given nav item.
-   */
```
IrinaWei, line 106: Could we add these JSDoc comments back here?

packages/bpk-component-horizontal-nav/src/BpkHorizontalNav.tsx
```diff
@@ -0,0 +1,196 @@
...
+  if (autoScrollToSelected || type === HORIZONTAL_NAV_TYPES.light) {
+    children = Children.map(rawChildren, (child) => {
+      const childProps: { type?: string } = {};
+      let childRef: ((ref: HTMLDivElement | null) => void) | undefined;
+      if (autoScrollToSelected) {
+        if (
+          child &&
+          typeof child === 'object' &&
+          'props' in child &&
+          child.props &&
```
IrinaWei: Since now child is already object and has `props` we might delete `child.props &&`

IrinaWei: Or we might extract a comment variable
```
const isElement = typeof child === 'object' && child !== null && 'props' in child;
```
Since L169 did the same

IrinaWei (approved):
LGTM

PR #4390, opened 2026-04-09
[ST-7745] [BpkThumbButton] Add small size variant  

mariamuzas:
**What** 
Adds a size prop ('default' | 'small') and an iconColor prop ('default' | 'primary') to `BpkThumbButton` component : 

**size**
- small uses sm icons, no fixed dimensions, and `bpk-text-xs font size` — designed for inline use alongside text
- default preserves existing behaviour (no breaking change)
- Disables the `bpk-touch-tappable 48px pseudo-element on --small` to prevent touch targets overlapping adjacent buttons

**iconColor**
- primary renders the icon in `$bpk-text-primary-day`
- default uses `$bpk-text-disabled-day`— default for both sizes
- Decouples colour from size so either can be combined independently

**Why**                                                                                                                                                                        
Needed by BpkAiBlurb component (https://github.com/Skyscanner/backpack/pull/4338) where thumb buttons sit inline next to caption text, making the default 24px lg icons visually oversized. 

[Figma design](https://www.figma.com/design/KXf2gHNLDe2cXWUoHl4cTX/Backpack%E2%80%A8Foundations---Components?node-id=15094-15634&m=dev)

**Notes**
- ThumbsButtonSize, ThumbsButtonColour exported as a named type alongside the existing ThumbsButtonType
-  New Small Storybook story added; VisualTest updated with small and small-selected rows
-  iconColor defaults in destructuring (consistent with size and selected)

- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests


packages/bpk-component-thumb-button/src/BpkThumbButton.tsx
```diff
@@ -52,13 +59,19 @@ const BpkThumbButton = ({
...
   selected = false,
+  size = 'default',
   type,
 }: BpkThumbButtonProps) => {
-  const Icon = type === 'up' ? ThumbsUpIcon : ThumbsDownIcon;
+  const icons: Record<ThumbsButtonSize, Record<ThumbsButtonType, ComponentType>> = {
+    small: { up: ThumbsUpIconSm, down: ThumbsDownIconSm },
+    default: { up: ThumbsUpIconLg, down: ThumbsDownIconLg },
+  };
+  const Icon = icons[size][type];
```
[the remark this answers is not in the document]
mariamuzas: If I understand correctly, not needed as size and type are both typed (`ThumbsButtonSize` and `ThumbsButtonType`), so invalid values cannot be passed by consumers. The default value on size also ensures it's always defined.

packages/bpk-component-thumb-button/src/BpkThumbButton.module.scss
```diff
@@ -48,6 +49,10 @@
     transition: none;
   }
 
+  &:active {
+    color: tokens.$bpk-text-secondary-day;
+  }
+
```
mariamuzas, line 55: Fix a small bug: https://skyscanner.slack.com/archives/C099ZFP0U83/p1775753159048179?thread_ts=1775635520.043069&cid=C099ZFP0U83

packages/bpk-component-thumb-button/index.ts
```diff
@@ -18,5 +18,5 @@
 
 import BpkThumbButton from './src/BpkThumbButton';
 
-export type { BpkThumbButtonProps, ThumbsButtonType } from './src/BpkThumbButton';
+export type { BpkThumbButtonProps, ThumbsButtonType, ThumbsButtonSize } from './src/BpkThumbButton';
```
RichardSyq: Looks like `ThumbsButtonColor` didn't make it into the exports here — only `ThumbsButtonSize` was added. The PR description says both types should be named exports, and consumers who want to type the `iconColor` prop value won't be able to import it from the package. Should be:

```ts
export type { BpkThumbButtonProps, ThumbsButtonType, ThumbsButtonSize, ThumbsButtonColor } from './src/BpkThumbButton';
```

packages/bpk-component-thumb-button/src/BpkThumbButton.module.scss
```diff
@@ -48,6 +49,10 @@
     transition: none;
   }
 
+  &:active {
+    color: tokens.$bpk-text-secondary-day;
+  }
```
RichardSyq, line 54: Think there might be a CSS specificity issue here. `.bpk-thumb-button:active` has specificity (0,1,1), which beats `.bpk-thumb-button--selected` at (0,1,0). So when a selected button is actively pressed, this rule wins and the icon will flash `$bpk-text-secondary-day` rather than keeping the accent colour from `--selected`.

Interestingly the `--color-primary + --selected` combination is fine since two classes give it (0,2,0), but plain `--selected` on its own isn't covered.

Suggestion: add `&--selected:active { color: tokens.$bpk-core-accent-day; }` inside the `--selected` block.

packages/bpk-component-thumb-button/src/BpkThumbButton.module.scss
```diff
@@ -60,4 +65,31 @@
...
+
+  &--small {
+    width: auto;
+    height: auto;
+
+    @include typography.bpk-text-xs;
+
+    &::before {
+      content: none;
```
RichardSyq, line 96: This disables the `::before` touch target that was added to meet the WCAG 44×44px minimum (the original commit for `bpk-touch-tappable` cited this requirement explicitly). I understand why — overlapping targets are a real UX problem in an inline context — but the README and the `size` prop JSDoc don't currently mention that the small variant drops the touch target.

Could we add a note to `### Small size` in the README (and ideally the JSDoc too) that this variant isn't suitable for touch-primary surfaces?

packages/bpk-component-thumb-button/src/BpkThumbButton.tsx
```diff
@@ -42,6 +50,10 @@ export type BpkThumbButtonProps = {
    * Whether the thumb is in selected state.
    */
   selected?: boolean;
+  /**
+   * Size variant. 'small' uses sm icons with no fixed dimensions, matching inline text use cases.
+   */
+  size?: ThumbsButtonSize;
```
RichardSyq, line 57: The `accessibility-test.tsx` file isn't updated in this PR — it still only runs axe against the default props. The `size="small"` path is structurally different (different icon component, touch target suppressed) and worth covering with a dedicated axe test. Would be good to add cases for at least `size="small"`, `iconColor="primary"`, and their combination before merging.

packages/bpk-component-thumb-button/src/BpkThumbButton.module.scss
mariamuzas: aaah interesting! Will add that!

RichardSyq (approved):
lgtm

PR #4411, opened 2026-04-16
[QUAR-1552] Add wrapTitle prop to BpkNavigationBar to prevent title truncation

villanuevadani:
## Summary

- Adds a `wrapTitle` boolean prop to `BpkNavigationBar` that disables `white-space: nowrap` / `text-overflow: ellipsis` on the title element, allowing long titles to wrap onto multiple lines
- **`BpkBottomSheet` now passes `wrapTitle` unconditionally so bottom sheet titles always wrap (fixes truncation of long localised titles e.g. `de-DE` DSA modal title)**
- Adds a Storybook story to `BpkNavigationBar` demonstrating long-title wrapping behaviour
- Adds unit tests for the new prop in both `BpkNavigationBar` and `BpkBottomSheet`

Jira: https://skyscanner.atlassian.net/browse/QUAR-1552

## Test plan

- [ ] `BpkNavigationBar` renders title without ellipsis when `wrapTitle` is set
- [ ] `BpkBottomSheet` titles wrap onto multiple lines instead of truncating
- [ ] Storybook story for `BpkNavigationBar` shows long-title wrapping
- [ ] Unit tests pass for both `BpkNavigationBar` and `BpkBottomSheet`

### DIFF
--- a/packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
+++ b/packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
@@ -68,6 +68,12 @@
     &--on-dark {
       color: tokens.$bpk-text-primary-inverse-day;
     }
+
+    &--wrap {
+      white-space: normal;
+      overflow: visible;
+      text-overflow: unset;
+    }
   }
 
   &__title-container {
--- a/packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
+++ b/packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
@@ -70,9 +70,9 @@
     }
 
     &--wrap {
+      text-overflow: unset;
       white-space: normal;
       overflow: visible;
-      text-overflow: unset;
     }
   }
 
--- a/packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
+++ b/packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
@@ -38,6 +38,7 @@ export type Props = {
   title: ReactNode;
   titleTextStyle?: TextStyle;
   titleTagName?: Tag;
+  wrapTitle?: boolean;
   className?: string;
   leadingButton?: ReactElement | null;
   trailingButton?: ReactElement | null;
@@ -57,6 +58,7 @@ const BpkNavigationBar = (props: Props) => {
     titleTagName = "span",
     titleTextStyle = TEXT_STYLES.heading5,
     trailingButton,
+    wrapTitle = false,
     ...rest
   } = props;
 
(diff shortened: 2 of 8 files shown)


packages/bpk-component-navigation-bar/src/BpkNavigationBar.module.scss
```diff
@@ -68,6 +68,12 @@
     &--on-dark {
       color: tokens.$bpk-text-primary-inverse-day;
     }
+
+    &--wrap {
+      white-space: normal;
+      overflow: visible;
```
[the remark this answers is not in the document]
villanuevadani: `overflow: visible` is added on purpose for the scenario

kirstybryce (dismissed):
LGTM 🚀

packages/bpk-component-navigation-bar/src/BpkNavigationBar.tsx
```diff
@@ -91,6 +93,7 @@ const BpkNavigationBar = (props: Props) => {
         <span className={getClassNames(
           'bpk-navigation-bar__title',
           `bpk-navigation-bar__title--${barStyle}`,
+          wrapTitle && 'bpk-navigation-bar__title--wrap',
```
RichardSyq, line 98: Could this also handle the case where `title` is a `ReactNode`?

Looking at the `else` branch just below, when a non-string title is passed the component renders `bpk-navigation-bar__title-container` instead, and `wrapTitle` is never consulted there — so the `--wrap` modifier never gets applied.

Since `BpkBottomSheet` now passes `wrapTitle` unconditionally, a consumer who passes a `ReactNode` title to `BpkBottomSheet` would still see truncation. Worth extending the modifier to the `else` branch as well, or documenting the limitation in the prop's JSDoc?

villanuevadani: Good catch, I actually missed it. What would you do? I think both are valid options

RichardSyq: Looking at the SCSS, __title-container (the ReactNode path) only has justify-self: center — there’s no overflow: hidden, white-space: nowrap, or text-overflow: ellipsis, so ReactNode titles already wrap naturally without needing wrapTitle.

I’d suggest going with Option 2: add a JSDoc comment on the wrapTitle prop noting it only applies to string titles. Adding the modifier to the ReactNode path would be dead code since there’s no truncation to undo.

villanuevadani: Cool, I'll do so!

villanuevadani: Done 5c62ecb166d55f171b3431ac013b4957976bd89a

RichardSyq (approved):
lgtm

PR #4412, opened 2026-04-16
[ST-7669][BpkComparisonTable] Add BpkComparisonTable component

mariamuzas:
## BpkComparisonTable

Adds **BpkComparisonTable**, a modal-based comparison table for viewing up to 3 items side by side. The consumer owns the columns state and controls the open/close behaviour.

### API

Namespace component with three sub-components:

- **`BpkComparisonTable.Root`** — modal shell wrapping `BpkModalV3`; accepts `isOpen` / `onClose`.
- **`BpkComparisonTable.Header`** — header row with optional title, close trigger, and an optional slot for `BpkAiBlurb`.
- **`BpkComparisonTable.Content`** — scrollable comparison table. Accepts 1–3 `BpkCompareColumn` columns and handles layout, callbacks and placeholder rendering.

### Key behaviours

- **Automatic placeholder columns:** the table always renders 3 columns. Empty slots are filled with "Add more" placeholders that trigger `onAddMoreClick`, guiding users to add items.
- **Column removal:** each column header includes a Remove button that fires `onRemove(itemId)`, letting the consumer update their state.
- **Scroll-driven collapse animation:** as the user scrolls, the image area and remove button smoothly fade out and collapse via CSS variables (`--bpk-image-opacity`, `--bpk-rows-offset`) synchronised from a JS scroll handler. Rows stay locked in place during the animation phase, then normal scrolling resumes. The header cells are sticky so column context is always visible.
- **Best-deal badge:** optional `BpkBadge type="brand"` overlay on any column header via `bestTag`.
- **AI blurb slot:** optional `children` on `Header` for composing `BpkAiBlurb` or any summary content.

### Included

- Unit tests (column rendering, callbacks, column count validation, open/close)
- Accessibility tests (axe, 1/2/3 column configurations)
- Storybook examples (standalone modal + combined with `BpkComparisonTray`)
- Component README with full usage guide and prop tables

---

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests


packages/bpk-component-comparison-table/src/BpkComparisonTray/BpkComparisonTray.module.scss
```diff
@@ -66,13 +66,13 @@
...
   position: absolute;
   top: 0;
-  right: 0;
+  inset-inline-end: calc(-8 * #{tokens.$bpk-one-pixel-rem});
   display: flex;
   width: 16 * tokens.$bpk-one-pixel-rem;
   height: 16 * tokens.$bpk-one-pixel-rem;
   justify-content: center;
   align-items: center;
-  transform: translate(50%, -50%);
```
mariamuzas, line 75: This fix a RTL bug in BpkComparisonTray close button, so the button positions correctly in both LTR and RTL layouts.

packages/bpk-component-comparison-table/src/BpkCompareModal/BpkCompareModalContent.tsx
```diff
@@ -0,0 +1,115 @@
...
+        />
+
+        <BpkTableBody type={TABLE_BODY_TYPES.striped}>
+          {rowIds.map((rowId) => (
+            <BpkTableRow key={rowId}>
+              {displayColumns.map((column, index) => (
+                <BpkTableCell
+                  key={column ? `${rowId}-${column.itemId}` : `${rowId}-placeholder-${index}`}
+                  // eslint-disable-next-line @skyscanner/rules/forbid-component-props
+                  className={column ? undefined : getClassName('bpk-compare-modal__placeholder-cell')}
```
mariamuzas: placeholder need a distinct background colour not available via BpkTableCell props

packages/bpk-component-comparison-table/src/BpkComparisonTable/README.md
```diff
@@ -0,0 +1,183 @@
+# BpkComparisonTable
```
dominicfraser:
```suggestion
# BpkComparisonTable (experimental) 
```

packages/bpk-component-comparison-table/src/BpkComparisonTable/README.md
```diff
@@ -0,0 +1,183 @@
...
+  removeLabel: 'Remove',
+  bestTagLabel: 'Best',
+  addMoreDescription: 'Add up to 3 deals to compare your favourites',
+  addMoreLinkText: 'Add more',
+};
+
+<BpkComparisonTable.Root isOpen={isOpen} onClose={() => setIsOpen(false)}>
+  <BpkComparisonTable.Header
+    title="Compare cars"
+    translations={translations}
```
dominicfraser: In GCs we use a key called `strings` - I'm thinking we want to be consistent between Backpack and GCs and also call this prop strings?

packages/bpk-component-comparison-table/src/BpkComparisonTable/README.md
```diff
@@ -0,0 +1,183 @@
...
+  <BpkComparisonTable.Content
+    columns={columns}
+    onRemove={handleRemove}
+    onAddMoreClick={() => setIsOpen(false)}
+    translations={translations}
+  />
+</BpkComparisonTable.Root>
+```
+
+If you don't need the AI blurb, omit the children from `BpkComparisonTable.Header`:
```
dominicfraser:
```suggestion
If you don't need the AI Blurb, omit the children from `BpkComparisonTable.Header`:
```

packages/bpk-component-comparison-table/src/BpkComparisonTray/BpkComparisonTray.stories.tsx
```diff
@@ -145,7 +145,7 @@ const InteractiveExample = () => {
 };
 
 const meta = {
-  title: 'bpk-component-comparison-table',
+  title: 'bpk-component-comparison-table/BpkComparisonTray',
```
dominicfraser: Now that we've got the main component we could consider just deleting this Story, unless we think it provides anything valuable that the overall Stories don't? 

(every story adds to Storybook build time and storage size)

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import STYLES from './BpkComparisonTable.module.scss';
+
+const getClassName = cssModules(STYLES);
+
+function BpkComparisonTableHeader({ children, title, translations }: BpkComparisonTableHeaderProps) {
+  return (
+    <>
+      <BpkModalV3.Header>
+        {title && <BpkModalV3.Title>{title}</BpkModalV3.Title>}
```
dominicfraser: We want to make sure there is an title accessible - either as a visible one or a non visible one.

Currently we only cater to there being a visible one, but if someone doesn't pass a title then there would be nothing for a screen reader to read.

If we look at `NoHeaderExample` in `examples/bpk-component-modal-v3/examples.tsx` we can see it uses `BpkVisuallyHidden` when not wanting to visibly have a title.

How could we use that here?

Not one to block merging this first PR, can be thought about in a sequential follow up.

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <BpkModalV3.CloseTrigger label={translations.closeLabel} />
+      </BpkModalV3.Header>
+      {children && (
+        <div className={getClassName('bpk-comparison-table__header-slot')}>
```
dominicfraser: We could possibly use a Bpk Layout component here? It's still very fuzzy when to prefer Bpk Layout in JS vs layout in CSS when _inside_ of Backpack, so not a blocking comment.

```
 padding-bottom: tokens.bpk-spacing-md();
 padding-inline: tokens.bpk-spacing-base();
```

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableContent.tsx
```diff
@@ -0,0 +1,115 @@
...
+          translations={translations}
+        />
+
+        <BpkTableBody type={TABLE_BODY_TYPES.striped}>
+          {rowIds.map((rowId) => (
+            <BpkTableRow key={rowId}>
+              {displayColumns.map((column, index) => (
+                <BpkTableCell
+                  key={column ? `${rowId}-${column.itemId}` : `${rowId}-placeholder-${index}`}
+                  // eslint-disable-next-line @skyscanner/rules/forbid-component-props
```
dominicfraser, line 112: Could we add a comment here explaining why this is necessary?

As a sequential thing to look at, we probably want to note that BpkTableCell needs this ability natively without us using a className prop

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableColumnPlaceholder.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const getClassName = cssModules(STYLES);
+
+type BpkComparisonTableColumnPlaceholderProps = {
+  addMoreDescription: string;
+  addMoreLinkText: string;
+  onAddMoreClick: () => void;
+};
+
+function BpkComparisonTableColumnPlaceholder({
```
dominicfraser: Is this `BpkComparisonTableColumnHeaderPlaceholder` - a placeholder specifically for the header?

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableColumnPlaceholder.tsx
```diff
@@ -0,0 +1,67 @@
...
+}: BpkComparisonTableColumnPlaceholderProps) {
+  return (
+    <BpkFlex
+      direction="column"
+      align="center"
+      gap={BpkSpacing.MD}
+      paddingTop={BpkSpacing.LG}
+      paddingInline={BpkSpacing.XL}
+    >
+      <div className={getClassName('bpk-comparison-table__placeholder-text')}>
```
dominicfraser: Could we do:

```
<BpkText
  tagName="p"
  textStyle={TEXT_STYLES.bodyDefault}
  color={TEXT_COLORS.textSecondary}
  textAlign="center"
  className={getClassName('bpk-comparison-table__placeholder-text')}
>
```

moving textAlign to the prop, and making it clear that white-space is a missing capability of BpkText? Or does it not render as expected?

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableHeaderRow.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from './common-types';
+
+type BpkComparisonTableHeaderRowProps = {
+  displayColumns: Array<BpkCompareColumn | null>;
+  onRemove: (itemId: string) => void;
+  onAddMoreClick: () => void;
+  translations: BpkComparisonTableTranslations;
+};
+
+function BpkComparisonTableHeaderRow({
```
dominicfraser: I think the more normal style in Backpack would be

```
const BpkComparisonTableHeaderRow = ({
```

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.module.scss
```diff
@@ -0,0 +1,151 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@use '../../../bpk-mixins/tokens';
+@use '../../../bpk-mixins/breakpoints';
+
+$image-area-height: 83 * tokens.$bpk-one-pixel-rem;
+
+// ─── Scroll container ────────────────────────────────────────────────────────
```
Faye-Xiao: nit: could you remove these dividers?

packages/bpk-component-comparison-table/index.ts
```diff
@@ -25,4 +26,12 @@ export type {
...
-export default BpkComparisonTray;
+export type {
+  BpkCompareRow,
+  BpkCompareColumn,
+  BpkComparisonTableRootProps,
+  BpkComparisonTableTranslations,
+  BpkComparisonTableNamespace,
+} from './src/BpkComparisonTable/common-types';
+
+export { BpkComparisonTable, BpkComparisonTray };
```
Faye-Xiao: The `BpkComparisonTray` is an existing component in the product version.  https://backpack.github.io/storybook/?path=/docs/bpk-component-comparison-table--documentation

If we change the default import, it would be a breaking change.  Besides that, currently, every backpack provides a default import. 

According to the folder name, it looks like the default one should be `BpkComparisonTable`. 🤔

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableContent.tsx
mariamuzas: Looks that the comment on the PR got outdated: https://github.com/Skyscanner/backpack/pull/4412#discussion_r3125388218

You mean to add it as a comment in the file?

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.module.scss
```diff
@@ -0,0 +1,151 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@use '../../../bpk-mixins/tokens';
+@use '../../../bpk-mixins/breakpoints';
+
+$image-area-height: 83 * tokens.$bpk-one-pixel-rem;
```
Faye-Xiao, line 22: nit: could you add a note for this CSS variable in case someone modifies the `IMAGE_FADE_THRESHOLD_PX` in JSX? 

e.g.
```
// SYNC: keep in sync with IMAGE_FADE_THRESHOLD_PX in BpkComparisonTableContent.tsx
```

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableColumnPlaceholder.tsx
mariamuzas: `BpkComparisonTableColumnPlaceholder` is the placeholder rendered inside the column header (<th>) when a slot/column is empty, if nnot `BpkComparisonTableColumnHeader` renders this. 
Maybe better to rename it to sth like: `BpkComparisonTableColumnHeaderPlaceholder`

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableColumnPlaceholder.tsx
mariamuzas: I did try this, it does work but we need to add className as prop in `BpkText` which triggers the eslint forbidden prop rule.
I've tried to minimised eslint rule exceptions

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableContent.tsx
```diff
@@ -0,0 +1,115 @@
...
+  onAddMoreClick,
+  onRemove,
+  translations,
+}: BpkComparisonTableContentProps) {
+  const scrollContainerRef = useRef<HTMLDivElement>(null);
+
+  // Pad with nulls so the table always renders exactly MAX_COLUMNS slots.
+  const displayColumns = [...columns, null, null, null].slice(0, MAX_COLUMNS);
+
+  const rowIds = columns[0]?.rows.map((row) => row.rowId) ?? [];
```
Faye-Xiao, line 54: nit： The `rowIds` is derived from ·columns[0]·, and cells for other columns are looked up by `rowId` via a Map.  I suppose, this works great when the consumer keeps all columns in sync — but if the data comes from an API and one column ends up with a different rowId order or a missing entry, the affected cells just render as `null`. No error, no warning, just mysteriously empty cells that could take a while to track down.

The README does call this out clearly, which is appreciated! 😄 But I wonder if it's worth adding a small dev-mode guard to surface the problem earlier:

e.g.
```
if (process.env.NODE_ENV !== 'production' && columns.length > 1) {
  const baseIds = columns[0].rows.map((r) => r.rowId).join(',');
  columns.forEach((col) => {
    const ids = col.rows.map((r) => r.rowId).join(',');
    if (ids !== baseIds) {
      console.warn(
        `[BpkComparisonTable] Column "${col.itemId}" rowId sequence differs from column 0. Rows may misalign.`,
      );
    }
  });
}
```

packages/bpk-component-comparison-table/src/BpkComparisonTray/BpkComparisonTray.stories.tsx
mariamuzas: Yeah, I think we can delete it, the main story use the tray in the same way  that the individual story.

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.stories.tsx
```diff
@@ -0,0 +1,340 @@
...
+    setItems(remaining);
+  };
+
+  useEffect(() => {
+    if (pendingFocusIndexRef.current === null) return;
+    const idx = pendingFocusIndexRef.current;
+    pendingFocusIndexRef.current = null;
+    const tray = document.querySelector('[data-backpack-ds-component="ComparisonTray"]');
+    const removeButtons = tray?.querySelectorAll<HTMLButtonElement>('button[aria-label^="Remove"]');
+    removeButtons?.[idx]?.focus();
```
Faye-Xiao, line 247: nit: 
Small thing in the story — the focus-restoration after item removal queries the DOM directly:
```
const tray = document.querySelector('[data-backpack-ds-component="ComparisonTray"]');
const removeButtons = tray?.querySelectorAll<HTMLButtonElement>('button[aria-label^="Remove"]');
removeButtons?.[idx]?.focus();
```
This leans on two internal details of `BpkComparisonTray` — the data-backpack-ds-component attribute and the `aria-label^="Remove"` selector. If either of those changes down the line, the focus shift quietly stops working.

I get that there's no clean way to thread a ref through the Tray without adding API surface, so this might be a pragmatic call for story code. **If that's the reasoning, a short comment explaining the trade-off would make it much easier for the next person** who reads this to understand why the DOM query is there rather than flagging it as a bug.

packages/bpk-component-comparison-table/src/BpkComparisonTable/README.md
```diff
@@ -0,0 +1,183 @@
...
+      { rowId: 'rating', cell: <span>3.8 — Good</span> },
+    ],
+    removeA11yLabel: 'Remove Hertz deal',
+  },
+]);
+
+const handleRemove = (itemId: string) => {
+  const remaining = columns.filter((column) => column.itemId !== itemId);
+  setColumns(remaining);
+  if (remaining.length < 2) setIsOpen(false);
```
Faye-Xiao: Small question about intended behaviour — the README example closes the modal when `remaining.length < 2`, but the stories seem to follow a slightly different rule:

**StandaloneExample** doesn't close on remove at all
**CombinedExample** only auto-closes when `items.length === 0`

I'm not sure which is intentional. The component does render a valid 1-column + 2-placeholder state, so maybe keeping the modal open with one item (so the user can add more) is actually the desired UX? If so, the README example might be a bit too prescriptive with < 2.

Either way, it'd be great to have the stories and the README tell the same story — even if the answer is "1-item comparison is valid, consumer decides when to close". What was the design intent here? 🤔

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableContent.tsx
dominicfraser: Please! Missed it as an original comment.

General thing of 'whenever we use a suppression, leave in code why to justify to the next person'

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableColumnPlaceholder.tsx
dominicfraser: I think what we have is equally as 'bad' from the point of the eslint rule, its just that the rule isn't clever enough to know.

Optional, but I'd be more explicit - prefer to be obviously 'bad' than have someone assume we're doing a 'good' thing. Subjective, not a blocker.

packages/bpk-component-comparison-table/index.ts
mariamuzas: Yeah, this would be in theory a breaking change, but this is not currently used by service. 
If we need a default,  agree the default should be `BpkComparisonTable`, I can make the changes

dominicfraser: > If we change the default import, it would be a breaking change. Besides that, currently, every backpack provides a default import.

The entire component is experimental right now - its not complete or open to anyone using it. That means no breaking changes apply - its being split into incremental PRs, but until it has been validated in car hire it counts as 'experimental' and can be changed completely without anything other than a patch version. 

The previous code was merged early to allow for separate review, but is all the same thing.

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.stories.tsx
```diff
@@ -0,0 +1,340 @@
...
+        />
+      </BpkComparisonTable.Root>
+    </BpkVStack>
+  );
+};
+
+// ─── Meta ─────────────────────────────────────────────────────────────────────
+
+const meta = {
+  title: 'bpk-component-comparison-table',
```
Faye-Xiao, line 327: The `BpkComparisonTray` stories already use` bpk-component-comparison-table/BpkComparisonTray `as their title, but `BpkComparisonTable` sits at the root level (`bpk-component-comparison-table`). This makes the table stories act as the "parent" group in Storybook, with the tray appearing as a child — which may be a bit misleading.

Bt the way, for the special case, layout components, does intentionally use the root level for `BpkProvider` (as a package-level entry point), but that's a deliberate exception. For this package I don't think we want the same treatment — both components are peers. ( Correct me if I missed something) 

Would it make sense to mirror the same pattern here?

```
- title: 'bpk-component-comparison-table',
+ title: 'bpk-component-comparison-table/BpkComparisonTable',
```

That way both components sit as siblings under the same folder, which feels more consistent with how the rest of Backpack structures multi-component packages.

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableContent.tsx
mariamuzas: Good catch! I got this actually in the first iteration from Claude. Decided to remove it, as the main responsibility to pass the correct correct and consistent column data comes from the consumer, and and felt this might be adding more complexity. 
But I can add it back, I can see this can be useful maybe through the integration of the component

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.module.scss
```diff
@@ -0,0 +1,151 @@
...
+
+// ─── Scroll container ────────────────────────────────────────────────────────
+
+.bpk-comparison-table__scroll-container {
+  min-height: 0;
```
Faye-Xiao: ℹ️  The CI physical-property check is flagging a number of properties in this file. The Tray's SCSS already went through this migration (e.g. right → inset-inline-end in this PR), so it'd be good to bring the Table's SCSS in line too before merging. The main ones to address:

- margin-bottom / margin-top → margin-block-end / margin-block-start
- padding-bottom → padding-block-end
- top / bottom → inset-block-start / inset-block-end
- width / height / min-width / min-height / max-width / max-height → inline-size / block-size / min-inline-size / min-block-size / max-inline-size / max-block-size

The top/bottom/height ones are arguably not RTL-sensitive, but the project is migrating to logical properties across the board so the check flags them regardless.

packages/bpk-component-comparison-table/src/BpkComparisonTable/README.md
mariamuzas: Good spot! Yeah, the readme is outdated, will update that.

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.stories.tsx
mariamuzas: (Ahh thought I fixed that...) In any case that was be the original intention, but I we are going to remove the `BpkComparisonTray` as it feels there is a bit of duplication with the Tray in the stories: https://github.com/Skyscanner/backpack/pull/4412#discussion_r3129437655

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.stories.tsx
mariamuzas: Agreed, the focus management of the tray, in this case after remove, is the consumer's responsibility, and this is just one way to do it in the story. I'll add a comment clarifying the trade-off so it doesn't get flagged.

packages/bpk-component-comparison-table/index.ts
Faye-Xiao: Ha, thanks for the context! That actually ties directly into my separate comment on the README — if the whole thing is experimental, it'd be great to surface that at the API level too, not just in the README.  Looks like we missed that.

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableColumnPlaceholder.tsx
mariamuzas: Fair point

mariamuzas: Looking into it more, I think we can get rid of the classname all together, I think it was a left over from first iterations and I was testing some bits with the sentences on the placeholder, but it's not needed anymore 🥳

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableHeader.tsx
mariamuzas: Yes I think so, we can use BpkBox here

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableHeader.tsx
mariamuzas: Ah, thank you! `BpkVisuallyHidden` works pretty well, tested locally and seems good. Added now with the some other changes: [737cecd](https://github.com/Skyscanner/backpack/pull/4412/commits/737cecddedcaead792700f012dce75c3bb6c9831)

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { BpkBox, BpkSpacing } from '../../../bpk-component-layout';
+import { BpkModalV3 } from '../../../bpk-component-modal';
+import  BpkVisuallyHidden  from '../../../bpk-component-visually-hidden';
```
Faye-Xiao, line 21: nit: There's a double space in the BpkVisuallyHidden import on this line

packages/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTableColumnHeaderPlaceholder.tsx
```diff
@@ -0,0 +1,64 @@
...
+    textStyle={TEXT_STYLES.bodyDefault}
+    color={TEXT_COLORS.textSecondary}
+    >
+    {addMoreDescription}
+  </BpkText>
+  <BpkButton
+    type={BUTTON_TYPES.link}
+    size={SIZE_TYPES.small}
+    onClick={onAddMoreClick}
+  >
```
Faye-Xiao, line 58: nit: `BpkText` and `BpkButton` are indented at 2 spaces instead of 4, and the closing tags are inconsistent

Faye-Xiao (approved):
LGTM. 
Confirmed with the PR maker that the left-nit comment will be resolved in the following PR.

PR #4419, opened 2026-04-20
[NO-JIRA] Fix :focus-visible ring in Firefox for <button> elements

MortenSorth:
### What
Extends the global `:focus-visible` rule in `bpk-stylesheets/index.scss` to also target `button` and `[type='button'|'reset'|'submit']` at matching specificity, so Backpack's focus indicator wins in Firefox.

### Why
Bundled `normalize.css@4.2.0` ships:
```css
button:-moz-focusring,
[type=button]:-moz-focusring,
[type=reset]:-moz-focusring,
[type=submit]:-moz-focusring { outline: 1px dotted ButtonText; }
```
At specificity `0-1-1`, this beats Backpack's bare `:focus-visible` rule (`0-1-0`). `:-moz-focusring` matches wherever `:focus-visible` would in Firefox, so normalize wins on every unstyled `<button>`. On dark surfaces the dotted `ButtonText` outline is near-invisible — fails WCAG 2.4.7 (Focus Visible). Firefox-only; Chrome / Safari / Edge are unaffected.

Components that ship their own `&:focus-visible` rule (BpkCardV2, BpkSwitch, BpkInsetBanner, BpkSegmentedControl, BpkComparisonTray, BpkCheckboxCard, etc.) are unaffected — their scoped rule already matches normalize's specificity. BpkButton relies on the global rule and is the primary surface hit.

This was first patched locally in [Skyscanner/global-components#3259](https://github.com/Skyscanner/global-components/pull/3259); moving the fix to the source so every Backpack consumer benefits.

### Why this approach

- **Bump `normalize.css` 4 → 8** would drop the legacy rule entirely but also changes form controls, tables, `sub`/`sup` etc. Too much unrelated blast radius for a focus-ring bug. Worth doing as a separate PR with full Percy review — tracking as a follow-up.
- **Reset `:-moz-focusring`** as a single bare rule looks tempting but sits at `0-1-0` and loses to normalize's `0-1-1`. A working reset would need the same four selectors listed here, so stating the positive intent (`:focus-visible` applies) reads better and survives future normalize churn.

### Testing

- Storybook in Firefox 145: tabbed through BpkButton on light and dark surfaces — confirmed 2px solid blue outline appears as designed.
- Storybook in Chrome and Safari: unchanged.
- `npm run build` green. `stylelint` clean via pre-commit hook.

### Follow-ups (out of scope)

- Bump `normalize.css` 4.x → 8.x with Percy review across the estate.
- Add a Firefox-specific focus-ring Percy scenario on a dark surface.
- Once this lands and publishes, remove the local workaround in [Skyscanner/global-components#3259](https://github.com/Skyscanner/global-components/pull/3259).


xiaogliu (approved):
Hey, Morten, just a comment issue, after apply Copilot suggestion, it LGTM, thanks for your contribution!

PR #4433, opened 2026-04-23
[BpkAccordion][BpkAriaLive][BpkAutosuggest] Use logical properties for RTL

gert-janvercauteren:
Replace physical CSS properties with logical equivalents so these components flow correctly in RTL without needing an explicit \`bpk-rtl\` mixin override.

- **BpkAccordionItem**: \`text-align: left\` → \`text-align: start\`; removed \`utils.bpk-rtl\` override and \`utils\` import.
- **BpkAriaLive stories**: \`margin-right\` → \`margin-inline-end\` on the chip spacing.
- **BpkAutosuggest**: \`padding-left\` → \`padding-inline-start\`; \`margin-right\` → \`margin-inline-end\` on the suggestion icon; removed the \`utils.bpk-rtl\` block and \`utils\` import.

Add \`patch\` label — visual behaviour in LTR is unchanged; RTL now relies on native logical-property flipping.

- [x] Tests (existing SCSS snapshot/visual coverage)
- [ ] Accessibility tests — keyboard/zoom/screen reader unchanged; no interaction or semantics modified

### DIFF
--- a/packages/bpk-component-autosuggest/src/BpkAutosuggest.module.scss
+++ b/packages/bpk-component-autosuggest/src/BpkAutosuggest.module.scss
@@ -20,7 +20,6 @@
 @use '../../bpk-mixins/borders';
 @use '../../bpk-mixins/layers';
 @use '../../bpk-mixins/typography';
-@use '../../bpk-mixins/utils';
 
 $arrow-size: tokens.bpk-spacing-sm() * 3;
 
@@ -104,19 +103,14 @@ $arrow-size: tokens.bpk-spacing-sm() * 3;
     padding: tokens.bpk-spacing-base();
 
     &--indent {
-      padding-left: tokens.bpk-spacing-xl();
+      padding-inline-start: tokens.bpk-spacing-xl();
     }
 
     &-icon {
       display: flex;
-      margin-right: tokens.bpk-spacing-md();
       vertical-align: top;
       fill: tokens.$bpk-text-secondary-day;
-
-      @include utils.bpk-rtl {
-        margin-right: 0;
-        margin-left: tokens.bpk-spacing-md();
-      }
+      margin-inline-end: tokens.bpk-spacing-md();
     }
 
     &-content {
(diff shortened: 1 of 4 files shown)


packages/bpk-component-autosuggest/src/BpkAutosuggest.module.scss
```diff
@@ -104,19 +103,14 @@ $arrow-size: tokens.bpk-spacing-sm() * 3;
...
     &-icon {
       display: flex;
-      margin-right: tokens.bpk-spacing-md();
       vertical-align: top;
       fill: tokens.$bpk-text-secondary-day;
-
-      @include utils.bpk-rtl {
-        margin-right: 0;
-        margin-left: tokens.bpk-spacing-md();
-      }
```
Faye-Xiao, line 119: The same physical-property + `bpk-rtl `override pattern exists in `BpkAutosuggestV2` and wasn't caught by this PR:
https://github.com/Skyscanner/backpack/blob/af3c529b70ae4da72a3bf87246d6786334e73b7a/packages/bpk-component-autosuggest/src/BpkAutosuggestV2/BpkAutosuggest.module.scss#L82-L94
```
&--indent {
  padding-left: tokens.bpk-spacing-xl();  // → padding-inline-start
}
&-icon {
  margin-right: tokens.bpk-spacing-md();  // → margin-inline-end
  @include utils.bpk-rtl {               // → can be removed
    margin-right: 0;
    margin-left: tokens.bpk-spacing-md();
  }
}
```
Would be great to include this in the same PR since it's an identical pattern in the V2 of the same component. Also, the utils import on line 23 of that file can be dropped once the bpk-rtl block is removed (assuming bpk-visually-hidden on line 144 is the only other utils usage — worth double-checking).

Separately, there are similar physical property + `bpk-rtl `override patterns in a handful of other components (`BpkDescriptionList`, `BpkPaginationList`, `BpkModalInner`, `BpkRating`) — those could follow in a separate PR if you'd prefer to keep the scope here focused.

gert-janvercauteren: Good catch — applied the same change to BpkAutosuggestV2 in bf8f85f. Confirmed utils.bpk-visually-hidden is the only other usage so the import stays. The other components (BpkDescriptionList, BpkPaginationList, BpkModalInner, BpkRating) will follow in a separate PR.

Supremeyh (approved):
lgtm

PR #4434, opened 2026-04-23
[MAT2867][BpkScrim] fix: prevent scroll-jump when opening/closing modal on iOS

lynn-sc:
Remove `requestAnimationFrame` / `setTimeout` wrappers in `withScrim` so that `storeScroll`, `fixBody`, and `lockScroll` run synchronously on mount. This ensures the body is locked before the first paint, eliminating the visible scroll-jump when opening a modal on iOS. On unmount, `unfixBody` is now called before `restoreScroll` to prevent a second visual jump on close. `lockScroll` also gains `touch-action: none` and `overscroll-behavior: contain` to block iOS rubber-band overscroll.

**Before**

https://github.com/user-attachments/assets/317fa87c-4bfa-451f-86e9-8636a9e92ed6

**After**

https://github.com/user-attachments/assets/184ac1eb-0b64-436e-9919-55a8e11361b8

- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/bpk-scrim-utils/src/scroll-utils.tsx
+++ b/packages/bpk-scrim-utils/src/scroll-utils.tsx
@@ -77,6 +77,10 @@ export const fixBody = () => {
     return;
   }
 
+  // Set top before position:fixed so the browser doesn't jump to scrollY=0.
+  // scrollOffset is captured by storeScroll() immediately before this call.
+  body.style.top = `-${scrollOffset}px`;
+  body.style.width = '100%';
   body.style.position = 'fixed';
 };
 
@@ -88,6 +92,8 @@ export const unfixBody = () => {
   }
 
   body.style.position = '';
+  body.style.top = '';
+  body.style.width = '';
 };
 
 export const lockScroll = () => {
@@ -101,6 +107,11 @@ export const lockScroll = () => {
 
   body.style.overflow = 'hidden';
   body.style.paddingRight = paddingRight;
+  // On iOS, `overflow: hidden` alone does not prevent touch-scroll or rubber-band
+  // overscroll on the body. `touch-action: none` blocks scroll gestures and
+  // `overscroll-behavior: contain` stops overscroll chaining to the viewport.
+  body.style.touchAction = 'none';
+  body.style.overscrollBehavior = 'contain';
 };
 
 export const unlockScroll = () => {
@@ -112,4 +123,6 @@ export const unlockScroll = () => {
 
   body.style.overflow = '';
   body.style.paddingRight = '';
+  body.style.touchAction = '';
+  body.style.overscrollBehavior = '';
 };
--- a/packages/bpk-scrim-utils/src/scroll-utils.tsx
+++ b/packages/bpk-scrim-utils/src/scroll-utils.tsx
@@ -98,6 +98,12 @@ export const unfixBody = () => {
   body.style.width = '';
 };
 
+// Locks background scroll on the body. Safe to call on all platforms.
+//
+// Sets `overflow: hidden` to prevent the page behind a modal from scrolling,
+// compensates for the hidden scrollbar with `padding-right`, and applies
+// `overscroll-behavior: contain` to stop scroll chaining from modal content to
+// the viewport. None of these block user touch gestures.
 export const lockScroll = () => {
   const body = getBodyElement();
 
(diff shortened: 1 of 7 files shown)


packages/bpk-scrim-utils/src/withScrim.tsx
```diff
@@ -126,8 +118,10 @@ const withScrim = <P extends object>(
       const applicationElement = getApplicationElement();
 
       if (isIphone || isIpad) {
-        setTimeout(restoreScroll, 0);
+        // unfixBody before restoreScroll: restoring scroll while body is still fixed
+        // prevents a second visual jump on close.
         unfixBody();
+        restoreScroll();
       }
       unlockScroll();
```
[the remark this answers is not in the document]
lynn-sc: https://github.com/Skyscanner/backpack/pull/4434/changes#diff-e0df0a54c1f3ab77718a532ab6ac6a4896e0cc19d34ee9a86a751e976e005a44R73

packages/bpk-scrim-utils/src/scroll-utils.tsx
```diff
@@ -112,4 +129,33 @@ export const unlockScroll = () => {
...
+  body.style.overscrollBehavior = previousOverscrollBehavior;
+};
+
+// Blocks touch gestures on the body via `touch-action: none`. iOS-only.
+//
+// On iOS Safari, `overflow: hidden` alone does not stop touch-scroll or
+// rubber-band overscroll on the body — `touch-action: none` is needed. Do NOT
+// call this on non-iOS platforms: `touch-action` combines with descendants'
+// effective touch-action, so setting `none` on body blocks touch scrolling in
+// any modal content that doesn't explicitly declare its own `touch-action`.
```
Faye-Xiao, line 163: The comment here already mentions that modal content needs to explicitly declare its own `touch-action` to stay scrollable. But I think `BpkModalInner.module.scss` and `BpkDrawerContent.module.scss` are not updated in this PR — both `__content` containers only have `overflow-y: auto` without `touch-action: pan-y`. So on iOS, users won't be able to scroll content inside a modal by touch after this change. Could we add `touch-action: pan-y t`o those scroll containers in the same PR, or at least track it as a follow-up?

packages/bpk-scrim-utils/src/scroll-utils.tsx
```diff
@@ -17,6 +17,8 @@
  */
 
 let scrollOffset = 0;
+let previousTouchAction = '';
+let previousOverscrollBehavior = '';
```
Faye-Xiao: These module-level variables work fine for a single scrim, but there is a case where they can go wrong. Backpack has an official `Nested` story in `BpkModal.stories.tsx` where two modals are open at the same time. When the inner modal mounts, it overwrites `previousTouchAction` with 'none' (because the outer modal already set it). Then when the outer modal unmounts, it restores `body.touchAction` to `'none' `instead of ''. After both modals are closed, `touch-action: none` stays on the body permanently and the whole page stops responding to touch. Is this a known limitation, or should we handle nested scrims here?

packages/bpk-scrim-utils/src/scroll-utils.tsx
lynn-sc: Added intentionally to counteract a side effect of lockTouchAction(). That function sets touch-action: none on <body> to block iOS rubber-band overscroll on the background — but touch-action propagates down the ancestor chain, which would also block touch-scrolling inside the modal content area. Adding pan-y on .bpk-scrim-content re-enables vertical scrolling there. It works because .bpk-scrim-content has overflow: auto, making it a scroll container and a boundary in the touch-action inheritance chain — the pan-y overrides the none from body for this subtree only.

packages/bpk-scrim-utils/src/scroll-utils.tsx
lynn-sc: Good catch — this is a real bug. Fixed in the latest commit with reference counting: each lock type (**lockTouchAction, lockScroll, fixBody**) now has its own depth counter. Save/restore only runs at the 0→1 and 1→0 transition, so nested mounts just bump the counter without clobbering the outer scrim's saved value. Also added a guard in **storeScroll** to skip when **fixBodyDepth > 0** (body is already **position: fixed** so **pageYOffset** reads 0 and would corrupt the outer scroll offset). Regression tests covering the Nested story scenario are added in **scroll-utils-test.ts.**

Faye-Xiao (approved):
LGTM

PR #4446, opened 2026-04-29
[CLOV-1402][token-sync] Transform Figma variables into DTCG token files

gc-skyscanner:
Adds the Figma → DTCG stage on top of CLOV-1401's Figma client.

## Summary

- `npm run sync` writes `primitives.json`, `backpack.day.json`, `backpack.night.json`, and `manifest.json` under `token-sync/tokens/` (committed, so PRs surface real Figma-side changes in the diff).
- Alias resolution: same-collection aliases inlined, cross-collection aliases preserved as `{Group.Token}` references. Cycle detection and subscribed-id fallback included.
- Deterministic output (locale-independent sort, 2-space JSON, stable manifest) so CI diffs are meaningful.

image

Test in CI 
https://github.com/Skyscanner/backpack/actions/runs/25480788610/job/74764216502 
## Test plan

- [x] `npx jest token-sync/` — all tests pass (transformer, writer, pipeline, formatter)
- [x] `npm run sync` locally with a valid Figma token produces the four files
- [x] Trigger **Actions → Sync Figma variables → Run workflow** in CI and confirm green


token-sync/src/dtcg-transformer.ts
```diff
@@ -0,0 +1,731 @@
...
+    if (existing) {
+      existing.push(variable);
+    } else {
+      map.set(variable.key, [variable]);
+    }
+  }
+  return map;
+}
+
+export function buildPreservedReferenceKeys(
```
gc-skyscanner, line 714: Returns the variable keys that should be written as {Group.Token} references rather than inlined. Any variable in another selected collection is preserved so cross-collection aliases survive for style-dictionary to resolve at build time.

Example — `Primitives (primitive)` + `Backpack (semantic)`:
When building Backpack, Primitives' keys are preserved, so:
  `Backpack/Canvas/Default` → alias → `Primitives/Colour/Pink`  becomes  `{ $value: "{Colour.Pink}" }`   (preserved reference) not  ` { $value: "#ff66b3" }  `       (inlined literal)

token-sync/src/build-dtcg.ts
```diff
@@ -0,0 +1,364 @@
...
+    const preservedReferenceKeys = buildPreservedReferenceKeys(
+      classified,
+      collection.id,
+      localVariables,
+    );
+    const context: ResolveContext = {
+      localVariablesById: localVariables,
+      localVariablesByKey,
+      localCollectionsById,
+      preservedReferenceKeys,
```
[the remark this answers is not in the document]
gc-skyscanner: In the current Backpack setup both target collections (`Primitives` and `Backpack`) are local, and `filterLocalTargets` already enforces this — so any alias from a target variable into a `remote: true ` variable would be a designer mistake against an unrelated subscribed library. That scenario doesn’t exist in our Figma file.

Even if it did, `meta.variables` reflects Figma’s current authoritative snapshot, so inlining that value isn’t “stale” — it’s what Figma says the value is. Cross-library aliases that genuinely can’t be resolved still surface via `skipUnresolvedAliases` in the build summary.

We’d revisit this if Backpack ever started subscribing to external libraries, but for now an extra filter would be defensive code with no real failure to defend against.

token-sync/src/dtcg-writer.ts
```diff
@@ -0,0 +1,213 @@
...
+//
+// The Figma file key is deliberately NOT included: this directory is
+// committed to the open-source repo, and the key identifies the private
+// source file. Downstream stages (style-dictionary etc.) only read the
+// token files, not the manifest.
+export function buildManifest(
+  outputs: DTCGModeOutput[],
+  modeCounts: ReadonlyMap<string, number>,
+  generatedAt: string = new Date().toISOString(),
+): DTCGManifest {
```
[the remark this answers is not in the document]
gc-skyscanner: Keeping `generatedAt` as-is — its value to humans reading the file outweighs the diff noise.  Will handle “is there a real change?” in the auto-PR CI workflow when we build it (e.g. `jq 'del(. generatedAt)'` before comparing, or scoping the diff to token files only).

token-sync/README.md
```diff
@@ -1,41 +1,23 @@
 # token-sync
 
-A small CLI that authenticates against Figma's [Variables REST API](https://www.figma.com/developers/api#variables)
-and fetches variable collections from the Backpack Foundations & Components file. It prints the matched
-collection names and modes — a minimal, verifiable foundation for downstream design-token sync.
+Fetches Figma variables from the Backpack Foundations & Components file and
+writes deterministic [DTCG](https://design-tokens.github.io/community-group/format/)
+JSON files under `token-sync/tokens/`.
```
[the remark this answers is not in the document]
gc-skyscanner: Updated the PR description — script is npm run sync. The tokens/ directory is committed by design so PRs surface real Figma-side changes in the diff; for the timestamp noise concern see the reply on [dtcg-writer.ts](https://github.com/Skyscanner/backpack/pull/4446#discussion_r3194310413).

token-sync/src/sync-helpers.ts
```diff
@@ -22,6 +22,16 @@ import { FigmaApiError, type LocalVariableCollection } from './figma-api';
 
 export const TARGET_COLLECTION_NAMES = ['Primitives', 'Backpack'] as const;
 
+export function sortBy<T>(items: T[], key: (item: T) => string): T[] {
```
gc-skyscanner, line 25: didn't add unit test for this function since it is covered by the cases in ` build-dtcg-test.ts` and ` dtcg-transformer-test.ts`

token-sync/src/sync-helpers.ts
```diff
@@ -63,16 +73,12 @@ export function filterLocalTargets(
   allCollections: LocalVariableCollection[],
   targetNames: readonly string[],
 ): LocalTargetFilterResult {
-  const matched = allCollections.filter(
-    (collection) =>
-      !collection.remote && targetNames.includes(collection.name),
-  );
-  const foundNames = new Set(matched.map((collection) => collection.name));
+  const localCollections = allCollections.filter((collection) => !collection.remote);
```
gc-skyscanner, line 76: no logic change here, just optimize and rename `matched` to `matchedCollections` for easier reading

token-sync/src/dtcg-transformer.ts
```diff
@@ -0,0 +1,701 @@
...
+  }
+  if (typeof value === 'string' || typeof value === 'boolean') {
+    return value;
+  }
+  throw new Error(
+    `Unsupported literal value for DTCG token type "${tokenType}": ${JSON.stringify(value)}`,
+  );
+}
+
+// Classification
```
gc-skyscanner, line 225: Maps each collection to its role (primitive | semantic) using `COLLECTION_ROLES`.
 Throws if a collection name is not in `COLLECTION_ROLES` — callers should only pass collections that have already been validated against the known target list.

token-sync/src/dtcg-transformer.ts
```diff
@@ -0,0 +1,701 @@
...
+        ).join(', ')}.`,
+      );
+    }
+    classified.push({ collection, role });
+  }
+  return classified;
+}
+
+// Alias resolution
+export function resolveAliasTarget(
```
gc-skyscanner, line 246: `resolveAliasTarget` resolves a Figma `VARIABLE_ALIAS ` id to its `LocalVariable`, handling two id schemes:

- Direct id (`"variable-colour-pink"`) — looked up in `localVariablesById`
- Subscribed id (`"VariableID:<key>/<sub>"`) — key-based fallback via `localVariablesByKey`, used for cross-library references

Returns `undefined` if no match is found; the caller decides whether to skip or throw.

token-sync/src/dtcg-transformer.ts
```diff
@@ -0,0 +1,701 @@
...
+
+  throw new Error(
+    `Could not resolve mode "${preferredModeName}" for collection "${collection.name}"`,
+  );
+}
+
+// Resolve a variable's value for a mode, following alias chains. Aliases
+// pointing at a preserved-reference variable return a `{Group.Token}` DTCG
+// reference; other aliases are inlined all the way to the literal value.
+export function resolveVariableValue(
```
gc-skyscanner, line 292: `resolveVariableValue` converts a Figma variable's value for a given mode into its DTCG representation. Three cases:

- **Literal** (colour, number, etc.) — normalised directly, e.g. `#ff66b3`, `8px`
- **Cross-collection alias** (points to a primitive) — preserved as a reference, e.g. `{Colour.Pink}`
- **Same-collection alias** (points to another semantic token) — recursively followed until a literal is reached, then inlined; `seen` tracks visited nodes to catch cycles

**Example:**
```
Canvas/Contrast (Day) → Canvas/Default (Day) → Colour/Pink  (primitive)
                         ↑ same collection, inline            ↑ cross-collection, keep as {Colour.Pink}

result: { $value: '{Colour.Pink}', inlinedFrom: 'Canvas.Default' }
```

token-sync/src/dtcg-transformer-test.ts
```diff
@@ -0,0 +1,644 @@
...
+      response.meta.variables,
+    );
+    // Primitives keys should be preserved when building Backpack outputs.
+    expect(keysForBackpack.has(KEY_COLOUR_PINK)).toBe(true);
+    expect(keysForBackpack.has(KEY_COLOUR_BERRY)).toBe(true);
+    // Backpack keys should not preserve themselves.
+    expect(keysForBackpack.has('key-semantic-canvas-default')).toBe(false);
+  });
+
+  it('returns an empty set when there is only one collection', () => {
```
gc-skyscanner, line 447: When there is only one collection, there is no "other collection", so the preserved keys set is empty.

token-sync/src/dtcg-transformer.ts
```diff
@@ -0,0 +1,701 @@
...
+  }
+  const canonicalCandidate = candidates.find((candidate) =>
+    isCanonicalLocalId(candidate.id),
+  );
+  return canonicalCandidate ?? candidates[0];
+}
+
+// Pick the mode id for a collection by preferred name, falling back to the
+// collection's default mode or (for single-mode collections) its only mode.
+export function getCollectionModeId(
```
gc-skyscanner, line 271: `getCollectionModeId` translates a mode name like "Day" into the Figma modeId
needed to read valuesByMode.

The fallback matters for cross-collection alias resolution: when "Day" tokens in Backpack point at `Primitives` variables, Primitives has no "Day" mode — so we fall back to its `default/only` mode instead of throwing.

Priority: exact name match → defaultModeId → single-mode shortcut → error.

token-sync/src/dtcg-transformer.ts
```diff
@@ -0,0 +1,701 @@
...
+}
+
+function ownerOf(tree: DTCGTree, pathKey: string): string | undefined {
+  return PATH_OWNERS.get(tree)?.get(pathKey);
+}
+
+// Breadth-first search under `node` for any recorded path owner. Used to
+// name a concrete variable when a leaf collision hits a group that was
+// created implicitly (no owner was recorded at the exact prefix path).
+function findOwnerBelow(
```
gc-skyscanner, line 395: `findOwnerBelow` improves the collision error message for a specific edge case:
when a token tries to occupy a path that already exists as an implicit group.

Groups are created silently as variables are written — "Colour/Brand/Pink" creates the group at "Colour/Brand" without recording an owner for it. So when "Colour/Brand" (a flat token) collides with that group, ownerOf() returns `undefined`, and the error would unhelpfully say "vs <existing group>".

findOwnerBelow BFS-walks the group to find any descendant with a recorded owner, giving the error a real variable name to blame instead.

token-sync/src/dtcg-writer.ts
```diff
@@ -0,0 +1,207 @@
...
+): string {
+  if (isMultiMode) {
+    return `${slugify(collectionName)}.${slugify(modeName)}.json`;
+  }
+  return `${slugify(collectionName)}.json`;
+}
+
+// Build the manifest records in the same order as the outputs so callers
+// control determinism.
+export function buildManifest(
```
gc-skyscanner, line 124: The Figma file key is deliberately NOT included: this directory is committed to the open-source repo, and the key identifies the private source file. Downstream stages (style-dictionary etc.) only read the token files, not the manifest.

token-sync/tokens/backpack.day.json
```diff
@@ -0,0 +1,645 @@
+{
```
gc-skyscanner, line 1: The files inside token-sync/ are auto-generated by running `npm run sync`

IrinaWei (commented):
### Code review

Found 1 issue (reviewed by 5/6 agents — Sass agent skipped, no SCSS files).

🤖 Generated with [Claude Code](https://claude.ai/code)

token-sync/src/dtcg-writer-test.ts
```diff
@@ -0,0 +1,329 @@
...
+      assertSafeOutputDir(path.join(process.cwd(), 'token-sync', 'tokens')),
+    ).not.toThrow();
+  });
+
+  it('rejects an empty string', () => {
+    expect(() => assertSafeOutputDir('')).toThrow(/empty/);
+  });
+
+  it('rejects the filesystem root', () => {
+    const root = path.parse(process.cwd()).root;
```
IrinaWei: Could you please take a look at this object destructuring issue?

token-sync/tokens/backpack.day.json
```diff
@@ -0,0 +1,640 @@
...
+        "$value": 900,
+        "$type": "fontWeight"
+      },
+      "Label": {
+        "$value": "700px",
+        "$type": "dimension"
+      },
+      "Subhead": {
+        "$value": "300px",
+        "$type": "dimension"
```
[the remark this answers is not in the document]
IrinaWei: Any future FLOAT weight whose scope isn’t set to `FONT_STYLE` will silently fall through to dimension + "px" with no warning. Worth a follow-up to skip or warn on ambiguous FLOAT?

token-sync/src/dtcg-writer.ts
```diff
@@ -0,0 +1,213 @@
...
+): Promise<DTCGManifest> {
+  // Guard the destructive `rm` below against obviously-wrong output dirs,
+  // and fail fast on filename collisions so we never silently overwrite.
+  assertSafeOutputDir(outputDir);
+
+  const modeCounts = countModesPerCollection(outputs);
+  const manifest = buildManifest(outputs, modeCounts, now().toISOString());
+  assertUniqueFileNames(manifest);
+
+  await rm(outputDir, { force: true, recursive: true });
```
IrinaWei: `writeDTCGFiles` does rm -rf then writes — if anything between fails (disk full, signal, EACCES), tokens/ is left half-written and git status shows a partial diff that’s easy to accidentally commit. Would we consider writing to a temp dir + rename so the old state stays intact on failure, especially before the auto-PR workflow lands?

token-sync/src/dtcg-transformer.ts
```diff
@@ -0,0 +1,731 @@
...
+            variableName: variable.name,
+            variableId: variable.id,
+            variableKey: variable.key,
+            reason: 'path-collision',
+            collidingVariableName: error.collidingVariableName,
+          });
+          // Undo the alias counters we bumped for a write that didn't land.
+          if (resolved.preservedAliasTo) preservedAliasCount -= 1;
+          if (resolved.inlinedFrom) inlinedAliasCount -= 1;
+          shouldWriteVariable = false;
```
IrinaWei: Nit: this loop tracks “did the write succeed” in two parallel ways — the shouldWriteVariable flag and the increment-then-undo on stats. Could collapse both by making setTokenAtPath return a result + only committing stats after success, then continue in each error branch. Adding a new stat field later wouldn’t need a paired undo. Not blocking.

token-sync/src/dtcg-writer-test.ts
gc-skyscanner: do you mean `path.parse(process.cwd()).root`  in the test? https://github.com/Skyscanner/backpack/pull/4446/changes#diff-f21368b5e093865f5a6a8245dc67e66956f33eab406f715be402bd539916e87aR203
it is a test input for ensuring the output dir is not the filesystem root

token-sync/src/dtcg-transformer.ts
gc-skyscanner: Good catch, updated in https://github.com/Skyscanner/backpack/pull/4446/changes/e4bac2da54313a8e613e391ba9f2b6262342f9bc

token-sync/src/dtcg-writer.ts
gc-skyscanner: fair point, updated in https://github.com/Skyscanner/backpack/pull/4446/changes/e4bac2da54313a8e613e391ba9f2b6262342f9bc

token-sync/src/dtcg-writer-test.ts
IrinaWei: Oh I just saw there's a build failure about object destructuring brought by Github. I rerun the job and it's gone now

token-sync/tokens/backpack.day.json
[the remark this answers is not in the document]
gc-skyscanner: added a log and added a field `ambiguousFloatVariables`  to record it https://github.com/Skyscanner/backpack/pull/4446/changes/cd8396ce9bbea702ed55362b944f5272c515d6da

IrinaWei (approved):

token-sync/src/dtcg-writer-test.ts
gc-skyscanner: I think you were watching the outdated code 😄

PR #4449, opened 2026-04-30
[NO-JIRA][BpkCarousel] Always show page indicator nav in stories

gert-janvercauteren:
Adds `showPageIndicatorNav` to the `Default` and `WithCarouselPageIndicator` story examples so the previous/next navigation buttons always render regardless of viewport size, making the stories (and the `VisualTest`/`VisualTestWithZoom` variants that reuse them) cover the nav UI consistently.

- [x] Storybook examples updated

Label: `skip-changelog` (stories-only change).


xiaogliu (approved):
LGTM

Supremeyh (approved):

PR #4466, opened 2026-05-08
[CLOV-1403][token-sync] Stage 2: DTCG → CSS custom properties

gc-skyscanner:
## Summary

Adds Stage 2 of the Figma → DTCG → CSS pipeline: reads the DTCG JSON files from Stage 1 and emits two theme CSS files via Style Dictionary v5.

- **CSS generation** (`tokens:build-css`) — emits `token-sync/css/theme-backpack-light.css` (`:root`) and `theme-backpack-dark.css` (`:root[data-theme="dark"]`), with `px → rem` conversion and `Component` prefix stripped from variable names
- **Safety checks** — aborts on Light/Dark asymmetry, CSS name collisions after prefix stripping, and non-`px` dimension values
- **Atomic writes** — CSS output staged in a sibling directory and renamed into place, so a mid-build failure leaves the previous committed CSS intact (mirrors the Stage 1 pattern from #4446)
- **Script names** — renamed to `tokens:fetch` / `tokens:build-css` / `tokens:sync` to make the Figma-token context clear in the monorepo

## Test plan

- [ ] `npm run tokens:build-css` builds successfully after `npm run tokens:fetch`
- [ ] `npm run tokens:sync` runs both stages end-to-end
- [ ] `npx jest token-sync/` — 171 tests pass


package.json
```diff
@@ -48,7 +48,9 @@
     "start": "npm run build && npm run storybook",
     "storybook": "storybook dev -p 9001",
     "storybook:dist": "storybook build -c .storybook -o dist-storybook",
-    "sync": "tsx token-sync/src/index.ts",
+    "tokens:fetch": "tsx token-sync/src/index.ts",
+    "tokens:build-css": "tsx token-sync/src/build-css-cli.ts",
+    "tokens:sync": "npm run tokens:fetch && npm run tokens:build-css",
     "sassdoc": "sassdoc {packages/bpk-mixins/**/*,node_modules/@skyscanner/bpk-svgs/dist/*,node_modules/@skyscanner/bpk-foundations-web/tokens/base.default}.scss -d dist-sassdoc -v --strict",
```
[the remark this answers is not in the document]
gc-skyscanner: fixed in https://github.com/Skyscanner/backpack/pull/4466/changes/426e332326e6731c2e5c17bec66f2314be9450f6

token-sync/src/build-css.ts
```diff
@@ -0,0 +1,277 @@
...
+        missing.push(filePath);
+      }
+    }),
+  );
+  if (missing.length > 0) {
+    throw new Error(
+      `Missing Stage 1 DTCG file(s):\n  ${missing.join(
+        '\n  ',
+      )}\nRun \`npm run sync\` to (re)generate them before building CSS.`,
+    );
```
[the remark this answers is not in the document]
gc-skyscanner: fixed in https://github.com/Skyscanner/backpack/pull/4466/changes/426e332326e6731c2e5c17bec66f2314be9450f6

token-sync/src/build-css.ts
```diff
@@ -0,0 +1,277 @@
...
+          outputs.push(path.join(buildDir, file.destination));
+        }
+      }
+    }, Promise.resolve());
+
+    await rm(buildDir, { force: true, recursive: true });
+    await rename(stagingDir, buildDir);
+  } catch (error) {
+    await rm(stagingDir, { force: true, recursive: true });
+    throw error;
```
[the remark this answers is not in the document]
gc-skyscanner: fixed in https://github.com/Skyscanner/backpack/pull/4466/changes/426e332326e6731c2e5c17bec66f2314be9450f6

token-sync/src/style-dictionary-config.ts
```diff
@@ -0,0 +1,523 @@
...
+// top-level `Foo` group co-existing with a `Component.Foo` group: both
+// produce `--bpk-foo-…` after stripping. SD itself wouldn't error on this —
+// it would just emit both, with the second silently winning the cascade.
+export function findCssNameCollisions(
+  tree: unknown,
+  // Reserved for future cross-file collision detection. Unused today —
+  // collisions are scoped per file because each mode's CSS is built
+  // independently — but kept on the signature so callers don't need to
+  // change shape later.
+
```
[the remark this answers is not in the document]
gc-skyscanner: fixed in https://github.com/Skyscanner/backpack/pull/4466/changes/426e332326e6731c2e5c17bec66f2314be9450f6

token-sync/src/build-css-cli.ts
```diff
@@ -0,0 +1,54 @@
...
+const scriptDir = path.dirname(fileURLToPath(import.meta.url));
+
+// Default layout:
+//   token-sync/tokens/   ← Stage 1 DTCG output (regenerated by `npm run sync`)
+//   token-sync/css/      ← Stage 2 CSS output (regenerated by `npm run build:tokens`)
+//
+// CSS lives outside `tokens/` deliberately: Stage 1's DTCG writer clears the
+// whole `tokens/` directory before each sync to drop stale JSON, and we don't
+// want that to wipe the committed CSS as a side-effect. Both paths can be
+// overridden via env vars to support custom build pipelines.
```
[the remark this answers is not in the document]
gc-skyscanner: fixed in https://github.com/Skyscanner/backpack/commit/426e332326e6731c2e5c17bec66f2314be9450f6

token-sync/src/build-css.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  await assertInputsExist(tokensDir);
+  await assertInputsAreBuildable(tokensDir);
+
+  // Stage in a sibling directory so any SD failure leaves the existing
+  // `buildDir` (committed to git) untouched. The rename at the end is the
+  // only step that mutates `buildDir`.
+  const stagingDir = `${buildDir}.staging-${process.pid}-${Date.now()}`;
+  assertSafeOutputDir(stagingDir);
+
```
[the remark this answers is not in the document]
gc-skyscanner: fixed in fixed in https://github.com/Skyscanner/backpack/pull/4466/changes/426e332326e6731c2e5c17bec66f2314be9450f6

token-sync/src/build-css-test.ts
```diff
@@ -0,0 +1,488 @@
...
+      expect(result.stderr).toContain('--bpk-button-dimension-padding-h');
+      // Both source paths should be cited so the designer can find them.
+      expect(result.stderr).toContain('Button.Dimension.padding-h');
+      expect(result.stderr).toContain('Component.Button.Dimension.padding-h');
+    } finally {
+      await cleanup();
+    }
+  });
+
+  it('rejects DTCG inputs containing non-px dimension values', async () => {
```
gc-skyscanner, line 277: `size/pxToRem` only handles px; non-px units must be caught at build time.

token-sync/src/build-css.ts
```diff
@@ -0,0 +1,302 @@
...
+        '\n  ',
+      )}\nRun \`npm run tokens:fetch\` to (re)generate them before building CSS.`,
+    );
+  }
+}
+
+// Run all Stage 2 structural checks (dimension units, CSS name collisions,
+// semantic token symmetry) off a single parse. Returns the discovered semantic
+// file names for later use in SD config building.
+async function assertInputsAreBuildable(
```
gc-skyscanner, line 169: Parse every input file once and run all the structural checks Stage 2 cares about: dimension values must be Xpx (or a DTCG alias) so the px→rem transform can handle them, and no two tokens may collapse to the same CSS variable name after Component stripping. Doing both walks off a single parse keeps the I/O cost down and lets a bad input file surface every issue it has at once instead of one per build attempt.

token-sync/src/build-css-test.ts
```diff
@@ -0,0 +1,342 @@
...
+  BACKPACK_DARK_FILE,
+  BACKPACK_LIGHT_FILE,
+  DARK_OUTPUT_FILE,
+  DARK_SELECTOR,
+  LIGHT_OUTPUT_FILE,
+  LIGHT_SELECTOR,
+  PRIMITIVES_FILE,
+} from './style-dictionary-config';
+
+// CLI is spawned via tsx (not imported) to keep SD's ESM runtime out of
```
gc-skyscanner, line 44: We invoke the CLI via tsx in a child process rather than calling unBuildCSS() directly. Style Dictionary v5 ships ESM-only and pulls in a chain of ESM-only deps (chalk, change-case, path-unified, …) that would otherwise force the whole jest suite to babel-transform every nested `node_modules`. Spawning keeps the contract end-to-end (env vars + exit code + filesystem output) and isolates SD's runtime from the test runner.

token-sync/src/style-dictionary-config.ts
```diff
@@ -0,0 +1,453 @@
...
+  return VALID_PX_LITERAL.test(value) || DTCG_ALIAS.test(value);
+}
+
+// Walk a parsed DTCG tree and return every dimension token whose `$value`
+// would not survive the px→rem assumption. `$type` cascades from the
+// nearest ancestor group that declared one (DTCG spec §7.1) — leaf tokens
+// rarely repeat `$type` themselves, so the cascade matters in practice.
+//
+// Pure: no I/O, no SD. Caller is responsible for reading + parsing the JSON.
+export function findInvalidDimensions(
```
gc-skyscanner, line 284: `findInvalidDimensions` walks the parsed DTCG tree and collects every `$type: dimension` token whose `$value` is neither a `px` literal (e.g. "`16px`") nor a DTCG alias (e.g. `{Spacing.md}`). It handles `$type` inheritance per [DTCG spec](https://www.designtokens.org/tr/drafts/format/#type-0) — leaf tokens rarely repeat `$type` themselves, so the type is tracked from the nearest ancestor group that declared it.

The result feeds `formatDimensionViolations`, which surfaces a human-readable error before Style Dictionary runs. This short-circuits the build instead of letting `size/pxToRem` silently mishandle units like `em` or `%`.

token-sync/src/build-css.ts
```diff
@@ -0,0 +1,302 @@
...
+  return { semanticFileNames };
+}
+
+// Run the SD builds for all discovered semantic token themes (light, dark, etc).
+// No logging / no process.exit — the CLI wrapper (build-css-cli.ts) handles presentation.
+export async function runBuildCSS({
+  buildDir,
+  tokensDir,
+}: BuildCSSOptions): Promise<BuildCSSResult> {
+  if (!path.isAbsolute(tokensDir)) {
```
gc-skyscanner, line 257: Absolute paths only. 
Relative paths would be resolved against  `process.cwd()` at every I/O call (readdir, mkdir, rename, rm…), and the CWD can change mid-run (e.g. SD's internal config loading). That would make the staging/backup dance and the final atomic swap point at different directories — staging written here, then `rename` from there resolved against a different CWD. Forcing absolute paths upstream keeps every subsequent path operation deterministic.

token-sync/src/build-css.ts
```diff
@@ -0,0 +1,302 @@
...
+  });
+  const outputs: string[] = [];
+
+  try {
+    await rm(stagingDir, { force: true, recursive: true });
+    await mkdir(stagingDir, { recursive: true });
+
+    // Run sequentially via reduce (avoids no-await-in-loop). See PR thread
+    // on why parallelism here offers no real speedup.
+    await configs.reduce<Promise<void>>(async (prev, { config }) => {
```
gc-skyscanner, line 289: SD’s buildAllPlatforms mutates internal state on the instance, and parallelism here would offer no measurable speedup for two small theme builds while making any failure logs interleave.  Reduce serialises the awaits without tripping the no-await-in-loop ESLint rule.

token-sync/src/style-dictionary-config.ts
```diff
@@ -0,0 +1,453 @@
...
+export const CSS_PREFIX = 'bpk';
+
+// Top-level DTCG group names that should NOT show up in the CSS variable
+// name. These come from Figma's organisational hierarchy ("Component/" is a
+// folder there) but carry no useful information for consumers — every token
+// underneath is already namespaced by the component name (Badge, Button, …).
+//
+// Compared case-insensitively to the first path segment. Strip-then-kebab
+// so e.g. `Component.Badge.Colour.bg-default` → `--bpk-badge-colour-bg-default`.
+export const STRIPPED_TOP_LEVEL_SEGMENTS: ReadonlySet<string> = new Set([
```
gc-skyscanner: Top-level DTCG group names that should NOT show up in the CSS variable name. These come from Figma's organisational hierarchy ("Component/" is a folder there) but carry no useful information for consumers — every token underneath is already namespaced by the component name (Badge, Button, …).
Compared case-insensitively to the first path segment. Strip-then-kebab so e.g. `Component.Badge.Colour.bg-default` → `--bpk-badge-colour-bg-default`.

token-sync/src/style-dictionary-config.ts
```diff
@@ -0,0 +1,453 @@
...
+    `Two or more DTCG tokens map to the same CSS custom property. Rename ` +
+    `one of the colliding sources in Figma, or extend STRIPPED_TOP_LEVEL_SEGMENTS ` +
+    `with care so the stripping no longer produces a clash.`
+  );
+}
+
+// Walk a parsed DTCG tree and collect every leaf token's dotted path.
+// `Component.Badge.Colour.bg-default` is one entry. Used by the symmetry
+// check below — pure, no I/O.
+function collectTokenPaths(tree: unknown): string[] {
```
gc-skyscanner, line 181: `collectTokenPaths` recursively walks a parsed DTCG tree and returns the dotted path of every leaf token (nodes that have a `$value`). DTCG metadata keys (`$type`, `$value`, etc.) are skipped — only structural group keys become path segments.

It’s a pure helper used exclusively by `findAsymmetricSemanticTokens`: convert each theme tree to a `Set<string>`, then diff the two sets to surface tokens that exist in one theme but not the other.

input
```
{
  "Component": {
    "Badge": {
      "Colour": {
        "$type": "color",
        "bg-default": { "$value": "#fff" },
        "bg-hover":   { "$value": "#eee" }
      }
    }
  },
  "Primitives": {
    "Blue": { "$value": "#0070f3" }
  }
}
```
output
```
[
  "Component.Badge.Colour.bg-default",
  "Component.Badge.Colour.bg-hover",
  "Primitives.Blue"
]
```

token-sync/src/style-dictionary-config.ts
```diff
@@ -0,0 +1,453 @@
...
+// Filter that keeps tokens authored in `sourceFileName` and drops the rest.
+// We use this to omit Primitives from the per-mode CSS output: they're loaded
+// as a `source` so cross-collection aliases resolve, but we don't want raw
+// primitives leaking into the theme stylesheets — only the semantic Backpack
+// tokens should produce `--bpk-*` variables.
+//
+// `filePath` is set to the absolute path SD read the file from. We compare on
+// basename so the filter remains correct regardless of how callers spell the
+// source paths (relative vs absolute, different working dir, etc.).
+export function makeBackpackTokenFilter(
```
gc-skyscanner, line 351: Filter that keeps tokens authored in `sourceFileName` and drops the rest.  We use this to omit Primitives from the per-mode CSS output: they're loaded as a `source` so cross-collection aliases resolve, but we don't want raw primitives leaking into the theme stylesheets — only the semantic Backpack tokens should produce `--bpk-*` variables. `filePath` is set to the absolute path SD read the file from. We compare on  basename so the filter remains correct regardless of how callers spell the source paths (relative vs absolute, different working dir, etc.).

token-sync/src/style-dictionary-config.ts
```diff
@@ -0,0 +1,453 @@
...
+// If semanticFileNames is omitted, reads tokensDir to auto-discover them.
+export function buildStyleDictionaryConfigs({
+  buildDir,
+  cssTransforms,
+  semanticFileNames,
+  tokensDir,
+}: BuildConfigOptions): Array<{ name: string; config: Config }> {
+  // Trailing slash is required: SD's `buildPath` is concatenated raw with the
+  // file destination, so omitting it produces paths like `/tmp/cssfile.css`.
+  const buildPath = buildDir.endsWith(path.sep) ? buildDir : `${buildDir}${path.sep}`;
```
gc-skyscanner, line 402: Trailing slash is required: SD's `buildPath` is concatenated raw with the file destination, so omitting it produces paths like `/tmp/cssfile.css`.

token-sync/css/theme-backpack-dark.css
```diff
@@ -0,0 +1,197 @@
...
+  --bpk-speech-bubble-colour-bg-default: #131d2b;
+  --bpk-switch-android-switch-default-disabled: rgba(255, 255, 255, 0.1);
+  --bpk-switch-android-switch-knob-disabled: rgba(255, 255, 255, 0.2);
+  --bpk-switch-android-switch-on-contrast-fill-disabled: rgba(255, 255, 255, 0.1);
+  --bpk-switch-android-switch-on-contrast-off: rgba(255, 255, 255, 0.8);
+  --bpk-switch-ios-switch-default-off: rgba(255, 255, 255, 0.2);
+  --bpk-switch-ios-switch-disabled: #e0e4e9;
+  --bpk-switch-ios-switch-on-contrast-off: rgba(255, 255, 255, 0.2);
+  --bpk-trip-advisor: #9ce6c0;
+  --bpk-i-os-tab-bar-fill: #243346;
```
gert-janvercauteren: We probably want to filter out iOS/Android specific tokens

token-sync/css/theme-backpack-dark.css
```diff
@@ -0,0 +1,197 @@
...
+  --bpk-badge-dimension-padding-horizontal-subtle: 0.5rem;
+  --bpk-badge-typography-tmp-badge-label: 0.875rem;
+  --bpk-badge-typography-tmp-badge-line-height: 0;
+  --bpk-button-colour-bg-destructive: #e0e4e9;
+  --bpk-button-colour-bg-destructive-pressed: #ff649c;
+  --bpk-button-colour-bg-disabled: #0b121d;
+  --bpk-button-colour-bg-feature-pressed: #d1f7ff;
+  --bpk-button-colour-bg-featured: #d1f7ff;
+  --bpk-button-colour-bg-footer: rgba(255, 255, 255, 0);
+  --bpk-button-colour-bg-footer-pressed: rgba(255, 255, 255, 0);
```
gert-janvercauteren: Should we add a `_` in front of the var name to 'mark' them private? Ideally, we have these all grouped at the bottom together or in a different :root { } tag :D (For all component tokens)

token-sync/css/theme-backpack-dark.css
gc-skyscanner: Good catch! I added the iOS / Android filters when generating CSS in https://github.com/Skyscanner/backpack/pull/4466/commits/8f2732ced9d70d9192c1a113b9be1b84b8e6f8b3 
However, I haven’t applied the same filters to the DTCG JSON files yet. Because I think the DTCG JSON should be a faithful snapshot of Figma variables, and the symmetry / collision checks only run in Stage 2 (CSS generation) anyway. What do you think?

token-sync/css/theme-backpack-dark.css
gc-skyscanner: Good point. I personally lean toward a `bpk-private-` prefix (e.g. `--bpk-private-button-...`) since CSS custom properties with `_` are technically valid but unusual, and tooling support is inconsistent. A `bpk-private-` prefix is still human-readable, grep-friendly, and signals intent clearly. I’m happy to go with this approach if you agree. Alternatively, if you prefer the separate `:root {}` block grouping instead, I can follow that way too.
What would you be more comfortable with?

token-sync/src/build-css-test.ts
```diff
@@ -0,0 +1,387 @@
...
+      expect(result.code).not.toBe(0);
+      expect(result.stderr).toMatch(
+        /semantic token\(s\) defined in only one of/,
+      );
+      expect(result.stderr).toContain(fileName);
+      expect(result.stderr).toContain('Surface.Highlight');
+    },
+  );
+
+  it('rejects DTCG inputs whose tokens collide on CSS name after Component → private rename', async () => {
```
gc-skyscanner, line 248: Top-level `Button` collides with `Component.Button` after stripping — both become `--bpk-private-button-…`. Build must refuse, not silently overwrite.

token-sync/src/style-dictionary-config.ts
```diff
@@ -0,0 +1,449 @@
...
+    tokenPath.length > 0 &&
+    typeof tokenPath[0] === 'string' &&
+    tokenPath[0].toLowerCase() === COMPONENT_SEGMENT
+      ? [PRIVATE_SEGMENT_RENAME, ...tokenPath.slice(1)]
+      : tokenPath;
+  const withPrefix =
+    typeof prefix === 'string' && prefix.length > 0
+      ? [prefix, ...renamed]
+      : renamed;
+  return withPrefix.map(kebabSegment).filter(Boolean).join('-');
```
kerrie-wu: **Nit — degenerate empty CSS name slips past the collision check**

`kebabSegment` replaces every character outside `[a-z0-9-]` with `-` and trims, so any path segment that's all non-alphanumeric kebabs down to `""` and gets dropped by `.filter(Boolean)`. If *every* segment in the path degenerates that way (e.g. `["💎"]` or `["!@#"]`), `kebabBpkName` returns `""` and SD will emit `--bpk-: <value>;` — invalid CSS that browsers silently ignore.

`findCssNameCollisions` only flags `sources.length > 1`, so a single degenerate token slips through undetected and the failure mode shifts from "build refuses" to "value silently missing in the browser."

Realistically won't trigger with current Figma authoring conventions, but a small guard turns it into a loud failure:

```ts
const result = withPrefix.map(kebabSegment).filter(Boolean).join('-');
if (!result || result === prefix) {
  throw new Error(
    `Token path ${tokenPath.join('.')} produced an empty CSS name after kebab-casing.`,
  );
}
return result;
```

(`result === prefix` covers the edge case where every original segment degenerates but the `bpk` prefix survives, leaving `--bpk:` — same outcome.)

token-sync/src/build-css.ts
```diff
@@ -0,0 +1,348 @@
...
+    await rm(stagingDir, { force: true, recursive: true });
+    throw error;
+  }
+
+  // Atomic swap: move buildDir aside (backup), promote staging to buildDir,
+  // then drop the backup. Each rename failure path cleans up its own mess so
+  // we never leak orphaned `.staging-*` / `.backup-*` dirs.
+  let backupCreated = false;
+  try {
+    await rename(resolvedBuildDir, backupDir);
```
kerrie-wu, line 315: **Nit — doc the brief ENOENT window**

Between this rename and the one at L322, `buildDir` does not exist on disk for a microsecond-scale window. The comment block above explains the failure-cleanup contract but not this one timing detail, so a future reader could be tempted to "simplify" the two-step swap (e.g. `rm -rf buildDir && rename(staging, buildDir)`) and accidentally widen the window from microseconds to the full SD build time.

The two-step approach is the right one — POSIX `rename` can't atomically overwrite a non-empty directory cross-platform — so worth a one-liner pinning down the intent:

```ts
// Note: between the two renames, `buildDir` briefly does not exist on disk.
// Fine for offline CSS builds; not safe for live-served paths. Keep the
// two-step swap so non-empty directory replacement stays portable.
```

Non-blocking — purely a future-proofing comment.

gc-skyscanner: good point, updated in https://github.com/Skyscanner/backpack/pull/4466/changes/b1807027369c1d3c9cbb6276c7838fdc65d986ad

token-sync/src/style-dictionary-config.ts
gc-skyscanner: good catch, updated in https://github.com/Skyscanner/backpack/pull/4466/changes/b1807027369c1d3c9cbb6276c7838fdc65d986ad

kerrie-wu (approved):
LGTM

PR #4475, opened 2026-05-12
[BpkRadio] Migrate JS to TypeScript

xiaogliu:
## Summary
- Convert the `bpk-component-radio` package from JavaScript/Flow to TypeScript
- Rename source/test files (`.js` → `.ts`/`.tsx`) and update the snapshot file name so Jest still finds it
- Replace `propTypes` + Flow `Props` with a typed `Props` export based on `Omit<InputHTMLAttributes<HTMLInputElement>, 'type' | 'className'>`
- Re-export `BpkRadioProps` from the package entrypoint, matching the pattern used by `bpk-component-checkbox`

## Notes
- `aria-label={(ariaLabel || label) as string}` keeps the original behavior where a `ReactNode` label is coerced to string (existing snapshot shows `aria-label="[object Object]"` for non-string labels)
- `BpkRadio.stories.tsx`: drop the now-unused `@ts-expect-error Untyped import` for `BpkRadio`
- `form-test.tsx`: add `label` (now required by TS) and drop the redundant `type="radio"` prop; cast `getByTestId('form')` to `HTMLFormElement` for the `FormData` constructor

## Test plan
- [x] `npm run jest -- packages/bpk-component-radio` (4 suites, 13 tests, 9 snapshots — all pass)
- [x] `npx tsc --noEmit` — no errors in `bpk-component-radio`
- [x] `npx eslint packages/bpk-component-radio` — no errors, no warnings


packages/bpk-component-radio/src/BpkRadio.tsx
```diff
@@ -65,13 +64,12 @@ const BpkRadio = ({
...
       despite not being in the spec. */}
       {/* eslint-disable jsx-a11y/role-supports-aria-props */}
-      {/* $FlowFixMe[cannot-spread-inexact] - inexact rest. See decisions/flowfixme.md */}
       <input
         type="radio"
         className={getClassName('bpk-radio__input')}
         name={name}
         disabled={disabled}
-        aria-label={ariaLabel || label}
+        aria-label={(ariaLabel || label) as string}
```
xiaogliu: This cast only satisfies TypeScript; it still sends non-string React nodes to `aria-label` at runtime, producing names like `[object Object]`. For the TS migration, can we model the props so non-string labels require `ariaLabel`, or otherwise only derive `aria-label` from string labels? `BpkCheckbox` already avoids this pattern.

xiaogliu (commented):
Left one inline comment on the `aria-label` cast.

packages/bpk-component-radio/src/BpkRadio.tsx
xiaogliu: Good catch — switched to the BpkCheckbox pattern in 0f4449fae: `aria-label={ariaLabel || (typeof label === 'string' ? label : undefined)}`. When the label is a node and no `ariaLabel` is supplied, no `aria-label` attribute is emitted (so the input falls back to the visible text instead of `[object Object]`). Snapshot updated accordingly.

packages/bpk-component-radio/src/__snapshots__/BpkRadio-test.tsx.snap
```diff
@@ -56,7 +56,6 @@ exports[`BpkRadio should render correctly when label is a node 1`] = `
   >
     <input
       aria-invalid="false"
-      aria-label="[object Object]"
```
xiaogliu: Removed in 0f4449fae alongside the BpkRadio change to only derive `aria-label` from string labels.

Previously, when `label` was a ReactNode (e.g. `<span>Direct</span>`) and no `ariaLabel` was provided, the node was passed straight to `aria-label` and the DOM stringified it to the literal text `"[object Object]"` — which is what this snapshot was capturing.

After the fix (`aria-label={ariaLabel || (typeof label === 'string' ? label : undefined)}`), no `aria-label` attribute is emitted in this case, so the input falls back to the visible label text via the wrapping `<label>` instead of an unhelpful `"[object Object]"` accessible name.

gert-janvercauteren: @xiaogliu did/can we do the same for BpkCheckbox?

packages/bpk-component-radio/src/BpkRadio.stories.tsx
kerrie-wu: nit : this event: any is pre-existing, but now that BpkRadio is typed it can just be removed — TS will infer it as ChangeEvent<HTMLInputElement>.

xiaogliu: Done in a3bc7537e — TS now infers `event` as `ChangeEvent<HTMLInputElement>` from the typed `onChange` prop. Thanks!

packages/bpk-component-radio/src/__snapshots__/BpkRadio-test.tsx.snap
xiaogliu: Yes, `BpkCheckbox` already uses `aria-label={typeof label === 'string' ? label : undefined}` https://github.com/Skyscanner/backpack/blob/main/packages/bpk-component-checkbox/src/BpkCheckbox.tsx#L86 (it doesn't expose an `ariaLabel` prop)

kerrie-wu (approved):
LGTM

PR #4478, opened 2026-05-13
[NO-JIRA][BpkCheckboxV2] Allow safe prop forwarding on atomic parts

IrinaWei:
## Summary

Open up safe prop forwarding on `BpkCheckboxV2`'s atomic parts (`Root`, `Control`, `Label`, `Description`, `HiddenInput`) so consumers can wire up `data-testid`, `aria-label`/`aria-labelledby`/`aria-describedby`, `id`, and other native HTML attributes — while keeping `className` and `style` strictly locked down so Backpack retains ownership of the visual layer.

## What changed

- Each part's prop type is now built on the matching `HTMLAttributes` / `InputHTMLAttributes` interface with `className` and `style` (and the props the V2 wrapper already manages) removed via `Omit`.
- A defensive runtime strip + dev-only `console.warn` catches any `as any` bypass that smuggles `className` / `style` through.
- `BpkCheckboxV2HiddenInputProps` is now exported from the package index alongside the existing part types.
- `Indicator` is unchanged — it returns `null`, so there is no element to forward props to.
- No shorthand `<BpkCheckboxV2 label="..." />` API in this PR — tracked separately.
- V1 `BpkCheckbox` is untouched.

## Test plan

- [x] `jest packages/bpk-component-checkbox/src/BpkCheckboxV2/BpkCheckboxV2-test.tsx` — 15/15 pass, including 3 new tests covering forwarding + className/style stripping
- [x] `eslint packages/bpk-component-checkbox/src/BpkCheckboxV2 --ext .ts,.tsx` clean
- [x] `tsc --noEmit` clean for the checkbox package
- [ ] Reviewer to sanity-check that consumer-facing types disallow `className` / `style` (the test file exercises this with `@ts-expect-error`)

### DIFF
--- a/packages/bpk-component-checkbox/index.ts
+++ b/packages/bpk-component-checkbox/index.ts
@@ -32,3 +32,4 @@ export type { BpkCheckboxV2RootProps, BpkCheckboxV2CheckedState } from './src/Bp
 export type { BpkCheckboxV2ControlProps } from './src/BpkCheckboxV2/BpkCheckboxV2Control';
 export type { BpkCheckboxV2LabelProps } from './src/BpkCheckboxV2/BpkCheckboxV2Label';
 export type { BpkCheckboxV2DescriptionProps } from './src/BpkCheckboxV2/BpkCheckboxV2Description';
+export type { BpkCheckboxV2HiddenInputProps } from './src/BpkCheckboxV2/BpkCheckboxV2HiddenInput';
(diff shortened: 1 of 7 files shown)


IrinaWei:
Closing in favour of a more minimal scope — see follow-up PR which only adds `data-testid` on Root and `aria-label` / `aria-labelledby` / `aria-describedby` on HiddenInput, instead of opening up safe pass-through on every part.

PR #4486, opened 2026-05-14
[CLOV-119] Migrate bpk-component-progress JS to TypeScript

xiaogliu:
## Summary
- Convert the `bpk-component-progress` package from JavaScript/Flow to TypeScript
- Rename source/test files (`.js` → `.ts`/`.tsx`) and the snapshot file so Jest still finds it
- Replace `propTypes` + Flow `Props` with a typed `Props` export, re-exported as `BpkProgressProps`
- Type `Props` as `Omit<HTMLAttributes<HTMLDivElement>, 'className'> & { ... }` so native div attrs flow through `...rest` while keeping `className?: string | null`
- Replace `propTypes` + `defaultProps` with `static defaultProps`; destructure `onComplete`/`onCompleteTransitionEnd` out of `rest` instead of the legacy `delete rest.x` workaround
- `tabIndex="0"` → `tabIndex={0}` so the prop type-checks; rendered DOM is unchanged (`tabindex="0"`)
- Type both class containers in the stories file with explicit `Props`/`State` generics
- Narrow `getValueText` return type from `unknown` to `string` (addresses review feedback) — non-string returns would have been DOM-stringified to `"[object Object]"` and announced by screen readers

## Notable type narrowings (theoretical breaking, runtime unchanged)
- `tabIndex={null}` no longer type-checks — the new `Props` inherits `tabIndex?: number` from `HTMLAttributes<HTMLDivElement>` (was Flow `?number`). Pass `undefined` or omit the prop instead. Runtime DOM behavior is unchanged.
- `getValueText` must now return `string` (was Flow `mixed`). Existing valid usages already do.

## Test plan
- [x] `npm run jest -- packages/bpk-component-progress` (3 suites, 10 tests, 5 snapshots all pass)
- [x] `npx tsc --noEmit` — no errors in `bpk-component-progress`
- [x] `npx eslint packages/bpk-component-progress` — no errors, no warnings


packages/bpk-component-progress/src/BpkProgress.tsx
```diff
@@ -132,19 +119,17 @@ class BpkProgress extends Component<Props> {
...
-    delete rest.onCompleteTransitionEnd;
-
     return (
-      // $FlowFixMe[cannot-spread-inexact] - inexact rest. See 'decisions/flowfixme.md'.
       <div
         className={classNames}
         role="progressbar"
-        aria-valuetext={getValueText ? getValueText(value, min, max) : null}
+        aria-valuetext={
+          getValueText ? (getValueText(value, min, max) as string) : undefined
```
kerrie-wu: **`as string` masks an a11y contract gap** *(recurring pattern)*

`getValueText` is typed `(...) => unknown` at L58, so `as string` here only silences TS. At runtime a consumer returning a non-string (number, object, ReactNode) is coerced to `"[object Object]"` and announced to assistive tech — visually invisible, but an a11y regression.

Same anti-pattern flagged on #4475 (BpkRadio): _"This cast only satisfies TypeScript; it still sends non-string React nodes to `aria-label` at runtime, producing names like `[object Object]`."_

**Fix** — narrow the source type instead of casting at the use site:

```tsx
// L58
getValueText?: ((value: number, min: number, max: number) => string) | null;

// L127
aria-valuetext={getValueText ? getValueText(value, min, max) : undefined}
```

The README example already returns a string, so no break for valid usage; bad callers like `() => 42` now fail at compile time instead of producing `[object Object]` at runtime.

kerrie-wu (commented):
🤖 Generated with [Claude Code](https://claude.ai/code)

packages/bpk-component-progress/src/BpkProgress.tsx
```diff
@@ -47,42 +44,30 @@ const renderSteps = (numberOfSteps) => {
...
 };
 
-type Props = {
-  max: number,
-  min: number,
-  value: number,
-  stepped: boolean,
-  small: boolean,
-  className: ?string,
-  tabIndex: ?number,
```
kerrie-wu, line 57: Nit: tabIndex no longer accepts null.

Old Flow type was tabIndex: ?number (accepts null); the new TS Props inherits tabIndex?: number from HTMLAttributes<HTMLDivElement>, so tabIndex={null} is now a TS error. Unlikely to bite in practice — tabIndex={null} is uncommon — but worth a one-liner in the PR description for completeness

packages/bpk-component-progress/src/BpkProgress.tsx
xiaogliu: Fixed in c54f20e14 — narrowed `getValueText` return type to `string` and dropped the `as string` cast. Non-string returns now fail at compile time.

packages/bpk-component-progress/src/BpkProgress.tsx
xiaogliu: Good catch — added a note to the PR description under "Notable type narrowings". Decided not to widen `Props.tabIndex` back to `number | null` so the type stays consistent with the inherited `HTMLAttributes<HTMLDivElement>` shape (matches BpkRadio etc.). Runtime behavior is unchanged.

kerrie-wu (approved):
LGTM

PR #4491, opened 2026-05-18
[NO-JIRA][BpkCheckboxCard] Cars variant pixel parity with origin BpkPanel

RichardSyq:
## Summary

Brings `BpkCheckboxCard` cars variant to pixel-parity with the legacy `BpkPanel`-based carhire chip, so consumers can drop their workaround SCSS overrides. Scoped to the `cars` variant — other variants are unchanged.

## Changes

- **Indicator alignment** — Removed `withButtonAlignment` from the corner tick. The HOC injects an inline `margin-top: 0.25rem` meant for inline button icons, which pushes the absolutely-positioned indicator down 4px from where SCSS expects it.
- **Stroke via inset `box-shadow`, not `border`** — Default unchecked / hover / disabled / focus all use inset shadows so the chip's outer width matches origin and the stroke is never clipped by ancestor `overflow: hidden` (e.g. inside `BpkMobileScrollContainer`).
- **Touch checked / `:active` replicates origin's `inset 2px + 0.2px border`** — On `@media (hover: none) and (pointer: coarse)` the selected stroke is the same 0.5px-rendered border + 2px inset combo that origin ships on mobile/tablet. Pixel-perfect carhire parity, including origin's intentional 1px width shift on selection. Touch detection avoids the trap of viewport-based queries (zooming a desktop into <1024px CSS would otherwise trigger mobile styling).
- **`:active` paints the checked stroke immediately** — At touchstart, masking the 200ms `box-shadow` transition so React state updates feel instant.
- **`:focus-visible` becomes an inset `text-link` shadow** — Replaces the base outline that gets clipped by `overflow: hidden` and stays sticky on iOS Safari after a touch tap.
- **New Storybook story** — `WithCarVariantMobile` shows the row-layout cars chip inside `BpkMobileScrollContainer`, the canonical carhire mWeb deployment.

| Before (origin/main) | After (this PR) |
|---|---|
| image| image|

## Consumer impact

After this PR, carhire's `.CategoryChip` SCSS workarounds (`box-shadow` overrides on `__selected`, the `> span { margin-top: 0 !important }` reset) can be deleted.

## Validation

- `npx stylelint`, `npx eslint`, `npx jest` on the package — all green.
- Chrome DevTools MCP at mWeb 3x DPR: origin (port 5000) `Medium` chip checked = `outerW 159.141px`, `border 0.5px solid rgb(5, 32, 60)`, `box-shadow inset 0 0 0 2px rgb(5, 32, 60)`. Storybook with this patch produces identical values.

## Test plan

- [ ] `WithCarVariantMobile` story at mobile viewport: 2.5px navy stroke (border + inset), tick at top-right, no clipping inside scroll container.
- [ ] Same story at desktop viewport with zoom 150% / 200%: stroke stays at 2px (touch detection is viewport-independent).
- [ ] Real iOS device tap → no sticky blue ring after deselection.
- [ ] `WithIndicator` story still renders the manual `Indicator` correctly.


packages/bpk-component-checkbox-card/src/BpkCheckboxCard.module.scss
```diff
@@ -119,26 +119,68 @@
...
   &--cars {
-    // White background, border changes to primary color on hover/checked
+    // Inset box-shadow instead of border: outer width matches origin, never
+    // clipped by ancestor overflow:hidden (e.g. BpkMobileScrollContainer).
+    transition:
+      background-color tokens.$bpk-duration-sm ease-in-out,
+      box-shadow tokens.$bpk-duration-sm ease-in-out,
+      color tokens.$bpk-duration-sm ease-in-out;
+    border: none;
+    box-shadow: inset 0 0 0 tokens.$bpk-border-size-sm tokens.$bpk-line-day;
```
kerrie-wu: The `box-shadow` values here (and in `[data-state='checked']` and `:has(:focus-visible)` below) hardcode `_day` tokens:

```scss
box-shadow: inset 0 0 0 tokens.$bpk-border-size-sm tokens.$bpk-line-day;
// ...
box-shadow: inset 0 0 0 (2 * tokens.$bpk-border-size-sm) tokens.$bpk-core-primary-day;
// ...
box-shadow: inset 0 0 0 (2 * tokens.$bpk-border-size-sm) tokens.$bpk-text-link-day;
```

The original implementation used CSS custom properties (`--bpk-checkbox-card-default-border-color` etc.) intentionally — they allow consumers to override values via `BpkThemeProvider` (e.g. night-mode tokens). Switching to hardcoded `_day` tokens means the border colour will stay light regardless of the active theme.

Suggest keeping the CSS variable + fallback pattern:

```scss
box-shadow: inset 0 0 0 tokens.$bpk-border-size-sm
  var(--bpk-checkbox-card-default-border-color, #{tokens.$bpk-line-day});
```

and setting the corresponding CSS variables inside `--cars` so a future night-mode extension only needs to override those variables.

packages/backpack-web/src/bpk-component-checkbox-card/src/BpkCheckboxCard.module.scss
```diff
@@ -119,26 +119,68 @@
...
+        tokens.$bpk-core-primary-day;
+
+      @media (hover: none) and (pointer: coarse) {
+        border: (tokens.$bpk-one-pixel-rem * 0.2) solid
+          tokens.$bpk-core-primary-day;
+      }
+    }
+
+    // Paint the checked stroke on touchstart so the transition doesn't lag.
+    &:active:not([data-disabled]):not([data-loading]) {
```
kerrie-wu, line 182: The PR description says `:active` makes the checked stroke appear instantly on touchstart, but the component-level `transition: box-shadow ${bpk-duration-sm} ease-in-out` is still in effect here — there's no override to clear it. The box-shadow will still animate in, not snap.

To actually achieve instant feedback, add `transition: none` inside this block:

```scss
&:active:not([data-disabled]):not([data-loading]) {
  transition: none; // suppress animation on touchstart
  box-shadow: inset 0 0 0 (2 * tokens.$bpk-border-size-sm) tokens.$bpk-core-primary-day;
  ...
}
```

packages/bpk-component-checkbox-card/src/BpkCheckboxCard.module.scss
RichardSyq: Good catch — restored. Latest commit re-introduces `--bpk-checkbox-card-{default,hover,checked}-border-color` declarations on the cars block and routes the inset `box-shadow` / touch border colours through `var(..., fallback)`. Default visual unchanged because each fallback resolves to the previously hardcoded token; consumers can now override the cars stroke colours via `BpkThemeProvider`. Focus ring kept hardcoded (`text-link-day`) since there is no existing component-level CSS variable for the focus colour and it is intentionally a system-level a11y indicator.

packages/backpack-web/src/bpk-component-checkbox-card/src/BpkCheckboxCard.module.scss
RichardSyq: You are right — the parent `transition: box-shadow ${bpk-duration-sm}` was still in effect, so the `:active` shadow was animating in despite the rule. Latest commit adds `transition: none` inside the `:active` block so the border (which was already instant — not in the transition list) and the inset shadow now both paint together at touchstart. The PR description claim of instant feedback is now actually true.

kerrie-wu (approved):

PR #4496, opened 2026-05-19
[CLOV-1409] Automate Figma variable sync PR creation

IrinaWei:
## Summary
- Add daily and manual Figma variable sync automation that skips runs with no meaningful token changes.
- Cancel in-flight runs when a new sync triggers (`concurrency.cancel-in-progress: true`), so an urgent token update can preempt a stale run instead of queueing behind it.
- Build CSS only when fetched token values changed, then create a fresh automated PR for generated token/CSS updates.
- Open the new automated PR before closing superseded `figma-token-sync/*` PRs, so a failed push or PR creation can't leave the repo with no open token-sync PR.
- Add a release-label classifier for generated token diffs, using `minor` only when the diff is purely additive (new token paths), and `major` when an existing token path was changed, removed, or renamed.
- Default the generated PR to `major` if label classification fails, so the sync can still reach human review.
- Document the CI credentials and sync behaviour in `token-sync/README.md`.

### Trigger workflow manually
Set ref to current ref to trigger workflow dispatch (**would change back to main before merge**)
Manually triggered action: https://github.com/Skyscanner/backpack/actions/runs/26266827654/job/77311851919
Generated PR：https://github.com/Skyscanner/backpack/pull/4531

### Trigger workflow by schedule
Need to be tested after this PR is merged

## Notes
- Checkout currently uses `${{ github.ref }}` so this branch can be manually selected in Actions for workflow testing. We can switch it back to `main` before merge if we want the final scheduled workflow to always force a main checkout.
- The schedule cron is temporarily set to `*/30 * * * *` for validation. The production cadence will be **daily at 18:00 GMT** (`0 18 * * *`)

### DIFF
--- a/.github/workflows/sync-figma-variables.yml
+++ b/.github/workflows/sync-figma-variables.yml
@@ -145,7 +145,8 @@ jobs:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
           NEW_BRANCH: ${{ steps.commit.outputs.branch_name }}
         run: |
-          open_prs="$(gh pr list --state open --limit 100 --json number,headRefName,url --jq --arg new "$NEW_BRANCH" '.[] | select(.headRefName | startswith("figma-token-sync/")) | select(.headRefName != $new) | [.number, .headRefName, .url] | @tsv')"
+          open_prs="$(gh pr list --state open --limit 100 --json number,headRefName,url \
+            | jq -r --arg new "$NEW_BRANCH" '.[] | select(.headRefName | startswith("figma-token-sync/")) | select(.headRefName != $new) | [.number, .headRefName, .url] | @tsv')"
 
           if [ -z "$open_prs" ]; then
             echo "No superseded Figma token sync pull requests."
--- a/.github/workflows/sync-figma-variables.yml
+++ b/.github/workflows/sync-figma-variables.yml
@@ -1,11 +1,18 @@
 name: Sync Figma variables
 
 on:
+  schedule:
+    # Weekly polling sync, every Monday at 09:00 UTC.
+    - cron: '0 9 * * 1'
   workflow_dispatch:
     inputs:
       file_key:
         description: 'Optional Figma file key override (defaults to the FIGMA_FILE_KEY repo secret).'
         required: false
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
 
 permissions: {}
 
@@ -17,6 +24,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          ref: ${{ github.ref }}
           persist-credentials: false
 
       - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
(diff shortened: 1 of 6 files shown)


.github/workflows/sync-figma-variables.yml
```diff
@@ -1,12 +1,19 @@
 name: Sync Figma variables
 
 on:
+  schedule:
+    # Temporarily poll every 30 minutes for validation; the production cadence will be daily at 18:00 GMT.
+    - cron: '*/30 * * * *'
```
IrinaWei, line 6: I can't test scheduled work unless it's merged

.github/workflows/sync-figma-variables.yml
```diff
@@ -17,6 +24,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          ref: ${{ github.ref }}
```
IrinaWei: Would change it to main before merge

.github/workflows/sync-figma-variables.yml
```diff
@@ -1,12 +1,19 @@
...
+    - cron: '*/30 * * * *'
   workflow_dispatch:
     inputs:
       file_key:
         description: 'Optional Figma file key override (defaults to the FIGMA_FILE_KEY repo secret).'
         required: false
 
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
```
gc-skyscanner: Out of curiosity, why don't we set it as true here?

IrinaWei: 🤔 Since it will run on a relatively low cadence, there’s little benefit in cancelling an in-progress run. We can still manually cancel a long-running workflow from the GitHub Actions UI.

gc-skyscanner: but when there is a running workflow, and we want to manual trigger a new one since there is another new important token updated. then there will be two running workflow. and we have to manual cancel the previous one.
what is the negative effect of setting it as true?

.github/workflows/sync-figma-variables.yml
```diff
@@ -32,5 +40,139 @@ jobs:
           FIGMA_FILE_KEY: ${{ github.event.inputs.file_key || secrets.FIGMA_FILE_KEY }}
         run: npm run tokens:fetch
 
+      - name: Detect meaningful fetched token changes
+        id: token_changes
+        run: |
+          token_status="$(git status --porcelain --untracked-files=all -- token-sync/tokens)"
+          untracked_tokens="$(git ls-files --others --exclude-standard -- token-sync/tokens)"
```
gc-skyscanner: is it possible that we do this? seems we will already get the log when getting `token_status` and can just grep  untracked_tokens from the log
```suggestion
untracked_tokens="$(printf '%s\n' "$token_status" | grep '^?? ' || true)"
```

.github/workflows/sync-figma-variables.yml
```diff
@@ -32,5 +40,139 @@ jobs:
...
+
+          if [ -z "$token_status" ]; then
```
gc-skyscanner: I would think, every time we run the token `tokens:fetch`, `generatedAt` will be changed. so I suspect if we need to verify `token_status`, since I assume it would be `true` everytime. and we could just skip this check and verify the change when ignoring `generatedAt`

.github/workflows/sync-figma-variables.yml
```diff
@@ -32,5 +40,139 @@ jobs:
...
+            echo "No superseded Figma token sync pull requests."
+            exit 0
+          fi
+
+          while IFS=$'\t' read -r number _branch _url; do
+            gh pr close "$number" --comment "Closing this automated token sync PR because a newer run will open a fresh PR with the latest cumulative Figma token changes against main."
+            echo "Closed superseded Figma token sync pull request: ${_url}"
+          done <<< "$open_prs"
+
+      - name: Commit generated token changes
```
gc-skyscanner, line 77: seens we close a PR first, then genereate a new one. what will happen if we get error bwteen closing a PR and creating a new PR? would it be safer if we close the PR after creating the new one?

.github/workflows/sync-figma-variables.yml
RichardSyq, line 6: Curious what cadence we want to merge with here. The comment/README says the production cadence will be daily at 18:00 GMT, while the PR summary says weekly, but the current cron would keep running every 30 minutes after this lands. Could we switch this to the final cadence before merging, e.g. `0 18 * * *` if daily is intended?

.github/workflows/sync-figma-variables.yml
RichardSyq: I know this was already called out in the thread, but could we make the checkout ref `main` before merging? Since the generated PR is always opened with `--base main`, a manual dispatch from a non-main ref would create the `figma-token-sync/*` branch from that branch and could include unrelated changes in the automated PR. Gating PR creation to `main` would also work if we still want branch-based testing.

.github/workflows/sync-figma-variables.yml
```diff
@@ -32,5 +40,139 @@ jobs:
...
+
+          if [ -z "$open_prs" ]; then
+            echo "No superseded Figma token sync pull requests."
+            exit 0
+          fi
+
+          while IFS=$'\t' read -r number _branch _url; do
+            gh pr close "$number" --comment "Closing this automated token sync PR because a newer run will open a fresh PR with the latest cumulative Figma token changes against main."
+            echo "Closed superseded Figma token sync pull request: ${_url}"
+          done <<< "$open_prs"
```
RichardSyq: One thing I'm wondering about: should we close superseded `figma-token-sync/*` PRs only after the new PR has been created successfully? Right now the workflow closes the existing PRs before pushing/opening the replacement, so a later failure in `git push`, label application, or `gh pr create` could leave us with no open token-sync PR. Creating the new PR first, then closing the old ones while excluding the new branch, would make the handoff a bit safer.

.github/workflows/sync-figma-variables.yml
```diff
@@ -32,5 +40,139 @@ jobs:
...
+
       - name: Stage 2 - Build CSS from DTCG tokens
+        if: steps.token_changes.outputs.changed == 'true'
         run: npm run tokens:build-css
+
+      - name: Detect generated token changes
+        id: generated_changes
+        if: steps.token_changes.outputs.changed == 'true'
+        run: |
+          changed_files="$(git status --porcelain -- token-sync/tokens token-sync/css)"
```
gc-skyscanner: A stupid question, why do we check again for the `token-sync/tokens`? since we already checked the changes states at last step?

.github/workflows/sync-figma-variables.yml
IrinaWei: Sorry I forgot to update the PR summary. I think whether would it be changed to the final cadence depends on when it will be merged. Because the final cadence is daily and is midnight 2:00 for China, if it would be merged tomorrow morning, I would have to wait for another day to get the result. For that case, I won't change to the final cadence.

.github/workflows/sync-figma-variables.yml
IrinaWei: Yeah I think we can change it to true as well. We won't get throughout log if canceled in the middle but that's OK

.github/workflows/sync-figma-variables.yml
IrinaWei: Changed the order. Now we have new pr first then would close the suspended PRs

RichardSyq (approved):
LGTM

PR #4498, opened 2026-05-19
fix(BpkInsetBannerSponsored): remove per-item icons from bottom sheet, make title optional

villanuevadani:
## Summary

- Removes the hardcoded `ViewIcon` (item 0) and `InfoIcon` (subsequent items) that were rendered inside each bottom-sheet section of `BpkInsetBannerSponsored` V2. The DSA Legal Names spec calls for icon-free section content across all surfaces.
- Makes `bottomSheetContent[].title` optional (`title?: string`) so a title-less first section (e.g. the DSA intro body) is a valid shape.
- Replaces `key={item.title}` with a stable `key={item.title ?? \`section-\${index}\`}` to avoid React key collisions when a title is absent.
- The CTA trigger `(i)` icon that opens the bottom sheet is **unchanged** — this PR only affects the per-item icons inside the sheet.
- Deletes the `&--bottom-sheet-icon` SCSS rule since the wrapper it styled is gone.
- Updates tests to assert icon absence and adds a DSA-shape story (title-less intro + two titled sections).

## Visual change

Each bottom-sheet item previously rendered a leading icon (`ViewIcon` for the first, `InfoIcon` for the rest). After this PR, items render text only. This is a visible change for all consumers.

## Known consumers

| Repo | File | Impact |
| --- | --- | --- |
| `carhire-homepage` | `libs/cars-results/features/src/Adverts/InlinePlusAd/InlinePlusAd.tsx` | Currently on a local V3 fork (QUAR-1531 work). PR B of that work restores V2 and uses the new title-optional shape. Aligns with DSA Legal Names epic. |
| `web-platform` | `libs/flights-booking-panel/.../SponsoredPricingItem.tsx` | Passes a single bottom-sheet item — today renders `ViewIcon`. After this PR it shows title + description only. Cross-team heads-up sent to Sponsored Airlines team. |

## Checklist

- [x] Tests updated and passing (`npm test -- src/bpk-component-inset-banner`)
- [x] Snapshots updated for iconless variants
- [x] Story added: `WithDsaStyleBottomSheetExampleV2Light`
- [x] `common-types.ts` — `title` widened to optional
- [x] SCSS — `&--bottom-sheet-icon` rule removed
- [x] Lint clean

### DIFF
--- a/packages/backpack-web/src/bpk-component-inset-banner/src/BpkInsetBannerV2/BpkInsetBannerSponsored-test.tsx
+++ b/packages/backpack-web/src/bpk-component-inset-banner/src/BpkInsetBannerV2/BpkInsetBannerSponsored-test.tsx
@@ -111,7 +111,7 @@ describe('BpkInsetBanner', () => {
     expect(screen.getByText('Consectetur adipiscing elit')).toBeInTheDocument();
   });
 
-  it('should not render an info icon when there is only one item in bottom sheet content', () => {
+  it('should not render view or info icons in bottom sheet items', () => {
     render(
       <BpkInsetBannerSponsored
         title="Lorem ipsum"
@@ -120,14 +120,15 @@ describe('BpkInsetBanner', () => {
           text: 'Sponsored',
           bottomSheetContent: [
             {
-              title: 'Single item title',
-              description: 'Single item description',
-            }
+              title: 'First item title',
+              description: 'First item description',
+            },
+            {
+              title: 'Second item title',
+              description: 'Second item description',
+            },
           ],
           bottomSheetTitle: 'About this advert',
-          closeBtnIcon: true,
-          labelTitle: true,
-          bottomSheetLabel: 'Info',
           buttonCloseLabel: 'Close',
         }}
         backgroundColor="#FFE300"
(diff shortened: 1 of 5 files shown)


gert-janvercauteren:
Hi @villanuevadani, thanks for the PR. Do you think you could use the newer V3 banner component and wire up your own Bottomsheet in your local codebase? 

This sponsored version where we have a hard link to a bottomsheet is not where we plan to be in the future and we would like to (as soon as feasible) move away from it.

villanuevadani:
Hey @gert-janvercauteren , thanks for your look! 

Actually good point, this is something I wanted to share with my team before moving on, but I appreciate your view here!

Would you be good with updating the `Sponsored` one to use the V3 approach? Or you would discourage having this `Sponsored` version completely?

If you would like us to remove the `Sponsored` and go for this approach, we have two consumers of this:
- `carhire-homepage` -> Inline Plus Ad
- `web-platform` -> Booking Panel Inline Ad

So the path to follow from our side would be:
1. Integrate InlinePlus Ad in CarHire with the V3 custom usage (no creation of V3 `Sponsored I mean`)
2. Integrate in Booking Panel Ad (same)
3. Once these are away from SponsoredV2, we could get rid of it in Backpack

WDYT?

villanuevadani:
Won't do, thanks @gert-janvercauteren !

PR #4500, opened 2026-05-20
[CLOV-106][BpkPagination] Migrate JS to TypeScript

xiaogliu:
## Summary
- Convert the `bpk-component-pagination` package from JavaScript to TypeScript
- Replace `propTypes` + `defaultProps` with typed `Props` exports across the four sub-components (`BpkPagination`, `BpkPaginationList`, `BpkPaginationPage`, `BpkPaginationNudger`, `BpkPaginationBreak`)
- Rename source/test files (`.js` → `.ts`/`.tsx`) and the snapshot files so Jest still finds them
- Re-export `BpkPaginationProps` from `index.ts`
- Drop the now-unused `@ts-expect-error` suppression on the `BpkPagination` import in the stories file

JIRA: [CLOV-106](https://skyscanner.atlassian.net/browse/CLOV-106)

## Test plan
- [x] `npm run jest -- packages/backpack-web/src/bpk-component-pagination` — 6 suites, 15 tests, 4 snapshots, all green
- [x] `npx tsc --noEmit` — no errors in `bpk-component-pagination`
- [x] `npx eslint packages/backpack-web/src/bpk-component-pagination` — no errors, no warnings

[CLOV-106]: https://skyscanner.atlassian.net/browse/CLOV-106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/backpack-web/src/bpk-component-pagination/src/BpkPagination.tsx
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import PropTypes from 'prop-types';
+import type { HTMLAttributes } from 'react';
```
RichardSyq, line 19: Curious about the decision to drop `prop-types` entirely here — Constitution §V mentions keeping both TypeScript types and prop-types during migration, so I'm wondering if that was intentional or just missed? Happy to be corrected if there's a reason it's safe to remove them in one go (e.g. no JS consumers left for this package). If not, might be worth keeping them alongside the new interfaces and cleaning them up in a follow-up PR 🙂

xiaogliu: Good catch on the Constitution reference. A bit of context on why I dropped both prop-types and defaultProps here:

React 19 has removed them. Per the [React 19 upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops):

> In React 19, we’re removing the propType checks from the React package, and using them will be silently ignored. […] We’re also removing defaultProps from function components in place of ES6 default parameters.

So keeping them would be either no-op or actively broken under React 19. Recent migrations (BpkProgress, BpkRadio, BpkFormValidation, BpkAutosuggest) all dropped them in one step for the same reason.

Constitution §V looks out of date. The “maintain both” bullet predates React 19 and the recent precedent. Suggested path:

Leave this PR as-is.
I’ll open a follow-up PR to update Constitution §V so the policy matches reality.
WDYT?

RichardSyq: Makes sense — React 19 context is a good call and the precedent from recent migrations checks out. LGTM. Looking forward to the Constitution §V update!

RichardSyq (approved):
lgtm

PR #4505, opened 2026-05-20
[CLOV-1571][BpkLayout] Expand layout component API with opacity, as, dir, and Stack aliases

Faye-Xiao:
## Summary

During Backpack adoption in [carhire-homepage], an audit of layout component usage revealed props that were silently passing through Chakra but were not formally declared in the Backpack layout API. This PR adds official support for those props.

### Changes

**\`BpkCommonLayoutProps\` (all layout components):**
- \`opacity?: number\` — CSS opacity value (0–1), used for fade/disabled effects
- \`as?: ElementType\` — polymorphic rendering, allows changing the underlying HTML element (e.g. \`as="section"\`, \`as="nav"\`)
- \`dir?: 'ltr' | 'rtl' | 'auto'\` — text direction, for embedding bidirectional content within an LTR page (e.g. rendering Arabic numerals inside an LTR layout)
- \`marginBlock?: BpkResponsiveValue<BpkSpacingValue>\` — CSS logical property for block-axis (vertical) margin; RTL-safe equivalent of \`marginTop\` + \`marginBottom\`; supports responsive values
- \`paddingBlock?: BpkResponsiveValue<BpkSpacingValue>\` — CSS logical property for block-axis (vertical) padding; RTL-safe equivalent of \`paddingTop\` + \`paddingBottom\`; supports responsive values

**\`BpkStack\` / \`BpkHStack\` / \`BpkVStack\`:**
- \`alignItems\` — semantic alias for \`align\` (maps to CSS \`align-items\`); supports responsive values
- \`justifyContent\` — semantic alias for \`justify\` (maps to CSS \`justify-content\`); supports responsive values
- When both alias and canonical prop are provided, canonical (\`align\`/\`justify\`) takes precedence — specifically, canonical wins when it is non-nullish; if canonical is \`undefined\`, the alias value is used
- Bug fix: alias props no longer override HStack/VStack built-in defaults when not provided
- \`BpkStackSpecificProps\` uses \`Omit<BpkFlexGridProps, 'alignItems' | 'justifyContent'>\` to make the alias declarations explicit (no reliance on declaration-order shadowing)

**\`forwardRef\` type:**
[...]

PR #4506, opened 2026-05-20
[CLOV-1571][BpkLayout] Fix layout component type system to enforce allowlist-only props

Faye-Xiao:
## Summary

Fixes a type system bug where Backpack layout components silently accepted any Chakra UI prop due to a denylist-based TypeScript approach. This PR replaces the denylist with an explicit allowlist, so unsupported props now cause compile-time errors in consumer codebases — enabling teams to discover and migrate away from unsupported usage via `tsc --noEmit`.

**Depends on:** #4505 (must be merged first)

### Root cause

The previous approach used `extends Omit<RemoveCommonProps<BoxProps>, ...>` which inherited all Chakra props and only blocked specific ones. Any prop not on the denylist was silently valid TypeScript — and was also passed through to Chakra at runtime by `processBpkComponentProps`.

### Changes

**`types.ts`:**
- Removed `RemoveCommonProps<T>`, `LayoutEventProps`, `DisallowedShorthandProps` helper types
- `BpkBoxSpecificProps`, `BpkFlexSpecificProps`, `BpkGridSpecificProps`, `BpkGridItemSpecificProps`, `BpkStackSpecificProps` — all changed from `extends RemoveCommonProps<ChakraType>` to explicit allowlist-only interfaces

**`BpkCardV2.stories.tsx`:**
- Removed invalid `columns` prop (not a `BpkGrid` prop — only exists on Chakra's `SimpleGrid`)
- Changed `alignItems="end"` → `align="end"` on `BpkVStack`

### Verification

After applying this PR on top of #4505, running `tsc --noEmit` in carhire-homepage surfaces the remaining unsupported prop usages that teams need to migrate.

---

- [x] Tests
- [ ] Component `README.md`
- [ ] Accessibility tests

### DIFF
--- a/packages/backpack-web/src/bpk-component-card/src/BpkCardV2.stories.tsx
+++ b/packages/backpack-web/src/bpk-component-card/src/BpkCardV2.stories.tsx
@@ -254,7 +254,6 @@ const FlightLeg = ({
   stops: string;
 }) => (
   <BpkGrid
-    columns={4}
     templateColumns="1fr 1fr 1fr 1fr"
     gap={BpkSpacing.Base}
     align="center"
--- a/packages/backpack-web/src/bpk-component-layout/src/types.ts
+++ b/packages/backpack-web/src/bpk-component-layout/src/types.ts
@@ -30,57 +30,7 @@ import type {
 } from '@chakra-ui/react';
 
 
-/**
- * Layout-level event props that should not be exposed on layout components.
- * onClick is handled via BpkCommonLayoutProps; onFocus/onBlur are reintroduced
- * on BpkBoxProps only.
- */
-type LayoutEventProps =
-  | 'onMouseEnter'
-  | 'onMouseLeave'
-  | 'onMouseOver'
-  | 'onMouseOut'
-  | 'onMouseDown'
-  | 'onMouseUp'
-  | 'onFocus'
-  | 'onBlur'
-  | 'onKeyDown'
-  | 'onKeyUp'
-  | 'onKeyPress';
 
-/**
- * Shorthand props from the underlying layout system that we do NOT expose on
- * Backpack layout components. These mostly mirror longer-form spacing,
- * sizing and visual props that we already model explicitly via
- * BpkCommonLayoutProps and BpkFlexGridProps.
- */
-type DisallowedShorthandProps =
-  // Spacing shorthands
-  | 'p'
-  | 'pt'
-  | 'pr'
-  | 'pb'
-  | 'pl'
-  | 'px'
-  | 'py'
-  | 'm'
-  | 'mt'
-  | 'mr'
-  | 'mb'
-  | 'ml'
-  | 'mx'
-  | 'my'
-  // Size shorthands
-  | 'w'
-  | 'h'
-  | 'minW'
-  | 'maxW'
-  | 'minH'
-  | 'maxH'
-  // Visual shorthands that map to props we have intentionally excluded
-  | 'bg'
-  | 'rounded'
-  | 'shadow';
 
 /**
  * Flexbox & grid layout props that we explicitly support on Backpack layout
@@ -162,28 +112,11 @@ type BpkBoxResponsiveLayoutProps = {
 type BpkBoxResponsiveLayoutPropKeys = keyof BpkBoxResponsiveLayoutProps;
 
 /**
- * Base type that removes common layout props, reserved props (className,
- * children) and all layout-level event props from Chakra UI props.
- *
- * These will be replaced with Backpack-specific types.
- */
-export type RemoveCommonProps<T> = Omit<
-  T,
-  | keyof BpkCommonLayoutProps
-  | 'className'
-  | 'children'
-  | LayoutEventProps
-  | FlexGridPropKeys
-  | DisallowedShorthandProps
->;
-
-/**
- * Component-specific props for BpkBox
- * Includes all Box props except those in BpkCommonLayoutProps
+ * Component-specific props for BpkBox.
+ * Explicit allowlist — does NOT inherit from Chakra BoxProps.
  */
 export interface BpkBoxSpecificProps
-  extends Omit<RemoveCommonProps<BoxProps>, BpkBoxResponsiveLayoutPropKeys>,
-    BpkBoxResponsiveLayoutProps,
+  extends BpkBoxResponsiveLayoutProps,
     Omit<BpkFlexGridProps, BpkBoxResponsiveLayoutPropKeys> {}
 
 /**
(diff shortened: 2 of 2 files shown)


packages/backpack-web/src/bpk-component-card/src/BpkCardV2.stories.tsx
```diff
@@ -254,13 +254,12 @@ const FlightLeg = ({
...
 }) => (
   <BpkGrid
-    columns={4}
     templateColumns="1fr 1fr 1fr 1fr"
     gap={BpkSpacing.Base}
     align="center"
   >
     <AirlineLogo />
-    <BpkVStack alignItems="end" gap={BpkSpacing.None}>
+    <BpkVStack align="end" gap={BpkSpacing.None}>
```
Supremeyh: it was changed to `align` but [Line283](https://github.com/Skyscanner/backpack/blob/5b666a6e87d5c2a1cfefe13a9ab09ff1aa911198/packages/backpack-web/src/bpk-component-card/src/BpkCardV2.stories.tsx#L283) still use `alignItems`, might bring confusion

packages/backpack-web/src/bpk-component-layout/src/types.ts
```diff
@@ -258,7 +191,7 @@ export type BpkVesselProps = {
  * Component-specific props for BpkFlex
  * Includes all Flex props except those in BpkCommonLayoutProps
  */
-export interface BpkFlexSpecificProps extends RemoveCommonProps<FlexProps> {
+export interface BpkFlexSpecificProps {
```
Supremeyh: Stale doc comments on the renamed interfaces 
- [types.ts:191-192](packages/backpack-web/src/bpk-component-layout/src/types.ts#L191-L192) (`BpkFlexSpecificProps`) still says `Includes all Flex props except those in BpkCommonLayoutProps` — false now (zero inheritance).
- Same staleness on `BpkGridSpecificProps`, `BpkGridItemSpecificProps`, `BpkStackSpecificProps`.
- Only `BpkBoxSpecificProps` got the new `Explicit allowlist — does NOT inherit from Chakra BoxProps.` line. Update the others to match — this is exactly the kind of type-vs-comment drift that tripped up the original denylist design.

packages/backpack-web/src/bpk-component-layout/src/types.ts
```diff
@@ -341,14 +274,12 @@ type BpkStackOptions = {
...
  * Component-specific props for BpkStack
- * Includes all Stack props except those in BpkCommonLayoutProps
  * Overrides StackOptions to support BpkResponsiveValue.
  * `alignItems` and `justifyContent` are accepted as semantic aliases for `align` and `justify`.
  * If both are provided, `align`/`justify` take precedence.
  */
 export interface BpkStackSpecificProps
-  extends Omit<RemoveCommonProps<StackProps>, StackOptionKeysType>,
-    BpkStackOptions,
+  extends BpkStackOptions,
```
Supremeyh, line 287: After this PR, `BpkStackSpecificProps` ([types.ts:282-291](packages/backpack-web/src/bpk-component-layout/src/types.ts#L282-L291)) extends:
- `BpkStackOptions.align` (responsive)
- `BpkFlexGridProps.alignItems` (non-responsive, from the explicit declaration in [types.ts:96](packages/backpack-web/src/bpk-component-layout/src/types.ts#L96))
- own `alignItems?: BpkStackOptions['align']` alias (responsive — overrides the one inherited from `BpkFlexGridProps` because of declaration order)

Same goes for `justifyContent`. This works, but it's subtle and depends on declaration order beating inheritance. Two suggestions:
1. `Omit` the colliding keys from `BpkFlexGridProps` when extending — makes it explicit.
2. Add a one-liner comment on the alias declarations explaining "this intentionally shadows `BpkFlexGridProps.alignItems` to make the prop responsive."

Faye-Xiao:
Addressed all three review comments from @Supremeyh in commit [147b5a295](https://github.com/Skyscanner/backpack/commit/147b5a295):

1. **BpkCardV2.stories.tsx consistency** — Changed `alignItems="start"` → `align="start"` on line 283 to match `align="end"` on line 262.

2. **Stale doc comments** — Updated `BpkFlexSpecificProps`, `BpkGridSpecificProps`, `BpkGridItemSpecificProps`, and `BpkStackSpecificProps` to say *"Explicit allowlist — does NOT inherit from Chakra XProps."*, matching the `BpkBoxSpecificProps` pattern.

3. **`BpkStackSpecificProps` type collision** — Changed `extends ... BpkFlexGridProps` to `extends ... Omit<BpkFlexGridProps, 'alignItems' | 'justifyContent'>` so the responsive alias declarations below unambiguously replace the non-responsive `BoxProps` variants, with an explanatory comment.

packages/backpack-web/src/bpk-component-card/src/BpkCardV2.stories.tsx
Faye-Xiao: Good catch\! Fixed — the fix for line 283 was actually moved to the base branch (`layout-expand-unsupported-props`, [PR #4505](https://github.com/Skyscanner/backpack/pull/4505)) since that's where the `alignItems` alias was first introduced. Now both lines 262 and 283 consistently use the canonical `align` prop.

packages/backpack-web/src/bpk-component-layout/src/types.ts
Faye-Xiao: Fixed — updated all four interfaces (`BpkFlexSpecificProps`, `BpkGridSpecificProps`, `BpkGridItemSpecificProps`, `BpkStackSpecificProps`) to use the same `Explicit allowlist — does NOT inherit from Chakra XProps.` pattern as `BpkBoxSpecificProps`. Good point about comment drift being exactly the kind of thing that caused confusion with the original denylist design.

packages/backpack-web/src/bpk-component-layout/src/types.ts
Faye-Xiao: Both suggestions taken — went with option 1 (`Omit`) and also updated the JSDoc on the alias declarations. To keep the PR boundary clean, the `Omit` change was actually moved to the base branch (`layout-expand-unsupported-props`, [PR #4505](https://github.com/Skyscanner/backpack/pull/4505)) since that's where the aliases were first introduced. This PR just carries the updated comment explaining the intent.

packages/backpack-web/src/bpk-component-card/src/BpkCardV2.stories.tsx
Faye-Xiao: Actually reverted this change — since `alignItems` is now an officially supported alias in the base PR (#4505), having one line use `align` and another use `alignItems` is no longer inconsistent. Both are valid, and keeping the original form avoids unnecessary churn.

Supremeyh (dismissed):

Supremeyh (approved):

PR #4509, opened 2026-05-20
[NO-JIRA][nx] Promote backpack-web to a buildable Nx project

gert-janvercauteren:
## Summary

Implements **Issue 3.1** of the Nx adoption plan — replaces the gulp/run-s orchestrated build with a buildable Nx project for `backpack-web` and wires up its quality-gate targets in one go.

### Build pipeline

- New `packages/backpack-web/project.json` with the full Nx target chain: `generate-svgs → copy-normalize-css → build-sass → build-stylesheets → build`, plus `nx-release-publish` rooted at `packages/backpack-web/dist`.
- Build outputs now land inside the package (`packages/backpack-web/dist`) so the published artifact matches `nx release` semantics. Legacy `tsconfig.declaration.json` `outDir`/`rootDir` and the four post-build transpilation scripts are updated to follow.
- Adds an `exports` map on `packages/backpack-web/package.json` (source-time only) so sibling Nx projects can resolve `@skyscanner/backpack-web/...` against `./src/...` before the package is built.
- New `scripts/transpilation/strip-exports.js` runs at the end of the build and **deletes** the `exports` field from the dist `package.json`. The published artifact has historically had no `exports` field and relies on Node's legacy filesystem resolution; stripping the field keeps that contract bug-compatible with the v21.0.1 published shape.
- Root `package.json` `build` script now bridges to `nx run backpack-web:build`; legacy `build:*` and `transpile:*` sub-scripts are removed (subsumed by the Nx target).
- `.github/workflows/release.yml` updated to publish from `packages/backpack-web/dist` and skip the dead `npm run transpile` step.

### Quality gates

- Adds `typecheck`, `lint-js`, `lint-scss`, composite `lint`, and `test` Nx targets.
- New `packages/backpack-web/{tsconfig.json,tsconfig.lib.json}` enable composite-mode TypeScript project references; `tsconfig.solutions.json` references the new `tsconfig.lib.json`.
- New per-project `packages/backpack-web/jest.config.js` (TZ=Etc/UTC, `rootMode: 'upward'` Babel resolution, project-relative module name mappers).
- `.eslintignore` and `.stylelintignore` extended to exclude the new `packages/backpack-web/dist` build output so lint passes after a build.

### CI

- Aligns `_build.yml`'s PercyTests dist-storybook cache restore key with the upload key in `pr.yml` / `main.yml` / `release.yml` (both now hash `packages/backpack-web/**` and `libs/backpack-storybook-utils/**`). Pre-existing drift introduced when storybook-utils was extracted; surfaced by this PR.

This corresponds to [Issue 3.1](../blob/main/.context/bpk-nx-adoption-plan.md) in `.context/bpk-nx-adoption-plan.md`.

## Test plan

- [x] \`nx run backpack-web:build\` produces \`packages/backpack-web/dist/\` with transpiled JS, \`.d.ts\`, copied CSS, and copied bpk-mixins/bpk-stylesheets.
- [x] A second \`nx run backpack-web:build\` invocation is a full cache hit (5 / 5 cached).
- [x] Build does not mutate any tracked files (\`git status\` clean against \`HEAD\` for tracked paths).
- [x] \`packages/backpack-web/dist/package.json\` has **no** \`exports\` field (matches v21.0.1's published shape; legacy filesystem resolution).
- [x] \`packages/backpack-web/package.json\` (source-time) keeps an \`exports\` map pointing at \`./src/...\` so workspace-internal bare imports resolve to source.
- [x] \`nx run backpack-web:typecheck\` passes (composite \`tsc --build\`).
- [x] \`nx run backpack-web:lint-js\` passes (0 errors).
- [x] \`nx run backpack-web:lint-scss\` passes (excludes built dist).
- [x] \`nx run backpack-web:test\` runs Jest against \`packages/backpack-web/jest.config.js\` with \`TZ=Etc/UTC\`.
- [ ] CI \`Build Backpack\` step (\`npm run build\`) runs the new Nx pipeline.
- [ ] PercyTests cache restore matches Create-Build-Cache key after the \`_build.yml\` fix.
- [ ] Release dry-run from \`packages/backpack-web/dist\` succeeds (deferred to 5.2 reworking the workflow).


scripts/transpilation/strip-exports.js
```diff
@@ -0,0 +1,48 @@
...
+const distPackageJsonPath = path.join(
+  'packages',
+  'backpack-web',
+  'dist',
+  'package.json',
+);
+
+const pkg = JSON.parse(fs.readFileSync(distPackageJsonPath, 'utf8'));
+
+delete pkg.exports;
```
RichardSyq, line 39: Quick check on this script — the description and test plan describe a different operation from what the body actually does, so wanted to make sure I'm reading it right.

The PR description says the script "rewrites the `./src/...` source-time `exports` map into the published `./...` shape (with extension swap and explicit icon/stylesheet wildcards)", and the test-plan checkbox says "exports point at compiled `.js` (e.g. `./bpk-component-icon` → `./bpk-component-icon/index.js`) and include the icon/stylesheet wildcard fallbacks". But the body is a one-liner — `delete pkg.exports` — and the inline comment above says the goal is to strip the field so the published package keeps the legacy filesystem resolution contract.

Just confirming: was the strategy intentionally changed from "rewrite to compiled paths" to "delete entirely so consumers fall back to filesystem resolution"? On `main`, `packages/backpack-web/package.json` has no `exports` / `main` / `module`, so deleting the field does keep things bug-compatible — that part looks fine. The mismatch is just between the PR description / test-plan checkbox / file name (which all imply rewriting) and the actual implementation.

Would be good to either update the description + test-plan checkbox + maybe rename the file to `strip-exports.js`, or actually rewrite if that was the original plan and we lost it somewhere. No blocker either way — just want it not to look misleading to the next person reading git blame.

RichardSyq (commented):
One question on `rewrite-exports.js` — see inline.

scripts/transpilation/strip-exports.js
gert-janvercauteren: Yep, deliberate switch — the original plan was rewrite-to-compiled-paths, but during review we realised the published package has no `exports` field, so adopting one would be a behaviour change for consumers (with `exports` Node skips legacy filesystem resolution). Stripping keeps the published shape bug-compatible while still letting the source-time map drive workspace-internal resolution.

Sorry for the doc/code drift. Pushed:

- 6bf72f8b9f — renamed `scripts/transpilation/rewrite-exports.js` → `strip-exports.js` and updated the `build` target.
- PR description and test-plan checkboxes updated to describe stripping (and call out the source-time `exports` map being kept).

RichardSyq (approved):
LGTM

PR #4521, opened 2026-05-21
[CLOV-102] BpkPrice Migrate JS to TypeScript

xiaogliu:
## Summary
- Convert the `bpk-component-price` package from JavaScript/Flow to TypeScript
- Rename source/test files (`.js` → `.ts`/`.tsx`) and the snapshot file so Jest still finds it
- Replace `propTypes` + Flow `Props` with a typed `Props` export, re-exported from `index.ts` as `BpkPriceProps`
- `Props` extends `Omit<HTMLAttributes<HTMLDivElement>, 'className'>` so native div attrs flow through `...rest` while keeping `className?: string \| null`
- `SIZES` / `ALIGNS` use `as const` so consumers get literal-typed values
- Drop the now-unused `@ts-expect-error` suppressions on the `BpkPrice` and `common-types` imports in the stories file

## Notes for reviewer
- In `BpkPrice-test.tsx`, the existing `const icon = NewWindowIcon` passes a component reference (not a JSX element). Flow accepted this; TS does not. To keep the migration pure, I cast to `ReactNode` (`as unknown as ReactNode`) so the existing snapshots — which already render an empty `<span class="bpk-price__icon ...">` because React silently does nothing with a function child — continue to match. Worth a follow-up to either drop the icon test cases or pass `<NewWindowIcon />` and refresh the snapshots.

## Test plan
- [x] `npm run jest -- packages/backpack-web/src/bpk-component-price` (4 suites, 92 tests, 64 snapshots all pass)
- [x] `npx tsc --noEmit` — no errors in `bpk-component-price`
- [x] `npx eslint packages/backpack-web/src/bpk-component-price` — no errors, no warnings


packages/backpack-web/src/bpk-component-price/src/BpkPrice-test.tsx
```diff
@@ -15,21 +15,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/* @flow strict */
+
+import type { ReactNode } from 'react';
 
 import { render, screen } from '@testing-library/react';
 
+
```
Faye-Xiao: nit: Line 18,23 --  blank line

packages/backpack-web/src/bpk-component-price/src/BpkPrice.tsx
```diff
@@ -27,26 +25,28 @@ import { SIZES, ALIGNS } from './common-types';
...
+  trailingText?: string | null;
+  previousPrice?: string | null;
+  icon?: ReactNode;
+  dataAttributes?: Record<string, string>;
 };
 
 const getClassName = cssModules(STYLES);
 
-const getPriceTextStyle = (size: $Values<typeof SIZES>) => {
+const getPriceTextStyle = (size: Props['size']) => {
```
Faye-Xiao: Nit: `Props['size']` resolves to `'xsmall' | 'small' | 'medium' | 'large' | undefined`,
but neither function handles `undefined` — passing it silently falls through to the
default return value. Since `size` already has a default param in the component,
`undefined` will never actually reach here, but the signature is wider than needed.

Consider tightening to `NonNullable<Props['size']>` (or the raw union directly) so
TS catches any future callsite that passes `undefined` at compile time rather than
silently returning `heading5` / `xs` at runtime.

packages/backpack-web/src/bpk-component-price/src/BpkPrice.tsx
```diff
@@ -27,26 +25,28 @@ import { SIZES, ALIGNS } from './common-types';
...
 
-type Props = {
-  price: string,
-  size: $Values<typeof SIZES>,
-  align: $Values<typeof ALIGNS>,
-  className: ?string,
-  leadingText: ?string,
+type NativeDivProps = HTMLAttributes<HTMLDivElement>;
+
+export type Props = Omit<NativeDivProps, 'className'> & {
```
Faye-Xiao, line 30: Question: are we intentionally expanding the public API surface here?

Previously the Flow type only described BpkPrice-specific props, while this now inherits all native `HTMLAttributes<HTMLDivElement>` props. Since `BpkPriceProps` is exported from the package, consumers can now rely on native props like `role`, `tabIndex`, `onClick`, `aria-*`, etc.

That may be fine if this matches Backpack’s TS migration pattern, but it would make those pass-through props part of the public contract rather than just an implementation detail. Could we confirm that’s intentional?

packages/backpack-web/src/bpk-component-price/src/BpkPrice-test.tsx
```diff
@@ -15,21 +15,23 @@
...
+import BpkPrice, { type Props as BpkPriceProps } from './BpkPrice';
 import { ALIGNS, SIZES } from './common-types';
 
 const price = '£1,830';
 const previousPrice = '£2,000';
 const leadingText = 'from';
 const trailingText = 'per day';
-const icon = NewWindowIcon;
-let props;
+const icon = NewWindowIcon as unknown as ReactNode;
```
Faye-Xiao: nit: Could we avoid casting the icon component to `ReactNode` here?

`BpkPrice` renders `icon` as a React child, so the prop should ideally receive an element, e.g. `<NewWindowIcon />`, rather than the component function itself. The `as unknown as ReactNode` cast makes the test pass type-checking but could hide an invalid usage pattern.

Faye-Xiao:
Hi @xiaogliu , Snapshots are discouraged — since this file is already being touched, can we convert the `toMatchSnapshot()` calls to explicit `screen.getBy*` / `toHaveClass` assertions?

packages/backpack-web/src/bpk-component-price/src/BpkPrice.tsx
xiaogliu: Yes, intentional — this matches the established Backpack TS migration pattern. [BpkPagination](https://github.com/Skyscanner/backpack/blob/main/packages/backpack-web/src/bpk-component-pagination/src/BpkPagination.tsx#L33) and BpkCheckbox both extend native HTML attribute types via `Omit<HTMLAttributes<...>, 'className'>`. The runtime behaviour is unchanged (the component already `{...rest}`-spread onto the root `<div>` under Flow); the type just makes that contract explicit so callers can pass `role`, `tabIndex`, `onClick`, `aria-*`, etc. without a `@ts-expect-error`.

packages/backpack-web/src/bpk-component-price/src/BpkPrice-test.tsx
xiaogliu: Done — removed the extra blank lines on lines 18 and 23.

packages/backpack-web/src/bpk-component-price/src/BpkPrice.tsx
xiaogliu: Done — tightened both helper signatures to `NonNullable<Props['size']>`. TS will now catch any future callsite that passes `undefined` at compile time instead of silently falling through to the default return value.

packages/backpack-web/src/bpk-component-price/src/BpkPrice-test.tsx
xiaogliu: Done — replaced the `NewWindowIcon as unknown as ReactNode` cast with `<NewWindowIcon />` rendered inline. The test now passes an actual React element, which matches the intended usage of the `icon` prop.

xiaogliu:
Done — replaced all `toMatchSnapshot()` calls with explicit `screen.getByText`, `toHaveClass`, and `toBeInTheDocument` assertions, and deleted the snapshot file. Each test now documents exactly what it checks.

Faye-Xiao:
Hi @xiaogliu, I checked the Percy build and found a slight change on the bpk-price-ranger visual test, as shown in the screenshot below. Could you double-check if it's a flaky test or a real dom change, as price ranger uses the BpkPrice? 

Screenshot 2026-06-05 at 18 55 06

xiaogliu:
> Hi Vincent Liu (@xiaogliu), I checked the Percy build and found a slight change on the bpk-price-ranger visual test, as shown in the screenshot below. Could you double-check if it's a flaky test or a real dom change, as price ranger uses the BpkPrice?
> 
> Screenshot 2026-06-05 at 18 55 06

I didn’t see there is any difference, just Percy flaky.

Faye-Xiao:
Approved the precy failure https://percy.io/09e69e44/web/backpack/builds/50523019?utm_campaign=09e69e44&utm_content=backpack&utm_source=github_status_public as the failure came from flaky tests.

Faye-Xiao (approved):

PR #4523, opened 2026-05-21
[CLOV-64] BpkStarRating Migrate JS to TypeScript

xiaogliu:
## Summary
- Convert the `bpk-component-star-rating` package from JavaScript/Flow to TypeScript
- Rename source/test files (`.js` -> `.ts`/`.tsx`) and snapshot files (`*-test.js.snap` -> `*-test.tsx.snap`) so Jest still finds them under the new test filenames
- Replace `propTypes` + Flow `Props` with typed `Props` exports for `BpkStar`, `BpkStarRating`, `BpkInteractiveStar`, `BpkInteractiveStarRating`, re-exported from `index.ts` as `BpkStarProps`, `BpkStarRatingProps`, `BpkInteractiveStarProps`, `BpkInteractiveStarRatingProps`
- Convert `STAR_TYPES` to `as const` and derive a `StarType` union
- Migrate `withInteractiveStarRatingState` to a generic class component (`P extends object`) with a `public static displayName: string` field
- Drop the now-unused `@ts-expect-error` suppressions on the index/component imports in the stories file
- No business-logic, JSX, or rendering changes - existing snapshots remain valid

## Test plan
- [x] `npm run jest -- packages/backpack-web/src/bpk-component-star-rating` - 7 suites, 42 tests, 30 snapshots all pass
- [x] `npx tsc --noEmit` - no errors in `bpk-component-star-rating`
- [x] `npx eslint packages/backpack-web/src/bpk-component-star-rating` - no errors, no warnings


packages/backpack-web/src/bpk-component-star-rating/src/withInteractiveStarRatingState.tsx
```diff
@@ -16,47 +16,51 @@
...
     }
 
-    onRatingSelect = (rating: number, event: Function) => {
+    onRatingSelect = (
+      rating: number,
+      event?: MouseEvent<HTMLButtonElement> | null,
+    ) => {
       if (event) {
-        event.persist();
+        (event as SyntheticEvent).persist();
```
Faye-Xiao: nit: `event.persist() `was needed in React 16 to opt out of event pooling, but React 17 removed synthetic event pooling entirely — it's been a no-op ever since.

The as SyntheticEvent cast is only here to make TS accept .persist() on a MouseEvent. Both the cast and the call can go:

```
onRatingSelect = (
  rating: number,
  event?: MouseEvent<HTMLButtonElement> | null,
) => {
  const callback = () => {
    if (this.props.onRatingSelect) {
      this.props.onRatingSelect(rating, event);
    }
  };

  this.setState(() => ({ rating }), callback);
};
```

packages/backpack-web/src/bpk-component-star-rating/src/withInteractiveStarRatingState.tsx
```diff
@@ -79,7 +83,7 @@ const withInteractiveStarRatingState = (
     render() {
       return (
         <InteractiveStarRating
-          {...this.props}
+          {...(this.props as P)}
```
Faye-Xiao: Nit: worth destructuring out the HOC-owned prop before spreading, so it's clear only `P` props are passed down to InteractiveStarRating:

```
render() {
  const { onRatingSelect, ...rest } = this.props;
  return (
    <InteractiveStarRating
      {...(rest as P)}
      rating={this.state.rating}
      hoverRating={this.state.hoverRating}
      onRatingHover={this.onRatingHover}
      onMouseLeave={this.onMouseLeave}
      onRatingSelect={this.onRatingSelect}
    />
  );
}
```
`as P` is still needed (TS can't infer `Omit<P & WithInteractiveStarRatingStateProps, 'onRatingSelect'> = P`), but the intent is more explicit.

Faye-Xiao:
Hi @xiaogliu, we're moving away from snapshot tests — they tend to pass silently after an accidental `--updateSnapshot` and don't communicate intent. Since this is a JS→TS migration with no behaviour changes, the snapshots are still valid, but it's a good opportunity to replace them with explicit assertions.

For example, in BpkStar-test.tsx:


// instead of
`expect(asFragment()).toMatchSnapshot();`

// prefer
```
const { getByRole } = render(<BpkStar type={STAR_TYPES.FULL} />);
expect(getByRole('img', { hidden: true })).toHaveClass('bpk-star--filled');
30 occurrences across 5 test files — not asking to do all of them in this PR, but worth a follow-up ticket if not already tracked.
```

packages/backpack-web/src/bpk-component-star-rating/src/BpkStar.tsx
```diff
@@ -98,8 +96,7 @@ const BpkStar = ({
 
   if (type === STAR_TYPES.HALF) {
     return (
-      // $FlowFixMe[cannot-spread-inexact] - inexact rest. See decisions/flowfixme.md
-      <span className={[containerClassNames, halfIconClassNames]} {...rest}>
+      <span className={[containerClassNames, halfIconClassNames] as unknown as string} {...rest}>
```
Faye-Xiao: Hi @xiaogliu , could you improve this code as it's also a type question? 
 
`as unknown as string is hiding a real type error` — React's className expects a string, not an array. The cast makes TS happy but at runtime this becomes `bpk-star__container bpk-star__container--half-star,bpk-star bpk-star--half bpk-star--filled` (comma-separated), which would break the styling.

packages/backpack-web/src/bpk-component-star-rating/src/BpkStarRating.tsx
```diff
@@ -47,17 +45,23 @@ export const ROUNDING_TYPES = {
...
-  className: ?string,
-  large: boolean,
-  extraLarge: boolean,
-  maxRating: number,
-  rating: number,
-  rounding:
-    | typeof ROUNDING_TYPES.down
-    | typeof ROUNDING_TYPES.up
-    | typeof ROUNDING_TYPES.nearest,
+type RatingLabel = string | ((rating: number, maxRating: number) => string);
```
Faye-Xiao, line 48: Small question on the typing here: Flow previously allowed the callback to return `mixed`, while this TS migration narrows it to `string`.

Given the value is used as `aria-label`, `string` seems semantically correct and probably better than preserving the overly-wide Flow type. Could we just confirm this is an intentional tightening rather than an accidental change from the migration? If yes, I think this is fine as-is.

packages/backpack-web/src/bpk-component-star-rating/src/BpkStarRating.tsx
```diff
@@ -47,17 +45,23 @@ export const ROUNDING_TYPES = {
...
+
+type Rounding =
+  | typeof ROUNDING_TYPES.down
+  | typeof ROUNDING_TYPES.up
+  | typeof ROUNDING_TYPES.nearest;
+
+type NativeDivProps = HTMLAttributes<HTMLDivElement>;
+
+export type Props = Omit<NativeDivProps, 'className' | 'aria-label'> & {
```
Faye-Xiao, line 57: 🤔  Question: are we intentionally expanding the public TS surface to include native div/button props here?

Flow previously described only the component-specific props, while the new TS Props extends `HTMLAttributes<HTMLDivElement>` / `ButtonHTMLAttributes<HTMLButtonElement>`. Since these Props are now exported from the package, consumers can rely on native props such as `role`, `tabIndex`, `onClick`, `aria-*`, etc. That may be fine if Backpack’s convention is to type all pass-through props this way, but it would make this a deliberate API contract rather than just an implementation detail.

Could we confirm this matches the TS migration pattern for Backpack components?

packages/backpack-web/src/bpk-component-star-rating/src/withInteractiveStarRatingState.tsx
xiaogliu: Done in 83a76582b — destructured `onRatingSelect` out of `this.props` before spreading.

packages/backpack-web/src/bpk-component-star-rating/src/BpkStar.tsx
xiaogliu: Good catch — fixed in 83a76582b by switching to `.join(' ')`. Worth flagging that this comma-separated class string was actually a pre-existing bug from the JS code (the original snapshot had `bpk-icon--rtl-support,bpk-star ...` too); my `as unknown as string` cast was just hiding it. The half-star snapshots are updated to the corrected, space-separated form.

packages/backpack-web/src/bpk-component-star-rating/src/BpkStarRating.tsx
xiaogliu: Yes, intentional. The return value is only ever used as the `aria-label` on the rendered `<div role="img">`, so `string` is the right type. The Flow `mixed` was wider than the actual contract.

packages/backpack-web/src/bpk-component-star-rating/src/BpkStarRating.tsx
xiaogliu: Yes, intentional — native pass-through props (`data-testid`, `role`, `aria-*`, etc.) flow through `...rest` onto the underlying DOM element, so the type should reflect that. Happy to discussion more as a separate convention discussion.

xiaogliu:
Agreed on the direction — explicit assertions are clearer about intent. I'd prefer to keep this PR scoped to the JS→TS rename (no behaviour change) and tackle the snapshot rewrite as a follow-up because more files need to update for BpkStarRating. I created another PR to track the 30 occurrences across the 5 test files. https://github.com/Skyscanner/backpack/pull/4670

packages/backpack-web/src/bpk-component-star-rating/src/withInteractiveStarRatingState.tsx
xiaogliu: Done in 83a76582b — dropped both the cast and the persist call. Thanks for the catch.

Faye-Xiao:
Approved the Percy failure as it seems the failure came from some flaky tests. https://percy.io/09e69e44/web/backpack/builds/50522997/changed/2719051046?browser=chrome&browser_ids=73&currentBuildItemIndex=1&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_public&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

Faye-Xiao (approved):
LGTM.

PR #4520, opened 2026-05-21
[NO-JIRA][nx] Migrate Storybook config into backpack-storybook-host lib

gert-janvercauteren:
## Summary

Implements Issue 2.2 of the Nx adoption plan: relocates the root `.storybook/` config into a new internal Nx lib at `libs/backpack-storybook-host/` (tags `scope:internal`, `type:storybook`). The `@nx/storybook` plugin auto-infers `serve-storybook`/`build-storybook`/`static-storybook`/`build-deps`/`watch-deps`; the project adds explicit `lint-js`/`lint`, a `typecheck` that runs `tsc --build` against both `tsconfig.lib.json` and `tsconfig.storybook.json`, and a `visual-test` target that wraps the Percy run.

The host's `main.ts` discovers stories under `packages/backpack-web/src/*/src/**/*.stories.*`; `preview.tsx` and `themeableAttributes.js` import from the bare `@skyscanner/backpack-web/...` package path; `webpack.config.js` aliases that path to source for fast dev resolution and runs `babel-loader` with `rootMode: 'upward'`. Root scripts (`storybook`, `storybook:dist`, `percy-test`) now invoke the corresponding `nx` targets, and `tsconfig.solutions.json` references the new lib so project references compose end-to-end. `nx graph` confirms `backpack-storybook-host → backpack-web → backpack-storybook-utils` with no cycle; `typecheck` and `lint` pass locally.

## Checklist

- [ ] `README.md` (If you have created a new component) — n/a
- [ ] Component `README.md` — n/a
- [ ] Tests — n/a (storybook host has no test target per plan)
- [ ] Accessibility tests — n/a
- [ ] Storybook examples created/updated — n/a (config relocation only)
- [ ] Migration guide — n/a (internal tooling)

skip-changelog

PR #4525, opened 2026-05-21
[NO-JIRA][nx] Module boundaries + nx affected CI

gert-janvercauteren:
## Summary

- Register `@nx/enforce-module-boundaries` in `.eslintrc` so `scope:publishable` (`backpack-web`) can only depend on other publishable libs and `scope:internal` (`backpack-storybook-utils`, `backpack-storybook-host`) can depend on either; dev-only files (tests, stories, figma helpers, snapshots, webpack/gulp configs) are exempt via an overrides block.
- Switch the CI Build job to `nx affected` for build/lint/typecheck/test, add `nrwl/nx-set-shas@v4.3.3` with the required `actions: read` permission, and propagate that permission through the `pr.yml` / `main.yml` / `release.yml` callers.
- Drop the obsolete `packages/backpack-web/{node_modules,package-lock.json}` cache paths (npm workspaces hoists) and bump `CACHE_NAME` to `node-modules-cache-v2` across all four workflows.

This implements Issues 4.1 and 5.1 of the Nx adoption plan. The `backpack-harness:lint` / `check-pristine` consolidation is gated behind Issue 6.1 — the legacy `check-react-versions` / `check-bpk-dependencies` / `check-pristine-state` steps remain in `_build.yml` with a `TODO(Issue 6.1)` marker.

## Test plan

- [x] `nx run backpack-web:lint-js`, `backpack-storybook-utils:lint-js`, `backpack-storybook-host:lint-js` — all pass with the new rule active (0 errors).
- [x] `nx show projects --affected` correctly resolves projects from `.eslintrc` edits.
- [ ] CI Build job picks up base/head SHAs from `nx-set-shas` and runs `nx affected` against the right baseline.
- [ ] Touching one component should rebuild only the affected chain in CI.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -13,7 +13,7 @@ defaults:
     shell: bash -l {0}
 
 env:
-  CACHE_NAME: node-modules-cache
+  CACHE_NAME: node-modules-cache-v2
   BUILD_CACHE_NAME: build-cache
 
 permissions: {}
@@ -37,8 +37,7 @@ jobs:
         with:
           path: |
             node_modules/
-            packages/backpack-web/node_modules/
-          key: ${{ env.CACHE_NAME }}-${{ hashFiles('package-lock.json', 'packages/backpack-web/package-lock.json') }}
+          key: ${{ env.CACHE_NAME }}-${{ hashFiles('package-lock.json') }}
 
       - name: Install dependencies
         if: ${{ steps.npm-cache.outputs.cache-hit != 'true' }}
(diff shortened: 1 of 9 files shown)


.github/workflows/pr.yml
```diff
@@ -83,6 +81,9 @@ jobs:
 
   Build:
     permissions:
+      actions: read
+      contents: read
+      id-token: write
```
RichardSyq, line 86: Quick security ask on the new Build permissions: `nrwl/nx-set-shas` only needs `actions: read`, but the block also adds `id-token: write`, and the same propagates through `_build.yml` / `main.yml` / `release.yml`.

The pr.yml one is the one I'd flag — `pull_request` is the only event in the set that any contributor can trigger, so anything that runs in a PR (a compromised dev-dep, an install script, a hijacked third-party action) could request an OIDC token representing this repo. Today there's no consumer in any of these workflows (no `--provenance`, no `aws-actions/configure-aws-credentials`, no sigstore), so the token is currently harmless — but once anything starts trusting this repo's OIDC (even later, in an unrelated PR), the residual `id-token: write` on PR-event workflows quietly becomes real attack surface.

If it's reserved for upcoming npm provenance (Issue 5.2?), would it make sense to scope `id-token: write` to `release.yml` only and call it out in the PR description? Otherwise probably worth dropping it from pr.yml / main.yml / _build.yml here. Non-blocking — just principle of least privilege.

RichardSyq (approved):
LGTM — module boundaries config + nx affected wiring all hang together cleanly, and the cache pipeline updates from #4520 carry through correctly. Left one non-blocking security note on the new Build permissions.

PR #4541, opened 2026-05-22
[NO-JIRA] Surface Percy silent failures as CI errors

RichardSyq:
## Summary
- Percy CLI exits 0 even when it fails internally (missing token, API errors, build not created), causing the PercyTests job to show success while actually failing silently
- The Percy Test step now asserts a `https://percy.io/.../builds/` URL appears in the output; if it doesn't, the step fails with a clear error
- This ensures CI correctly shows a red cross when Percy cannot complete its work

## What changed
The Percy Test step now:
1. Captures Percy output to `${RUNNER_TEMP}/percy.log`
2. Greps for a percy.io build URL — only printed when a build is actually finalized
3. Exits with code 1 if no URL is found, surfacing it as a proper CI failure

## Why URL-grep instead of failure-string grep
Earlier iterations of this PR matched failure strings like `Build not created` / `Failed to create build` / `Error:`. That approach has two problems:
- **Failure space is open** — any new wording in a future Percy CLI release silently breaks the check, putting us right back in the silent-failure hole this PR is meant to fix
- **Success space is closed** — a build is finalized exactly one way, and the printed URL contains `percy.io`, which is part of Percy's product contract (the domain rarely changes; CLI wording often does)

The token-empty pre-check was also dropped: a missing/invalid token produces no build URL, so the URL assertion already covers it.

## Test plan
- [ ] Verify normal Percy runs (with valid token) still pass
- [ ] Verify that an empty/invalid `PERCY_TOKEN` causes the job to fail (no URL → exit 1)
- [ ] Verify that a Percy API/quota error during the run causes the job to fail


.github/workflows/_build.yml
```diff
@@ -186,7 +186,17 @@ jobs:
           key: ${{ env.BUILD_CACHE_NAME }}-${{ hashFiles('packages/backpack-web/**', 'libs/backpack-storybook-utils/**', 'libs/backpack-storybook-host/**') }}
 
       - name: Percy Test
-        run: npm run percy-test
         if: ( github.ref == 'refs/heads/main' || github.repository == github.event.pull_request.head.repo.full_name) && github.actor != 'dependabot[bot]'
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+        run: |
+          set -o pipefail
```
xiaogliu: `set -o pipefail` doesn't actually buy us anything here. The whole premise of this PR is that `npm run percy-test` exits 0 even on failure, so the left side of the pipe never returns non-zero from Percy. `pipefail` only catches `tee` itself failing, which won't happen on a hosted runner. Not harmful, but it can mislead future readers into thinking it provides real protection — consider removing it or adding a brief comment explaining what it's actually for.

.github/workflows/_build.yml
```diff
@@ -186,7 +186,17 @@ jobs:
...
+          if [ -z "$PERCY_TOKEN" ]; then
+            echo "::error::PERCY_TOKEN is not set or empty"
+            exit 1
+          fi
+          npm run percy-test 2>&1 | tee /tmp/percy.log
```
xiaogliu: Minor: `${{ runner.temp }}/percy.log` is more portable if this workflow ever runs on Windows or self-hosted runners. Not a blocker on the current Linux setup.

.github/workflows/_build.yml
```diff
@@ -186,7 +186,17 @@ jobs:
...
+          if grep -q "\[percy\] Build not created" /tmp/percy.log; then
```
xiaogliu: Two things here:

1. The PR description promises `Failed to create build` and `Error:` are also caught, but only `Build not created` is checked. Looking at the failing run, `Failed to create build` actually appears *before* `Build not created`, so it's the earlier and more useful signal. Suggest:

```bash
if grep -qE '\[percy\] (Build not created|Failed to create build|Error:)' /tmp/percy.log; then
```

2. This string-matching approach is inherently fragile — any wording change in a future Percy CLI release will silently break the check, putting us right back in the silent-failure hole this PR is meant to fix. A short comment in the workflow noting this brittleness (and pointing to where the strings come from) would help the next maintainer. Even better long-term would be opening an upstream issue on `@percy/cli` asking for a non-zero exit on build creation failure.

xiaogliu (approved):
LGTM

PR #4553, opened 2026-05-22
[NO JIRA][BpkCardList] Remove initiallyInViewCardIndex prop

jimmycook:
## Summary
- Removes `initiallyInViewCardIndex` from `BpkCardList` and the row/rail container, plus the supporting `initialPageIndex` plumbing through `BpkCardListCarousel` and `usePageScrollSync`.
- Drops the dedicated story, tests, and prop-passes; tests still pass (60/60 in `bpk-component-card-list`).

## Why
The prop ran `scrollIntoView` from a mount-time `useEffect`. When the carousel mounts inside a parent that hasn't laid out yet (e.g. `AnimateHeight` at `display: none`), the target card has a zero-size box and the scroll silently no-ops, leaving DOM and React state out of sync.

Its sole intended consumer — ExpandedPricingOption in web-platform — hit exactly this and has since switched to a manual ref-based scroll triggered after its `AnimateHeight` animation completes (IRN-6568, web-platform commit `29b3db2687`). Codebase search shows no other consumers, so the prop is dead weight with a misleading API.

Drafted as part of an internal decision doc: `~/.claude/tmp/remove-bpkcardlist-initial-selection.md`.

## Test plan
- [x] `npx jest packages/backpack-web/src/bpk-component-card-list` — 60/60 passing
- [x] `tsc --noEmit` clean for `packages/backpack-web`
- [ ] Storybook smoke check on `RowToRail` / `RowToStack` examples
- [ ] Confirm with consuming teams (none expected) that no app references the prop


xiaogliu (dismissed):
LGTM

PR #4634, opened 2026-05-25
[CLOV-1621][BpkComparisonTable] Bottom-anchorable headerContent

Supremeyh:
## Summary

Ticket: https://skyscanner.atlassian.net/browse/CLOV-1621

When one column's `headerContent` (e.g. a long product description) wraps to more lines than its neighbours, table semantics equalise the `<th>` cell heights but the inner header-content flex column doesn't fill the cell. Direct children — typically the price + CTA — sat at their natural y-coordinate in each cell, so columns with shorter content showed misaligned CTAs vs. the tall column.

This PR ships in **4 commits**:

### 1. `[CLOV-1621][BpkComparisonTable] Bottom-anchorable headerContent`

Establish a vertical block-size chain so consumers can bottom-anchor any direct child of `headerContent` with `margin-block-start: auto`:

- `<table>` gets `block-size: 1px` — a CSS spec workaround so percentage block-size on descendants resolves (table/tr/th default to auto, which makes child %s collapse to auto). 1px is well below natural content height, so content wins for actual sizing; the value just flips `<table>` into "specified" mode.
- `thead` / `thead tr` / `thead th` propagate `block-size: 100%` down so header-content has a determinate base to fill.
- `header-content` gets `block-size: 100%` to fill the equalised cell.
- `header-id-section` becomes a flex column with `flex: 1` so it claims the remaining space below the image wrapper.
- `overflow-anchor: none` is added to the scroll container — the scroll handler drives `margin` / `max-block-size` via `--bpk-image-opacity` (layout-affecting props), and without this the browser's scroll anchoring adjusts `scrollTop` to compensate for the header height change, re-fires the scroll handler, and causes jitter on tablet viewports.

A new "BpkComparisonTable - Uneven Header Content" story intentionally gives one column a multi-line description to visually demonstrate the bottom alignment.

### 2. `[CLOV-1621][BpkLayout] Allow 'auto' on margin props`

The chain above is dead weight without a way for consumers to actually trigger it. BpkStack's allowlist strips both `style` and `className`, so the natural `style={{ marginBlockStart: 'auto' }}` doesn't reach the DOM. Spacing-token validation also rejected `'auto'`.

This commit adds first-class support:

- New `BpkMarginValue = BpkSpacingValue | 'auto'` type.
- New `isValidMarginValue` validator.
- `processSpacingProps` routes **11 margin keys** (`margin`, `marginTop` / `Right` / `Bottom` / `Left`, `marginStart` / `End`, `marginInline`, `marginBlockStart` / `BlockEnd`, `marginBlock`) to it.
- `marginBlockStart` and `marginBlockEnd` are added as first-class `BpkSpacingProps` — the logical block-direction longhands, symmetric with the existing `marginStart` / `marginEnd` inline-direction longhands.
- Padding and gap intentionally still reject `'auto'` — CSS ignores it there, so the DS surfaces the misuse at compile time (TS error) rather than letting it silently no-op.
- New `BpkMarginValue` and `isValidMarginValue` are exported from the `bpk-component-layout` barrel.

Tests assert `marginTop="auto"` and `marginBlockStart="auto"` are accepted; `padding="auto"` is rejected.

### 3. `[CLOV-1621][BpkComparisonTable] Use marginBlockStart=auto in Uneven story`

Switches the story from the (silently-stripped) `style={{ marginBlockStart: 'auto' }}` to `marginBlockStart="auto"` now that the layout package supports the logical block-start prop. Using `marginBlockStart` rather than the physical `marginTop` is more semantically correct and keeps the SCSS comment (`consumers bottom-anchor children via margin-block-start: auto`) accurate.

## Migration / Breaking changes

**None.** All changes are additive:

- The SCSS chain (commit 1) is opt-in; existing consumers see no visual difference unless they set `margin-*: auto` on a `headerContent` child.
- The `BpkMarginValue` extension (commit 2) only adds `'auto'` to the margin allowlist and exposes two new optional props (`marginBlockStart` / `marginBlockEnd`). Existing `BpkSpacingValue` margin usage continues to type-check.

## Screenshots
| Platform | Before | After |
| :------: | :----: | :---: |
| Desktop  |   |  |
| mWeb  |  |  |

## Test plan

- [x] `npm run typecheck` clean
- [x] Jest: `npx jest packages/backpack-web/src/bpk-component-layout/src/BpkStack-test.tsx` — `marginTop="auto"` and `marginBlockStart="auto"` accepted; `padding="auto"` rejected
- [x] Jest: `npx jest packages/backpack-web/src/bpk-component-comparison-table` — existing snapshots unaffected
- [x] Storybook: `BpkComparisonTable - Uneven Header Content` — CTAs in all 3 columns bottom-align across row-equalised cell heights
- [x] Storybook: existing `CompareModalDefault` story — no visual regression on equal-height columns
- [x] Manual: no scroll jitter on tablet viewport
- [x] Manual: keyboard navigation through the comparison-table modal still works
- [x] Percy / Chromatic visual diff reviewed

---

- [x] Ensure the PR title includes the name of the component you are changing
- [ ] Component `README.md` — *worth noting bottom-anchor pattern in BpkComparisonTable README; defer to follow-up if too much for this PR*
- [x] Tests — added for `BpkLayout` margin auto (`marginTop` and `marginBlockStart`)
- [ ] Accessibility tests — keyboard & screen-reader nav unchanged; reflow at 200%/400% zoom unaffected (chain only changes height behavior)
        - [ ] Zoom functionality
            - [ ] 200% text zoom — page remains functional and readable
            - [ ] 400% page zoom — content reflows to one column
- [x] No breaking changes (additive)

[CLOV-1621]: https://skyscanner.atlassian.net/browse/CLOV-1621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLOV-1621]: https://skyscanner.atlassian.net/browse/CLOV-1621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/backpack-web/src/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.module.scss
```diff
@@ -54,9 +54,28 @@ $image-area-height: 83 * tokens.$bpk-one-pixel-rem;
     border-spacing: 0;
     box-shadow: none;
 
+    // 1px (block-size) is a CSS spec workaround: percentage heights only resolve when the parent has an *explicitly* specified size, and table/tr/th default to auto (content-based). The 1px is well below natural content height so content wins for actual sizing, but it flips <table>'s height into "specified" mode so descendants' block-size: 100% can cascade through thead → tr → th → header-content. This lets consumers bottom-align CTAs (or any other element) inside variable-height headerContent across columns by setting margin-block-start: auto on the desired child.
+    block-size: tokens.$bpk-one-pixel-rem;
```
Supremeyh, line 62: 1px (block-size) is a CSS spec workaround: percentage heights only resolve when the parent has an *explicitly* specified size, and table/tr/th default to auto (content-based). The 1px is well below natural content height so content wins for actual sizing, but it flips table's height into "specified" mode so descendants' block-size: 100% can cascade through thead → tr → th → header-content. This lets consumers bottom-align CTAs (or any other element) inside variable-height headerContent across columns by setting margin-block-start: auto on the desired child.

packages/backpack-web/src/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.module.scss
```diff
@@ -99,11 +119,16 @@ $image-area-height: 83 * tokens.$bpk-one-pixel-rem;
 
 // ID section — slides up over the fading image.
 // Negative margin-block-start collapses layout space (unlike translateY) so the sticky header cell actually shrinks in height and rows are no longer hidden behind invisible space.
+// flex column with flex: 1 + min-block-size: 0 lets the section claim the remaining vertical space inside header-content (after the fixed-height image wrapper), so consumers can bottom-align any direct child of headerContent with margin-block-start: auto. Works in concert with the block-size trick on <table> above to give the percentage chain a determinate base.
```
Supremeyh: flex column with flex: 1 + min-block-size: 0 lets the section claim the remaining vertical space inside header-content (after the fixed-height image wrapper), so consumers can bottom-align any direct child of headerContent with margin-block-start: auto. Works in concert with the block-size trick on <table> above to give the percentage chain a determinate base.

packages/backpack-web/src/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.stories.tsx
```diff
@@ -349,3 +354,70 @@ export const CompareModalWithTray = {
   name: 'BpkComparisonTable + BpkComparisonTray',
   render: () => <CombinedExample />,
 };
+
+// ─── Uneven header content (CTA bottom-alignment) ────────────────────────────
+
+// Demonstrates that CTAs (or any direct child of headerContent that opts in
+// via margin-block-start: auto) stay bottom-aligned across columns even when
+// one column has a much taller description than its neighbours.
+const UnevenHeaderContentExample = () => {
```
Supremeyh, line 514: Demonstrates that CTAs (or any direct child of headerContent that opts in via margin-block-start: auto) stay bottom-aligned across columns even when one column has a much taller description than its neighbours.

packages/backpack-web/src/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.module.scss
```diff
@@ -99,11 +119,16 @@ $image-area-height: 83 * tokens.$bpk-one-pixel-rem;
 
 // ID section — slides up over the fading image.
 // Negative margin-block-start collapses layout space (unlike translateY) so the sticky header cell actually shrinks in height and rows are no longer hidden behind invisible space.
+// flex: 1 claims space after the image wrapper; consumers bottom-align children via margin-block-start: auto.
```
Faye-Xiao, line 126: The comment says `margin-block-start: auto` but nothing in this PR actually sets that — the story uses `marginTop="auto"` which emits `margin-top: auto (physical)`. Works fine for our supported locales since RTL doesn't affect block direction, but the comment is a bit misleading.

Minor nit: 
worth either fixing the comment to say `margin-top: auto`, or — if we want to be more correct — adding `marginBlockStart/marginBlockEnd` to `BpkSpacingProps` (Chakra already supports them) and using that in the story instead. We already expose `marginStart`/`marginEnd` for the inline direction, so leaving out the block equivalents feels a bit asymmetric.

Faye-Xiao:
Checked the storybook examples and found a weird thing on this case https://backpack.github.io/storybook-prs/4634/?path=/story/bpk-component-comparison-table--compare-modal-uneven-header-content&globals=viewport.value:tablet
When I switched it to the tablet and swiped the content to the bottom, it kept shaking as the gif shows.
 
after-fix-comparision

It occurs on both the PR storybook example and the local machine

packages/backpack-web/src/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.module.scss
```diff
@@ -25,6 +25,10 @@ $image-area-height: 83 * tokens.$bpk-one-pixel-rem;
   flex: 1;
   overflow: auto;
   min-block-size: 0;
+
+  // The scroll handler drives margin/max-block-size via --bpk-image-opacity (layout-affecting props).
+  // Scroll anchoring would adjust scrollTop to compensate for the size change, re-firing the handler and causing jitter.
+  overflow-anchor: none;
```
Supremeyh, line 31: The scroll handler drives margin/max-block-size via --bpk-image-opacity (layout-affecting props). Scroll anchoring would adjust scrollTop to compensate for the size change, re-firing the handler and causing jitter.

packages/backpack-web/src/bpk-component-comparison-table/src/BpkComparisonTable/BpkComparisonTable.module.scss
Supremeyh: good catch, update in [CSS logical shorthands marginBlockStart and marginBlockEnd](https://github.com/Skyscanner/backpack/pull/4634/changes/0a49c5cbd5279dafe9c5c83a884b2af005545987)

Supremeyh:
> Checked the storybook examples and found a weird thing on this case https://backpack.github.io/storybook-prs/4634/?path=/story/bpk-component-comparison-table--compare-modal-uneven-header-content&globals=viewport.value:tablet When I switched it to the tablet and swiped the content to the bottom, it kept shaking as the gif shows.

---

Cause: The scroll handler drives margin/max-block-size via --bpk-image-opacity (layout-affecting props). Scroll anchoring would adjust scrollTop to compensate for the size change, re-firing the handler and causing jitter. fix in [commit](https://github.com/Skyscanner/backpack/pull/4634/changes/3b62a09fba2145f7cbc92b7df2b8caa13bacb219)

Faye-Xiao (approved):
LGTM.
The tablet shaking issue has been improved, but it still has a slight shake on the first render. Accept it and keep an eye on real use case

PR #4635, opened 2026-05-25
[CLOV-1606][BpkModalV3] Fix infinite focus redirect loop when BpkModalV3 and BpkDrawer coexist

Faye-Xiao:
## Description

When a `BpkDrawer` (which uses the legacy `withScrim`/`focusScope` focus management system) is open at the same time as a `BpkModalV3` (which uses ark-ui's `@zag-js/focus-trap`), both systems simultaneously listen to `focusin` on `document`. This causes an infinite focus redirect loop that overflows the call stack.

**Root cause**: `focusScope` registers a bubble-phase `focusin` listener that synchronously calls `element.focus()`. `@zag-js/focus-trap` registers a capture-phase listener but does not call `stopImmediatePropagation` when focus is already inside the modal. The result is a recursive `focusin` dispatch loop (~8 frames per iteration) that overflows the call stack after ~100–200 iterations.

**Fix**: When `BpkModalV3Root` opens, it calls `focusScope.unscopeFocus()` to deactivate the legacy listener before ark-ui's focus trap takes over. This is a compatibility patch that keeps both systems stable while `BpkDrawer` has not yet been migrated to ark-ui.

## Changes

- **`BpkModalV3Root.tsx`**: Call `focusScope.unscopeFocus()` in a `useEffect` when `isOpen` becomes `true`, with a comment explaining why.
- **`BpkModalV3-test.tsx`**: Mock `focusScope` and add 3 test cases covering: modal opens (`unscopeFocus` called), modal closed (`unscopeFocus` not called), and closed → open transition.
- **`BpkModalV3.stories.tsx`**: Add `WithDrawer` story demonstrating `BpkModalV3` opened from within a `BpkDrawer`.

## Test Screenshots 

Tested it in local carhire
|  Platform  |  Before  |  After  |
| :--------: | :------: | :-----: | 
|  mWeb  |   before-fix-mweb  |   after-fix-mweb  | 
|  destop  |   before-fix-desktop  |   after-fix-desktop  | 

- no error in mWeb 
- no change in desktop, no issue with keyboard management in desktop

## Checklist

- [x] Ensure the PR title includes the name of the component you are changing so it's clear in the release notes for consumers of the changes in the version e.g `[CLOV-1606][BpkModalV3] Fix infinite focus redirect loop when BpkModalV3 and BpkDrawer coexist`
- [ ] `README.md` (If you have created a new component) — N/A, no new component
- [ ] Component `README.md` — N/A
- [x] Tests
- [ ] Accessibility tests
        - [x] Ability to navigate using a [keyboard only](https://webaim.org/techniques/keyboard/) — this fix restores correct keyboard focus when both components coexist
- [x] Storybook examples created/updated — `WithDrawer` story added to `bpk-component-modal-v3`
- [ ] For breaking changes or deprecating components/properties, migration guides added to the description of the PR — N/A, this is a bug fix

[CLOV-1606]: https://skyscanner.atlassian.net/browse/CLOV-1606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3-test.tsx
+++ b/packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3-test.tsx
@@ -136,6 +136,7 @@ describe('BpkModalV3', () => {
     });
 
     it('should call focusScope.resumeFocus when modal transitions from open to closed', () => {
+      jest.useFakeTimers();
       const { rerender } = render(
         <BpkModalV3.Root open onOpenChange={jest.fn()}>
           <BpkModalV3.Content>
@@ -152,7 +153,12 @@ describe('BpkModalV3', () => {
           </BpkModalV3.Content>
         </BpkModalV3.Root>,
       );
+      // Lock stays active during the exit animation — not released yet.
+      expect(focusScope.resumeFocus).not.toHaveBeenCalled();
+
+      act(() => { jest.runAllTimers(); });
       expect(focusScope.resumeFocus).toHaveBeenCalledTimes(1);
+      jest.useRealTimers();
     });
 
     it('should not call focusScope.resumeFocus when the modal is initially open', () => {
@@ -171,6 +177,7 @@ describe('BpkModalV3', () => {
     });
 
     it('should call portalLock.unlock when modal transitions from open to closed', () => {
+      jest.useFakeTimers();
       const { rerender } = render(
         <BpkModalV3.Root open onOpenChange={jest.fn()}>
           <BpkModalV3.Content>
@@ -187,7 +194,12 @@ describe('BpkModalV3', () => {
           </BpkModalV3.Content>
         </BpkModalV3.Root>,
       );
+      // Lock stays active during the exit animation — not released yet.
+      expect(portalLock.unlock).not.toHaveBeenCalled();
+
+      act(() => { jest.runAllTimers(); });
       expect(portalLock.unlock).toHaveBeenCalledTimes(1);
+      jest.useRealTimers();
     });
 
     it('should call portalLock.unlock on unmount while open', () => {
--- a/packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3-test.tsx
+++ b/packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3-test.tsx
@@ -27,8 +27,10 @@ import BpkModalV3 from './BpkModalV3';
 jest.mock('../../../bpk-scrim-utils/src/focusScope', () => ({
   __esModule: true,
   default: {
-    unscopeFocus: jest.fn(),
     scopeFocus: jest.fn(),
+    unscopeFocus: jest.fn(),
+    pauseFocus: jest.fn(),
+    resumeFocus: jest.fn(),
   },
 }));
 
@@ -86,20 +88,22 @@ describe('BpkModalV3', () => {
 
   describe('Root', () => {
     beforeEach(() => {
-      (focusScope.unscopeFocus as jest.Mock).mockClear();
+      (focusScope.pauseFocus as jest.Mock).mockClear();
+      (focusScope.resumeFocus as jest.Mock).mockClear();
+      (focusScope.scopeFocus as jest.Mock).mockClear();
     });
 
-    it('should call focusScope.unscopeFocus when the modal opens', () => {
+    it('should call focusScope.pauseFocus when the modal opens', () => {
       renderModal({ open: true });
-      expect(focusScope.unscopeFocus).toHaveBeenCalledTimes(1);
+      expect(focusScope.pauseFocus).toHaveBeenCalledTimes(1);
     });
 
-    it('should not call focusScope.unscopeFocus when the modal is closed', () => {
+    it('should not call focusScope.pauseFocus when the modal is closed', () => {
       renderModal({ open: false });
-      expect(focusScope.unscopeFocus).not.toHaveBeenCalled();
+      expect(focusScope.pauseFocus).not.toHaveBeenCalled();
     });
 
-    it('should call focusScope.unscopeFocus when modal transitions from closed to open', () => {
+    it('should call focusScope.pauseFocus when modal transitions from closed to open', () => {
       const { rerender } = render(
         <BpkModalV3.Root open={false} onOpenChange={jest.fn()}>
           <BpkModalV3.Content>
@@ -107,7 +111,7 @@ describe('BpkModalV3', () => {
           </BpkModalV3.Content>
         </BpkModalV3.Root>,
       );
-      expect(focusScope.unscopeFocus).not.toHaveBeenCalled();
+      expect(focusScope.pauseFocus).not.toHaveBeenCalled();
 
       rerender(
         <BpkModalV3.Root open onOpenChange={jest.fn()}>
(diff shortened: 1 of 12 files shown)


packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3Root/BpkModalV3Root.tsx
```diff
@@ -57,6 +58,16 @@ const BpkModalV3Root = ({
 
   useBodyLock(type === MODAL_V3_TYPES.chatbot && bodyLockOpen);
 
+  // When this modal opens, deactivate any active focusScope (used by legacy
+  // withScrim components such as BpkDrawer). Both systems listen to 'focusin'
+  // on document, and their simultaneous presence causes an infinite focus
+  // redirect loop that overflows the call stack.
+  useEffect(() => {
+    if (isOpen) {
+      focusScope.unscopeFocus();
```
Supremeyh: This is the main concern with the current implementation.

Looking at withScrim.tsx:114-115, the drawer only calls focusScope.scopeFocus(this.dialogElement) once — in componentDidMount. It has no mechanism to re-register after an external caller clears the global state.

When BpkModalV3 closes (drawer still mounted), the sequence is:

1. BpkModalV3Root opens → focusScope.unscopeFocus() called ✅
2. ark-ui focus trap takes over the modal ✅
3. User closes BpkModalV3 → BpkModalV3Root unmounts / isOpen becomes false
4. focusScope listener is gone and never restored ❌
5. The drawer is still visually open but focus can now escape it ❌
The PR trades a crash for a silent a11y regression. On desktop this matters most: keyboard users can now Tab out of the drawer after the modal closes.

Suggested fix — save and restore the scope element:

```
// BpkModalV3Root.tsx
const scopedElementRef = useRef<HTMLElement | null>(null);

useEffect(() => {
  if (isOpen) {
    // Save current focus scope element before clearing it
    // (focusScope doesn't expose its element directly — see note below)
    focusScope.unscopeFocus();
  } else if (scopedElementRef.current) {
    // Re-activate the saved scope when modal closes
    focusScope.scopeFocus(scopedElementRef.current);
    scopedElementRef.current = null;
  }
}, [isOpen]);
```
However, focusScope.ts doesn't expose the currently scoped element, so this would require either:

- Option A: Add getScopedElement(): HTMLElement | null to focusScope API, or
- Option B: Add pauseFocus() / resumeFocus() to focusScope that internally preserves teardownFn, or
- Option C (minimal): withScrim re-calls scopeFocus in componentDidUpdate when the drawer is still open

packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3Root/BpkModalV3Root.tsx
```diff
@@ -23,6 +23,7 @@ import { Dialog } from '@ark-ui/react';
 import { durationBase } from '@skyscanner/bpk-foundations-web/tokens/base.es6';
 
 import { getDataComponentAttribute, useBodyLock } from '../../../../bpk-react-utils';
+import focusScope from '../../../../bpk-scrim-utils/src/focusScope';
```
Supremeyh: It imports focusScope directly from a sibling package's src/ path. This creates a hard coupling between bpk-component-modal and the internals of bpk-scrim-utils. The import path ../../../../bpk-scrim-utils/src/focusScope bypasses the package boundary. If bpk-scrim-utils reorganizes its internals, BpkModalV3Root breaks silently.

Minor suggestion: at minimum, export focusScope from bpk-scrim-utils' package index so this reads as an inter-package dependency, not an internal-path hack.

packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3-test.tsx
```diff
@@ -75,6 +85,40 @@ describe('BpkModalV3', () => {
   });
 
   describe('Root', () => {
+    beforeEach(() => {
```
Supremeyh, line 100: Tests are correct for what they test. Three concerns:

- There is no test for the close path: the tests verify unscopeFocus is called on open, but there's no test asserting behaviour when the modal closes while a drawer is active. This is precisely the gap where the a11y regression lives.

- scopeFocus mock is registered but never asserted in any test — suggests the author considered testing restore behavior but didn't implement it.

- The beforeEach clears only unscopeFocus but not scopeFocus. If a future test for restore is added, the mock should clear both:


```
beforeEach(() => {
  (focusScope.unscopeFocus as jest.Mock).mockClear();
  (focusScope.scopeFocus as jest.Mock).mockClear(); // add this
});
```

packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3.stories.tsx
```diff
@@ -691,6 +692,76 @@ const LogoExample = () => (
   </ModalContainer>
 );
 
+const WithDrawerExample = () => {
```
Supremeyh: The WithDrawerExample story is excellent — it clearly demonstrates the exact failure scenario and serves as a regression canary. One nit: the story doesn't close the modal and verify the drawer still traps focus, which would expose the regression mentioned above.

Supremeyh (commented):
Check on [story](https://backpack.github.io/storybook-prs/4635/?path=/story/bpk-component-modal-v3--with-drawer), it seems cannot close drawer with keyboard Esc and  when close the modal, it would flash fast, is it expected?

https://github.com/user-attachments/assets/af9df71f-3639-413c-b1cd-4dcb8a0b6726

packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3Root/BpkModalV3Root.tsx
Faye-Xiao: Good catch. Fixed in the latest commit by replacing `unscopeFocus()` with `pauseFocus()` / `resumeFocus()` pattern.

When the modal opens, `pauseFocus()` removes the focusin listener but keeps the element reference. When the modal closes, `resumeFocus()` re-registers the listener **without** calling `focus()` immediately, so ark-ui's own focus-return logic runs first and there is no focus jump.

The key change is in `focusScope.ts`: extracted a `registerListener()` helper that only attaches the listener (no initial focus steal), which `resumeFocus()` uses.

packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3Root/BpkModalV3Root.tsx
Faye-Xiao: Fixed. Added `focusScope` as a named export in `bpk-scrim-utils/index.ts`, and updated the import in `BpkModalV3Root` to use the package index instead of the src path.

packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3-test.tsx
Faye-Xiao: All three points fixed:

- Added a test for the close path: `should call focusScope.resumeFocus when modal transitions from open to closed`
- The `scopeFocus` mock is now also cleared in `beforeEach` alongside `pauseFocus` and `resumeFocus`
- Updated all existing tests from `unscopeFocus` to `pauseFocus` / `resumeFocus` to match the new implementation

packages/backpack-web/src/bpk-component-modal/src/BpkModalV3/BpkModalV3.stories.tsx
Faye-Xiao: Added verification steps to the modal body text: after closing the modal, press Tab inside the drawer and check that focus stays trapped. This makes it easy to test the post-close behaviour manually in Storybook.

Faye-Xiao (commented):
Thanks for the review\! The Esc key and flash issue — I will verify manually in the Storybook. Will update here after testing.

Faye-Xiao:
Hey @Supremeyh , I tested this — you're right, both issues are real. Let me explain what's going on.

**Why Esc closes both at the same time:** `BpkDrawer` uses a legacy `Portal` component that adds its own `keydown` listener directly on `document`. When Esc is pressed, both the modal (handled by ark-ui) and the drawer (handled by Portal) pick up the same event at the same time. So both start their close animations simultaneously — that's the flash you see.

**Why this is hard to fix here:** `BpkDrawer`, `BpkModal`, and `BpkDialog` all use the `withScrim` HOC, which has no concept of overlay stacking. There's no way for a lower overlay to know that a higher one is already handling Esc. This is a fundamental limitation of the legacy architecture, and not something this PR can really fix cleanly.

**What this PR does cover:** The infinite focus redirect loop (the stack overflow crash) is fixed. The Esc/flash issue is a separate, deeper problem that lives in the legacy system.

**Proposal:** `BpkDrawer`, `BpkModal`, and `BpkDialog` are planned to be deprecated and replaced by `BpkModalV3`. Once that happens, this mixed usage won't exist anymore, so the Esc/flash issue disappears on its own. Given that, I'd suggest we just remove the `WithDrawer` story for now — keeping it might make people think this combination is fully supported, but it isn't. The actual fix (preventing the crash) is already covered by unit tests. Happy to remove it if you agree!

PR #4637, opened 2026-05-25
[IRN-6612]: use 1fr columns in grid layout to fix oversized single-item cards

jimmycook:
## Problem

`BpkCardListGridStack` uses `grid-template-columns: repeat(N, auto)` for its grid layout. When there are fewer items than columns (e.g. a single card in a 3-column grid), the empty columns have no content to size against and effectively collapse, allowing the occupied column to grow beyond its intended `1/N` share of the container.

This produces oversized cards — a single item nearly fills half the viewport instead of one third.

## Fix

Change `auto` to `1fr` for grid column tracks:

```scss
- grid-template-columns: repeat(var(--initially-shown-cards, 3), auto);
+ grid-template-columns: repeat(var(--initially-shown-cards, 3), 1fr);
```

`1fr` tracks always distribute available space equally, regardless of how many items are present — matching the behaviour of the row/rail carousel which hard-codes `flex-basis: calc(100% / N)`.

## Before / After

**Before** (`auto` — single card bloated):
before

**After** (`1fr` — single card correctly sized to 1/3):
after

> Screenshots taken from the new `GridToStackWithSingleItem` Storybook story added in this PR.

## Changes

- `BpkCardListGridStack.module.scss` — `auto` → `1fr` for grid column tracks
- `BpkCardList.stories.tsx` — adds `GridToStackWithSingleItem` and `RowToRailWithSingleItem` stories to surface this edge case

## Test plan

- [ ] `GridToStackWithSingleItem` story: single card is ~1/3 container width
- [ ] `GridToStack` / `GridToStackWithExpand` stories with multiple items: no regression in column sizing
- [ ] `RowToRailWithSingleItem` story: single card sized correctly in carousel

### DIFF
--- a/packages/backpack-web/src/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.module.scss
+++ b/packages/backpack-web/src/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.module.scss
@@ -24,7 +24,10 @@
 
   &__grid {
     display: grid;
-    grid-template-columns: repeat(var(--initially-shown-cards, 3), minmax(0, 1fr));
+    grid-template-columns: repeat(
+      var(--initially-shown-cards, 3),
+      minmax(0, 1fr)
+    );
     gap: tokens.bpk-spacing-lg();
   }
 
--- a/packages/backpack-web/src/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.module.scss
+++ b/packages/backpack-web/src/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.module.scss
@@ -24,7 +24,7 @@
 
   &__grid {
     display: grid;
-    grid-template-columns: repeat(var(--initially-shown-cards, 3), 1fr);
+    grid-template-columns: repeat(var(--initially-shown-cards, 3), minmax(0, 1fr));
     gap: tokens.bpk-spacing-lg();
   }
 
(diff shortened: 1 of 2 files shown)


packages/backpack-web/src/bpk-component-card-list/src/BpkCardListGridStack/BpkCardListGridStack.module.scss
```diff
@@ -24,7 +24,7 @@
 
   &__grid {
     display: grid;
-    grid-template-columns: repeat(var(--initially-shown-cards, 3), auto);
+    grid-template-columns: repeat(var(--initially-shown-cards, 3), 1fr);
```
xiaogliu:
```suggestion
    grid-template-columns: repeat(var(--initially-shown-cards, 3), minmax(0, 1fr));
```
Hey, Jimmy, thanks for the contribution, can we change to `minmax(0, 1fr)`

`1fr` is shorthand for `minmax(auto, 1fr)`, where auto resolves to each track’s min-content width. If a card contains unbreakable content wider than 1/N of the container (long URLs, unbroken strings, etc.), the track will exceed its share and the columns will no longer be evenly distributed.

`minmax(0, 1fr)` forces the minimum to 0, guaranteeing strict equal distribution regardless of content. Overflow handling becomes the card’s responsibility, which is the standard pattern.

jimmycook: image
committed that and tested locally

xiaogliu (approved):
LGTM

PR #4641, opened 2026-05-27
[NO-JIRA] drop legacy fallback

IrinaWei:

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/backpack-web/src/bpk-mixins/_utils.scss
+++ b/packages/backpack-web/src/bpk-mixins/_utils.scss
@@ -198,7 +198,6 @@
 ///   }
 
 @mixin bpk-themeable-property($property, $variableName, $fallback) {
-  #{$property}: $fallback;
   #{$property}: var($variableName, $fallback);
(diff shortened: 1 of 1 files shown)


packages/backpack-web/src/bpk-mixins/_utils.scss
```diff
@@ -198,7 +198,6 @@
 ///   }
 
 @mixin bpk-themeable-property($property, $variableName, $fallback) {
-  #{$property}: $fallback;
```
IrinaWei, line 201: The first declaration was a fallback for browsers without CSS custom property support (IE 11 and earlier). Backpack's browserslist targets are now Chrome 109+, Edge 142+, Firefox 145+, Safari 16+, Samsung 29+ — every supported browser implements CSS custom properties natively, so the bare-fallback line is dead weight.

xiaogliu (approved):
LGTM

PR #4650, opened 2026-05-29
[CLOV-1412][bpk-stylesheets] Ship token CSS variables in base.css

gc-skyscanner:
## Summary

Restores [#4492](https://github.com/Skyscanner/backpack/pull/4492) (reverted in #4648) and fixes the Storybook build error it introduced.

### Root cause

PR #4492 moved the three token CSS imports (`primitives.css`, `theme-backpack-light.css`, `theme-backpack-dark.css`) from `index.scss` into `index.js`. This was the first time raw `.css` files were imported via JS in this repo, so they began flowing through Storybook's webpack pipeline.

Storybook's webpack5 builder already ships a default `.css` rule. The host's `webpack.config.js` was pushing a second `/\.css/` rule on top. Webpack doesn't pick one — it concatenates every matching rule's loader chain:

1. Our rule: `postcss-loader → css-loader → style-loader` → outputs **JS**
2. Storybook's default rule: `css-loader` receives that **JS** as input → `SyntaxError: Unknown word import`

### Fix

Following the same pattern used in storybook-web-platform for SVG handling: find the existing `.css` rule and add an `exclude` so it no longer matches `.css` files, then let our rule handle them exclusively.

```js
const cssRule = config.module.rules.find(
  (rule) => rule?.test instanceof RegExp && rule.test.test('a.css'),
);
if (cssRule) cssRule.exclude = /\.css$/i;
```

## Test plan

- [ ] Run `npm run storybook` locally and confirm the three `Unknown word import` errors are gone
- [ ] Verify component styles still render correctly in Storybook
- [ ] Confirm CSS custom property token variables (from `primitives.css` / theme files) are present at runtime

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


libs/backpack-storybook-host/.storybook/webpack.config.js
```diff
@@ -82,6 +82,14 @@ module.exports = ({ config }) => {
...
     },
   });
+  // Storybook's default webpack5 config already has a .css rule; if we push
+  // ours on top the two chains concatenate and css-loader receives style-loader
+  // JS output, causing "SyntaxError: Unknown word import". Exclude .css from
+  // the default rule so only ours handles it.
+  const cssRule = config.module.rules.find(
+    (rule) => rule?.test instanceof RegExp && rule.test.test('a.css'),
+  );
+  if (cssRule) cssRule.exclude = /\.css$/i;
```
gc-skyscanner, line 92: |before|after|
|---|---|
|Image|Image|

IrinaWei (approved):
LGTM

PR #4661, opened 2026-06-02
[SDBCK-281] Fix carousel nav button specificity

xiaogliu:
## Summary

Fix the carousel page indicator nav button so its circular arrow size is stable in production Storybook builds.

## Root cause

`BpkPageIndicator` passes a carousel-specific class to `BpkButton` for the circular nav arrows. The old selector for that override was a single class, so it had the same specificity as `BpkButton`'s `iconOnly` padding rule. In local Storybook the page-indicator CSS happened to be injected after the button CSS, so the circular button kept `padding: 0`. In production Storybook, CSS is split into chunks and `BpkButton` styles can be loaded later, causing `.bpk-button--icon-only-*` to override the carousel nav padding and stretch the circular arrows.

## Change

Nest the carousel nav button rule under the carousel nav wrapper using `& &-button`. This compiles to a two-class selector, making the page-indicator override more specific than the button icon-only padding regardless of CSS chunk load order. A short comment was added to explain why the repeated `&` is intentional.

Before
image

After
image

## Validation

- `npx stylelint packages/backpack-web/src/bpk-component-page-indicator/src/BpkPageIndicator.module.scss --fix`
- `node -e "require('./scripts/scss/style-compiler.js').compile('./packages/backpack-web/src/bpk-component-page-indicator/src/BpkPageIndicator.module.scss')"`
- `TZ=Etc/UTC npx jest packages/backpack-web/src/bpk-component-page-indicator/src/BpkPageIndicator-test.tsx packages/backpack-web/src/bpk-component-page-indicator/src/accessibility-test.tsx --runInBand`

### DIFF
--- a/packages/backpack-web/src/bpk-component-page-indicator/src/BpkPageIndicator.module.scss
+++ b/packages/backpack-web/src/bpk-component-page-indicator/src/BpkPageIndicator.module.scss
@@ -111,15 +111,17 @@ $container-width: $dot-size * $displayed-total + $dot-margin *
     --bpk-button-secondary-on-dark-text-color: #{tokens.$bpk-text-primary-day};
     --bpk-button-secondary-on-dark-hover-text-color: #{tokens.$bpk-text-primary-day};
     --bpk-button-secondary-on-dark-active-text-color: #{tokens.$bpk-text-primary-day};
-  }
 
-  &__nav-carousel-button {
-    display: inline-flex;
-    width: tokens.bpk-spacing-xl();
-    height: tokens.bpk-spacing-xl();
-    min-height: tokens.bpk-spacing-xl();
-    padding: 0;
-    justify-content: center;
-    align-items: center;
+    // Keep this selector more specific than BpkButton's icon-only padding,
+    // which may be loaded later in production CSS chunks.
+    & &-button {
+      display: inline-flex;
+      width: tokens.bpk-spacing-xl();
+      height: tokens.bpk-spacing-xl();
+      min-height: tokens.bpk-spacing-xl();
+      padding: 0;
+      justify-content: center;
+      align-items: center;
+    }
   }
 }


packages/backpack-web/src/bpk-component-page-indicator/src/BpkPageIndicator.module.scss
```diff
@@ -111,15 +111,17 @@ $container-width: $dot-size * $displayed-total + $dot-margin *
...
+    // which may be loaded later in production CSS chunks.
+    & &-button {
+      display: inline-flex;
+      width: tokens.bpk-spacing-xl();
+      height: tokens.bpk-spacing-xl();
+      min-height: tokens.bpk-spacing-xl();
+      padding: 0;
+      justify-content: center;
+      align-items: center;
+    }
```
Faye-Xiao, line 125: nit: The `& &-button` pattern works but creates a hidden dependency on the DOM structure in `NavButton` — if the wrapping <span> is ever removed, this silently breaks.

A more robust fix would follow the pattern already used in this block: expose a CSS custom property on BpkButton (e.g. `--bpk-button-icon-only-padding`) and override it here alongside the other button variables. That removes the specificity hack and stays consistent with the existing theming approach.

Worth a follow-up ticket rather than blocking this PR.

Faye-Xiao (approved):
This fix looks good to me, but left a nit comment for further improvement from the long-term consideration.

Faye-Xiao:
Approved the percy failure https://percy.io/09e69e44/web/backpack/builds/50370434/changed/2708682854?browser=chrome&browser_ids=73&currentBuildItemIndex=5&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_public&viewLayout=side-by-side&viewMode=new&width=375&widths=375%2C1280 as there are some flaky test.

PR #4673, opened 2026-06-05
[LOOM-2442] React 19 Migration: Part 2 - Migrate class-component defaultProps (re-targeted)

Sybsw:
## Description

Step 2 of the React 19 migration. PR #4455 set up the toolchain and migrated function-component `defaultProps` via codemod; this PR finishes the manual migration of class-component `static defaultProps`, which the codemod can't safely transform automatically (class semantics differ — defaults need to be threaded through `this.props` destructures and method signatures, not a parameter default).

React 19 removes `defaultProps` support for function components and (per the React 18.3 deprecation) discourages it for classes. Without this PR, every remaining class component that relies on `static defaultProps` will start receiving `undefined` for unspecified optional props on R19, breaking render logic, event handlers, and lifecycle methods.

## Approach

Each class is migrated by:
1. Removing the `static defaultProps = { ... }` block.
2. Marking the corresponding props optional in the TS interface (or Flow type for `.js` files).
3. Destructuring with defaults at every read site — `render`, lifecycle methods, and any other method that reads from `this.props`.
4. Where a default is referenced in multiple methods, hoisting it to a module-level `const` so the value isn't duplicated.

Commits are grouped by migration shape (render-only vs. multi-method vs. twin components) so reviewers can verify the pattern once per group rather than re-deriving it for every file.

## Changes (31 files across 6 commits)

1. **BpkCalendar / BpkDatepicker** (6 files) — extracted `DEFAULT_MARK_TODAY` / `DEFAULT_MAX_DATE` / `DEFAULT_MIN_DATE` to module-level constants in `BpkCalendarContainer`, migrated `BpkCalendarWeek`, `BpkDatepicker`, and dropped now-redundant `defaultProps` spreads in the Calendar story and `BpkCalendarWeek-test`. `BpkDatepicker` also rewires its `inputComponent` default to a function reference rather than a JSX literal so the type narrows correctly.
2. **Render-only batch** (8 files) — classes whose only `this.props` consumer is `render()`. Mechanical destructure-with-defaults: `withAccordionItemState`, `BpkCalendarDate`, `BpkCalendarGridHeader`, `withLazyLoading`, `BpkInput`, `BpkInputField`, `BpkSplitInput`, `withInteractiveStarRatingState`.
3. **Twin-pair** (4 files) — `bpk-component-banner-alert` and `bpk-component-info-banner` ship structurally identical `AnimateAndFade.tsx` + `withBannerAlertState.tsx`. Migrated symmetrically so the two stay in lock-step.
4. **Multi-method batch** (11 files) — classes that read `this.props` in `render` plus one or more lifecycle methods (`componentDidMount`, `componentDidUpdate`, `componentWillUnmount`) or instance handlers. Each default hoisted once to module scope and re-used across all consumers: `BpkBarchart`, `BpkCalendarGrid`, `BpkCalendarGridTransition`, `BpkGridToggle`, `BpkBackgroundImage`, `BpkImage`, `withInfiniteScroll`, `withOpenEvents`, `BpkMobileScrollContainer`, `BpkProgress`, `withScrim`.
5. **BpkPortal** (1 file) — largest single migration; defaults referenced in `render`, `componentDidMount`, `componentDidUpdate`, `componentWillUnmount`, and three private handlers. Kept in its own commit because the diff dwarfs anything else.
6. **BpkCollapsible test** (1 file) — wrap an `aria-expanded` assertion in `waitFor` for React 19's revised state-batching timing. Not a `defaultProps` change, but it surfaced when running the suite under R19 during this work and belongs with the migration.

## Notable callouts

- **BpkProgress bug fix**: previously had `static defaultProps = { onComplete: () => null }`, and `componentDidUpdate` truthy-checked `onComplete` before calling `onCompleteTransitionEnd`. The truthy check could never fail because the default itself was truthy — meaning the guard was dead code and the prop's contract was unclear. Restructured to call the destructure-defaulted `onComplete` unconditionally; behaviour is unchanged but the intent is now expressed in the type.
- **BpkCalendarGrid cycle**: can't import `DEFAULT_MAX_DATE` / `DEFAULT_MIN_DATE` from `BpkCalendarContainer` (would create a cycle — `BpkCalendarContainer` already depends on `BpkCalendarGrid`). Defined a local copy at module scope. The two constants are kept identical by convention; if either side changes, both must update.

## Deferred to follow-up PRs

Picking up from PR #4455's deferred list, this PR resolves item 1. Still outstanding:

- **13 `.js` story / HOC files with leftover `.defaultProps`** — these are non-class instances (function components and stories) that the codemod missed, mostly because they assign `.defaultProps` externally rather than inline. Mechanical, but separate from the class migration.
- **Full `.js` (Flow) prop-types removal** — Flow files still carry runtime `prop-types` declarations alongside their Flow types. Removing them is gated on the Flow → TS conversion progressing.
- **`@types/react@19` bump** — currently pinned to `19.0.14` against `@types/react-dom@19.0.6` for the React19 CI overlay. Bumping the workspace `@types/react` unlocks the two `types-react-codemod` preset-19 sub-transforms that were filtered out in PR #4455 (they only typecheck against `@types/react@19`).
- **Flip `Build / React19` to required** — currently `continue-on-error: true` in `.github/workflows/_build.yml:115`. Once green on this PR, a one-line follow-up flips it. Tracked as the "What's next" below.
- **Move codemod to `web-migration-scripts/migrations/2026-05-react-19/`** — the custom jscodeshift transform from PR #4455 currently lives in `scripts/react-19/`. Move is a chore, blocked on nothing, but unrelated to the migration itself.

## What's next

Once the `Build / React19` CI job is green on this PR, a follow-up commit flips `continue-on-error: true` to required in `.github/workflows/_build.yml` per `REACT_19_MIGRATION.md` step 12.

## Stacks onto

#4455 (`LOOM-2442` base PR)

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests


packages/backpack-web/src/bpk-component-accordion/src/withAccordionItemState.tsx
gc-skyscanner, line 65: Do we need to assign the default values for `expanded, initiallyExpanded` here?

packages/backpack-web/src/bpk-component-banner-alert/src/withBannerAlertState.tsx
gc-skyscanner, line 119: do we need to assign the default value for `show`?

packages/backpack-web/src/bpk-component-grid-toggle/src/BpkGridToggle.js
gc-skyscanner: should we remove this?

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarContainer.tsx
```diff
@@ -373,16 +382,20 @@ const withCalendarState = <P extends object>(Calendar: ComponentType<P>) => {
...
-        maxDate,
-        minDate,
+        fixedWidth = true,
+        initiallyFocusedDate = null,
+        markOutsideDays = true,
+        markToday = DEFAULT_MARK_TODAY,
+        maxDate = DEFAULT_MAX_DATE,
+        minDate = DEFAULT_MIN_DATE,
         onDateSelect,
         onMonthChange,
```
gc-skyscanner: Maybe?
```suggestion
        onMonthChange = null
```

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarContainer.tsx
gc-skyscanner, line 232: To be honest, I am a bit concerned about if we could get the right "default value of `selectionConfiguration`" here after changing when `selectionConfiguration` is not set since we will use `selectionConfiguration` inside  `focusedDateHasChanged` directly 
https://github.com/Skyscanner/backpack/blob/af158e7559b6f241e0e131697ffccdc1f168679b/packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarContainer.tsx#L112-L113

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarDate.tsx
```diff
@@ -187,6 +174,7 @@ class BpkCalendarDate extends PureComponent<Props> {
 
     delete buttonProps.preventKeyboardFocus;
     delete buttonProps.isoLabel;
+    delete buttonProps.isToday;
```
gc-skyscanner, line 177: Could I ask why do we delete this prop in this PR?

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarWeek.tsx
gc-skyscanner, line 287: Same concern here
https://github.com/Skyscanner/backpack/pull/4673/changes#r3386072854

gc-skyscanner, line 188: could you help check if we will have runtime error when 
the original default props are set here? 
Same concern with
https://github.com/Skyscanner/backpack/pull/4673/changes#r3386072854

gc-skyscanner, line 216: could you help check if we will have runtime error when 
the original default props are set here? 
Same concern with
https://github.com/Skyscanner/backpack/pull/4673/changes#r3386072854

packages/backpack-web/src/bpk-component-datepicker/src/BpkDatepicker.tsx
gc-skyscanner: Do we need to set the default value here?

gc-skyscanner: Same with `prevProps.isOpen` do we need to set the default value for it?

packages/backpack-web/src/bpk-component-image/src/BpkImage.tsx
gc-skyscanner: Seems we should also set the default value for suppressHydrationWarning?

packages/backpack-web/src/bpk-component-info-banner/src/withBannerAlertState.tsx
gc-skyscanner, line 119: may be we should set the default value for `show`?

packages/backpack-web/src/bpk-component-input/src/BpkInput.tsx
```diff
@@ -22,7 +22,7 @@ import { withButtonAlignment } from '../../bpk-component-icon';
 import ClearIcon from '../../bpk-component-icon/sm/close-circle';
 import { cssModules, getDataComponentAttribute } from '../../bpk-react-utils';
 
-import { CLEAR_BUTTON_MODES, defaultProps } from './common-types';
```
gc-skyscanner, line 25: could we remove the defaultProps in `common-types`?

packages/backpack-web/src/bpk-component-input/src/withOpenEvents.tsx
```diff
@@ -150,7 +138,7 @@ const withOpenEvents = <P extends object>(WithOpenEventsInputComponent: Componen
     render(): ReactElement<any> {
       const {
         className,
-        hasTouchSupport,
+        hasTouchSupport = HAS_TOUCH_SUPPORT,
         isOpen,
```
gc-skyscanner, line 142: do we need to set the default value for `isOpen`?

packages/backpack-web/src/bpk-component-accordion/src/withAccordionItemState.tsx
Sybsw: No, neither default matters at this destructure site.

- `expanded` and `initiallyExpanded` are pulled out of `this.props` only to **strip them from `...rest`** before forwarding to `ComposedComponent`. Their values aren't read here; the wrapped component receives `expanded={this.state.expanded}` (managed state) instead.
- The previously-defaulted `initiallyExpanded` is read in the constructor (`const { initiallyExpanded = false } = props`), where the default is applied. The previously-defaulted `expanded` was an inert default — never read in `render`, never seeded into state — so dropping that default is intentional.
- The third old default (`onClick: null`) is also unused: the HOC's own `onClick` always overrides it on the wrapped element.

packages/backpack-web/src/bpk-component-banner-alert/src/withBannerAlertState.tsx
Sybsw: No, `show` is destructured here only to strip it from `...rest` before forwarding to `WrappedComponent`. The wrapped component is passed `show={this.state.show}` (managed state, seeded to `true` in the constructor), so the prop default is irrelevant for rendering.

The pre-migration `static defaultProps = { show: true }` was equally inert — `show` was never read in render, only state was — so the behaviour is unchanged.

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarContainer.tsx
Sybsw: Good catch — done in `0dea87351`. Defaulted both `onMonthChange` and `onDateSelect` to `null` in the render destructure, matching the pattern already used in `handleDateFocus` / `handleDateSelect` and the prior `static defaultProps` semantics.

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarContainer.tsx
Sybsw: Good catch, fixed in `0dea87351`.

You're right: `focusedDateHasChanged` reads `prevProps.selectionConfiguration!` / `currentProps.selectionConfiguration!` directly, so it doesn't see the destructure default applied in `componentDidUpdate`. With `static defaultProps` gone, an unset `selectionConfiguration` would crash on `.type` read.

Replaced both `!` assertions with `?? DEFAULT_SELECTION_CONFIGURATION` so the function is safe regardless of what the caller passes. Behaviour is identical to the pre-migration default.

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarDate.tsx
Sybsw: Filtering `isToday` out of `...buttonProps` before spreading onto the native `<button>` — without this, React would warn about an unknown `istoday` DOM attribute.

This wasn't required *before* this PR because `static defaultProps = { isToday: false }` was filling the prop in, then it was being read inside `render` (e.g. for the today modifier) — so it never landed in `...buttonProps`. After the migration, `isToday` isn't destructured in `render` (it's not read locally; it's a passthrough to internal modifier logic via `this.props`), which means it falls into `...buttonProps` and would leak to the DOM.

Matches the existing `delete buttonProps.preventKeyboardFocus` / `delete buttonProps.isoLabel` pattern.

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarWeek.tsx
Sybsw: The cross-referenced concern (`focusedDateHasChanged`) is fixed in `0dea87351`. For `BpkCalendarWeek` itself: `shouldComponentUpdate` reads `this.props.selectionConfiguration?.type === ...` with optional chaining (line 311), so an unset `selectionConfiguration` is safe — it just falls through to the next branch. No runtime change needed here.

Sybsw: `singleDateHandler` is only called from `shouldComponentUpdate` after the gate `this.props.selectionConfiguration?.type === CALENDAR_SELECTION_TYPE.single && this.props.selectionConfiguration.date` (line 311–313). That gate ensures `selectionConfiguration` is defined and is a `single`-type config with a non-null `date` before this function ever runs.

The cast inside (`as SelectionConfigurationSingle`) is therefore safe at runtime. With `static defaultProps` removed, the gate still does the work — same behaviour as before.

Sybsw: Same pattern as `singleDateHandler`. `rangeDateHandler` is gated by `this.props.selectionConfiguration?.type === CALENDAR_SELECTION_TYPE.range` (line 318) before being called. The optional chain handles the unset case (returns `false` from `sCU`), and inside the function `selectionConfiguration` is guaranteed range-typed.

The destructured `startDate` / `endDate` may legitimately be `undefined` for a partial range, which the function handles via the truthy `&&` chain (`startDate && nextStartDate && ...`). No runtime error.

packages/backpack-web/src/bpk-component-datepicker/src/BpkDatepicker.tsx
Sybsw: Done in `0dea87351`. Defaulted `isOpen = false` in the destructure to match the constructor (`const { isOpen = false } = props;`).

Sybsw: Also fixed in `0dea87351`: `const { isOpen: prevIsOpen = false } = prevProps;` and rewired the comparison to use `prevIsOpen`. Without the defaults, an undefined-on-first-render → false-on-second-render transition would have spuriously triggered `onOpen()`.

packages/backpack-web/src/bpk-component-image/src/BpkImage.tsx
Sybsw: The DOM attribute reads `this.props.suppressHydrationWarning` directly without destructuring it — and an undefined value here is fine: React only opts into the hydration-warning suppression when the prop is **truthy**, and the JSX serialises an undefined `suppressHydrationWarning={undefined}` as a no-op. Behaviour matches the prior `static defaultProps = { suppressHydrationWarning: false }` (false and undefined are equivalent for this attribute).

No functional change. Happy to add `suppressHydrationWarning = false` for symmetry if you'd prefer, let me know.

packages/backpack-web/src/bpk-component-info-banner/src/withBannerAlertState.tsx
Sybsw: Same answer as the banner-alert twin (#2): `show` is destructured only to keep it out of `...rest`. The wrapped component receives `show={this.state.show}` (managed state, seeded to `true` in the constructor), not the destructured prop. The pre-migration `defaultProps.show: true` was inert here too — no behavioural change.

packages/backpack-web/src/bpk-component-input/src/BpkInput.tsx
Sybsw: Not in this PR — `defaultProps` is **re-exported from the package's public entry point** (`bpk-component-input/index.ts:22,35`) so external consumers may rely on it. Removing it is a breaking-change candidate that should land alongside the wider Flow-→-TS / prop-types cleanup tracked in the PR description ("Full `.js` (Flow) prop-types removal" follow-up).

Inside this PR, `BpkInput` no longer reads `defaultProps` (defaults are inlined into the destructure), so the symbol is decoupled from the component. The lingering internal use is `BpkInput.stories.tsx`'s `ClearableInput` example, which I left as-is for the same external-consumer reason.

**I will make sure I clarify soon whether this can be included in a 3rd PR before time runs out for this release**

packages/backpack-web/src/bpk-component-input/src/withOpenEvents.tsx
Sybsw: No need here. `isOpen` is destructured only to strip it from `...rest` before forwarding (it's not a valid `<input>` DOM attribute and would otherwise leak as `isopen=""`). The component's own logic reads `this.props.isOpen` directly inside `handleBlur`, where `undefined` is correctly treated as falsy — same behaviour as the pre-migration `defaultProps.isOpen: false`.

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarDate.tsx
gc-skyscanner: That makes sense then. Just walk through the file, seems  `isToday` is pulled out of `this.props` only to strip them from `...rest` before forwarding to `button`, like the comment above

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarWeek.tsx
gc-skyscanner: you are right, but inside `singleDateHandler`, the `nextSelectConfig` (const nextSelectConfig =
    nextProps.selectionConfiguration as SelectionConfigurationSingle;) will be destructured directly that seems it assumed `nextSelectConfig` will not be undefined (line [195-197](https://github.com/Skyscanner/backpack/blob/0dea87351e87c97fdcdc4647b90f8c2755b36f1d/packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarWeek.tsx#L195-L197)). is it safe?

gc-skyscanner: Same concerned about the next props desctructuring https://github.com/Skyscanner/backpack/pull/4673#discussion_r3393013153

packages/backpack-web/src/bpk-component-image/src/BpkImage.tsx
gc-skyscanner: or maybe we could set `{!!this.props.suppressHydrationWarning}` here? to keep the type as boolean 😄

packages/backpack-web/src/bpk-component-calendar/src/BpkCalendarWeek.tsx
Sybsw: Agreed, the cast as SelectionConfigurationSingle/Range doesn't protect against undefined at runtime, and the destructure/property access inside each handler would throw if nextProps.selectionConfiguration was unset. Now both handlers fall back to DEFAULT_SELECTION_CONFIGURATION before the cast, restoring the pre-defaultProps safety.

Sybsw: Same reply https://github.com/Skyscanner/backpack/pull/4673#discussion_r3394417096

packages/backpack-web/src/bpk-component-image/src/BpkImage.tsx
Sybsw: Thanks, done

gc-skyscanner (dismissed):

PR #4690, opened 2026-06-11
[NO-JIRA]: fix(BpkProvider): validate html[lang] before passing to Ark to prevent crash loop

bjoern-wahle:
## Summary

- Fixes an indefinite crash loop introduced by https://github.com/Skyscanner/web-platform/pull/26006 on Jun 10 that caused ~55,000 errors across ~20 sessions before it was identified
- `getArkLocale()` now validates `html[lang]` with `new Intl.Locale()` before returning it, falling back to `'en-US'` for any value that throws — matching the pattern already used in `getLangDir()`

## Root cause

`getArkLocale()` reads `document.documentElement.getAttribute('lang')` and returns it directly if truthy (`return lang || 'en-US'`). That value is passed to Ark's `<LocaleProvider locale={locale}>`, which calls `@zag-js/i18n-utils` `isRTL(locale)` → `new Intl.Locale(locale)` with no try/catch.

Two truthy values that throw a `RangeError`:
- `""` (empty string) — `getAttribute('lang')` returns `""` not `null` when `<html lang="">` is present
- `"123"` (numeric string) — produced by some Skyscanner market locale configs

Confirmed affected markets: skyscanner.nl, skyscanner.pl, skyscanner.net, tianxun.com.

## Why it loops

The `RangeError` is thrown during `BpkProvider`'s re-render (triggered by `useEffect → setLocale(getArkLocale())`). The App `ErrorBoundary` catches it and renders its fallback — which #26006 also wrapped in `BpkProvider`. That fallback `BpkProvider` mounts, its `useEffect` fires, reads the same bad `html[lang]`, throws again. Repeat indefinitely. Each cycle fires `componentDidCatch` → one New Relic error event → Houston redirect on first cycle. Observed sessions generating 1,000–6,401 error events each.

## Validation

**Red/green TDD:**
1. Added three failing tests first (covering `lang="123"`, `lang=""`, and `lang="en-GB"` regression)
2. Confirmed `lang="123"` test fails against unfixed code: `new Intl.Locale("123")` throws `"Incorrect locale information provided"`
3. Applied fix → all 16 tests pass

**Loop regression test:**

`LoopBoundary` is an `ErrorBoundary` whose fallback also mounts `BpkProvider` — the exact production topology. The test renders `LoopBoundary > BpkProvider` with `html[lang]="123"` and asserts `catchCount === 0`.

<details>
<summary>What happens without the fix</summary>

Running the loop test against the unfixed code never returns — it exhausts the Node.js heap:

```
FATAL ERROR: Ineffective mark-compacts near heap limit
Allocation failed - JavaScript heap out of memory
```

The full React + jsdom cycle runs unbounded: `BpkProvider` renders → Ark's `LocaleProvider` calls `isRTL("123")` → `new Intl.Locale("123")` throws `RangeError: Incorrect locale information provided` → `LoopBoundary.componentDidCatch` fires → fallback renders another `BpkProvider` → same `html[lang]` → throws again → repeat.

The crash stack confirms the real production call path:

```
RangeError: Incorrect locale information provided
    at new Locale (<anonymous>)
    at isRTL (@zag-js/i18n-utils/dist/is-rtl.js)
    at LocaleProvider (@ark-ui/react/dist/providers/locale/locale-provider.cjs)
    at BpkProvider (BpkProvider.tsx)
```

</details>

<details>
<summary>What happens with the fix</summary>

`getArkLocale()` catches the `RangeError` before it reaches Ark and returns `'en-US'`. `BpkProvider` renders cleanly, `LoopBoundary` never catches, `catchCount` stays at 0. The test completes in ~1s.

</details>

## Test plan

- [ ] `lang="123"` → BpkProvider renders without crashing, Ark receives `"en-US"`
- [ ] `lang=""` → BpkProvider renders without crashing, Ark receives `"en-US"`  
- [ ] `lang="en-GB"` → Ark still receives `"en-GB"` (no regression for valid locales)
- [ ] ErrorBoundary+BpkProvider loop topology → no catch, no OOM, `catchCount === 0`
- [ ] Existing RTL tests unchanged (all 16 pass)

### DIFF
--- a/packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
+++ b/packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
@@ -223,7 +223,7 @@ describe('BpkProvider - RTL support', () => {
     expect(getByTestId('locale').textContent).toBe('en-US');
   });
 
-  it('falls back to en-US and does not crash when html[lang] is an invalid locale (numeric string)', async () => {
+  it('falls back to en-US and does not crash when html[lang] is an invalid locale (numeric string)', () => {
     document.documentElement.setAttribute('lang', '123');
 
     const { getByTestId } = render(
@@ -235,7 +235,7 @@ describe('BpkProvider - RTL support', () => {
     expect(getByTestId('locale').textContent).toBe('en-US');
   });
 
-  it('falls back to en-US and does not crash when html[lang] is an empty string', async () => {
+  it('falls back to en-US and does not crash when html[lang] is an empty string', () => {
     document.documentElement.setAttribute('lang', '');
 
     const { getByTestId } = render(
--- a/packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
+++ b/packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
@@ -16,6 +16,9 @@
  * limitations under the License.
  */
 
+import type { ReactNode } from 'react';
+import { Component } from 'react';
+
 import { useLocaleContext } from '@ark-ui/react';
 import { act, render } from '@testing-library/react';
 import '@testing-library/jest-dom';
@@ -23,6 +26,39 @@ import '@testing-library/jest-dom';
 import { BpkBox } from './BpkBox';
 import { BpkProvider } from './BpkProvider';
 import { BpkSpacing } from './tokens';
+
+// Mirrors the production topology: error boundary whose fallback also mounts BpkProvider.
+// This is exactly the structure that causes the infinite remount loop.
+class LoopBoundary extends Component<
+  { children: ReactNode },
+  { errored: boolean }
+> {
+  constructor(props: { children: ReactNode }) {
+    super(props);
+    this.state = { errored: false };
+  }
+
+  static getDerivedStateFromError() {
+    return { errored: true };
+  }
+
+  componentDidCatch() {
+    this.catchCount += 1;
+  }
+
+  catchCount = 0;
+
+  render() {
+    if (this.state.errored) {
+      return (
+        <BpkProvider>
+          <div data-testid="fallback">fallback</div>
+        </BpkProvider>
+      );
+    }
+    return this.props.children;
+  }
+}
 
 const LocaleReader = () => {
   const { locale } = useLocaleContext();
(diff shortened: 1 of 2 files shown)


AyaMallah (dismissed):

mungodewar (dismissed):
LGTM

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -90,7 +90,15 @@ const getArkLocale = (): string => {
     return FALLBACK_LOCALE_BY_DIRECTION[explicitDir];
   }
 
-  return lang || 'en-US';
+  if (lang) {
+    try {
+      const locale = new Intl.Locale(lang);
+      if (locale) return lang;
```
Supremeyh, line 96: seems the `if (locale)` guard is dead code:

```
const locale = new Intl.Locale(lang);
if (locale) return lang;   // ← always true; Intl.Locale never returns falsy
```
`new Intl.Locale()` either constructs a Locale object (always truthy) or throws. The if check adds noise and implies the object could be falsy, which it can't. could simplify to:

```
try {
  new Intl.Locale(lang);
  return lang;
} catch {
  // Invalid locale string — fall through to default
}
```

packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
```diff
@@ -187,6 +223,66 @@ describe('BpkProvider - RTL support', () => {
     expect(getByTestId('locale').textContent).toBe('en-US');
   });
 
+  it('falls back to en-US and does not crash when html[lang] is an invalid locale (numeric string)', async () => {
```
Supremeyh: Two new tests are marked async without any `await` and the empty-string test below it. They work, but `async` is misleading — can remove it.

AyaMallah: Cleaned up 🧹

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
AyaMallah: Nit: `if (locale)` is needed to satisfy the no-new ESLint rule, so will keep this as is...

mungodewar (dismissed):
LGTM

Supremeyh (approved):
lgtm

PR #4693, opened 2026-06-12
[CLOV-1642] feat(BpkProvider): add Emotion cache boundary to fix SSR hydration and Cypress visual tests

RichardSyq:
## Context

According to https://skyscanner.atlassian.net/wiki/spaces/~71202057130c3157a14c3bb0e231b5236fb88e/pages/2652111273/Decision+-+BPK+2.0+SSR+Hydration+Fix
Two related issues drove this change:

1. **SSR hydration** ([CLOV-1642](https://skyscanner.atlassian.net/browse/CLOV-1642)): BPK 2.0 components inside `BpkProvider` were breaking SSR hydration in `local-components-client` apps — the SSR skeleton stayed visible alongside real content after hydration. Root cause: without an Emotion cache boundary, Chakra v3 is the first Emotion boundary in the tree. Client classNames diverge from SSR output before Emotion registers the server styles, causing `ReactDOM.hydrate()` to silently bail the subtree.

2. **Cypress / Percy visual tests** ([CLOV-1667](https://skyscanner.atlassian.net/browse/CLOV-1667), [hotels-website#12025](https://github.com/Skyscanner/hotels-website/pull/12025)): Emotion's speedy mode inserts styles via `CSSStyleSheet.insertRule()`, hiding them from Percy's DOM serialiser. `hydrateRoot()` can also strip Emotion's SSR style nodes before screenshots are captured. hotels-website worked around this with a custom `EmotionCacheProvider` wrapper — this PR moves that fix into `BpkProvider` so all consumers get it for free.

## Changes

`BpkProvider` now wraps its tree in an Emotion `CacheProvider` (`key: 'css'`) so the cache is ready before Chakra's inner boundary.

- **Nesting safety**: nested `BpkProvider` instances reuse the parent cache via `BpkEmotionCacheContext`
- **Cypress / Percy**: detects `window.Cypress` / `window.hotelsDisableEmotionSpeedy` and uses `speedy: false` + recreates the cache after hydration via `useEffect`
- **SSR safe**: cache created via `useState` lazy initialiser, not a module-level singleton
- Adds `@emotion/cache` and `@emotion/react` as direct deps of `backpack-web`

## Test plan

- [ ] `TZ=Etc/UTC npx jest .../BpkProvider-test.tsx` — all tests pass
- [ ] `npx eslint .../BpkProvider.tsx` — no errors
- [ ] `npm run check-bpk-dependencies` — passes
- [ ] Verify SSR skeleton disappears after hydration in `cars-webapp` with BPK 2.0 components
- [ ] Verify Percy snapshots in Cypress pick up Backpack styles

The bundle size was checked, and it did not change significantly after the PR update.https://github.com/Skyscanner/carhire-homepage/pull/5517#issuecomment-4705812098
image

test in carhire
Before
image

After
image

[CLOV-1642]: https://skyscanner.atlassian.net/browse/CLOV-1642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLOV-1667]: https://skyscanner.atlassian.net/browse/CLOV-1667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -37,6 +41,27 @@ export interface BpkProviderProps {
  */
 const bpkSystem = createSystem(defaultBaseConfig, createBpkConfig());
 
+type CypressWindow = Window & {
+  Cypress?: unknown;
+  bpkDisableEmotionSpeedy?: boolean;
+};
+
+const isCypressEnv = (): boolean => {
```
kerrie-wu, line 53: The reference PR (hotels-website#12025) carried a full WORKAROUND comment block above this logic explaining: the root cause (full-document `hydrateRoot` strips SSR `<style>` tags → Emotion falls back to `insertRule()` → Percy's DOM serialiser can't read them), why it must live in production code, and a `Remove once Emotion / React Router 7 provides a cleaner solution` note. That context is *more* important here than in the app layer, because `BpkProvider` is depended on by every BPK 2.0 consumer. Without it, a future maintainer seeing `isCypressEnv` + the post-mount `setOwnCache` has no way to judge whether this can be deleted. Please port that comment block across.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -37,6 +41,27 @@ export interface BpkProviderProps {
...
+  if (win.Cypress || win.bpkDisableEmotionSpeedy) return true;
+  try {
+    return Boolean((win.parent as CypressWindow | undefined)?.Cypress);
+  } catch {
+    return false;
+  }
+};
+
+const createBpkEmotionCache = (speedy?: boolean) =>
+  createCache({ key: 'css', ...(speedy !== undefined && { speedy }) });
```
kerrie-wu: When `speedy === undefined` (the non-Cypress production path) this evaluates to `createCache({ key: 'css', ...false })`. Spreading a boolean primitive is legal (it yields `{}`), so it works — but it relies on a JS quirk that reads like a bug and some lint rules flag it. Prefer an explicit form, and add the return type while here:

```tsx
const createBpkEmotionCache = (speedy?: boolean): EmotionCache =>
  createCache(speedy === undefined ? { key: 'css' } : { key: 'css', speedy });
```

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -37,6 +41,27 @@ export interface BpkProviderProps {
...
+  const win = window as CypressWindow;
+  if (win.Cypress || win.bpkDisableEmotionSpeedy) return true;
+  try {
+    return Boolean((win.parent as CypressWindow | undefined)?.Cypress);
+  } catch {
+    return false;
+  }
+};
+
+const createBpkEmotionCache = (speedy?: boolean) =>
```
kerrie-wu: `'css'` is both Emotion's default key and the key Chakra v3 uses internally — sharing it is intentional so this boundary intercepts before Chakra registers its own. That's an implicit contract; a one-line comment here would stop someone from "cleaning it up" to a custom key and silently breaking the fix.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -135,14 +160,41 @@ const useArkLocale = (): string => {
...
- * @returns {JSX.Element} The provider wrapping its children with Chakra and Ark context.
+ * @returns {ReactElement} The provider wrapping its children with Chakra and Ark context.
  */
-export const BpkProvider = ({ children }: BpkProviderProps): JSX.Element => {
+export const BpkProvider = ({ children }: BpkProviderProps): ReactElement => {
+  const parentCache = useContext(BpkEmotionCacheContext);
+  const isNested = parentCache !== null;
+
+  const [isCypress] = useState(isCypressEnv);
+  const [ownCache, setOwnCache] = useState(() =>
```
kerrie-wu: **[nit]** When `isNested` is true, `ownCache` is initialised to `parentCache` but never read — the `if (isNested) return inner;` branch below doesn't render either provider. The `useState` + `useEffect` are dead weight in the nested case (harmless, since no extra `createCache` happens). Could short-circuit the state when nested. Non-blocking. (Note: *not* re-providing context when nested is the correct behaviour — a 3rd-level provider still reads the outermost cache. 👍)

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -37,6 +41,27 @@ export interface BpkProviderProps {
...
 
+type CypressWindow = Window & {
+  Cypress?: unknown;
+  bpkDisableEmotionSpeedy?: boolean;
+};
+
+const isCypressEnv = (): boolean => {
+  if (typeof window === 'undefined') return false;
+  const win = window as CypressWindow;
+  if (win.Cypress || win.bpkDisableEmotionSpeedy) return true;
```
kerrie-wu, line 58: **[suggestion]** Flag renamed from `hotelsDisableEmotionSpeedy` (what hotels sets in its `cypress/support`) to `bpkDisableEmotionSpeedy`. The primary Cypress path via `win.Cypress` is auto-injected and unaffected, but this *explicit flag* fallback won't fire for hotels until they update their key. Given the cross-repo move is already coordinated, a short inline comment naming the flag (so consumers know what to set) would help.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -37,6 +41,27 @@ export interface BpkProviderProps {
...
+  try {
+    return Boolean((win.parent as CypressWindow | undefined)?.Cypress);
+  } catch {
```
kerrie-wu, line 61: **[nit]** This swallows the `SecurityError` thrown when `win.parent` is a cross-origin frame. An empty `catch` reads like an oversight — add a one-liner, e.g. `// Cross-origin parent frame — SecurityError is expected`.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -135,14 +160,41 @@ const useArkLocale = (): string => {
...
+  const hasRecreated = useRef(false);
   const locale = useArkLocale();
 
-  return (
+  useEffect(() => {
+    if (isNested || !isCypress) return;
+    if (hasRecreated.current) return;
+    hasRecreated.current = true;
+    setOwnCache(createBpkEmotionCache(false));
+  }, [isCypress, isNested]);
```
kerrie-wu: **[nit]** Both `isCypress` (from `useState` with no setter) and `isNested` (from context, stable within this provider's lifetime) never change, so this effect already runs once. The reference PR used `[]` + the `hasRecreated` ref guard (the guard is needed to defend against StrictMode's double-invocation — keep it). Aligning to `[]` matches the validated pattern. Non-blocking.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
```diff
@@ -65,6 +79,91 @@ const LocaleReader = () => {
...
+
+    await waitFor(() => {
+      expect(mockCreateCache).toHaveBeenCalledWith({
+        key: 'css',
+        speedy: false,
+      });
+    });
+  });
+
+  it('recreates the cache after mount in Cypress to replace hydration-stripped styles', async () => {
```
kerrie-wu: **[nit]** The name says "in Cypress" but the test triggers via the `bpkDisableEmotionSpeedy` flag, not `window.Cypress`, and only asserts `>= 2` calls. Either rename to something like `...when speedy is disabled`, or add a separate case that drives recreation through `window.Cypress` so the flag path and the Cypress path are each locked down.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
```diff
@@ -65,6 +79,91 @@ const LocaleReader = () => {
...
 };
 
+describe('BpkProvider - Emotion cache', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+    delete (window as any).Cypress;
+    delete (window as any).bpkDisableEmotionSpeedy;
+  });
+
+  it('creates cache without speedy in a normal (non-Cypress) environment', () => {
```
kerrie-wu, line 89: This locks down "no `speedy: false`" but not "created exactly once / never recreated". Since the non-Cypress path still creates a cache, adding `expect(mockCreateCache).toHaveBeenCalledTimes(1);` would pin the "normal env never recreates" behaviour explicitly.

kerrie-wu (commented):
Functional logic looks correct — cache reference consistency, provider nesting order (CacheProvider outside Chakra), and context propagation all check out. Comments below are about value creation/passing clarity and missing context, not correctness blockers. Cross-repo coordination with hotels has been confirmed separately.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -37,6 +39,8 @@ export interface BpkProviderProps {
  */
 const bpkSystem = createSystem(defaultBaseConfig, createBpkConfig());
 
+const createBpkEmotionCache = () => createCache({ key: 'bpk' });
```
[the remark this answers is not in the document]
RichardSyq: Outdated — this comment was on the original commit (`4e921a9d`) before the redesign. The key is now `'css'` (intentionally shared with Chakra v3 so this boundary intercepts first), and a `BpkEmotionCacheContext` was introduced so nested `BpkProvider` instances reuse the parent cache instead of creating a new one. Resolving.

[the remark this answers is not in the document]
RichardSyq: Fixed in 41e412e6f. Added an optional `emotionCache` prop on `BpkProvider` and exported `createBpkEmotionCache` from the package. SSR entries can now do:

```tsx
const cache = createBpkEmotionCache();
const { extractCriticalToChunks } = createEmotionServer(cache);
const html = renderToString(<BpkProvider emotionCache={cache}>{app}</BpkProvider>);
const chunks = extractCriticalToChunks(html);
```

The Cypress post-mount recreate logic is skipped when an external cache is supplied (consumer owns the lifecycle).

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -127,14 +131,19 @@ const useArkLocale = (): string => {
  * tree render correctly in RTL without requiring additional wrapping or prop changes.
  *
  * @param {BpkProviderProps} props - The provider props.
- * @returns {JSX.Element} The provider wrapping its children with Chakra and Ark context.
+ * @returns {ReactElement} The provider wrapping its children with Chakra and Ark context.
  */
-export const BpkProvider = ({ children }: BpkProviderProps): JSX.Element => {
+export const BpkProvider = ({ children }: BpkProviderProps): ReactElement => {
```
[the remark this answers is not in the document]
RichardSyq: Leaving as-is. `ReactElement` is the more modern typing (React 19+ is moving away from `JSX.Element`), and the component does always return an element. Happy to revisit if there's a project-wide convention I missed.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
RichardSyq: Done in 41e412e6f — ported the context block. Trimmed it down to four lines (root cause + fix + remove-once condition + source PR) to keep it scannable, but all the key context is preserved.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
RichardSyq: Done in 41e412e6f, applied your suggested form verbatim:

```tsx
export const createBpkEmotionCache = (speedy?: boolean): EmotionCache =>
  createCache(speedy === undefined ? { key: 'css' } : { key: 'css', speedy });
```

Also exported now so SSR consumers can use it directly with `@emotion/server`.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
RichardSyq: Done in 41e412e6f.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
RichardSyq: Leaving as-is for now. Pulling the cache hooks behind an early return would either violate hooks rules or require splitting `BpkProvider` into a nested-vs-root pair of components, which felt heavier than the gain (a `useState` lazy initialiser that returns `parentCache` and a `useEffect` that early-returns). Happy to revisit if you'd like the split.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
RichardSyq: Done in 41e412e6f — added an inline comment noting it's an explicit escape hatch for non-Cypress Percy runs and that consumers set it in their test bootstrap.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
RichardSyq: Done in 41e412e6f.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
RichardSyq: Done in 41e412e6f — switched to `[]` with the `hasRecreated` ref guard for StrictMode. Added an `eslint-disable-next-line react-hooks/exhaustive-deps` plus a brief comment explaining the deps are stable for the provider's lifetime.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
RichardSyq: Done in 41e412e6f — split into two tests: `recreates the cache after mount when bpkDisableEmotionSpeedy is set` and `recreates the cache after mount when window.Cypress is set`, so both trigger paths are locked down independently.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider-test.tsx
RichardSyq: Tried `toHaveBeenCalledTimes(1)` first — it failed because React 18 StrictMode in tests can invoke the `useState` lazy initializer more than once. Kept the spirit of the suggestion via a different assertion: `not.toHaveBeenCalledWith({ speedy: false })` already proves no recreate happened (the recreate path is the only call site that passes `speedy: false`). Added a comment in the test explaining the choice. Let me know if you'd prefer a different shape.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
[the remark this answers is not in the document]
RichardSyq: Update: reverted in 31e860b22. After verifying that Emotion 11's inline SSR mode is sufficient for our hydration use case, the team decided to keep the cache strictly internal to `BpkProvider` as an intentional design choice — we'd rather not commit to a public cache surface that consumers would then depend on. If a future use case genuinely needs `<head>`-injected critical CSS, we can reopen the discussion then.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
RichardSyq: Update: the refactor (ternary + return type) is preserved, but the `export` is reverted in 31e860b22. The team decided to keep the helper internal — Emotion 11's inline SSR mode is sufficient for the hydration fix without exposing the cache to consumers.

kerrie-wu (approved):
Re-reviewed after the latest push (d245b64 → 31e860b). All nine comments from the previous round have been addressed:

- ✅ **WORKAROUND comment** ported back, with the `Ported from hotels-website#12025` + `Remove once Emotion / RR7...` notes — exactly the context the library layer needed.
- ✅ **spread-boolean** replaced with the explicit ternary and a `: EmotionCache` return type.
- ✅ **`key: 'css'`** now documents the intentional sharing with Chakra's internal cache.
- ✅ **flag rename** documented as an explicit escape hatch for non-Cypress Percy runs.
- ✅ **empty catch** annotated `cross-origin parent frame`.
- ✅ **useEffect deps** aligned to `[]` with the `hasRecreated` guard kept (StrictMode double-invoke) and the reasoning spelled out.
- ✅ **test rename + split**: the misleading `in Cypress` case is renamed, and a *separate* `window.Cypress`-driven recreation test was added — flag path and Cypress path are now each locked down.
- ✅ **'never recreates' assertion**: the non-Cypress test now proves the recreate path isn't taken via `not.toHaveBeenCalledWith({ speedy: false })`, with a clear comment.

The one remaining nit (#4: `ownCache` state is dead weight in the nested branch) is harmless and was flagged non-blocking — fine to leave. Value-passing data flow (cache reference consistency, CacheProvider-outside-Chakra ordering, context propagation, nested reuse) all verified correct. LGTM. 🚢

PR #4701, opened 2026-06-15
[LOOM-2442] React 19 Migration: Part 4 - Add nodeRef to remaining CSSTransition sites

Sybsw:
## Description

Step 4 of the React 19 migration. The audit on top of #4455/#4673/#4691 surfaced one remaining runtime blocker for shipping `@skyscanner/backpack-web` as React-19-forward-compatible: three production `react-transition-group@4` `CSSTransition` sites still relied on the library's `ReactDOM.findDOMNode(this)` fallback. React 19 removed `findDOMNode` from `react-dom`, so any of these sites would throw at runtime under an R19 host.

Jest masks the issue: the suite stubs `react-transition-group/CSSTransition` to a pass-through (`scripts/stubs/cssTransitionStub.js` via `package.json:85`), so neither the R18 baseline nor the new `Build / React19` overlay catches the gap. Storybook on R19 would.

## Why now

`react-transition-group@4.4.5` is the latest published version (Aug 2022). No v5 has shipped, no upstream R19 fix is available. The supported escape hatch is to pass a `nodeRef` to every `CSSTransition`/`Transition`/`TransitionGroup` site — which the team already did for four sites (`AnimateAndFade.tsx` × 2, `BpkDrawerContent.tsx`, `BpkFloatingNotification.tsx`). This PR closes the remaining three before the v43 release cut so consumers on R18 *or* R19 can install Backpack without surprises.

## Approach

Two patterns, both already used in the codebase:

1. **`TransitionInitialMount.tsx`** — children come from arbitrary callers (`BpkDialogInner`, `BpkPopover`). Mirrors the `cloneElement` pattern from `bpk-component-banner-alert/src/AnimateAndFade.tsx:148-166`: hold a `useRef<HTMLElement>(null)`, pass it as `nodeRef`, and `cloneElement(child, { ref: nodeRef })` to inject it into the existing single React-element child. **No DOM wrapper added.** Non-element children (`string`) fall through unchanged for type-compat.

2. **`BpkImage.tsx` and `BpkBackgroundImage.tsx`** — children of these `CSSTransition`s are already known host `<div>` elements (`bpk-image__spinner` / `bpk-background-image__spinner`). Mirrors `BpkFloatingNotification.tsx`: class-field `spinnerNodeRef = createRef<HTMLDivElement>()`, passed to `<CSSTransition nodeRef>` and to the inner `<div ref>`.

## Verification

- `npx eslint` on the three changed files — clean.
- `npm run typecheck` — clean.
- `TZ=Etc/UTC npx jest` on `TransitionInitialMount-test.tsx`, `BpkImage-test.tsx`, `BpkBackgroundImage-test.tsx`, `BpkDialog-test.tsx`, and the BpkPopover suite — **35/35 tests pass, 29/29 snapshots unchanged** (refs don't serialise; no DOM drift).

Pre-existing `act(...)` warnings inside BpkPopover (originating in `@floating-ui/react-dom`'s `flushSync`) are unrelated to this change.

## Stacks onto

#4691 (`LOOM-2442_C_v3`)

## Test plan

- [x] `npm run lint` — 0 errors
- [x] `npm run typecheck` — clean
- [x] `npx jest` for affected and consumer suites — 35/35 pass, snapshots stable
- [ ] R19 overlay: `npm install --no-save react@19.2.5 react-dom@19.2.5 @types/react@19.0.14 @types/react-dom@19.0.6 @types/prop-types`, copy into `packages/node_modules/`, then `npm run typecheck && npm run jest` — confirm zero `findDOMNode is not a function` errors trace to BpkImage / BpkBackgroundImage / TransitionInitialMount / BpkDialog / BpkPopover. (This will run in CI via `Build / React19`.)
- [ ] Storybook smoke under R19: `BpkImage` `WithAnimation` + `WithLazyLoadingAndAnimation`, `BpkBackgroundImage` `BackgroundImageWithLazyLoadingAndAnimation`, plus opening a `BpkDialog` and a `BpkPopover` — no `findDOMNode` console errors.


packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
```diff
@@ -38,18 +39,27 @@ const TransitionInitialMount = ({
...
-      appearActive: appearActiveClassName,
-    }}
-    in
-    appear
-    timeout={{ exit: 0, enter: 0, appear: transitionTimeout }}
-  >
-    {children}
-  </CSSTransition>
-);
+}: Props) => {
```
Supremeyh, line 48: The `Props` type retains `children: string | ReactNode`. When children is a string, `isValidElement(children)` returns false, so the string is passed directly to CSSTransition without a nodeRef-backed DOM node. On R19, CSSTransition would attempt `findDOMNode` on the string and throw.

This is almost certainly unreachable in practice (every real caller passes a dialog/popover element), but it's an invisible footgun. Options:

- Add a comment noting the string case is type-compat-only and never exercised with animation.
- Narrow the type to children: ReactElement and remove the string branch (breaking change, but would make the contract explicit).
- At minimum, the isValidElement guard silently no-ops on a string child, which could obscure a misconfigured caller.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
```diff
@@ -38,18 +39,27 @@ const TransitionInitialMount = ({
...
+      classNames={{
+        appear: appearClassName,
+        appearActive: appearActiveClassName,
+      }}
+      in
+      appear
+      timeout={{ exit: 0, enter: 0, appear: transitionTimeout }}
+    >
+      {isValidElement(children)
+        ? cloneElement(
```
Supremeyh: React docs flag `cloneElement` as discouraged. The PR description acknowledges this mirrors the AnimateAndFade.tsx pattern already in the codebase. Given that there's no `react-transition-group@5` and no better escape hatch, this is the correct pragmatic choice. Worth a // TODO: remove when react-transition-group v5 ships comment so it doesn't get forgotten.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
```diff
@@ -38,18 +39,27 @@ const TransitionInitialMount = ({
...
+            children as ReactElement<{ ref?: RefObject<HTMLElement | null> }>,
```
Supremeyh: it bypasses TS's prop-checking, at least need a comment explaining why it's safe would help future readers.

Supremeyh:
the [BpkDialogWrapper](https://github.com/Skyscanner/backpack/blob/46142a916059e85dd96c85428be89de18a9ee4c7/packages/backpack-web/src/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx#L157) also has CSSTransition, should it update too?

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
Sybsw: Agreed — narrowed in a60639580. Children is now typed as `ReactElement<{ ref?: RefObject<HTMLElement | null> }>`, which deletes the string branch entirely. The contract is now explicit: a single ref-accepting element is required. Knock-on benefit: the silent `isValidElement` no-op fallback is gone too, and the `as ReactElement<...>` cast is no longer needed.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
Sybsw: Agreed — added a `// TODO: revisit the cloneElement pattern when react-transition-group v5 ships; it is expected to remove the nodeRef requirement.` above the component in a60639580.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
Sybsw: Resolved by tightening the prop type rather than commenting around the cast. In a60639580, `children` is now typed as `ReactElement<{ ref?: RefObject<HTMLElement | null> }>` at the prop boundary, so `cloneElement(children, { ref: nodeRef })` typechecks directly — no `as` and no `isValidElement` runtime check. TS will now flag a misconfigured caller at compile time instead of silently no-op-ing.

Sybsw:
> the [BpkDialogWrapper](https://github.com/Skyscanner/backpack/blob/46142a916059e85dd96c85428be89de18a9ee4c7/packages/backpack-web/src/bpk-react-utils/src/BpkDialogWrapper/BpkDialogWrapper.tsx#L157) also has CSSTransition, should it update too?

Fixed in https://github.com/Skyscanner/backpack/commit/a6063958087d90dd839000a2b75d3c8eb474be20. Wired nodeRef={ref} on the CSSTransition reusing the existing useRef<HTMLDialogElement> that's already attached to the <dialog> for showModal()/close() — no extra ref needed and no DOM/snapshot impact. HTMLDialogElement extends HTMLElement so the types line up. Verified clean under both R18 and R19 (typecheck + full jest, 731 snapshots stable).

Supremeyh (dismissed):

PR #4708, opened 2026-06-16
[CLOV-1803] Export BpkEmotionCacheContext from BpkProvider for external cache injection

IrinaWei:
## Summary

- Export `BpkEmotionCacheContext` from `bpk-component-layout` so host apps can inject an external Emotion cache into `BpkProvider`
- When a non-null cache is provided via `BpkEmotionCacheContext`, `BpkProvider` operates in nested mode: skips creating its own cache and uses the external one for both `CacheProvider` and `BpkEmotionCacheContext`
- Add `@emotion/cache` and `@emotion/react` to `backpack-web` dependencies (Chakra v3 already brings these transitively; this makes the direct usage explicit)

## Why

hotels-website experiences a CSS flash when a browser extension (e.g. Google Translate) deletes SSR-injected emotion `<style>` tags before hydration completes. This triggers a React hydration error, after which all Backpack/Chakra styles are missing.

The fix in hotels-website is to recreate the Emotion cache after recovery and force all components to re-inject styles. But `BpkProvider` creates its own private cache — without this PR, there's no way for the host to take control of that cache.

With `BpkEmotionCacheContext` exported, hotels-website can wrap `BpkProvider` with a context value pointing to a fresh cache, so the full component tree (including Chakra internals) re-injects styles after recovery.

## Test plan

- [ ] Existing `BpkProvider` tests pass
- [ ] New test: `BpkProvider` uses own cache when no external cache provided
- [ ] New test: `BpkProvider` uses external cache injected via `BpkEmotionCacheContext`
- [ ] `BpkProvider` renders correctly in Storybook (no visual regression)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


Faye-Xiao (dismissed):
LGTM. Giving host applications the ability to inject and control the Emotion cache feels like the right level of flexibility, while keeping the default behaviour unchanged.

packages/backpack-web/src/bpk-component-layout/src/BpkProvider.tsx
```diff
@@ -160,23 +166,29 @@ const useArkLocale = (): string => {
...
-  const parentCache = useContext(BpkEmotionCacheContext);
-  const isNested = parentCache !== null;
+  const externalCache = useContext(BpkEmotionCacheContext);
+  const isNested = externalCache !== null;
 
   const [isCypress] = useState(isCypressEnv);
-  const [ownCache, setOwnCache] = useState(() =>
-    isNested ? parentCache : createBpkEmotionCache(isCypress ? false : undefined),
+  const [ownCache, setOwnCache] = useState<EmotionCache>(() =>
+    isNested ? externalCache : createBpkEmotionCache(isCypress ? false : undefined),
```
Faye-Xiao: One small nit: `isNested` reads a little implementation-focused to me — something like `hasExternalCache` might make the intent clearer.

Faye-Xiao (approved):

PR #4719, opened 2026-06-22
fix(BpkPopover, BpkDatepicker): fix isOpen prop sync and portal rendering

paulrose:
## Summary

These bugs were discovered while migrating [travel-widgets](https://github.com/Skyscanner/travel-widgets/pull/1077) from React 17 → 18 and bumping `@skyscanner/backpack-web` from v16 to v35. travel-widgets sits in a slightly unusual position: it renders Backpack components inside an **iframe** embedded in partner websites, and uses `renderTarget` to keep portals inside a specific container within that iframe (so that DOM scoping and post-message event handling work correctly). This combination exposed two bugs that likely go unnoticed in typical SPA usage.

A working patch (`patch-package`) is in place in the travel-widgets PR for now — these fixes replace it properly.

---

### Bug 1: `BpkDatepicker` — `renderTarget` not forwarded to `BpkPopover` on desktop

`renderTarget` was passed to `BpkModal` (mobile path) but silently dropped on the desktop (`BpkPopover`) path. Consumers that supply `renderTarget` to control where the calendar portal mounts had it ignored on desktop.

In the travel-widgets iframe case, this meant `FloatingPortal` rendered the calendar into `document.body` **outside** the `#skyscanner-travel-widgets-wrapper` element that Cypress tests scope their queries to, so no calendar element was ever found.

**Fix:** forward `renderTarget` to `BpkPopover` in the desktop branch:

```tsx
<BpkPopover
  ...
  renderTarget={renderTarget}  // ← was missing
  ...
>
```

**Test:** mocks the breakpoint to desktop, passes a `jest.fn()` as `renderTarget`, opens the datepicker via `isOpen`, and asserts `FloatingPortal` called `renderTarget()` — which only happens when `renderTarget` is forwarded to `BpkPopover`.

---

### Bug 2: `BpkPopover` — `null` renderTarget silently suppresses the portal

`@floating-ui/react`'s `FloatingPortal` treats `root={null}` as an explicit "don't render" signal — `useFloatingPortalNode` returns early and no portal node is ever created. The previous type signature allowed `null` to flow through to `FloatingPortal`, which meant any consumer passing `renderTarget={() => null}` (a valid "no custom target" sentinel) would silently prevent the popover from rendering.

**Fix:** coerce `null → undefined` so `FloatingPortal` falls back to `document.body`:

```tsx
<FloatingPortal root={renderElement ?? undefined}>
```

The `renderTarget` type is also widened to `null | HTMLElement | (() => HTMLElement | null | undefined)` to match what `BpkDatepicker` declares and what the runtime already handles.

---

**Reference:** https://github.com/Skyscanner/travel-widgets/pull/1077 — the patch-package workaround that prompted these fixes, and the e2e test failures it resolves.


packages/backpack-web/jest.config.js
```diff
@@ -43,6 +43,7 @@ module.exports = {
       '<rootDir>/../../scripts/stubs/cssTransitionStub.js',
     '@skyscanner/bpk-svgs/dist/svgs/^.+\\.svg$':
       '<rootDir>/../../scripts/stubs/fileStub.js',
+
```
xiaogliu: The lone blank-line change here looks unintentional — worth dropping.

packages/backpack-web/src/bpk-component-popover/src/BpkPopover.tsx
```diff
@@ -141,9 +141,7 @@ const BpkPopover = ({
   const [isOpenState, setIsOpenState] = useState(isOpen);
 
   useEffect(() => {
-    if (!isOpen) {
-      setIsOpenState(false);
-    }
+    setIsOpenState(isOpen);
```
xiaogliu: Hey Paul, the `isOpen` logic changed seems caused the testing failure. Could check in you microsite with the fixing? You can test by dev version https://github.com/Skyscanner/backpack/blob/main/DEV_RELEASE.md

paulrose:
Just pushed a regression test for Bug 1 (renderTarget not forwarded on desktop).

The test approach: mock the breakpoint to desktop (`useMediaQuery` returns `false`), pass a `jest.fn()` as `renderTarget`, open the datepicker via `isOpen`, then assert that `renderTargetFn` was called. `FloatingPortal` only calls `renderTarget()` when it actually receives it — so if `BpkDatepicker` ever stops forwarding `renderTarget` to `BpkPopover` on the desktop path, the mock won't be called and the test fails.

For Bug 2 (null → undefined coercion in `FloatingPortal root`): `@floating-ui/react` in JSDOM doesn't reproduce the portal-suppression behaviour from a real browser, so there's no reliable unit test for that one. The Cypress e2e tests in the travel-widgets PR cover it end-to-end.

IrinaWei (approved):
LGTM

PR #4725, opened 2026-06-23
[NO-JIRA][Storybook] Add dark/light mode toolbar toggle

gert-janvercauteren:
## Summary

- Adds a dark/light mode toolbar toggle to Storybook that sets `data-theme` on `:root` and updates the canvas background via `var(--bpk-canvas-default)` from Backpack's theme CSS files (already bundled via `bpk-stylesheets`).
- Individual stories can pin their canvas independent of the toolbar by setting `parameters: { bpkTheme: 'dark' | 'light' }` — applied to on-dark and on-light button stories.
- Registers a `dark-mode-compatible` Storybook tag for filtering stories by dark-mode support.

## Checklist

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests


packages/backpack-web/src/bpk-component-button/src/BpkButton.stories.tsx
```diff
@@ -502,6 +502,8 @@ export const BpkButtonPrimary = {
 
 export const BpkButtonPrimaryOnDark = {
   render: () => <PrimaryOnDarkExample />,
+  parameters: { bpkTheme: 'dark' },
+  tags: ['dark-mode-compatible'],
 };
 
 export const BpkButtonPrimaryOnLight = {
```
IrinaWei, line 509: PrimaryOnLightExample renders directly on the canvas background (no BpkDarkExampleWrapper), so --bpk-canvas-default from document.body is visible behind the button. When the toolbar is toggled to dark, data-theme="dark" is set on :root, the canvas CSS variable resolves to a dark colour, and this "on-light" variant ends up on a dark background — which is exactly the context it's not designed for.

The OnDark stories are pinned with parameters: `{ bpkTheme: 'dark' }` to prevent the inverse problem. This story needs the same treatment in the other direction:

```
export const BpkButtonPrimaryOnLight = {
  render: () => <PrimaryOnLightExample />,
  parameters: { bpkTheme: 'light' },
};
```

libs/backpack-storybook-host/.storybook/preview.tsx
```diff
@@ -25,8 +25,29 @@ import type { Preview } from '@storybook/react';
 import '@skyscanner/backpack-web/bpk-stylesheets';
 import '@skyscanner/backpack-web/bpk-stylesheets/font';
 import '@skyscanner/backpack-web/bpk-stylesheets/larken';
+import '@skyscanner/backpack-web/bpk-stylesheets/theme-backpack-dark';
+import '@skyscanner/backpack-web/bpk-stylesheets/theme-backpack-light';
```
IrinaWei: `bpk-stylesheets/index.js` already imports both theme CSS files unconditionally. Importing theme-backpack-dark and theme-backpack-light again here means the stylesheets are bundled/loaded twice. webpack deduplication may hide the symptom, but the intent is obscured and it creates a subtle ordering dependency between the two import sites.

Suggest removing lines 28–29 from preview.tsx and relying on the single import path through bpk-stylesheets (already imported on line 25).

IrinaWei:
The PR summary says individual stories pin their canvas using parameters: `{ background: 'dark' | 'light' }`, but the decorator reads `parameters.bpkTheme`, not parameters.background. A contributor following the PR description would write background: 'dark' and silently get no effect — the toolbar would still override the canvas.

Please update the PR description (and any inline docs) to use bpkTheme consistently.

IrinaWei (approved):
LGTM

PR #4729, opened 2026-06-24
[CLOV-1807] Update adoption guard release workflow

RichardSyq:
## Summary
https://skyscanner.atlassian.net/browse/CLOV-1807

- Replace manual adoption guard release versions with a `patch` / `minor` / `major` workflow input and compute the next immutable `backpack-adoption-guard/vMAJOR.MINOR.PATCH` tag from existing tags.
- Build `packages/backpack-adoption-guard/dist/` into a tag-only release commit so generated bundles stay out of `main` while released GitHub Action refs remain directly executable.
- Use a GitHub App installation token with Git HTTPS Basic auth for tag pushes, without persisting credentials in the repository remote.
- Skip husky hooks only for the automated release commit so `lint-staged` does not try to restage ignored generated `dist/` files.
- Restrict tag creation to workflow runs dispatched from the `main` branch; runs from other branches validate only.
- Document the adoption guard release type rules and exact-version pinning guidance.

## Validation

Release works in https://github.com/Skyscanner/backpack/actions/runs/28083958314/job/83145656306
New version work properly in https://github.com/Skyscanner/carhire-homepage/pull/5558

- Parsed `.github/workflows/backpack-adoption-guard-release.yml` with Ruby YAML.
- Ran `git diff --check` on the changed files.
- Ran Docker `actionlint` for `.github/workflows/backpack-adoption-guard-release.yml`.
- Ran Docker `zizmor` for `.github/workflows/backpack-adoption-guard-release.yml` with no findings.
- Verified current tag calculation: `patch -> 1.0.1`, `minor -> 1.1.0`, `major -> 2.0.0`.
- Ran a `dry_run=true` workflow dispatch on the PR branch; `Validate` passed and `CreateTags` was skipped.
- Ran a non-dry-run workflow dispatch while branch validation was temporarily enabled; `CreateTags` succeeded and pushed `backpack-adoption-guard/v1.0.1`.
- Restored the `main`-only guard for `CreateTags` after validating the tag push path.


.github/workflows/backpack-adoption-guard-release.yml
```diff
@@ -24,43 +24,86 @@ defaults:
...
         run: |
-          sha="$(git rev-parse --verify "${TARGET_REF}^{commit}")"
+          sha="$(git rev-parse HEAD)"
           echo "target_sha=${sha}" >> "$GITHUB_OUTPUT"
-          echo "Resolved ${TARGET_REF} -> ${sha}"
+          echo "Resolved ${GITHUB_REF} -> ${sha}"
+
+      - name: Compute next release tag
+        id: compute
+        env:
```
xiaogliu, line 56: **Suggestion**: Re-fetch tags here and re-verify `$RELEASE_TAG` does not exist before committing or pushing. The `Validate` job checked this earlier, but a concurrent dispatch could have pushed the same tag in the window between `Validate` and `CreateTags`. `cancel-in-progress: false` prevents cancellation but does not prevent two simultaneous runs from reaching this job.

```yaml
- name: Re-verify tag does not already exist
  env:
    RELEASE_TAG: ${{ needs.Validate.outputs.release_tag }}
  run: |
    git fetch --force --tags origin
    if git rev-parse "refs/tags/$RELEASE_TAG" >/dev/null 2>&1; then
      echo "::error::Tag $RELEASE_TAG already exists (concurrent release detected)"
      exit 1
    fi
```

.github/workflows/backpack-adoption-guard-release.yml
```diff
@@ -24,43 +24,86 @@ defaults:
...
+          RELEASE_TYPE: ${{ inputs.release_type }}
```
xiaogliu, line 57: **Suggestion**: Add a quick integrity check after the build to ensure `dist/index.js` was actually produced before staging and tagging. An empty or partially-written dist would produce a tagged-but-broken release.

```yaml
- run: test -f packages/backpack-adoption-guard/dist/index.js || (echo "::error::dist/index.js not found after build" && exit 1)
```

packages/backpack-adoption-guard/README.md
```diff
@@ -131,3 +133,20 @@ The generated `dist/` directory is ignored by git. Any ref used directly as a
 GitHub JavaScript action must still contain the generated bundle because
 `action.yml` runs `dist/index.js`; keep that packaging step separate from normal
 source changes.
+
+## Releasing
+
+Use the `Backpack Adoption Guard Release` workflow and choose the semver bump
```
xiaogliu, line 139: **Suggestion**: Add an explicit deprecation note for the old floating `v1` tag. Consumers who pinned to `@backpack-adoption-guard/v1` will get no warning that the tag is no longer updated. A one-liner migration note here would help:

```
> **Migration**: The floating `backpack-adoption-guard/v1` major tag is no longer updated.
> Pin to an immutable tag such as `backpack-adoption-guard/v1.0.0` going forward.
```

.github/workflows/backpack-adoption-guard-release.yml
RichardSyq: Addressed in `fa4140420`: `CreateTags` now re-fetches tags and re-checks `refs/tags/$RELEASE_TAG` before staging/committing/pushing. If the tag already exists, the release fails with `Tag $RELEASE_TAG already exists (concurrent release detected)`.

.github/workflows/backpack-adoption-guard-release.yml
RichardSyq: Addressed in `fa4140420`: added a `Verify adoption guard bundle` step immediately after `npx nx run backpack-adoption-guard:build` to fail if `packages/backpack-adoption-guard/dist/index.js` was not produced.

packages/backpack-adoption-guard/README.md
RichardSyq: Addressed in `fa4140420`: added a README migration note that the floating `backpack-adoption-guard/v1` major tag is no longer updated, and consumers should pin to an immutable tag such as `backpack-adoption-guard/v1.0.0` or newer.

xiaogliu (approved):
LGTM

PR #4730, opened 2026-06-24
[NO-JIRA][Storybook] Collapse iframe bundle for Percy to fix chunk-load flakiness

gert-janvercauteren:
## Summary

Percy visual tests have been intermittently failing with `Loading chunk N failed` errors on random stories (e.g. `render.percy.local/1437.b0636bd1.iframe.bundle.js`). Root cause: Storybook lazy-loads each story file with a dynamic `import()`, so webpack emits one async chunk per story (122 in this build). Under load a chunk request occasionally outruns Percy's asset-discovery proxy and the iframe renders an error frame instead of the story.

For the static build only, this PR forces webpack to merge every async chunk back into the main iframe bundle via `LimitChunkCountPlugin({ maxChunks: 1 })` plus `splitChunks: false` / `runtimeChunk: false`. With no dynamic imports left to fetch at snapshot time, the race disappears. `dist-storybook/*.iframe.bundle.js` goes from 122 chunks to 1 (~7.4 MB). Dev (`serve-storybook`) is untouched — default splitting stays on so HMR remains fast.

The `package-lock.json` churn is unrelated lockfile metadata flips that were already present in the workspace.

## Checklist

- [x] Storybook static build still produces a working `iframe.html` (verified locally — served the dist via http-server, iframe references the single bundle, bundle serves 200)
- [x] No component / API changes
- [ ] `skip-changelog` label — build-system tweak only


xiaogliu (approved):
LGTM.

PR #4732, opened 2026-06-24
[NOJIRA]add locale-aware line-break rules for non-Latin scripts

IrinaWei:
Add :lang() CSS rules to font.scss so kinsoku shori (JIS X 4051) and equivalent rules are enforced globally via the existing <html lang="..."> attribute — no per-component changes required.

Key decisions:
- line-break: strict for CJK/Korean fixes punctuation/particles appearing at line-start (を、に、の etc. must not begin a line)
- Japanese base rule wraps per-character under kinsoku (no `word-break: keep-all`) so it never overflows in browsers without `auto-phrase`; Chromium is upgraded to phrase-level breaks via `@supports (word-break: auto-phrase)`
- overflow-wrap: break-word for CJK as last-resort fallback; preserves Latin brand names and URLs on mixed-script pages unlike word-break: break-all
- word-break: keep-all for Korean preserves syllable-block boundaries; overflow-wrap: break-word fallback handles 360–390px narrow viewports
- bpk-breakpoint-query-small-tablet (48rem) loosens line-break to loose for CJK/Korean to avoid uneven rag on narrow screens
- Thai uses break-all + overflow-wrap: anywhere — no word spaces, ICU dictionary failure leaves zero break points otherwise
- Arabic/Hebrew: overflow-wrap only; direction left to existing html[dir] RTL system
- Russian/Ukrainian: hyphens: auto degrades silently in Chrome

## Japanese line breaking across browsers

Japanese has no word spaces, so there is no single CSS declaration that gives phrase-aware (文節) breaks everywhere. We handle it in layers:

- **Chromium (Chrome/Edge)** — `word-break: auto-phrase` breaks at natural phrase boundaries automatically. Best-in-class, but Chromium-only ([caniuse](https://caniuse.com/mdn-css_properties_word-break_auto-phrase)).
- **Firefox / Safari** — no `auto-phrase`, so they fall back to standard per-character wrapping under `line-break: strict` (kinsoku). This is the conventional way Japanese wraps in print and, crucially, **never overflows the container**. (An earlier revision used `word-break: keep-all` in the base rule; that forbids every break in space-less Japanese and caused horizontal overflow in Firefox/Safari — now fixed.)

### Optional `<wbr>` phrase hints (works in every browser)

For key marketing/UI copy where phrase-aware breaks are wanted in **all** browsers, the author inserts `<wbr>` at phrase boundaries. Backpack detects it via `:lang(ja):has(wbr)` and switches that text to `word-break: keep-all`, making the `<wbr>` marks the only break points — so Firefox and Safari honour the same boundaries as Chromium. **No consumer className is needed** — `<wbr>` is content, and its presence is itself the opt-in signal.

```jsx
// Safe everywhere; phrase-aware only in Chromium
<BpkText>プライスアラートを設定して東京行きのお得な航空券を見つけましょう</BpkText>

// Phrase-aware in every browser (incl. Firefox)
<BpkText>プライスアラートを<wbr />設定して<wbr />東京行きの<wbr />お得な航空券を<wbr />見つけましょう</BpkText>
```

Note: under `keep-all`, if the viewport is narrower than a single hinted phrase (e.g. narrower than `プライスアラートを`), that phrase still wraps mid-word (`プライスアラート` / `を`) rather than overflow the element — breaking is preferred over horizontal overflow.

Documented in the `bpk-stylesheets` README. See the [`bpk-stylesheets-fonts`](https://backpack.github.io/storybook/?path=/story/bpk-stylesheets-fonts--larken-font) Larken story ("Japanese line-break validation") — compare the default vs `<wbr>`-hinted samples in Firefox.

- [x] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/packages/backpack-web/src/bpk-stylesheets/README.md
+++ b/packages/backpack-web/src/bpk-stylesheets/README.md
@@ -161,13 +161,13 @@ By default, Japanese text (`:lang(ja)`) wraps safely in every browser:
 - **Chromium** (Chrome/Edge) uses `word-break: auto-phrase` to break at natural phrase (文節) boundaries automatically.
 - **Firefox / Safari** have no equivalent, so they fall back to standard per-character wrapping under kinsoku (`line-break: strict`). This is the conventional way Japanese wraps in print and never overflows the container.
 
-For key marketing/UI copy where you want phrase-aware breaks in **all** browsers (not just Chromium), insert `<wbr>` at the phrase boundaries. Backpack detects the presence of `<wbr>` via `:lang(ja):has(wbr)` and automatically switches that text to `word-break: keep-all`, making the `<wbr>` marks the only break points — so Firefox and Safari honour the same boundaries as Chromium.
+For key marketing/UI copy where you want phrase-aware breaks in **all modern browsers** (Chrome 105+, Safari 15.4+, Firefox 121+), insert `<wbr>` at the phrase boundaries. Backpack detects the presence of `<wbr>` via `:lang(ja):has(wbr)` and automatically switches that text to `word-break: keep-all`, making the `<wbr>` marks the only break points — so Firefox and Safari honour the same boundaries as Chromium.
 
 ```jsx
 // Without <wbr>: safe everywhere, phrase-aware only in Chromium.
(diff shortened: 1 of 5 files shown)


gc-skyscanner (approved):
Confirmed with localisation team, LGTM

PR #4738, opened 2026-06-25
[WOODPECKER-4406] Migrate package manager from npm to pnpm

viktor-yang:
https://skyscanner.atlassian.net/browse/WOODPECKER-4406

## What changed

This PR migrates the Backpack monorepo package manager from npm to pnpm.

It was generated by the [skill](https://github.com/Skyscanner/web-platform/blob/main/.claude/skills/monorepo-npm-to-pnpm9/SKILL.md), then reviewed and refined by a person.

## Why

pnpm provides faster installs, stricter dependency isolation (phantom dependency prevention), and better monorepo workspace support — aligning Backpack's toolchain with the direction set in monorepo

## Checklist

- [x] Build/CI infrastructure change only — no component, test, or Storybook changes
- [x] No `README.md` changes needed
- [x] No accessibility checks needed


package.json
```diff
@@ -83,7 +80,8 @@
       "^.+\\.(svg|png)$": "<rootDir>/scripts/stubs/fileStub.js",
       "react-transition-group/CSSTransition": "<rootDir>/scripts/stubs/cssTransitionStub.js",
       "@skyscanner/bpk-svgs/dist/svgs/^.+\\.svg$": "<rootDir>/scripts/stubs/fileStub.js",
-      "^react($|/.+)": "<rootDir>/node_modules/react$1"
+      "^react($|/.+)": "<rootDir>/node_modules/react$1",
+      "^react-dom($|/.+)": "<rootDir>/node_modules/react-dom$1"
```
viktor-yang, line 84: Two `react-dom` instances exist at runtime: the root is upgraded to `19.2.5` by `pnpm add -w`, but `@ark-ui/react` and `@zag-js/react` still load their lockfile-locked peer instance `react-dom@18.3.1`. When `react-dom@18` initialises, `moduleNameMapper` redirects its `require('react')` to `react@19`, whose restructured internals no longer expose `ReactCurrentDispatcher` — crash.

**Fixed** by adding `"^react-dom($|/.+)": "<rootDir>/node_modules/react-dom$1"` to `moduleNameMapper`, so all `require('react-dom')` calls — including those from the `@18` peer instances — are redirected to the root `react-dom@19.2.5`, eliminating the version mismatch.
image

package.json
```diff
@@ -98,7 +95,7 @@
     ],
     "testRegex": "(?:packages|token-sync)/.*-test\\.[jt]sx?$",
     "transformIgnorePatterns": [
-      "node_modules/(?!bpk|@skyscanner|d3-.*|internmap)"
+      "node_modules/(?!(.pnpm|bpk|@skyscanner|d3-.*|internmap))"
     ],
```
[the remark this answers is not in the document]
viktor-yang: (1) Yeah, need to fix
(2) No. It is the right way to use the `.pnpm` and it only match the scoped pacakges , according to the official document:
https://jestjs.io/zh-Hans/docs/configuration#transformignorepatterns-arraystring

.github/workflows/release.yml
```diff
@@ -22,22 +22,23 @@ jobs:
         with:
           persist-credentials: false
 
+      - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version-file: ".nvmrc"
```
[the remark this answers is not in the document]
viktor-yang: The `registry-url` still exist
https://github.com/Skyscanner/backpack/pull/4738/changes/BASE..82c97480ce28d5b9bb9a43a5d4e335a04b6efab1#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R109

.github/workflows/_build.yml
```diff
@@ -136,16 +137,16 @@ jobs:
           key: ${{ env.CACHE_NAME }}-${{ hashFiles('package-lock.json') }}
 
       - name: Override React to 19.2.5
-        run: npm install --no-save react@19.2.5 react-dom@19.2.5 @types/react@19.0.14 @types/react-dom@19.0.6 @types/prop-types
+        run: pnpm add --no-save react@19.2.5 react-dom@19.2.5 @types/react@19.0.14 @types/react-dom@19.0.6 @types/prop-types
```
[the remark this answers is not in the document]
viktor-yang: Fix by pnpm_config_lockfile

.github/workflows/_build.yml
```diff
@@ -13,7 +13,6 @@ defaults:
     shell: bash -l {0}
 
 env:
-  CACHE_NAME: node-modules-cache-v2
```
viktor-yang, line 16: Replace the mannual cache to the PNPM cache in  ` actions/setup-node`

.github/workflows/_build.yml
```diff
@@ -122,30 +118,27 @@ jobs:
...
-        with:
-          path: |
-            node_modules/
-          key: ${{ env.CACHE_NAME }}-${{ hashFiles('package-lock.json') }}
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
 
       - name: Override React to 19.2.5
-        run: npm install --no-save react@19.2.5 react-dom@19.2.5 @types/react@19.0.14 @types/react-dom@19.0.6 @types/prop-types
+        run: pnpm_config_lockfile=false pnpm add -w react@19.2.5 react-dom@19.2.5 @types/react@19.0.14 @types/react-dom@19.0.6 @types/prop-types @types/react-transition-group@4.4.12
```
viktor-yang, line 132: Declare the `@types/react-transition-group@4.4.12` to fix below type error:

image

Reason: Because npm uses flat hoisting, `npm install @types/react@19` overwrites the single shared `node_modules/@types/react/` that every package resolves to, so `@types/react-transition-group` automatically picks up the new version. With pnpm, `@types/react-transition-group@4.4.11` declares `@types/react` as a `dependency`, so pnpm installs a private isolated copy (`18.3.1`) inside `.pnpm/` that the root override can never touch — causing a type mismatch. `4.4.12` fixed this by moving `@types/react` to `peerDependencies`, so pnpm no longer installs a private copy and resolves directly to the root `19.0.14`.

```
npm (flat hoisting)
─────────────────────────────────────────────────────
node_modules/
└── @types/react/  ← 19.0.14 (overwritten, single copy)
                        ▲
                        │ both resolve to the same copy ✅
                        │
@types/react-transition-group ──────────────────────┘


pnpm + 4.4.11  (@types/react as "dependencies")
─────────────────────────────────────────────────────
node_modules/
├── @types/react/                          ← 19.0.14 (root, overridden)
└── .pnpm/
    └── @types+react-transition-group@4.4.11/
        └── node_modules/
            └── @types/react/              ← 18.3.1 (private copy, untouched)
                    ▲
                    │ resolves to OLD copy → type mismatch ❌
                    │
@types/react-transition-group ──────────────────────┘


pnpm + 4.4.12  (@types/react as "peerDependencies")
─────────────────────────────────────────────────────
node_modules/
├── @types/react/                          ← 19.0.14 (root)
└── .pnpm/
    └── @types+react-transition-group@4.4.12/
        └── node_modules/
            └── @types/react/              ← (no private copy, peer resolved from root)
                    ▲
                    │ resolves to root copy → types aligned ✅
                    │
@types/react-transition-group ──────────────────────┘
```

.claude/skills/backpack-external-component-migration/SKILL.md
```diff
@@ -587,7 +587,7 @@ Bpk[ComponentName].tsx             | 100 | 100 | 100 | 100 |
 
 **Solution:** For new components, this is expected on first run:
 ```bash
-npm run jest -- packages/backpack-web/src/bpk-component-[name] -u
+ppnpm run jest -- packages/backpack-web/src/bpk-component-[name] -u
```
kerrie-wu: There's a typo with a double "p".

.github/workflows/_build.yml
kerrie-wu, line 132: `pnpm_config_lockfile=false` is not a format pnpm recognises — only `npm_config_lockfile=false` (npm-compat) or `PNPM_CONFIG_LOCKFILE=false` (pnpm uppercase prefix) work. This prefix will be silently ignored, so `pnpm add` will still write the lockfile.

Suggest:
```yaml
run: PNPM_CONFIG_LOCKFILE=false pnpm add -w react@19.2.5 ...
```

package.json
```diff
@@ -199,7 +201,10 @@
     "webpack": "^5.107.2",
     "wrapper-webpack-plugin": "^2.2.2"
   },
-  "overrides": {
-    "braces": "3.0.3"
+  "pnpm": {
+    "overrides": {
+      "braces": "3.0.3"
+    },
+    "neverBuiltDependencies": ["node-sass"]
```
kerrie-wu, line 208: `engines.pnpm` allows `>=9.15.9`, but `package.json#pnpm.overrides` and `neverBuiltDependencies` are only read by pnpm 9 — pnpm 10+ moved these settings to `pnpm-workspace.yaml` and ignores the `package.json#pnpm` field with a warning.

This means on pnpm 10/11:
- `overrides.braces: 3.0.3` (security patch) is silently dropped
- `neverBuiltDependencies: ["node-sass"]` is ignored, triggering native compilation and likely breaking install

Suggest moving these to `pnpm-workspace.yaml`:
```yaml
overrides:
  braces: "3.0.3"
neverBuiltDependencies:
  - node-sass
```
Or constrain engines to `pnpm 9.x` if pnpm 10+ is not intended to be supported.

.github/workflows/_build.yml
viktor-yang: Thanks, Kerrie. I checked the [official document](https://pnpm.io/configuring#environment-variables):
> Environment variables whose names start with pnpm_config_ (or PNPM_CONFIG_) are loaded into configuration. These override settings from pnpm-workspace.yaml but not CLI arguments.

package.json
viktor-yang: Thanks, Kerrie. The `packageManager` is set to `pnpm@9.15.9` now:
https://github.com/Skyscanner/backpack/pull/4738/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519.

And `pnpm-workspace.yaml` is supported by pnpm 11: https://pnpm.io/package_json.
So I think we can move the configs to the `pnpm-workspace.yaml` when we migrate to the pnpm 11

kerrie-wu (approved):
LGTM

PR #4741, opened 2026-06-26
[BpkModal] Fix focus trap broken by React 19 ref injection

IrinaWei:
## Summary

`TransitionInitialMount` injected a `nodeRef` into its child via `cloneElement(children, { ref: nodeRef })`, which **overwrote** the child's existing ref.

`BpkModalInner` / `BpkDialogInner` render `<section ref={dialogRef}>`, where `dialogRef` comes from the `withScrim` HOC and is used to scope focus. With the ref clobbered, `dialogRef` never fired → `withScrim.dialogElement` stayed `undefined` → `focusScope.scopeFocus()` was never called → no focus trap → **Tab escapes the modal** (the failing E2E for the mobile datepicker in global-components, backpack#4701).

This was a regression from the React 19 migration, which switched `TransitionInitialMount` to the `nodeRef` + `cloneElement` pattern (react-transition-group@4 removed the `findDOMNode` fallback).

## Fix

Merge `nodeRef` with the child's own ref instead of overwriting it:

- A memoised callback ref (`useCallback`, keyed on the child ref) keeps a **stable identity** across renders, so the child's callback ref isn't churned every render.
- The child ref is read from `props.ref` first (React 19) then `element.ref` (React 18), so it keeps working across the upcoming React 19 upgrade — no silent re-break.

## Scope

| Consumer | Child of `TransitionInitialMount` | Affected? |
|---|---|---|
| `BpkModalInner` | `<section ref={dialogRef}>` | ✅ fixed |
| `BpkDialogInner` | `<section ref={dialogRef}>` | ✅ fixed |
| `BpkScrim` | `<div>` (no ref) | unaffected |
| `BpkTooltip` / `BpkPopover` | `<section>` (uses floating-ui `FloatingFocusManager`, no `dialogRef`) | unaffected |

## Testing

image

- New regression test in `TransitionInitialMount-test.tsx` asserting both callback-ref and object-ref children still receive the DOM node.
- All `TransitionInitialMount` / `withScrim` / `BpkModal` / `BpkDialog` suites pass (117 tests).
- `npm run lint` + `tsc --noEmit` clean.

---

- [x] Tests
- [x] Accessibility tests
        - [x] Ability to navigate using a [keyboard only](https://webaim.org/techniques/keyboard/) — Tab is now trapped inside the modal again
- [ ] Storybook examples created/updated — n/a, no API change

### DIFF
--- a/packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount-test.tsx
+++ b/packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount-test.tsx
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+import { createRef } from 'react';
+
 import { render } from '@testing-library/react';
 
 import TransitionInitialMount from './TransitionInitialMount';
@@ -33,5 +35,35 @@ describe('TransitionInitialMount', () => {
     );
 
     expect(asFragment()).toMatchSnapshot();
+  });
+
+  it("should preserve the child's own ref when injecting nodeRef", () => {
+    // Regression: cloneElement used to overwrite the child ref with nodeRef,
+    // which broke focus trapping in BpkModal/BpkDialog (the dialogRef from
+    // withScrim never fired). See backpack#4701.
+    const callbackRef = jest.fn();
+    const objectRef = createRef<HTMLElement>();
+
+    render(
+      <TransitionInitialMount
+        appearClassName="block--appear"
+        appearActiveClassName="block--appear-active"
+        transitionTimeout={250}
+      >
+        <section ref={callbackRef}>callback ref child</section>
+      </TransitionInitialMount>,
+    );
+    expect(callbackRef).toHaveBeenCalledWith(expect.any(HTMLElement));
+
+    render(
+      <TransitionInitialMount
+        appearClassName="block--appear"
+        appearActiveClassName="block--appear-active"
+        transitionTimeout={250}
+      >
+        <section ref={objectRef}>object ref child</section>
+      </TransitionInitialMount>,
+    );
+    expect(objectRef.current).toBeInstanceOf(HTMLElement);
   });
 });
--- a/packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount-test.tsx
+++ b/packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount-test.tsx
@@ -17,12 +17,35 @@
  */
 
 import { createRef } from 'react';
+import type { ReactNode, RefObject } from 'react';
 
 import { render } from '@testing-library/react';
 
 import TransitionInitialMount from './TransitionInitialMount';
 
+// Capture the nodeRef CSSTransition is given so we can assert it points at the
+// child node. The real CSSTransition drives its appear animation through a
+// DOM/raf pipeline jsdom does not run, so we render the children directly.
+let capturedNodeRef: RefObject<HTMLElement | null> | null = null;
+jest.mock('react-transition-group/CSSTransition', () => ({
+  __esModule: true,
+  default: ({
+    children,
+    nodeRef,
+  }: {
+    children: ReactNode;
+    nodeRef: RefObject<HTMLElement | null>;
+  }) => {
+    capturedNodeRef = nodeRef;
+    return children;
+  },
+}));
+
 describe('TransitionInitialMount', () => {
+  beforeEach(() => {
+    capturedNodeRef = null;
+  });
+
   it('should render correctly', () => {
     const { asFragment } = render(
       <TransitionInitialMount
@@ -65,5 +88,26 @@ describe('TransitionInitialMount', () => {
       </TransitionInitialMount>,
     );
     expect(objectRef.current).toBeInstanceOf(HTMLElement);
+  });
+
+  it('should populate the nodeRef that CSSTransition uses to drive the animation', () => {
+    // Covers the nodeRef side of the merged ref. CSSTransition reads nodeRef
+    // to drive the appear animation; that animation relies on a real DOM/raf
+    // pipeline jsdom cannot run, so we instead assert CSSTransition is handed a
+    // nodeRef pointing at the child node. Deleting `nodeRef.current = node`
+    // would null this and fail the test, where a snapshot test would not.
+    const { container } = render(
+      <TransitionInitialMount
+        appearClassName="my-appear"
+        appearActiveClassName="my-appear-active"
+        transitionTimeout={250}
+      >
+        <section>child</section>
+      </TransitionInitialMount>,
+    );
+
+    const section = container.querySelector('section');
+    expect(capturedNodeRef).not.toBeNull();
+    expect(capturedNodeRef?.current).toBe(section);
   });
 });
--- a/packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
+++ b/packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-import { cloneElement, useCallback, useRef } from 'react';
+import { cloneElement, useCallback, useRef, version } from 'react';
 import type { MutableRefObject, ReactElement, Ref } from 'react';
 
 // @ts-expect-error Untyped import. See `decisions/imports-ts-suppressions.md`.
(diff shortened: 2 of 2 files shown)


packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
```diff
@@ -47,6 +60,28 @@ const TransitionInitialMount = ({
...
 }: Props) => {
   const nodeRef = useRef<HTMLElement>(null);
+  // Read the child's own ref so injecting nodeRef does not clobber it.
+  // React 19 exposes ref as a normal prop (children.props.ref); React 18 keeps
+  // it on the element itself (children.ref). Check props.ref FIRST so that on
+  // React 19 we never touch element.ref (which logs a deprecation warning).
+  const childProps = children.props as { ref?: Ref<HTMLElement> };
+  const childRef =
+    childProps.ref ??
+    (children as ReactElement & { ref?: Ref<HTMLElement> }).ref;
```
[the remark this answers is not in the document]
IrinaWei: Good catch — fixed in 08dcffb. Switched from the `??` fallback to a React major-version gate (`parseInt(version, 10) >= 19`), so on React 19 we read only `children.props.ref` and never touch `element.ref`, even when the child has no ref of its own.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
```diff
@@ -35,7 +35,20 @@ type Props = {
...
+  if (!ref) {
+    return;
+  }
+  if (typeof ref === 'function') {
+    ref(node);
+  } else {
+    // eslint-disable-next-line no-param-reassign
+    (ref as MutableRefObject<HTMLElement | null>).current = node;
+  }
 };
```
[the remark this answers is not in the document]
IrinaWei: Partially applied in 08dcffb. nodeRef is now assigned directly (`nodeRef.current = node`) since it's an object ref we own, and the helper (renamed `assignChildRef`) only handles the child ref. The remaining `no-param-reassign` disable is unavoidable there — mutating `.current` on a caller-provided object ref is the documented contract for object refs, same as React's own `mergeRefs` implementations.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
```diff
@@ -35,7 +35,20 @@ type Props = {
   // findDOMNode fallback in react-transition-group@4, so we inject a nodeRef
   // into the child via cloneElement. A plain string/fragment cannot accept a
   // ref and would crash at runtime, so the type is narrowed accordingly.
-  children: ReactElement<{ ref?: RefObject<HTMLElement | null> }>;
+  // `Ref` (not `RefObject`) so the merged callback ref we inject is assignable.
+  children: ReactElement<{ ref?: Ref<HTMLElement> }>;
```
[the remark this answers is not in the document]
IrinaWei: Updated the comment in 08dcffb to spell out that the child must be a DOM element, class component, or forwardRef component, and that a function component which drops the ref leaves CSSTransition with no node to animate.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount-test.tsx
```diff
@@ -34,4 +36,34 @@ describe('TransitionInitialMount', () => {
...
+      </TransitionInitialMount>,
+    );
+    expect(callbackRef).toHaveBeenCalledWith(expect.any(HTMLElement));
+
+    render(
+      <TransitionInitialMount
+        appearClassName="block--appear"
+        appearActiveClassName="block--appear-active"
+        transitionTimeout={250}
+      >
```
kerrie-wu, line 86: The new test covers the **child-ref side** (both callback ref and object ref) — good. But the fix has two critical paths:

```tsx
const mergedRef = useCallback((node: HTMLElement | null) => {
  nodeRef.current = node;        // path A: CSSTransition reads this to animate
  assignChildRef(childRef, node); // path B: dialogRef / focus trap ← tested ✓
}, [childRef]);
```

Path A (`nodeRef.current`) is untested: if that line were deleted, CSSTransition would receive a perpetually-null ref and the appear animation would silently stop working — but the suite would still be green.

Suggest adding one assertion that covers path A, e.g.:

```tsx
const { container } = render(
  <TransitionInitialMount appearClassName="my-appear" appearActiveClassName="my-appear-active" transitionTimeout={250}>
    <section>child</section>
  </TransitionInitialMount>,
);
expect(container.querySelector('section')).toHaveClass('my-appear');
```

This only passes if `nodeRef.current` is correctly set.

kerrie-wu (commented):
The new test covers the **child-ref side** (both callback ref and object ref) — good. But the fix has two critical paths:

```tsx
const mergedRef = useCallback((node: HTMLElement | null) => {
  nodeRef.current = node;        // path A: CSSTransition reads this to animate
  assignChildRef(childRef, node); // path B: dialogRef / focus trap ← tested ✓
}, [childRef]);
```

Path A (`nodeRef.current`) is untested: if that line were deleted, CSSTransition would receive a perpetually-null ref and the appear animation would silently stop working — but the suite would still be green.

Suggest adding one assertion that covers path A, e.g.:

```tsx
const { container } = render(
  <TransitionInitialMount appearClassName="my-appear" appearActiveClassName="my-appear-active" transitionTimeout={250}>
    <section>child</section>
  </TransitionInitialMount>,
);
expect(container.querySelector('section')).toHaveClass('my-appear');
```

This only passes if `nodeRef.current` is correctly set.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
```diff
@@ -31,11 +31,32 @@ type Props = {
...
+  // receive nodeRef, so CSSTransition would have no node to animate.
+  // `Ref` (not `RefObject`) so the merged callback ref we inject is assignable.
+  children: ReactElement<{ ref?: Ref<HTMLElement> }>;
+};
+
+// Assigns a node to the child's own ref, supporting both callback refs and
+// object refs (e.g. createRef). nodeRef is handled separately below since it is
+// always an object ref we own.
+const assignChildRef = (
+  childRef: Ref<HTMLElement> | undefined,
```
kerrie-wu, line 48: Nit: `BpkDrawerContent` has an [identical inline implementation](https://github.com/Skyscanner/backpack/blob/main/packages/backpack-web/src/bpk-component-drawer/src/BpkDrawerContent.tsx#L101-L112) of this function/ref-type branching logic. Non-blocking for this PR, but worth extracting to `bpk-react-utils` as a shared utility so there is only one place to maintain.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount-test.tsx
IrinaWei: Great catch — added in fd75a1f. I couldn't use the `toHaveClass('my-appear')` assertion directly: react-transition-group drives the appear class through a real DOM/raf pipeline that jsdom doesn't run, so the class never lands in the test environment (the existing snapshot test shows the same — no appear class in the static output). Instead I mock CSSTransition and assert it's handed a nodeRef pointing at the child node. Verified it's a real mutation guard: deleting `nodeRef.current = node` makes the new test fail.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
IrinaWei: Agreed it's duplicated with BpkDrawerContent — both hand-roll the same nodeRef + consumer-ref merge. Since you flagged it non-blocking I'll keep this PR scoped to the focus-trap fix and raise a follow-up to extract a shared `mergeRefs`/`useMergedRef` util in bpk-react-utils and migrate both call sites. I'll link the ticket here once raised.

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount-test.tsx
IrinaWei: Add new test for it

packages/backpack-web/src/bpk-react-utils/src/TransitionInitialMount.tsx
IrinaWei: Maybe we can create a follow up ticket for it

kerrie-wu (approved):
LGTM

PR #4757, opened 2026-06-26
[bpk-component-infinite-scroll] Migrate to TypeScript

### LINKED ISSUE #4744: [bpk-component-infinite-scroll] Migrate to TypeScript
## Context

`bpk-component-infinite-scroll` is a pure-JS package with 4 source files. This package has a class-based HOC (`withInfiniteScroll`) that is also affected by the React 19 `defaultProps` migration.

## Files to migrate (4)

| File | Type |
|------|------|
| `src/withInfiniteScroll.js` | Class-based HOC — uses `prop-types` and `defaultProps` |
| `src/DataSource.js` | Data source utility |
| `src/intersection-observer.js` | Intersection observer wrapper |
| `src/withInfiniteScroll-test.flow.js` | Flow type test file — remove or convert |

## Migration guidance

1. **Rename** all `.js` → `.tsx` (components) or `.ts` (utilities).
2. **Replace `prop-types` imports** with TypeScript interfaces/types.
3. **Migrate `withInfiniteScroll` class component** — convert `static defaultProps` to destructure defaults or convert to a functional component with hooks. This is also tracked in the React 19 migration (`REACT_19_MIGRATION.md`).
4. **Remove `withInfiniteScroll-test.flow.js`** — Flow is no longer used in Backpack; convert any useful type assertions to TS.
5. **Update test files** and Storybook stories to `.tsx`.
6. **Update `index.js`** → `index.ts` and package entry points.
7. **Run** `npm run lint`, `npm run typecheck`, and `npm test`.

## ⚠️ React 19 interaction

The `withInfiniteScroll` HOC has `defaultProps` on a class component. Under React 19, class `defaultProps` still work but are deprecated for function components. A recent fix (#4689) corrected a bug where `initiallyLoadedElements` was used instead of `elementsPerScroll` — ensure the TS migration preserves this fix.

## References

- React 19 migration: `REACT_19_MIGRATION.md`
- Example completed migration: `BpkPrice` (#4521)

gert-janvercauteren:
## Summary
- Migrates `bpk-component-infinite-scroll` from JavaScript to TypeScript
- Replaces `prop-types` with TypeScript interfaces
- Converts `defaultProps` to ES6 destructure defaults
- Removes Flow type annotations
- Renames `.js` source/test/story files to `.tsx`/`.ts`

Closes #4744


packages/backpack-web/src/bpk-component-infinite-scroll/src/DataSource.ts
```diff
@@ -16,29 +16,27 @@
...
-  listeners: Array<() => mixed>;
+  listeners: Array<(...args: any[]) => void>;
 
   constructor() {
     this.listeners = [];
   }
 
-  /* eslint-disable-next-line no-unused-vars */
-  fetchItems(index: number, nElements: number): Promise<Array<T>> {
+
```
kerrie-wu: empty line should be deleted

packages/backpack-web/src/bpk-component-infinite-scroll/src/withInfiniteScroll.tsx
```diff
@@ -16,107 +16,105 @@
...
-type PropsWithDefault = Config<Props, typeof defaultProps>;
-
-const withInfiniteScroll = <T: ExtendedProps>(
-  ComponentToExtend: AbstractComponent<T>,
-): AbstractComponent<PropsWithDefault & $Diff<T, ExtendedProps>> =>
-  class WithInfiniteScroll extends Component<Props, State> {
-
-    handleKeyPress: (e: SyntheticKeyboardEvent<HTMLButtonElement>) => void;
-
+const OWN_PROP_KEYS = [
```
kerrie-wu: OWN_PROP_KEYS is inferred as a bare string[], so it's completely decoupled from the Props interface. If a new prop is added to Props but not mirrored here, omit won't strip it and it'll silently leak into ComponentToExtend via {...(rest as any)} — with no compiler warning. Annotating it with Array<keyof Props> at least catches typos and any renamed props at compile time:


```suggestion
const OWN_PROP_KEYS: Array<keyof Props> = [
```

packages/backpack-web/src/bpk-component-infinite-scroll/src/withInfiniteScroll.tsx
```diff
@@ -140,12 +138,14 @@ const withInfiniteScroll = <T: ExtendedProps>(
       this.fetchItems({
         elementsPerScroll:
           this.props.initiallyLoadedElements ?? defaultProps.initiallyLoadedElements,
-      }).then((newState) => {
-        this.setState(newState);
-      });
+      })
+        .then((newState) => {
+          this.setState(newState as State);
```
kerrie-wu: `React.Component.setState` already accepts `Partial<S>`, so `newState` (typed `Partial<State>`) can be passed directly — the `as State` cast is unnecessary here. Worse, it suppresses a check that could be useful: if `fetchItems` were ever changed to return an object genuinely missing a required field, `as State` would silence the error. Suggest dropping the cast:

```suggestion
          this.setState(newState);
```

packages/backpack-web/src/bpk-component-infinite-scroll/src/BpkInfiniteScroll.stories.tsx
```diff
@@ -333,9 +277,8 @@ const InferDatasourceWhenLessThanRequestElementsExample = () => {
   );
 };
 
-const meta = {
+const meta: Meta = {
   title: 'bpk-component-infinite-scroll',
-  component: WithInfiniteScrollMock,
```
kerrie-wu, line 338: Removing `WithInfiniteScrollMock` also dropped the `component:` field from `meta`, so the Docs page ArgTable will now be empty — a docs regression. The old mock's own comment noted it was only needed because HOC prop tables couldn't be generated in JS, and that it *"does work in TS, so we can remove this once we migrate to TS"*. So this is the right time to point `meta` at the real component instead:

```suggestion
const meta: Meta<typeof InfiniteList> = {
  title: 'bpk-component-infinite-scroll',
  component: InfiniteList,
  decorators: [withScrollReset],
};
```

packages/backpack-web/src/bpk-component-infinite-scroll/src/withInfiniteScroll.tsx
```diff
@@ -231,14 +234,14 @@ const withInfiniteScroll = <T: ExtendedProps>(
           }
           if (onScrollFinished && result.isListFinished) {
             onScrollFinished({
-              totalNumberElements: elementsToRender.length,
+              totalNumberElements: (result.elementsToRender || elementsToRender || []).length,
```
kerrie-wu, line 240: This changes observable behaviour: when the final batch is non-empty but smaller than `elementsPerScroll`, `onScrollFinished` now reports the total *including* that batch (e.g. 7 items / `elementsPerScroll=5` → fires with `7` instead of the old `5`). The new value is correct, but it slips in as a migration side effect and tests only assert `toHaveBeenCalled()`. Please call it out in the PR description and add a value assertion to lock it in, e.g. `expect(onScrollFinished).toHaveBeenCalledWith({ totalNumberElements: 7 })`.

xiaogliu: Thanks for flagging this. I checked the direct `BpkInfiniteScroll` consumers listed on the Backpack versions page, including:

- `Skyscanner/zensearch`
- `Skyscanner/partner-portal`
- `Skyscanner/web-platform`
- `Skyscanner/packages-website`
- `Skyscanner/hotels-website`
- `Skyscanner/carhire-homepage`
- `Skyscanner/devex-web`

My conclusion is:

- `onScrollFinished` is a behavior change, but I did not find any of the checked consumers relying on that callback in a way that would make this a breaking change.
- `.catch(console.error)` does change error propagation, but the checked consumers already handle fetch failures in their own data source or request layers, usually by logging the error and showing local error UI. I did not find a consumer that depends on `BpkInfiniteScroll` promise rejections as its primary monitoring signal.
- `renderSeeMoreComponent` is a type correction rather than a runtime break. The checked consumers already use `({ onSeeMoreClick }: { onSeeMoreClick: () => void }) => ...`, so this does not appear to break current usage.

So I think this PR introduces a few observable behavior/type changes, but I do not see evidence that it causes a breaking change for the repos currently consuming `BpkInfiniteScroll`. And it safe to mark as minor change.

packages/backpack-web/src/bpk-component-infinite-scroll/src/withInfiniteScroll.tsx
xiaogliu: updated in https://github.com/Skyscanner/backpack/pull/4757/changes/6f9bdc1482107f9e2ddc9c9f7ae353b563fd7718

xiaogliu:
This PR does change some component behavior, but after checking the current BpkInfiniteScroll consumers, I don't see a path that would affect real users in a breaking way. The changes are mostly a behavior correction and type tightening, so I think it's reasonable to ship this as a minor release.

Details:

> I dug into the current `BpkInfiniteScroll` consumers listed on the Backpack versions page, including `zensearch`, `partner-portal`, `web-platform`, `packages-website`, `hotels-website`, `carhire-homepage`, and `devex-web`.
>
> My takeaway is that this PR introduces a couple of observable behavior/type changes, but I did not find evidence of a breaking runtime change for the existing consumers:
> - `onScrollFinished` is now aligned with the actual rendered count, which is a behavior fix rather than a consumer-facing break.
> - `.catch(console.error)` changes how failed fetches are surfaced, but the consumers I checked already handle fetch errors in their own data-source/request layers, so they are not relying on `BpkInfiniteScroll` promise rejections as their main error signal.
> - The `renderSeeMoreComponent` type is narrower, but the current consumers already use the `({ onSeeMoreClick }: { onSeeMoreClick: () => void }) => ...` shape.
>
> So I think it is fair to call out these as behavior/type changes, but I don't see evidence that they are breaking changes for the repos currently consuming this component.

packages/backpack-web/src/bpk-component-infinite-scroll/src/BpkInfiniteScroll.stories.tsx
kerrie-wu: It's not showing up in the [Product Storybook](https://backpack.github.io/storybook/?path=/docs/bpk-component-infinite-scroll--documentation) either.

kerrie-wu: I checked production and the ArgTable is already missing there too, so this PR doesn't appear to introduce a docs regression here. Since the current published docs behave the same way, I don't think this should block the TS migration.

kerrie-wu (approved):
LGTM

PR #4761, opened 2026-06-26
[bpk-component-flare] Migrate to TypeScript

### LINKED ISSUE #4748: [bpk-component-flare] Migrate to TypeScript
## Context

`bpk-component-flare` is a pure-JS package with 2 source files using `prop-types`.

## Files to migrate (2)

| File | Type |
|------|------|
| `src/BpkContentBubble.js` | Content bubble component — uses `prop-types` (including `flareProps: PropTypes.object`) |
| `src/BpkFlareBar.js` | Flare bar component |

## Migration guidance

1. **Rename** both `.js` → `.tsx`.
2. **Replace `prop-types` imports** with TypeScript interfaces. Replace `flareProps: PropTypes.object` with a properly typed interface (e.g. `flareProps?: React.HTMLAttributes<HTMLDivElement>`).
3. **Migrate `defaultProps`** to ES6 destructure defaults.
4. **Update test files** (`.test.js` → `.test.tsx`) — replace snapshot tests with explicit `expect` assertions where possible.
5. **Update Storybook stories** (`.stories.js` → `.stories.tsx`).
6. **Update `index.js`** → `index.ts` and package entry points.
7. **Run** `npm run lint`, `npm run typecheck`, and `npm test`.

## References

- Example completed migration: `BpkPrice` (#4521), `BpkStarRating` (#4523)

gert-janvercauteren:
## Summary
- Migrates `bpk-component-flare` from JavaScript to TypeScript
- Replaces `prop-types` with TypeScript interfaces
- Converts `defaultProps` to ES6 destructure defaults
- Removes Flow type annotations
- Renames `.js` source/test/story files to `.tsx`/`.ts`

Closes #4748


packages/backpack-web/src/bpk-component-flare/src/BpkFlareBar.tsx
```diff
@@ -16,21 +16,30 @@
...
 import STYLES from './bpk-flare-bar.module.scss';
 
 const getClassName = cssModules(STYLES);
 
+type NativeDivProps = HTMLAttributes<HTMLDivElement>;
+
+export type Props = Omit<NativeDivProps, 'className'> & {
+  className?: string | null;
+  svgClassName?: string | null;
+  rounded?: boolean;
```
kerrie-wu: BpkFlareBar doesn't actually use rounded — it's not destructured, so it just gets spread onto the <div> (L55). Passing rounded={true} would render an invalid rounded="true" DOM attribute.

It was never in the old propTypes either, so it's always been a no-op. Suggest removing it here, plus the "rounded set to false" test in BpkFlareBar-test.tsx (it passes only because the prop does nothing).

Same dead rounded also appears on flareProps in BpkContentBubble.tsx — worth removing together.

packages/backpack-web/src/bpk-component-flare/src/BpkContentBubble.tsx
```diff
@@ -16,26 +16,38 @@
...
-  }) => {
+type NativeDivProps = HTMLAttributes<HTMLDivElement>;
+
+export type Props = Omit<NativeDivProps, 'className' | 'content'> & {
+  content?: ReactNode;
+  rounded?: boolean;
+  showPointer?: boolean;
+  className?: string | null;
+  contentClassName?: string | null;
+  flareProps?: (HTMLAttributes<HTMLDivElement> & { svgClassName?: string; rounded?: boolean }) | null;
```
kerrie-wu: flareProps gets spread onto BpkFlareBar (L108), but BpkFlareBar doesn't use rounded — so this rounded?: boolean is the same dead prop as the one on BpkFlareBar's type. If we drop it there, remove it here too, otherwise the type still advertises a prop that does nothing.

packages/backpack-web/src/bpk-component-flare/src/BpkFlareBar.tsx
xiaogliu: Good catch — removed `rounded` from `BpkFlareBar`'s `Props` and dropped the "rounded set to false" test in `BpkFlareBar-test.tsx` (plus its snapshot). Pushed in abb7241.

packages/backpack-web/src/bpk-component-flare/src/BpkContentBubble.tsx
xiaogliu: Removed `rounded` from the `flareProps` type too, and updated the two `BpkContentBubble-test.tsx` cases that were passing `flareProps={{ rounded: false }}`. Pushed in abb7241.

packages/backpack-web/src/bpk-component-flare/src/BpkContentBubble.tsx
```diff
@@ -16,26 +16,38 @@
...
-  }) => {
+type NativeDivProps = HTMLAttributes<HTMLDivElement>;
+
+export type Props = Omit<NativeDivProps, 'className' | 'content'> & {
+  content?: ReactNode;
+  rounded?: boolean;
+  showPointer?: boolean;
+  className?: string | null;
+  contentClassName?: string | null;
+  flareProps?: (HTMLAttributes<HTMLDivElement> & { svgClassName?: string }) | null;
```
gc-skyscanner: `flareProps` is spread directly onto `<BpkFlareBar {...flareProps} />`, so this inline type is essentially a manual copy of `BpkFlareBar`'s props — and already out of sync (`svgClassName: string` here vs `string | null` there). Suggest reusing the exported type:

```
import BpkFlareBar, { type Props as BpkFlareBarProps } from './BpkFlareBar';

flareProps?: BpkFlareBarProps | null;
```

xiaogliu: Good point — switched `flareProps` to reuse `BpkFlareBar`'s exported `Props` type directly instead of the manual copy. Pushed in 5f55847.

gc-skyscanner (approved):
LGTM

PR #4768, opened 2026-06-26
[CLOV-1812] Improve layout component API parity

RichardSyq:
## Summary

Implements CLOV-1812 by improving Backpack layout component API parity across \`BpkBox\`, \`BpkFlex\`, and \`BpkVessel\`.

## Changes

- Adds a focused set of mouse, pointer, and touch event handler props to the shared layout prop surface: down/up for mouse and pointer events, and start/end for touch events.
- Adds \`BpkFlex\` support for \`alignSelf\`, \`justifySelf\`, \`gridColumn\`, \`gridRow\`, \`rowGap\`, and \`columnGap\`.
- Adds responsive processing for the new \`BpkFlex\` self-placement and grid-placement props.
- Clarifies that \`BpkFlex\` grid placement props only take effect when the flex element is placed inside a grid parent.
- Adds \`ref\` forwarding to \`BpkVessel\`.
- Adds Jest coverage plus Storybook and README examples for the newly supported props.
- Keeps \`VisualTest\` initial snapshot text stable so that the new multi-counter interactive text does not cause Percy noise (shown only in the interactive \`Interactive\` story, not in the visual regression story).

## Validation

- \`TZ=Etc/UTC ./node_modules/.bin/jest packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx packages/backpack-web/src/bpk-component-layout/src/BpkFlex-test.tsx packages/backpack-web/src/bpk-component-layout/src/BpkVessel-test.tsx packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts --runInBand\`
- \`./node_modules/.bin/eslint packages/backpack-web/src/bpk-component-layout --ext .ts,.tsx\`
- \`npm run check-bpk-dependencies\`
- \`npm run check-react-versions\`
- \`git diff --check\`
- \`lint-staged\` / \`eslint --fix\` via commit hook

Note: full \`tsc --noEmit\` is currently blocked by existing missing generated icon/spinner modules outside this change; filtered output showed no \`bpk-component-layout\` type errors.

## Percy notes

Any \`bpk-component-price-range\` visual diff in Percy is **not caused by this PR**. It originates from \`05325deca\` (\`[NO-JIRA][Storybook] Add dark/light mode toolbar toggle\`) which was already on \`main\` before this branch was created and changes the Storybook canvas background for all stories. The \`bpk-component-layout\` stories in this PR do not affect \`bpk-component-price-range\` in any way.


packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
```diff
@@ -124,26 +124,42 @@ const PositionExample = () => (
...
         role="button"
         tabIndex={0}
-        onClick={increment}
+        onClick={() => increment('click')}
         onKeyDown={(e) => {
-          if (e.key === 'Enter' || e.key === ' ') increment();
+          if (e.key === 'Enter' || e.key === ' ') {
+            increment('click');
+          }
         }}
```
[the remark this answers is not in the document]
RichardSyq: Updated in `5d03747ef`: Enter still activates on keydown, while Space now calls `preventDefault()` on keydown and activates on keyup, which mirrors native button behavior more closely and avoids page scrolling in the Storybook example.

packages/backpack-web/src/bpk-component-layout/src/BpkVessel.tsx
```diff
@@ -63,14 +65,25 @@ import type { BpkVesselProps } from './types';
...
+  ) => {
+    const Component = Element as ElementType;
+
+    return (
+      <Component ref={ref} {...getDataComponentAttribute('Vessel')} {...restProps}>
+        {children}
+      </Component>
+    );
+  },
 );
```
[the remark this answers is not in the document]
RichardSyq: I am leaving this as `HTMLElement` for now. `BpkVessel` is deliberately a migration hatch with a constrained `as` union (`div`, `span`, `section`, etc.), and `HTMLElement` safely covers that current surface without adding a larger polymorphic ref typing pattern in this API-parity change. I agree an `as`-derived generic ref type could be a follow-up if we want stricter inference later; defaulting to `HTMLDivElement` would be less accurate for `as="section"` / `as="span"`.

packages/backpack-web/src/bpk-component-layout/src/BpkFlex-test.tsx
```diff
@@ -61,6 +61,43 @@ describe('BpkFlex', () => {
...
+        >
+          Content
+        </BpkFlex>
+      </BpkProvider>,
+    );
+
+    expect(container.firstChild).toHaveStyle('align-self: flex-end');
+    expect(container.firstChild).toHaveStyle('justify-self: center');
+    expect(container.firstChild).toHaveStyle('grid-column: 2/span 2');
+    expect(container.firstChild).toHaveStyle('grid-row: 1');
```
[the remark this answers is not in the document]
RichardSyq: Updated in `5d03747ef`: the test now reads `gridColumn` from computed style and normalizes whitespace before asserting, so equivalent serialization like `2 / span 2` vs `2/span 2` does not make the test brittle.

packages/backpack-web/src/bpk-component-layout/src/types.ts
```diff
@@ -199,6 +199,12 @@ export interface BpkFlexSpecificProps {
   grow?: BpkResponsiveValue<FlexProps['flexGrow']>;
   shrink?: BpkResponsiveValue<FlexProps['flexShrink']>;
   basis?: BpkResponsiveValue<BpkBasisValue>;
+  alignSelf?: BpkResponsiveValue<BoxProps['alignSelf']>;
```
RichardSyq, line 197: BpkFlex does not support alignSelf or justifySelf, while BpkBox and BpkStack do.

BpkFlex does not support gridRow or gridColumn, while BpkBox does.

packages/backpack-web/src/bpk-component-layout/src/BpkVessel.tsx
```diff
@@ -63,14 +65,25 @@ import type { BpkVesselProps } from './types';
...
-    {children}
-  </Element>
+export const BpkVessel = forwardRef<HTMLElement, BpkVesselProps>(
+  (
+    {
+      as: Element = 'div',
+      children,
+      ...restProps
+    },
+    ref,
```
RichardSyq, line 75: BpkVessel does not support ref.

packages/backpack-web/src/bpk-component-layout/src/types.ts
RichardSyq: BpkFlex does not support independent rowGap and columnGap control

packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
```diff
@@ -124,26 +124,51 @@ const PositionExample = () => (
...
+            e.preventDefault();
+          }
         }}
+        onKeyUp={(e) => {
+          if (e.key === ' ') {
+            e.preventDefault();
+            increment('click');
+          }
+        }}
+        onMouseDown={() => increment('mouseDown')}
```
RichardSyq, line 168: BpkBox does not support pointer/touch event handlers such as onMouseDown and onTouchStart.

packages/backpack-web/src/bpk-component-layout/src/commonProps.ts
gc-skyscanner: `onFocus`/`onBlur` remain BpkBox-only while the new mouse/pointer/touch events are in `BpkCommonLayoutProps`
`onFocus` and `onBlur` are defined only on `BpkBoxProps` (types.ts:155), while the newly added onMouseDown/onPointerDown/etc. land in `BpkCommonLayoutProps` (commonProps.ts:108-113), making them available on every layout component. The original comment in commonProps.ts:90-93 says `“BpkBox additionally exposes onFocus and onBlur on its own props type”`, but that asymmetry is now harder to justify when lower-level mouse/touch events are surfaced on all components. Was this intentional?

packages/backpack-web/src/bpk-component-layout/src/BpkFlex-test.tsx
```diff
@@ -61,6 +61,45 @@ describe('BpkFlex', () => {
...
+        </BpkFlex>
+      </BpkProvider>,
+    );
+
+    const element = container.firstChild as HTMLElement;
+
+    expect(element).toHaveStyle('align-self: flex-end');
+    expect(element).toHaveStyle('justify-self: center');
+    expect(element).toHaveStyle('grid-row: 1');
+    expect(getComputedStyle(element).gridColumn.replace(/\s/g, '')).toBe('2/span2');
```
gc-skyscanner, line 100: just out of curiocity, why don't we just `toHaveStyle` here🤔

packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
```diff
@@ -147,6 +147,35 @@ describe('BpkBox', () => {
...
 
+  it('supports pointer, mouse, and touch interaction props', () => {
+    const handleMouseDown = jest.fn();
+    const handlePointerDown = jest.fn();
+    const handleTouchStart = jest.fn();
+
+    const { getByText } = render(
+      <BpkProvider>
+        <BpkBox
+          onMouseDown={handleMouseDown}
```
gc-skyscanner, line 161: could we add more action/assertion for the rest new adding supporting event?

packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
```diff
@@ -124,26 +124,48 @@ const PositionExample = () => (
...
+        onClick={() => increment('click')}
         onKeyDown={(e) => {
-          if (e.key === 'Enter' || e.key === ' ') increment();
+          if (e.key === 'Enter' || e.key === ' ') {
+            increment('click');
+          }
         }}
+        onMouseDown={() => increment('mouseDown')}
+        onPointerDown={() => increment('pointerDown')}
+        onTouchStart={() => increment('touchStart')}
```
gc-skyscanner, line 172: could we add more case for the rest new adding event?

packages/backpack-web/src/bpk-component-layout/src/types.ts
```diff
@@ -199,6 +199,12 @@ export interface BpkFlexSpecificProps {
...
+  justifySelf?: BpkResponsiveValue<BoxProps['justifySelf']>;
```
gc-skyscanner, line 198: I am thinking if we should add these props to `BpkCommonLayoutProps` since we support them in BpkBox, BpkFlex. also,  Is it possible for a Grid container to be a child of BpkFlex, and vice versa?

packages/backpack-web/src/bpk-component-layout/src/types.ts
```diff
@@ -199,6 +199,12 @@ export interface BpkFlexSpecificProps {
...
+  gridColumn?: BpkResponsiveValue<BoxProps['gridColumn']>;
+  gridRow?: BpkResponsiveValue<BoxProps['gridRow']>;
+  rowGap?: BpkResponsiveValue<BpkSpacingValue>;
+  columnGap?: BpkResponsiveValue<BpkSpacingValue>;
```
gc-skyscanner, line 202: do we need to add `rowGap` and `columnGap`  to `BPK_RESPONSIVE_PROP_GROUPS_BY_COMPONENT` ?

packages/backpack-web/src/bpk-component-layout/src/commonProps.ts
RichardSyq: Good catch — this asymmetry is now resolved. In `b85cf019d` I promoted `onFocus`/`onBlur` from `BpkBoxProps` into `BpkCommonLayoutProps`, so they sit alongside the mouse/pointer/touch handlers and are exposed on all layout components (Box, Flex, Grid, Stack, …). This matches the documented intent that layout components support interactive container patterns. Added `onFocus`/`onBlur` coverage to the BpkFlex and BpkStack tests and updated the README to describe the shared interaction surface.

packages/backpack-web/src/bpk-component-layout/src/types.ts
RichardSyq: Good question. I looked into promoting `alignSelf` / `justifySelf` / `gridColumn` / `gridRow` into `BpkCommonLayoutProps`, but decided to hold off for now:

1. **Type surface and runtime are decoupled.** Adding them to `BpkCommonLayoutProps` only widens the type — the responsive object syntax (e.g. `gridColumn={{ base: 1, tablet: 2 }}`) is only processed if the key is also listed in that component's `item` group in `BPK_RESPONSIVE_PROP_GROUPS_BY_COMPONENT`. So a pure type move would give a partially-working API.
2. **Semantics differ per component.** `justifySelf` / `gridColumn` / `gridRow` only take effect when the parent is a grid (see the README note), and `BpkStack` isn't meant to be positioned as a grid/flex item — so these would surface as no-op props on Stack.

A cleaner de-dup would be a shared `BpkItemPlacementProps` mixin that only Box/Flex/Grid extend, but that means reworking the `BpkBoxSpecificProps` `Omit<…>` composition, touching the public type of every layout component just to remove 4 duplicated lines. I'd rather keep the explicit per-component declarations (precise, well-isolated) and leave the shared mixin as a possible follow-up if we add more item-placement props later.

On nesting: yes — a Grid can be a child of a Flex and vice versa, which is exactly why `gridColumn`/`gridRow` are useful on `BpkFlex` (a flex container placed as a grid item). The per-component allowlists already reflect this.

packages/backpack-web/src/bpk-component-layout/src/types.ts
RichardSyq: No — `rowGap`/`columnGap` are spacing props, so they go through the `spacingKeys` path in `tokenUtils.ts` (token→value conversion, including breakpoint handling), not the `BPK_RESPONSIVE_PROP_GROUPS_BY_COMPONENT` allowlist, which is for non-spacing enum/string layout props. Listing them there would be redundant and risk double-processing. The existing `gap` prop is handled the same way.

packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
RichardSyq: Done in `625e0663a` — the `supports pointer, mouse, and touch interaction props` test now fires and asserts `onMouseDown`/`onMouseUp`, `onPointerDown`/`onPointerUp` and `onTouchStart`/`onTouchEnd`, each expected to be called once.

packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
RichardSyq: Done in `625e0663a` — the interaction story now wires up all six new handlers and displays a live per-event counter (click / mouseDown / mouseUp / pointerDown / pointerUp / touchStart / touchEnd), so each new event is demonstrated in Storybook.

packages/backpack-web/src/bpk-component-layout/src/BpkFlex-test.tsx
RichardSyq: The other three assertions here do use `toHaveStyle`. `gridColumn` is the exception because it's a shorthand whose serialization is whitespace-sensitive (`2 / span 2` vs `2/span 2`), which makes `toHaveStyle` brittle across CSS-in-JS output. Reading the computed value and stripping whitespace keeps the assertion stable regardless of formatting. (This was also flagged by Copilot on the same line.)

packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
```diff
@@ -124,26 +124,54 @@ const PositionExample = () => (
...
+  };
+
   return (
     <LayoutWrapper>
       <BpkBox
         padding={BpkSpacing.MD}
         role="button"
         tabIndex={0}
-        onClick={increment}
+        onClick={() => increment('click')}
```
gc-skyscanner, line 160: Nit: could you help add `onBlur ` `onFocus` to storybook?

RichardSyq: Done in `e541ffbf1` — `onFocus`/`onBlur` are now wired into the interactive story's live event counter, alongside the click/mouse/pointer/touch handlers. The `VisualTest` story keeps `showEventCounts={false}`, so the Percy snapshot text is unaffected.

packages/backpack-web/src/bpk-component-layout/src/BpkVessel-test.tsx
```diff
@@ -63,6 +65,14 @@ describe('BpkVessel', () => {
     expect(getByText('Plain Content')).toBeInTheDocument();
   });
 
+  it('forwards ref to the underlying DOM element', () => {
```
gc-skyscanner, line 68: could we add accessibility test for BpkVessel?

RichardSyq: Done in `2c4691ffe`. BpkVessel was the only layout primitive missing from the existing `jest-axe` suite in `accessibility-test.tsx`, so I added it there alongside Box/Flex/Grid/Stack. Two cases, both asserting `toHaveNoViolations()`: a basic `as="section"` usage with an `aria-label`, and a semantic `as="nav"` landmark with links.

gc-skyscanner (approved):

PR #4772, opened 2026-06-29
feat(BpkCheckboxV2): add fullWidth prop to Root

IrinaWei:
## Summary

- Adds opt-in `fullWidth?: boolean` prop to `BpkCheckboxV2.Root`
- When `true`, applies `bpk-checkbox-v2--full-width` modifier → `width: 100%`
- Default `inline-flex` display **unchanged** — zero breaking changes for existing consumers (Overlay, carhire-homepage)
- Adds `FullWidth` Storybook story with unchecked / checked+description / disabled variants inside a constrained container

## Why not change `inline-flex` → `flex` globally?

Searched Skyscanner org — found two external consumers using `BpkCheckboxV2.Root` in contexts that would break if the root became block-level. An opt-in prop avoids all regressions.

## Test plan

- [ ] `BpkCheckboxV2-test` — 17 tests pass, snapshots unchanged
- [ ] Lint clean (lint-staged passed on commit)
- [ ] Storybook: `FullWidth` story renders at 300px container, all three variants fill width
- [ ] Existing stories unaffected

### DIFF
--- a/packages/backpack-web/src/bpk-component-checkbox/src/BpkCheckboxV2/BpkCheckboxV2.module.scss
+++ b/packages/backpack-web/src/bpk-component-checkbox/src/BpkCheckboxV2/BpkCheckboxV2.module.scss
@@ -36,6 +36,10 @@
   &[data-disabled] {
     cursor: not-allowed;
   }
+
+  &--full-width {
+    width: 100%;
+  }
 }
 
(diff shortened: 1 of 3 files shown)


IrinaWei:
Close it for now, this feature fix would be included in later BpkCheckBox upgrade

PR #4779, opened 2026-07-01
[BpkText] Migrate color prop to CSS custom properties for dark mode

gert-janvercauteren:
Migrates `BpkText`'s `color` prop and the base stylesheet body color to CSS custom properties, enabling proper dark mode support. The `$text-colors` SCSS map now uses `var(--bpk-text-*, fallback)` expressions instead of static Sass token values, so all named text colors respond to theme overrides. The `bpk-stylesheets/index.scss` body color is updated from a static `tokens.$bpk-font-color-base` to `var(--bpk-text-primary, tokens.$bpk-font-color-base)`, meaning default (uncolored) text also switches correctly in dark mode. The `dark-mode-compatible` tag is added to the BpkText Storybook meta.

- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests


packages/backpack-web/src/bpk-component-text/src/BpkText.module.scss
```diff
@@ -16,21 +16,55 @@
...
+  'text-error': (
+    --bpk-text-error,
+    tokens.$bpk-text-error-day,
+  ),
+  'text-hero': (
+    --bpk-text-hero,
+    tokens.$bpk-text-hero-day,
+  ),
+  'text-link': (
+    --bpk-text-link,
```
gc-skyscanner: Seems we don't have this variable in Figma variables 🤔 although it will fallback to `tokens.$bpk-text-link-day,` do we need to highlight it with our designer?

gert-janvercauteren: We have declared this 'dead' so probably better to not hook it up.

gc-skyscanner (approved):

PR #4789, opened 2026-07-02
[WOODPECKER-4406] Migrate package manager from npm to pnpm

viktor-yang:
Revert the revert of https://github.com/Skyscanner/backpack/pull/4738

Last time, we have merged that PR into main. But it throw an error:
image

Reason:
In the original PR, we have updated the pnpm cache for the `pr.yaml` and `_build.yml` and forgot to update the `main.yml` `release.yml` and `dev-release.yml`

In the `main.yml`, the cache key for the storybook build artifact is different:
* **Create-Build-Cache:** The cache key is computed **before** running `pnpm install`, so `hashFiles()` only sees the source files, resulting in the `76ad...` key.
* **`_build.yml` / `PercyTests`:** The cache key is computed **after** `pnpm install`, so `hashFiles()` also includes the package-local `node_modules`, resulting in the `c9bb...` key.

Fixed:
In this PR, we replaced the shared `node_modules` cache (the `Create-NPM-Cache` job and the `Restore NPM Cache` steps) with `setup-node`’s built-in pnpm store cache, together with an unconditional `pnpm install --frozen-lockfile` in every job, matching `pr.yml`.

This ensures that every job computes the `build-cache` `hashFiles()` key against the same on-disk state, which fixes the PercyTests cache miss in `main.yml`, where the cache-saving job previously skipped the install step.
See https://github.com/Skyscanner/backpack/pull/4789/changes/a73416364081830adce381a6c6c494726048b014


.specify/templates/plan-template.md
```diff
@@ -64,7 +64,7 @@ VALIDATION:
 **Build Tools**: Webpack 5, Babel 7
 **Linting**: ESLint (@skyscanner/eslint-config-skyscanner), Stylelint
 **Component Library**: Backpack Design System (Monorepo)
-**Package Manager**: npm >=10.7.0
+**Package Manager**: pnpm >=9.15.9
 **Node Version**: >=18.20.4
```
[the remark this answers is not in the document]
viktor-yang: No relative

.github/workflows/_build.yml
```diff
@@ -122,30 +118,27 @@ jobs:
...
-        with:
-          path: |
-            node_modules/
-          key: ${{ env.CACHE_NAME }}-${{ hashFiles('package-lock.json') }}
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
 
       - name: Override React to 19.2.5
-        run: npm install --no-save react@19.2.5 react-dom@19.2.5 @types/react@19.0.14 @types/react-dom@19.0.6 @types/prop-types
+        run: pnpm_config_lockfile=false pnpm add -w react@19.2.5 react-dom@19.2.5 @types/react@19.0.14 @types/react-dom@19.0.6 @types/prop-types @types/react-transition-group@4.4.12
```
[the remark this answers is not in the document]
viktor-yang: Fixed： https://github.com/Skyscanner/backpack/pull/4789/changes/93a7505ff4d396b21799fd248ccaa3037a9ced64

DEV_RELEASE.md
```diff
@@ -60,7 +60,7 @@ from.
    (no forks), and the user who triggered the workflow has write access.
 2. **Started comment** — a comment is posted on the PR linking to the
    workflow run.
-3. **Build** — `npm ci` and `npm run build` run against the PR's HEAD
+3. **Build** — `pnpm install --frozen-lockfile` and `pnpm run build` run against the PR's HEAD
    commit, producing `packages/backpack-web/dist`.
 4. **Publish** — only when `dry_run=false`. Bumps the version with
    `npm version --no-git-tag-version` and runs `npm publish --tag dev`. This
```
kerrie-wu, line 66: Why we keep  npm publish here, Could we change it to pnpm publish

.github/workflows/dev-release.yml
kerrie-wu, line 150: Could we change npm publish to pnpm publish ?

.github/workflows/release.yml
kerrie-wu, line 90: same here

DEV_RELEASE.md
viktor-yang: Hi Kerrie. Nice catch! Our goal is to switch the package manager from NPM to PNPM (`pnpm install`). `npm build` should still work, and we'll continue publishing to the npm registry.

If we also want to switch the publish command to `pnpm publish`, that's certainly possible, but it will require some extra work and involve a bit more risk.

Do you prefer `pnpm publish`?

kerrie-wu: Agree , keep `npm publish` first

kerrie-wu (approved):
LGTM

PR #4848, opened 2026-07-03
[HER-1012] Add secondary panel to BpkDrawer

lucia-jiang:
Adds a `secondaryPanel` prop that renders a second column inside the drawer. The drawer expands to calc(var(--dynamic-width) * 2) via CSS transition to reveal the secondary content alongside the primary panel. No breaking changes to existing API.

Video:

https://github.com/user-attachments/assets/9e8c2652-e878-4f80-bd26-02f0593c5199

- [X] Ensure the PR title includes the name of the component you are changing so it's clear in the release notes for consumers of the changes in the version e.g `[Clover-123][BpkButton] Updating the colour`
- [ ] Component `README.md`
- [X] Tests
- [ ] Accessibility tests
        - [X] Ability to navigate using a [keyboard only](https://webaim.org/techniques/keyboard/)
        - [X] Zoom functionality ([Deque University explanation](https://dequeuniversity.com/checklists/web/text)):
            - [X] The page SHOULD be functional AND readable when only the text is magnified to 200% of its initial size
        - [X] Ability to navigate using a [screen reader only](https://webaim.org/articles/screenreader_testing/)
- [X] Storybook examples created/updated


rhonasky (dismissed):
LGTM

rhonasky (dismissed):

rhonasky (dismissed):

packages/backpack-web/src/bpk-component-drawer/src/types.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ReactNode } from 'react';
+
+export type SecondaryPanelProps = {
```
gert-janvercauteren, line 21: Is there a way to keep the primary & secondary panel props in sync? Ideally we could have 'an array of drawers' that all render as 1 drawer one after the other :D

packages/backpack-web/src/bpk-component-drawer/src/BpkDrawerContent.module.scss
```diff
@@ -132,4 +134,48 @@
...
+    overflow-y: auto;
+  }
+
+  &__secondary {
+    display: flex;
+    width: var(--dynamic-width);
+    min-width: 0;
+    flex-direction: column;
+    flex: 1 1 var(--dynamic-width);
+    border-left: tokens.$bpk-border-size-sm solid tokens.$bpk-line-day;
```
Faye-Xiao: nit:   In RTL, the secondary panel is logically on the right, but `border-left` will put the border on the wrong side. We encourage using `border-inline-start` instead.

```
border-left: tokens.$bpk-border-size-sm solid tokens.$bpk-line-day;
```

packages/backpack-web/src/bpk-component-drawer/src/BpkDrawerContent.tsx
```diff
@@ -143,20 +167,32 @@ const BpkDrawerContent = ({
...
-                <BpkCloseButton label={closeLabel} onClick={onClose} />
+          {secondaryPanel?.isOpen ? (
+            <div className={getClassName('bpk-drawer__panels')}>
+              <div className={getClassName('bpk-drawer__primary')}>
+                {drawerHeader}
+                <div className={contentClassNames.join(' ')}>{children}</div>
+              </div>
+              <div
+                role="complementary"
+                aria-label="Secondary panel"
```
Faye-Xiao: Proposing adding an `ariaLabel` field to `SecondaryPanelProps` (optional, with a fallback). Otherwise, any non-English consumer silently ships incorrect screen reader text.

packages/backpack-web/src/bpk-component-drawer/src/BpkDrawer.stories.js
```diff
@@ -330,6 +330,68 @@ const DrawerWithTooltipExampleAbleToBeShown = () => (
   </DrawerContainer>
 );
 
+class DrawerWithSecondaryPanelContainer extends Component {
```
Faye-Xiao: nit: just curious if there’s a particular reason you're leaning towards class components? I noticed that most of the recent stories in this file are using functional components and hooks. It might be worth considering a switch for the sake of consistency. What do you think? 😊

packages/backpack-web/src/bpk-component-drawer/src/types.ts
lucia-jiang: Hi Gert! Thanks for your comment. It is a good idea to sync them, but!

The primary panel is structurally different from the secondary panel, a unified array shape would either force dialog-level props (dialogRef, closeOnScrimClick, mobileModalDisplay, isIpad, isIphone) onto every panel object where they'd never be used, or require a significant refactor of how BpkDrawer owns the dialog vs the panel content.

packages/backpack-web/src/bpk-component-drawer/src/BpkDrawer.stories.js
lucia-jiang: Thanks Faye, good point! I modified it now ☺️

packages/backpack-web/src/bpk-component-drawer/src/types.ts
lucia-jiang: This is how the first and second pannel would look like 
image

gert-janvercauteren: For visibility: agreed to keep a single sub-panel. Thanks!

packages/backpack-web/src/bpk-component-drawer/src/BpkDrawerContent.tsx
```diff
@@ -143,20 +166,43 @@ const BpkDrawerContent = ({
...
-                <BpkCloseButton label={closeLabel} onClick={onClose} />
+          {secondaryPanel?.isOpen ? (
+            <div className={getClassName('bpk-drawer__panels')}>
+              <div className={getClassName('bpk-drawer__primary')}>
+                {drawerHeader}
+                <div className={contentClassNames.join(' ')}>{children}</div>
+              </div>
+              <div
+                role="complementary"
+                aria-label={secondaryPanel.ariaLabel ?? 'Secondary panel'}
```
Faye-Xiao: Nit: the fallback `Secondary panel` is hardcoded English, which means consumers who don't pass `ariaLabel` silently get incorrect screen reader text in non-English contexts. Since a single complementary landmark doesn't require an accessible name, you can just drop the fallback:

```
- aria-label={secondaryPanel.ariaLabel ?? 'Secondary panel'}
+ aria-label={secondaryPanel.ariaLabel}
```

Faye-Xiao (approved):
LGTM.

PR #4870, opened 2026-07-06
[BpkLayout][CLOV-1823] Add layout positioning and spacing token support

Supremeyh:
## What

Extends the Backpack layout components with positioning, overlay, and spacing token improvements, split from CLOV-1812 for independent delivery.

**JIRA:** https://skyscanner.atlassian.net/browse/CLOV-1823

### Tokens (`tokens.ts`, `theme.ts`)
- Add `BpkSpacing.XXXL` (`4rem` / 64px)
- Update `BpkPositionValue` to accept BPK spacing tokens in addition to rem/% values
- Update `isValidPositionValue` and `isPercentage` (now accepts negative percentages, e.g. `-50%` for centering)
- Hoist `BPK_SPACING_TOKEN_SET` as a module-level `ReadonlySet` — O(1) token lookup, avoids per-call array allocation

### Common layout props (`commonProps.ts`)
- Add `insetInlineStart` / `insetInlineEnd` — RTL-safe logical equivalents of `left`/`right`; accept `BpkPositionValue` (rem, %, `'0'`, or BPK token); support responsive overrides
- Add `scrollMarginTop` / `scrollMarginBottom` — scroll snap margin props; scalar `BpkPositionValue` (accepts rem values like `'3.5rem'` for sticky-header offsets, plus BPK tokens and `'0'`)
- Add `pointerEvents?: 'none' | 'auto'` — click-through overlays; available on all layout components
- Remove `transform?: never` — `BpkCommonLayoutProps` is an explicit allowlist (never extends Chakra `BoxProps`), so the `never` was a convention marker only; `BpkBoxSpecificProps` adds `transform?: string` directly with no conflict

### `BpkBoxSpecificProps` (`types.ts`)
- Add `transform?: string` — BpkBox-only; supports any CSS transform string (e.g. `translateX`, `rotate`, `scale`)
- `BpkBoxProps` stays clean: `extends BpkCommonLayoutProps, BpkBoxSpecificProps` — no `Omit` needed

### `BpkFlexSpecificProps` (`types.ts`) + `BpkFlex.tsx`
- Add `flexShrink` as a direct CSS prop alias alongside existing `shrink`; `shrink` takes precedence if both are provided

### Token processing (`tokenUtils.ts`)
- Add `convertBpkPositionValue` — uses `BPK_SPACING_TOKEN_SET.has()` as the sole discriminator; no duplicated raw-value regex
- All position-family props (`top/right/bottom/left`, `insetInlineStart/End`, `scrollMarginTop/Bottom`) go through `convertBpkPositionValue`
- Update `processSpacingProps` JSDoc to list all handled prop groups

### Documentation
- `README.md` updated: `scrollMarginTop/Bottom` moved to Position offsets section with correct type and sticky-header example; `XXXL` token added; `transform` constraint clarified (BpkBox only); `pointerEvents` added to interaction surface; `flexShrink` alias noted

## Acceptance Criteria

- [x] Add `scrollMarginTop` and `scrollMarginBottom` support to common layout props
- [x] Add Backpack spacing token support for logical inset props (`insetInlineStart`, `insetInlineEnd`)
- [x] Add `transform` support for `BpkBox`
- [x] Add `pointerEvents` support for `BpkBox` and `BpkFlex`
- [x] Update `BpkPositionValue` so both arbitrary values and BPK tokens are accepted
- [x] Add `bpk-spacing-xxxl` where appropriate across tokens, docs, and types
- [x] Add tests and Storybook/docs examples for all new supported props and token behaviour

## Testing

- `tokenUtils-test.ts`: `bpk-spacing-xxxl` conversion; spacing tokens and raw rem in position props; `insetInlineStart/End` (scalar + responsive + raw); `scrollMarginTop/Bottom` (token + raw rem `'3.5rem'`); negative percentages (`-50%`); `convertBpkPositionValue` passthrough
- `BpkBox-test.tsx`: `transform`, `pointerEvents`, `insetInlineStart/End`, `scrollMarginTop/Bottom` (token + rem)
- `BpkFlex-test.tsx`: `pointerEvents`, `flexShrink` (direct prop + precedence over `shrink`)
- Storybook stories added: `Transform`, `PointerEvents`, `InsetInline` on `BpkBox`; `FlexShrink`, `FlexPointerEvents` on `BpkFlex`; `BpkSpacing.XXXL` folded into existing `Spacing` story

## Checklist

- [x] PR title includes component name and JIRA ticket
- [x] Component `README.md` updated
- [x] Tests added
- [ ] Accessibility tests (not applicable — structural token/prop changes only)

### DIFF
--- a/packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
+++ b/packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
@@ -491,105 +491,94 @@ describe('BpkBox', () => {
   });
 
   describe('transform prop (BpkBox only)', () => {
-    it('renders with a CSS transform string', () => {
+    it('applies the CSS transform to the DOM element', () => {
       const { container } = render(
         <BpkProvider>
           <BpkBox transform="translateX(10px)">Transformed</BpkBox>
         </BpkProvider>,
       );
-      expect(container.querySelector('div')).toBeInTheDocument();
+      expect(container.querySelector('div')).toHaveStyle('transform: translateX(10px)');
     });
 
-    it('produces different output with and without transform', () => {
+    it('applies a different transform value than no transform', () => {
       const { container: withTransform } = render(
         <BpkProvider>
           <BpkBox transform="rotate(45deg)">Rotated</BpkBox>
         </BpkProvider>,
       );
-      const { container: withoutTransform } = render(
-        <BpkProvider>
-          <BpkBox>No transform</BpkBox>
-        </BpkProvider>,
-      );
-      expect(withTransform.querySelector('div')?.className).not.toBe(
-        withoutTransform.querySelector('div')?.className,
-      );
+      expect(withTransform.querySelector('div')).toHaveStyle('transform: rotate(45deg)');
     });
   });
 
   describe('pointerEvents prop', () => {
-    it('renders with pointerEvents="none"', () => {
+    it('applies pointer-events: none to the DOM element', () => {
       const { container } = render(
         <BpkProvider>
           <BpkBox pointerEvents="none">Overlay</BpkBox>
         </BpkProvider>,
       );
-      expect(container.querySelector('div')).toBeInTheDocument();
+      expect(container.querySelector('div')).toHaveStyle('pointer-events: none');
     });
 
-    it('produces different output with pointerEvents="none" vs default', () => {
-      const { container: withNone } = render(
-        <BpkProvider>
-          <BpkBox pointerEvents="none">Click-through</BpkBox>
-        </BpkProvider>,
-      );
-      const { container: withAuto } = render(
+    it('applies pointer-events: auto to the DOM element', () => {
+      const { container } = render(
         <BpkProvider>
           <BpkBox pointerEvents="auto">Interactive</BpkBox>
         </BpkProvider>,
       );
-      expect(withNone.querySelector('div')?.className).not.toBe(
-        withAuto.querySelector('div')?.className,
-      );
+      expect(container.querySelector('div')).toHaveStyle('pointer-events: auto');
     });
   });
 
   describe('insetInlineStart / insetInlineEnd props', () => {
-    it('renders with a spacing token for insetInlineStart', () => {
+    it('applies inset-inline-start from a BPK spacing token (resolves to rem)', () => {
       const { container } = render(
         <BpkProvider>
           <BpkBox position="absolute" insetInlineStart={BpkSpacing.Base}>Content</BpkBox>
         </BpkProvider>,
       );
-      expect(container.querySelector('div')).toBeInTheDocument();
+      // BpkSpacing.Base = 1rem
+      expect(container.querySelector('div')).toHaveStyle('inset-inline-start: 1rem');
     });
 
-    it('renders with a raw rem value for insetInlineEnd', () => {
+    it('applies inset-inline-end from a raw rem value', () => {
       const { container } = render(
         <BpkProvider>
           <BpkBox position="absolute" insetInlineEnd="1.5rem">Content</BpkBox>
         </BpkProvider>,
       );
-      expect(container.querySelector('div')).toBeInTheDocument();
+      expect(container.querySelector('div')).toHaveStyle('inset-inline-end: 1.5rem');
     });
   });
 
   describe('scrollMarginTop / scrollMarginBottom props', () => {
-    it('renders with scrollMarginTop using a spacing token', () => {
+    it('applies scroll-margin-top from a BPK spacing token (resolves to rem)', () => {
       const { container } = render(
         <BpkProvider>
           <BpkBox scrollMarginTop={BpkSpacing.XXXL}>Scroll target</BpkBox>
         </BpkProvider>,
       );
-      expect(container.querySelector('div')).toBeInTheDocument();
+      // BpkSpacing.XXXL = 4rem
+      expect(container.querySelector('div')).toHaveStyle('scroll-margin-top: 4rem');
     });
 
-    it('renders with scrollMarginBottom using a spacing token', () => {
+    it('applies scroll-margin-bottom from a BPK spacing token', () => {
       const { container } = render(
         <BpkProvider>
           <BpkBox scrollMarginBottom={BpkSpacing.MD}>Scroll target</BpkBox>
         </BpkProvider>,
       );
-      expect(container.querySelector('div')).toBeInTheDocument();
+      // BpkSpacing.MD = .5rem
+      expect(container.querySelector('div')).toHaveStyle('scroll-margin-bottom: .5rem');
     });
 
-    it('renders with scrollMarginTop using a raw rem value (e.g. sticky-header height)', () => {
+    it('applies scroll-margin-top from a raw rem value (sticky-header height)', () => {
       const { container } = render(
         <BpkProvider>
           <BpkBox scrollMarginTop="3.5rem">Scroll target</BpkBox>
         </BpkProvider>,
       );
-      expect(container.querySelector('div')).toBeInTheDocument();
+      expect(container.querySelector('div')).toHaveStyle('scroll-margin-top: 3.5rem');
     });
   });
 
--- a/packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
+++ b/packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
@@ -490,4 +490,98 @@ describe('BpkBox', () => {
     });
   });
 
+  describe('transform prop (BpkBox only)', () => {
+    it('renders with a CSS transform string', () => {
+      const { container } = render(
+        <BpkProvider>
+          <BpkBox transform="translateX(10px)">Transformed</BpkBox>
+        </BpkProvider>,
+      );
+      expect(container.querySelector('div')).toBeInTheDocument();
+    });
+
+    it('produces different output with and without transform', () => {
+      const { container: withTransform } = render(
+        <BpkProvider>
+          <BpkBox transform="rotate(45deg)">Rotated</BpkBox>
+        </BpkProvider>,
+      );
+      const { container: withoutTransform } = render(
+        <BpkProvider>
+          <BpkBox>No transform</BpkBox>
+        </BpkProvider>,
+      );
+      expect(withTransform.querySelector('div')?.className).not.toBe(
+        withoutTransform.querySelector('div')?.className,
+      );
+    });
+  });
+
+  describe('pointerEvents prop', () => {
+    it('renders with pointerEvents="none"', () => {
+      const { container } = render(
+        <BpkProvider>
+          <BpkBox pointerEvents="none">Overlay</BpkBox>
+        </BpkProvider>,
+      );
+      expect(container.querySelector('div')).toBeInTheDocument();
+    });
+
+    it('produces different output with pointerEvents="none" vs default', () => {
+      const { container: withNone } = render(
+        <BpkProvider>
+          <BpkBox pointerEvents="none">Click-through</BpkBox>
+        </BpkProvider>,
+      );
+      const { container: withAuto } = render(
+        <BpkProvider>
+          <BpkBox pointerEvents="auto">Interactive</BpkBox>
+        </BpkProvider>,
+      );
+      expect(withNone.querySelector('div')?.className).not.toBe(
+        withAuto.querySelector('div')?.className,
+      );
+    });
+  });
+
+  describe('insetInlineStart / insetInlineEnd props', () => {
+    it('renders with a spacing token for insetInlineStart', () => {
+      const { container } = render(
+        <BpkProvider>
+          <BpkBox position="absolute" insetInlineStart={BpkSpacing.Base}>Content</BpkBox>
+        </BpkProvider>,
+      );
+      expect(container.querySelector('div')).toBeInTheDocument();
+    });
+
+    it('renders with a raw rem value for insetInlineEnd', () => {
+      const { container } = render(
+        <BpkProvider>
+          <BpkBox position="absolute" insetInlineEnd="1.5rem">Content</BpkBox>
+        </BpkProvider>,
+      );
+      expect(container.querySelector('div')).toBeInTheDocument();
+    });
+  });
+
+  describe('scrollMarginTop / scrollMarginBottom props', () => {
+    it('renders with scrollMarginTop using a spacing token', () => {
+      const { container } = render(
+        <BpkProvider>
+          <BpkBox scrollMarginTop={BpkSpacing.XXXL}>Scroll target</BpkBox>
+        </BpkProvider>,
+      );
+      expect(container.querySelector('div')).toBeInTheDocument();
+    });
+
+    it('renders with scrollMarginBottom using a spacing token', () => {
+      const { container } = render(
+        <BpkProvider>
+          <BpkBox scrollMarginBottom={BpkSpacing.MD}>Scroll target</BpkBox>
+        </BpkProvider>,
+      );
+      expect(container.querySelector('div')).toBeInTheDocument();
+    });
+  });
+
 });
--- a/packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
+++ b/packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
@@ -582,6 +582,15 @@ describe('BpkBox', () => {
       );
       expect(container.querySelector('div')).toBeInTheDocument();
     });
+
+    it('renders with scrollMarginTop using a raw rem value (e.g. sticky-header height)', () => {
+      const { container } = render(
+        <BpkProvider>
+          <BpkBox scrollMarginTop="3.5rem">Scroll target</BpkBox>
+        </BpkProvider>,
+      );
+      expect(container.querySelector('div')).toBeInTheDocument();
+    });
   });
 
 });
--- a/packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
+++ b/packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
@@ -714,3 +714,196 @@ const DirExample = () => (
 export const Dir = {
   render: () => <DirExample />,
 };
+
+/**
+ * Transform example – demonstrates the `transform` prop on BpkBox for translate, rotate, and scale.
+ *
+ * @returns {JSX.Element} Boxes with CSS transform applied.
+ */
+const TransformExample = () => (
+  <LayoutWrapper>
+    <BpkBox padding={BpkSpacing.SM} marginBottom={BpkSpacing.MD}>
+      <BpkText textStyle={TEXT_STYLES.label2}>translateX(16px) — horizontal shift</BpkText>
+      <BpkBox
+        transform="translateX(16px)"
+        padding={BpkSpacing.SM}
+        marginTop={BpkSpacing.SM}
+        backgroundColor={BACKGROUND_COLORS.surfaceElevated}
+      >
+        <BpkText>Shifted 16px to the right</BpkText>
+      </BpkBox>
+    </BpkBox>
+
+    <BpkBox padding={BpkSpacing.SM} marginBottom={BpkSpacing.MD}>
+      <BpkText textStyle={TEXT_STYLES.label2}>rotate(10deg) — slight rotation</BpkText>
+      <BpkBox
+        transform="rotate(10deg)"
+        padding={BpkSpacing.SM}
+        marginTop={BpkSpacing.SM}
+        backgroundColor={BACKGROUND_COLORS.surfaceElevated}
+      >
+        <BpkText>Rotated 10 degrees</BpkText>
+      </BpkBox>
+    </BpkBox>
+
+    <BpkBox padding={BpkSpacing.SM}>
+      <BpkText textStyle={TEXT_STYLES.label2}>scale(0.8) — scaled down</BpkText>
+      <BpkBox
+        transform="scale(0.8)"
+        padding={BpkSpacing.SM}
+        marginTop={BpkSpacing.SM}
+        backgroundColor={BACKGROUND_COLORS.surfaceElevated}
+      >
+        <BpkText>Scaled to 80%</BpkText>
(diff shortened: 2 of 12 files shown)


packages/backpack-web/src/bpk-component-layout/src/commonProps.ts
```diff
@@ -194,12 +204,12 @@ export interface BpkCommonLayoutProps extends BpkSpacingProps, AriaAttributes {
...
-  // Explicitly exclude transition & transform related props for performance reasons
+  // Explicitly exclude transition-related props — apply animations via CSS classes instead
   transition?: never;
   transitionProperty?: never;
   transitionDuration?: never;
   transitionTimingFunction?: never;
   transitionDelay?: never;
-  transform?: never;
+  // transform is excluded here so BpkBox can opt-in via its own specific props
   transformOrigin?: never;
```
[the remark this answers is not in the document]
Supremeyh: [specifi transform for BpkBox](https://github.com/Skyscanner/backpack/pull/4870/commits/6ba5efe52916bc71cf8e58887919ebdf1f518504)

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
```diff
@@ -205,15 +207,16 @@ export function isValidSizeValue(value: string): boolean {
...
  */
 export function isValidPositionValue(value: string): boolean {
   return (
     value === '0' || // bare zero — valid CSS without a unit
     /^-?\d+(\.\d+)?rem$/.test(value) || // rem values
-    isPercentage(value) // percentage values
+    isPercentage(value) || // percentage values
+    Object.values(BpkSpacing).includes(value as BpkSpacingToken) // BPK spacing tokens
   );
 }
```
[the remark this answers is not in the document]
Supremeyh: [specifi transform for BpkBox](https://github.com/Skyscanner/backpack/pull/4870/commits/6ba5efe52916bc71cf8e58887919ebdf1f518504)

packages/backpack-web/src/bpk-component-layout/src/tokenUtils.ts
```diff
@@ -267,6 +267,20 @@ export function processBpkComponentProps<T extends Record<string, any>>(
...
+ *
+ * @param {string} value - Position value or BPK spacing token
+ * @returns {string} CSS value
+ */
+export function convertBpkPositionValue(value: string): string {
+  if (value === '0' || /^-?\d+(\.\d+)?rem$/.test(value) || isPercentage(value)) {
+    return value;
+  }
+  return convertBpkSpacingToChakra(value);
+}
```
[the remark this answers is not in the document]
Supremeyh: [specifi transform for BpkBox](https://github.com/Skyscanner/backpack/pull/4870/commits/6ba5efe52916bc71cf8e58887919ebdf1f518504)

packages/backpack-web/src/bpk-component-layout/src/BpkFlex.tsx
```diff
@@ -42,7 +42,7 @@ export const BpkFlex = forwardRef<HTMLDivElement, BpkFlexProps>(
         alignItems: align,
         flexWrap: wrap,
         flexGrow: grow,
-        flexShrink: shrink,
+        flexShrink: shrink ?? flexShrink,
```
Supremeyh, line 45: `shrink` takes precedence over `flexShrink` if both are provided

packages/backpack-web/src/bpk-component-layout/src/types.ts
```diff
@@ -117,15 +117,18 @@ type BpkBoxResponsiveLayoutPropKeys = keyof BpkBoxResponsiveLayoutProps;
  */
 export interface BpkBoxSpecificProps
   extends BpkBoxResponsiveLayoutProps,
-    Omit<BpkFlexGridProps, BpkBoxResponsiveLayoutPropKeys> {}
+  Omit<BpkFlexGridProps, BpkBoxResponsiveLayoutPropKeys> {
+  // CSS transform — BpkBox only, for positioning and animation (e.g. translate, scale, rotate)
+  transform?: string;
```
Supremeyh, line 121: CSS transform — BpkBox only, for positioning and animation (e.g. translate, scale, rotate)

packages/backpack-web/src/bpk-component-layout/src/tokenUtils.ts
```diff
@@ -425,8 +445,11 @@ export function processSpacingProps<T extends Record<string, any>>(
       const isMarginProp = marginKeys.includes(key);
 
       let converter: (v: string) => string;
-      if (isSizeProp || isPositionProp) {
+      if (isSizeProp) {
         converter = (v: string) => v;
+      } else if (isPositionProp) {
+        // Resolves BPK spacing tokens to rem; raw rem/% values pass through without a warning.
+        converter = convertBpkPositionValue;
```
Supremeyh, line 461: Resolves BPK spacing tokens to rem; raw rem/% values pass through without a warning.

packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
```diff
@@ -714,3 +714,201 @@ const DirExample = () => (
...
+export const InsetInline = {
+  render: () => <InsetInlineExample />,
+};
+
+/**
+ * SpacingXXXL example – demonstrates the new `bpk-spacing-xxxl` (4rem / 64px) token.
+ *
+ * @returns {JSX.Element} A box using the xxxl spacing token for padding and margin.
+ */
+const SpacingXXXLExample = () => (
```
Faye-Xiao: nit: The BpkBox currently has a `Spacing` story. I suppose we don't need a separate one for the XXXL size; we can add it to the existing one.

packages/backpack-web/src/bpk-component-layout/src/types.ts
```diff
@@ -117,15 +117,17 @@ type BpkBoxResponsiveLayoutPropKeys = keyof BpkBoxResponsiveLayoutProps;
...
+}
 
 /**
  * Props for BpkBox component
  * Combines Box-specific props with Backpack common layout props.
  * onClick, onFocus and onBlur are inherited from BpkCommonLayoutProps.
  * textStyle maps to Chakra's `textStyle` theme prop for Backpack typography and supports responsive values.
  */
-export interface BpkBoxProps extends BpkCommonLayoutProps, BpkBoxSpecificProps {
+export interface BpkBoxProps extends Omit<BpkCommonLayoutProps, 'transform'>, BpkBoxSpecificProps {
```
Faye-Xiao: The ·Omit<BpkCommonLayoutProps, 'transform'>· approach works, but it's fighting against the design. The layout type system is an explicit allowlist — `BpkCommonLayoutProps` never extends `BoxProps`, so ·transform?: never· in `commonProps` isn't actually blocking a leaked Chakra prop; it's just a declarative intent marker. The never exists by convention, not necessity.

Given that, I would propose the simpler fix is:

Remove `transform?: never` from `commonProps.ts`
Add `transform?: string` directly to `BpkBoxSpecificProps`
Keep `BpkBoxProps` extends `BpkCommonLayoutProps`, `BpkBoxSpecificProps` — no Omit needed

```
// commonProps.ts — just remove the transform line
// transform?: never;  ← delete

// types.ts
export interface BpkBoxSpecificProps
  extends BpkBoxResponsiveLayoutProps,
  Omit<BpkFlexGridProps, BpkBoxResponsiveLayoutPropKeys> {
  transform?: string;
}

// BpkBoxProps stays clean
export interface BpkBoxProps extends BpkCommonLayoutProps, BpkBoxSpecificProps {
  children?: ReactNode;
}
```
Also worth noting: `transformOrigin?: never` is still in `BpkCommonLayoutProps` and is inherited by `BpkBoxProps` (it's not in the Omit). transform and transformOrigin are usually used together — either open both or document why transformOrigin isn't supported.

packages/backpack-web/src/bpk-component-layout/src/commonProps.ts
```diff
@@ -84,11 +84,18 @@ export interface BpkSpacingProps {
...
   right?: BpkResponsiveValue<BpkPositionValue>;
   bottom?: BpkResponsiveValue<BpkPositionValue>;
   left?: BpkResponsiveValue<BpkPositionValue>;
+  // CSS logical position props — RTL-safe equivalents of left/right
+  insetInlineStart?: BpkResponsiveValue<BpkPositionValue>;
+  insetInlineEnd?: BpkResponsiveValue<BpkPositionValue>;
+
+  // Scroll snap margin props — space between the scroll container's snap point and the element
+  scrollMarginTop?: BpkResponsiveValue<BpkSpacingValue>;
+  scrollMarginBottom?: BpkResponsiveValue<BpkSpacingValue>;
```
Faye-Xiao: `insetInlineStart`/`insetInlineEnd` as `BpkResponsiveValue` is fine — they're the logical equivalents of `left/right`, which already support responsive values, and the use cases are symmetric (e.g. an overlay that shifts position at different breakpoints).

🤔   `scrollMarginTop`/`scrollMarginBottom` as `BpkResponsiveValue` is questionable. `scroll-margin` is typically a fixed offset tied to something like a sticky header height — a value that doesn't change per breakpoint in practice. If the sticky header itself changes height across breakpoints, that's better handled via a CSS variable than a responsive prop object.

Supporting `BpkResponsiveValue` here adds API surface and processing overhead for a use case that's unlikely to exist. A plain `BpkSpacingValue` (scalar only) would be a safer, simpler type for this prop.

Could you double-check if we have a real use case that needs a responsive value for these 2 props?

packages/backpack-web/src/bpk-component-layout/src/tokenUtils.ts
```diff
@@ -405,13 +422,16 @@ export function processSpacingProps<T extends Record<string, any>>(
     'rowGap', 'columnGap',
     // Size props
     'width', 'height', 'minWidth', 'minHeight', 'maxWidth', 'maxHeight',
-    // Position props
+    // Position props (rem, %, '0', or BPK spacing tokens)
     'top', 'right', 'bottom', 'left',
+    'insetInlineStart', 'insetInlineEnd',
+    // Scroll snap margin props
+    'scrollMarginTop', 'scrollMarginBottom',
```
Faye-Xiao, line 439: nit: Could you update the JSDoc of the `processSpacingProps` function since we added the `insetInlineStart/End` and `scrollMarginTop/Bottom`?

packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
```diff
@@ -714,3 +714,201 @@ const DirExample = () => (
...
+  render: () => <TransformExample />,
+};
+
+/**
+ * PointerEvents example – demonstrates `pointerEvents="none"` for click-through overlays
+ * and `pointerEvents="auto"` for restoring interaction.
+ *
+ * @returns {JSX.Element} A stacked overlay with pointer-events control.
+ */
+const PointerEventsExample = () => {
```
Faye-Xiao, line 789: A typo? Looks like this is a position story. 🤔

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
```diff
@@ -205,15 +209,16 @@ export function isValidSizeValue(value: string): boolean {
...
- * @returns {boolean} True if the value is a valid rem or percentage
+ * @returns {boolean} True if the value is valid
  */
 export function isValidPositionValue(value: string): boolean {
   return (
     value === '0' || // bare zero — valid CSS without a unit
     /^-?\d+(\.\d+)?rem$/.test(value) || // rem values
-    isPercentage(value) // percentage values
+    isPercentage(value) || // percentage values
+    BPK_SPACING_TOKEN_SET.has(value) // BPK spacing tokens
```
Faye-Xiao, line 222: This is pre-existing, but this PR widens `BpkPositionValue` to include `${number}%`, which makes the gap more visible.

TypeScript's number includes negatives, so '-50%' is valid at the type level. But isPercentage uses `/^\d+(\.\d+)?%$/` which rejects negative values — so `isValidPositionValue`('-50%') returns false at runtime, triggering a warn and dropping the value.

Negative percentages are a real use case for position offsets (e.g. top: '-50%' to centre an absolutely-positioned element). Since this PR is already touching both `BpkPositionValue` and `isValidPositionValue`, it's a good moment to align them:

```
// isPercentage — allow optional leading minus
return /^-?\d+(\.\d+)?%$/.test(value);
```

packages/backpack-web/src/bpk-component-layout/src/commonProps.ts
Faye-Xiao: I have another question: 
the `scrollMarginTop/Bottom` property accepts the `BpkSpacingValue`, which means it only supports either `BpkSpacingToken` or a value formatted as `${number}%`. I'm not sure if there are any real cases for using a custom value with this property. 🤔 The reason I raised this question is that, unlike other spacing-related properties, `scrollMarginTop` is associated with scrolling behaviour and may not need to be as restrictive. I guess the practical use case for `scroll-margin` is offsetting by a fixed height — most commonly a sticky header, which is typically a rem value like '3.5rem', not a spacing token. Could you double-check this?  If yes, I would propose changing the type to `BpkPositionValue`, which would also unify the processing path — both pairs would go through convertBpkPositionValue instead of `scrollMarginTop/Bottom` silently taking the `convertBpkSpacingToChakra` fallback branch in `processSpacingProps`.

packages/backpack-web/src/bpk-component-layout/src/commonProps.ts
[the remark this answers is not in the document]
Supremeyh: Done — took the approach you suggested in the follow-up comment (r3534639158): removed `transform?: never` from `BpkCommonLayoutProps` entirely rather than using `Omit`. `BpkCommonLayoutProps` is an explicit allowlist and never extends Chakra's `BoxProps`, so `transform?: never` was a convention marker only — removing it creates no type-system gap. `BpkBoxSpecificProps` now declares `transform?: string` directly, and `BpkBoxProps extends BpkCommonLayoutProps, BpkBoxSpecificProps` stays clean with no `Omit`. `transformOrigin` stays as `never` on the shared interface with a comment explaining why (BpkBox's API isn't designed for full transform-origin control).

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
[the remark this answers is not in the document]
Supremeyh: Fixed — hoisted a module-level `export const BPK_SPACING_TOKEN_SET: ReadonlySet<string> = new Set(Object.values(BpkSpacing))` in `tokens.ts`. Both `isValidSpacingValue` and `isValidPositionValue` now use `BPK_SPACING_TOKEN_SET.has(value)` (O(1), no allocation) instead of `Object.values(BpkSpacing).includes()`. The set is also reused in `convertBpkPositionValue` in `tokenUtils.ts` as the sole discriminator between token and raw value, which also addressed the duplication comment.

packages/backpack-web/src/bpk-component-layout/src/tokenUtils.ts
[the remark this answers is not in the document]
Supremeyh: Fixed — `convertBpkPositionValue` no longer duplicates the raw-value regex. Since it is only called after `isValidPositionValue` has already accepted the value, we know the input is either a BPK spacing token or a valid raw value (rem/%, `'0'`). The function now uses `BPK_SPACING_TOKEN_SET.has(value)` as the sole discriminator: if it's a token, delegate to `convertBpkSpacingToChakra`; otherwise return the value unchanged. Any future changes to what constitutes a valid raw position value only need to happen in `isValidPositionValue`.

packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
Supremeyh: Done — removed the standalone `SpacingXXXL` story export and folded the XXXL demonstration into the existing `Spacing` story. `BpkSpacing.XXXL` (64px padding) now appears as the second item in `SpacingExample` alongside the base spacing example.

packages/backpack-web/src/bpk-component-layout/src/types.ts
Supremeyh: Agreed — this is the cleaner approach. `BpkCommonLayoutProps` is a hand-written allowlist and never extends Chakra's `BoxProps`, so `transform?: never` was only a declarative intent marker, not a real type guard against prop leakage. Without it there's no conflict: `BpkFlexSpecificProps`, `BpkGridSpecificProps`, etc. simply don't declare `transform`, so TypeScript rejects it on those components regardless. `BpkBoxSpecificProps` adds `transform?: string` directly, and `BpkBoxProps` now simply extends both without any `Omit`.\n\nAlso addressed the `transformOrigin` question: it stays as `never` on the shared interface (including `BpkBoxProps`, since `BpkBoxSpecificProps` doesn't override it). A comment in `commonProps.ts` explains the rationale: `transformOrigin` is only meaningful alongside `transform`, and BpkBox's API is not designed for full transform-origin control.

packages/backpack-web/src/bpk-component-layout/src/commonProps.ts
Supremeyh: Good catch on `scrollMarginTop/Bottom` — the responsive API was unnecessary overhead. Changed both to scalar only (removed `BpkResponsiveValue<>` wrapper).\n\nAlso took your follow-up suggestion (r3534871673) and changed the type from `BpkSpacingValue` to `BpkPositionValue`. `BpkSpacingValue` only accepts spacing tokens or percentages, which would reject the primary real-world use case: `scrollMarginTop="3.5rem"` (sticky-header height). `BpkPositionValue` (rem | % | `'0'` | `BpkSpacingToken`) fits better and unifies the processing path — `scrollMarginTop/Bottom` now go through `convertBpkPositionValue` alongside `insetInlineStart/End`, rather than silently falling through the `convertBpkSpacingToChakra` warning branch.

packages/backpack-web/src/bpk-component-layout/src/tokenUtils.ts
Supremeyh: Updated — the JSDoc for `processSpacingProps` now lists all handled prop groups explicitly:\n- Padding, margin, gap (unchanged)\n- Size (unchanged)\n- Position offsets: `top/right/bottom/left`, `insetInlineStart/End` — accept rem, %, `'0'`, or BPK spacing tokens\n- Scroll snap margin: `scrollMarginTop/Bottom` — scalar `BpkPositionValue` (rem for sticky-header heights, tokens, or `'0'`)

packages/backpack-web/src/bpk-component-layout/src/BpkBox.stories.tsx
Supremeyh: Not a typo — this is the `PointerEvents` story. The original version used `position="absolute"` scaffolding to build an overlay demo, which made it look like a position story. Simplified the example: it now shows two plain boxes, one with `pointerEvents="none"` (clicking on it does nothing) and one with `pointerEvents="auto"` (fully interactive). No position props, so the focal point of the story is clear.

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
Supremeyh: Fixed — updated `isPercentage` regex from `/^\d+(\.\d+)?%$/` to `/^-?\d+(\.\d+)?%$/`. Negative percentages like `'-50%'` (commonly used with `top: -50%` to vertically centre an absolutely-positioned element) now pass both the type check (`${number}%` already included negatives) and the runtime validator.\n\nAdded a test in `tokenUtils-test.ts` that passes `top: '-50%'` and `left: '-100%'` through `processBpkProps` and asserts they pass through unchanged.

packages/backpack-web/src/bpk-component-layout/src/commonProps.ts
Supremeyh: Good point — changed the type from `BpkSpacingValue` to `BpkPositionValue`.\n\n`BpkSpacingValue` only accepts spacing tokens or percentages, so `scrollMarginTop="3.5rem"` (the most common real use case: a sticky-header height that isn't a round spacing token) would be a TypeScript error. `BpkPositionValue` (`${number}rem` | `${number}%` | `'0'` | `BpkSpacingToken`) covers all valid inputs.\n\nThis also unifies the processing path: `scrollMarginTop/Bottom` now sit alongside `insetInlineStart/End` in the position keys group, both routed through `convertBpkPositionValue`. Previously they fell through the `convertBpkSpacingToChakra` fallback branch which would emit a dev warning for raw rem inputs.\n\nKept both scalar (no `BpkResponsiveValue<>` wrapper) per the earlier discussion — scroll-margin is a fixed offset, not a per-breakpoint value.

packages/backpack-web/src/bpk-component-layout/src/BpkBox-test.tsx
```diff
@@ -490,4 +490,107 @@ describe('BpkBox', () => {
...
   });
 
+  describe('transform prop (BpkBox only)', () => {
+    it('renders with a CSS transform string', () => {
+      const { container } = render(
+        <BpkProvider>
+          <BpkBox transform="translateX(10px)">Transformed</BpkBox>
+        </BpkProvider>,
+      );
+      expect(container.querySelector('div')).toBeInTheDocument();
```
Faye-Xiao: Just noticed a small question -  this always passes regardless of whether the prop was actually wired up or processed. All unit tests use this way 😢 
Can we adjust it to direct style assertions, as it is more readable and less brittle?

Faye-Xiao (approved):
LGTM

Faye-Xiao:
Approved the percy build https://percy.io/09e69e44/web/backpack/builds/51592075/changed/2787683847?browser=chrome&browser_ids=73&currentBuildItemIndex=4&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_public&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280 as some cases are flaky test. The BpkBox change is what we expected

PR #4873, opened 2026-07-07
[BpkPressable][CLOV-1813] Add unstyled clickable primitives for Backpack interactions

Supremeyh:
## What

Introduces `BpkPressable` — a polymorphic, ref-forwarding primitive for making arbitrary content interactive without inheriting any Backpack visual styling (button chrome, link colour, card elevation):

```tsx
// Button-like: keyboard-accessible, role=button, disabled support
<BpkPressable onClick={fn}>
  <AnyContent />
</BpkPressable>

// Link-like: anchor semantics, href required, blank/rel handling
<BpkPressable as="a" href="/flights">
  <AnyContent />
</BpkPressable>
```

## Why

Multiple teams hit a recurring gap during Backpack adoption — they needed click or navigation behaviour on arbitrary content but had no supported primitive for it:

- `BpkButton` injects visual styles that require `all: unset` to override (misuse)
- `BpkCard` adds elevation/padding and doesn't accept `href`/`onClick` directly
- `BpkVessel` renders a `<div>`, losing button semantics and keyboard activation
- Native `<button>`/`<a>` are explicitly off-limits per Backpack constitution

## API

| Prop | Mode | Type | Default | Notes |
|---|---|---|---|---|
| `as` | both | `'button' \| 'a'` | `'button'` | Selects rendered element |
| `href` | `as="a"` | `string` | — | **Required** in anchor mode |
| `blank` | `as="a"` | `boolean` | `false` | Sets `target="_blank"` + appends `rel="noopener noreferrer"` |
| `disabled` | both | `boolean` | `false` | Button: native disabled · Anchor: `aria-disabled` + prevents navigation |
| `ref` | both | `Ref<HTMLButtonElement \| HTMLAnchorElement>` | — | Forwarded via `forwardRef` |

All other standard `<button>` / `<a>` HTML attributes are forwarded via `...rest`. `className` and `style` are intentionally omitted (Backpack constitution §9 — components own their surface).

## Key design decisions

- **Single polymorphic component** over two separate ones — aligns with `BpkLink`'s `as` pattern, matches what teams expected from Confluence hard-case inventory
- **`blank` always appends `noopener noreferrer`** — merged with consumer `rel` if provided (e.g. `rel="sponsored"` + `blank` → `"sponsored noopener noreferrer"`), never silently dropped
- **`disabled` on anchor** — `aria-disabled="true"` + `e.preventDefault()` is the correct web pattern (native `disabled` attribute is not valid on `<a>`)
- **No `display` override in SCSS** — preserves browser defaults; consumers set layout via wrapper components
- **`forwardRef`** — unblocks floating-UI anchors, DOM measurement, and focus management
- **`{...rest}` before `className`** — runtime matches the TypeScript contract; `className` and `data-backpack-ds-component` always win
- **Naming: `BpkPressable`** — correct English, input-agnostic (covers mouse/touch/keyboard), aligns with React Native's `Pressable` convention

## Snapshots

image

## Storybook

Preview: https://backpack.github.io/storybook-prs/4873

Stories: Button mode · Anchor mode · Visual test · Visual test with zoom

---

- [x] PR title includes component name
- [x] `README.md` created for new component
- [x] Tests (30 unit tests)
- [x] Accessibility tests (5 jest-axe tests)
    - [x] Keyboard navigation: `type="button"` default, `:focus-visible` ring, Enter/Space activation
    - [x] Zoom: no fixed sizes, inherits font/layout from parent
    - [x] Screen reader: correct `role="button"` / `role="link"` semantics, `aria-disabled` for disabled anchor
- [x] Storybook examples created

[CLOV-1813]: https://skyscanner.atlassian.net/browse/CLOV-1813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.tsx
```diff
@@ -0,0 +1,90 @@
...
+    return (
+      <a
+        ref={ref as Ref<HTMLAnchorElement>}
+        href={href}
+        rel={resolvedRel}
+        target={target}
+        className={sharedClass}
+        {...getDataComponentAttribute('Pressable')}
+        {...rest}
+      >
```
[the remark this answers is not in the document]
Supremeyh: Fixed. `target` is now explicitly destructured from props and computed separately: `blank=true` always sets `_blank` via `resolvedTarget`, preventing consumer override. For `rel`: `rel || 'noopener noreferrer'` correctly treats an empty string as falsy, so `rel=""` safely falls back to `noopener noreferrer` — already secure.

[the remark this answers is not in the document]
Supremeyh: Fixed. `className` and `style` are now `Omit`-ted from both `AnchorProps` and `ButtonProps`, so consumers cannot pass them at the type level. This aligns with the Backpack constitution (components own their surface).

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  return (
+    <button
+      ref={ref as Ref<HTMLButtonElement>}
+      type="button"
+      disabled={disabled}
+      className={sharedClass}
+      {...getDataComponentAttribute('Pressable')}
+      {...rest}
+    >
```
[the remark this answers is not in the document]
Supremeyh: Fixed. Same as the anchor mode — `className` and `style` are `Omit`-ted from `ButtonProps`. The base `bpk-pressable` class is always applied and cannot be overridden.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  color: inherit;
+  text-align: inherit;
+  text-decoration: none;
+  cursor: pointer;
+  appearance: none;
+
+  &:disabled,
+  &[aria-disabled='true'] {
+    cursor: not-allowed;
+  }
```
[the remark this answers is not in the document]
Supremeyh: Fixed. `disabled` is now a first-class prop on `AnchorProps`. When `disabled=true`: `href` is removed (no navigation), `aria-disabled="true"` is set, `role="link"` restores the semantic, `tabIndex={-1}` removes from tab order, and `onClick` is suppressed (does not bubble). The SCSS `[aria-disabled='true']` selector is no longer orphaned.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+      <BpkPressable onClick={action('badge pressed')}>
+        <BpkBadge>Track prices</BpkBadge>
+      </BpkPressable>
+
+      <BpkPressable onClick={action('icon pressed')} aria-label="Favourite">
+        <SmallHeartIcon />
+      </BpkPressable>
+
+      <BpkPressable onClick={action('icon+text pressed')}>
+        <span style={{ display: 'flex', alignItems: 'center', gap: '0.25rem' }}>
```
Faye-Xiao: Nit: 0.25rem --> `BpkSpacing.XS`.
By the way, why do we have a raw span in this example? 🤔 didn't get the pont.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.tsx
```diff
@@ -0,0 +1,131 @@
...
+      tabIndex,
+      target,
+      ...rest
+    } = props;
+
+    const resolvedAriaDisabled = disabled ? true : ariaDisabled;
+    // href removed when disabled so the browser cannot navigate;
+    // role="link" restores the semantic lost when <a> has no href.
+    const resolvedHref = disabled ? undefined : href;
+    const resolvedRel = blank ? rel || 'noopener noreferrer' : rel;
```
Faye-Xiao: If a consumer accidentally passes ·rel=""· with `blank`, the `noopener noreferrer` default is silently dropped. The test even codifies this behaviour ("uses custom rel when provided with blank=true"). The PR description says "blank wins over raw target" but applies the opposite logic for rel. I'd flip the priority — either always enforce noopener noreferrer when `blank=true`, or merge it:

```
const resolvedRel = blank
  ? [rel, 'noopener noreferrer'].filter(Boolean).join(' ')
  : rel;
```

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+export const VisualTest = {
+  render: () => <MixedExample />,
+};
+
+export const VisualTestWithZoom = {
+  render: () => <MixedExample />,
+  args: {
+    zoomEnabled: true,
+  },
+};
```
Faye-Xiao, line 142: PR description lists "Stories: Button mode · Icon + text content · Anchor mode · When to use · Visual test" — but there's no WhenToUse export in the stories file. Either add it or remove it from the PR description.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.tsx
```diff
@@ -0,0 +1,131 @@
...
+        href={resolvedHref}
+        rel={resolvedRel}
+        role={resolvedRole}
+        tabIndex={resolvedTabIndex}
+        target={resolvedTarget}
+        aria-disabled={resolvedAriaDisabled}
+        className={sharedClass}
+        onClick={handleClick}
+        {...getDataComponentAttribute('Pressable')}
+        {...rest}
```
Faye-Xiao, line 106: Nit:   `{...rest}` is spread after `className={sharedClass}`, so JSX's last-wins rule means a JS consumer (or anyone using `as any`) can silently replace the base reset and focus ring by passing `className`. The TypeScript `Omit` only catches this at compile time.

Trivial fix — just move `{...rest}` before the protected attributes:
```
<a
  {...rest}
  href={resolvedHref}
  className={sharedClass}
  onClick={handleClick}
  {...getDataComponentAttribute('Pressable')}
>
```
Same change for the `<button>` branch. No behaviour change for TypeScript consumers, just makes the runtime match the type contract.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.stories.tsx
Supremeyh: Good spot on the span! The reason it's a raw `<span>` rather than `BpkFlex` is that `BpkFlex` (and all Backpack layout components) render a `<div>`, and HTML5 does not allow a `<div>` inside a `<button>`. Browsers would silently close the `<button>` before the `<div>`, breaking both layout and click area. The layout guidelines explicitly call this out: *"use `<span style={{ display: 'flex', ... }}>` for flex rows inside a button element."*

On `BpkSpacing.XS`: `BpkSpacing.XS` resolves to the string `'bpk-spacing-xs'`, not a CSS value. It's designed as a prop for Backpack layout components (`<BpkFlex gap={BpkSpacing.XS}>`), where the component maps it to the right CSS. Using it directly in `style={{ gap: BpkSpacing.XS }}` would produce `gap: bpk-spacing-xs` in the DOM — invalid CSS. Keeping `0.25rem` for this inline style.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.tsx
Supremeyh: You're right — good catch. Changed to always append `noopener noreferrer` when `blank=true`:

```ts
const resolvedRel = blank
  ? [rel, 'noopener noreferrer'].filter(Boolean).join(' ')
  : rel;
```

So `rel="sponsored"` + `blank` → `"sponsored noopener noreferrer"`. Updated the test name and expected value accordingly.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.stories.tsx
Supremeyh: Fixed — updated the PR description to match the current stories: Button mode · Anchor mode · Visual test · Visual test with zoom.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.tsx
Supremeyh: Fixed. Moved `{...rest}` before `className` and `getDataComponentAttribute` in both branches so the protected attributes always win at runtime, matching the TypeScript contract.

One nuance: `type` in the button branch is intentionally left overridable (consumer can pass `type="submit"`), so `{...rest}` sits between our `type="button"` default and the protected attrs — not fully at the top.

packages/backpack-web/src/bpk-component-pressable/src/BpkPressable.stories.tsx
Supremeyh: update: replace it with AlignedSearchIcon so we can not use span

Faye-Xiao (approved):

Faye-Xiao:
Approved the precy build as the failre came from the new cases and some flaky test https://percy.io/09e69e44/web/backpack/builds/51623275/changed/2789405161?browser=chrome&browser_ids=73&currentBuildItemIndex=1&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_public&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

PR #4886, opened 2026-07-13
[WOODPECKER-4447] Upgrade pnpm from 9 to 10

viktor-yang:
## Summary

Bump `packageManager` and `engines.pnpm` from `9.15.9` → `10.33.0`

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -4,3 +4,7 @@ engine-strict=true
 # Dependabot is not compatible with pnpm strict peer dependency installs.
 # Keep installs unblocked and enforce peer dependency health via a separate CI check.
 strict-peer-dependencies=false
+
+public-hoist-pattern[]=*eslint*
+public-hoist-pattern[]=*stylelint*
+public-hoist-pattern[]=*prettier*
--- a/package.json
+++ b/package.json
@@ -3,10 +3,10 @@
   "version": "0.0.1",
   "license": "Apache-2.0",
   "private": true,
-  "packageManager": "pnpm@9.15.9",
+  "packageManager": "pnpm@10.33.0",
   "engines": {
     "node": ">=22.9.0",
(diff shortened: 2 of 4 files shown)


.npmrc
```diff
@@ -4,3 +4,7 @@ engine-strict=true
 # Dependabot is not compatible with pnpm strict peer dependency installs.
 # Keep installs unblocked and enforce peer dependency health via a separate CI check.
 strict-peer-dependencies=false
+
+public-hoist-pattern[]=*eslint*
+public-hoist-pattern[]=*stylelint*
+public-hoist-pattern[]=*prettier*
```
viktor-yang, line 10: From [release document](https://github.com/pnpm/pnpm/releases/tag/v10.0.0):
public-hoist-pattern: nothing is hoisted by default. Packages containing eslint or prettier in their name are no longer hoisted to the root of node_modules. Related Issue: https://github.com/pnpm/pnpm/issues/8378

package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.0.1",
   "license": "Apache-2.0",
   "private": true,
-  "packageManager": "pnpm@9.15.9",
```
gc-skyscanner, line 6: don't we need to update package-lock.json? seems we locked the version 
https://github.com/Skyscanner/backpack/blob/d1d4e9d25105dbd6e009a39d4ce91bbc22cdc816/pnpm-lock.yaml#L1

viktor-yang: Thanks GC. I think no. According to [the spec](https://github.com/pnpm/spec/blob/master/lockfile/README.md), the lock version is 9 from `pnpm >= 9.0.0`

gc-skyscanner (approved):
LGTM

PR #4890, opened 2026-07-14
fix(ci): pin @figma/code-connect to 1.4.5 and scope PR validation to changed files

gc-skyscanner:
## Problem

The Code Connect CI job has been failing since 2026-07-09, when `@figma/code-connect@1.4.9` was released. The workflow was using `npx --package @figma/code-connect` without a version pin, so it picked up 1.4.9 at runtime.

**Root cause:** `@figma/code-connect@1.4.9` introduced a regression that causes a `TypeError: Cannot read properties of undefined (reading 'split')` when processing `BpkIcon.figma.tsx` (537 connect calls) and other files. This also blocks CI for unrelated PRs like #4849 that don't touch any broken files.

The last known stable version is **1.4.5** (confirmed working on main as of 2026-05-19).

## Fix

1. **Pin to `@figma/code-connect@1.4.5`** in both the dry-run and publish steps.
2. **Scope PR validation to changed files only** — instead of scanning the entire project (which hits pre-existing broken files like `BpkTicket`, `BpkCode`, `BpkCodeBlock`, `BpkJourneyArrow`), compute which `.figma.tsx` files changed in the PR and pass only those via `--file`.

## Why this unblocks #4849

PR #4849 only modifies `BpkPrice.figma.tsx`, which processes cleanly in CI. By scoping validation to changed files and pinning to 1.4.5, the CI will only validate that one file and succeed.

## Pre-existing issues (not fixed here)

The following files have stale Figma node IDs and will still fail when `figma connect publish` runs on main pushes — these need separate fixes:
- `BpkTicket.figma.tsx` — node `10911:51477` not found
- `BpkJourneyArrow.figma.tsx` — node `10885:8553` is not a top-level component
- `BpkCodeBlock.figma.tsx` — node `10858:59644` not found  
- `BpkCode.figma.tsx` — node `10858:59646` not found

### DIFF
--- a/.github/workflows/sync-figma-code-connect.yml
+++ b/.github/workflows/sync-figma-code-connect.yml
@@ -22,11 +22,16 @@ jobs:
           persist-credentials: false
       - name: Validate Code Connect (dry run)
         if: github.event_name == 'pull_request'
-        run: npx --package @figma/code-connect figma connect publish --dry-run
+        run: |
+          git fetch origin ${{ github.base_ref }} --depth=1
+          mapfile -t FILES < <(git diff --name-only origin/${{ github.base_ref }}...HEAD -- '*.figma.tsx')
+          if [ ${#FILES[@]} -gt 0 ]; then
+            npx --package @figma/code-connect@1.4.5 figma connect publish --dry-run --file "${FILES[@]}"
+          fi
         env:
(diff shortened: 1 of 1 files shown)


gc-skyscanner:
Superseded by the commit directly added to #4849.

PR #4896, opened 2026-07-14
[STAR-352] [BpkBadge] interactive variant

vivitt:
## Description
Add an new interactive variant to BpkBadge, using a polymorphic approach similar to BpkPressable. 
The badge renders a non-interactive span by default, but it can also be rendered as a button or a link using the new `as` prop.
Interactive badges render a trailing info-circle icon (see the Figma design linked below).
Screenshot 2026-07-15 at 16 40 55

### Notes
The interactive badge variants (as="button", as="a") are follow the pattern used in BpkPressable with two intentional simplifications:
- disabled is omitted from both variants, a disabled interactive badge is semantically a non-interactive badge so the default span variant should be used instead
- forwardRef is not used, interactive variants render as native <button> or <a> elements, which handle focus management natively without needing a DOM ref

## What changed:
- New `as` prop: 'span' (default, unchanged), 'button', or 'a'
- Interactive variants render a trailing info circle icon and apply bpk-badge--interactive styles (flex layout, focus ring via bpk-focus-indicator, cursor pointer)
- as="a" supports a blank convenience prop that sets target="_blank" and rel="noopener noreferrer" automatically
- Props are now a strict discriminated union, replacing the previous [rest: string]: any index signature. The span variant extends HTMLAttributes<HTMLSpanElement> so standard HTML attributes (aria-label, data-testid, etc.) are fully typed
- snapshot and axe tests added for both interactive variants.

### Useful links
[BpkBadge Interactive ticket](https://skyscanner.atlassian.net/jira/polaris/projects/BACKPACK/ideas/view/6690793?selectedIssue=BACKPACK-164)

[figma](https://www.figma.com/design/KXf2gHNLDe2cXWUoHl4cTX/branch/2SwQCUUg4jbxyoDq08FAC8/Backpack%E2%80%A8Foundations---Components?node-id=22394-5166&m=dev)

[Interstellar jira ticket](https://skyscanner.atlassian.net/browse/STAR-352)

### Interactive badge screenshot
example of usage in combined explore deal-badges
Screenshot 2026-07-15 at 16 37 36

- [ ] Component `README.md`
- [ ] Tests
- [ ] Accessibility tests


packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
```diff
@@ -50,38 +50,128 @@ const badgeTypeClassNames = {
 
 export type BadgeType = (typeof BADGE_TYPES)[keyof typeof BADGE_TYPES];
 
-export type Props = {
+type SpanProps = Omit<HTMLAttributes<HTMLSpanElement>, 'className' | 'style'> & {
+  as?: 'span';
+};
+
+type ButtonProps = Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'className' | 'style' | 'type' | 'disabled'> & {
```
vivitt, line 57: The interactive badge variants (as="button" and as="a") follow the pattern in BpkPressable, but unlike it they intentionally omit the disabled prop. The rationale behind this decision is that a disabled interactive badge is semantically a non-interactive badge, so the span variant shopuld be used instead. Also, there are no design specs for a disabled interactive state.
Based on the above, `disabled` is explicitly removed from both ButtonProps and AnchorProps via Omit so this is enforced at the type level rather than silently ignored.
Happy to revisit if there's a use case I'm missing.

packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
```diff
@@ -50,38 +50,128 @@ const badgeTypeClassNames = {
...
+  {
+    as = 'span',
+    centered = false,
+    children,
+    className = null,
+    docked = null,
+    type = BADGE_TYPES.normal,
+    ...rest
+  }: Props,
+  ref: Ref<HTMLButtonElement | HTMLAnchorElement >,
```
vivitt: ~~forwardRef is used so callers can receive a ref to the underlying DOM node for the interactive variants and can use it for programatic focus, or as a tooltip anchor. The span variant omits ref forwarding intentionally, since no interaction is expected on it.~~

packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
```diff
@@ -50,38 +50,128 @@ const badgeTypeClassNames = {
...
-  type = BADGE_TYPES.normal,
-  ...rest
-}: Props) => {
+export type Props = CommonProps & (
+  | SpanProps
+  | ButtonProps
+  | AnchorProps
+);
+
+const BadgeInfoCircleIcon = () => (
```
vivitt: Note that this is added instead of using BpkIcon because, according to the design specs, the icon is slightly different in size and shape, and this is an info logo rather than the Backpack icon. See https://skyscanner.atlassian.net/jira/polaris/projects/BACKPACK/ideas/view/6690793?selectedIssue=BACKPACK-164 for more details

packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
vivitt: Decided to keep it simple and remove the ref forwarding unless we see a clear need for it. Is there a particular reason BpkPressable forwards refs that we should consider here? 🙂

packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
```diff
@@ -50,37 +49,123 @@ const badgeTypeClassNames = {
 
 export type BadgeType = (typeof BADGE_TYPES)[keyof typeof BADGE_TYPES];
 
-export type Props = {
+type SpanProps = Omit<HTMLAttributes<HTMLSpanElement>, 'className' > & {
+  as?: 'span';
+  [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
```
vivitt, line 54: keeping `[rest: string]: any;` to avoid breaking existing users that pass arbitrary props

packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
```diff
@@ -50,37 +49,123 @@ const badgeTypeClassNames = {
...
-  type = BADGE_TYPES.normal,
-  ...rest
-}: Props) => {
+export type Props = CommonProps & (
+  | SpanProps
+  | ButtonProps
+  | AnchorProps
+  );
+
+const BadgeInfoCircleLogo = () => (
```
vivitt, line 82: Note that this is added instead of using BpkIcon because, according to the design specs, the icon is slightly different in size and shape, and this is an info logo rather than the Backpack icon. See [skyscanner.atlassian.net/jira/polaris/projects/BACKPACK/ideas/view/6690793?selectedIssue=BACKPACK-164](https://skyscanner.atlassian.net/jira/polaris/projects/BACKPACK/ideas/view/6690793?selectedIssue=BACKPACK-164) for more details

packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
```diff
@@ -50,37 +49,123 @@ const badgeTypeClassNames = {
...
+      <BadgeInfoCircleLogo />
+    </a>);
+  }
+
+  if (as === 'button') {
+    const buttonRest = rest as Omit<ButtonProps, 'as' | 'children'>;
+    return (
+      <button
+        type="button"
+        {...buttonRest}
```
gc-skyscanner: `type="button"` is set first, then `{...buttonRest}` is spread after it. `ButtonProps` omits type at the TypeScript level, but the cast rest as `Omit<ButtonProps, 'as' | 'children'>` provides no runtime enforcement — a JS caller or as any cast can pass `type="submit" ` and it will silently override the fixed value, causing the badge to submit a form. is it expected?

packages/backpack-web/src/bpk-component-badge/src/accessibility-test.tsx
```diff
@@ -33,4 +33,16 @@ describe('BpkBadge accessibility tests', () => {
     const results = await axe(container);
     expect(results).toHaveNoViolations();
   });
+
```
gc-skyscanner, line 36: The `blank` convenience prop (which auto-sets `target="_blank"` and `rel="noopener noreferrer"`) is a non-trivial piece of logic but has no dedicated test. There is no test that: (a) verifies `blank=true` renders `target="_blank" rel="noopener noreferrer"`, (b) verifies consumer-supplied `rel` is merged correctly (`rel="nofollow"` + `blank=true` → `"nofollow noopener noreferrer"`), or (c) confirms `blank` itself does not appear as a DOM attribute. A snapshot test using `<BpkBadge as="a" href="#" blank>` would catch regressions but only covers case (a) — explicit assertions on the rendered `rel` and `target` attributes are needed.

packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
```diff
@@ -15,8 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-import type { ReactNode } from 'react';
+import type { ReactNode, Ref, AnchorHTMLAttributes, ButtonHTMLAttributes, HTMLAttributes } from 'react';
```
gc-skyscanner: Seems `Ref` is unused. did i miss something🤔

vivitt: removed [`24cdb2f`](https://github.com/Skyscanner/backpack/pull/4896/changes/24cdb2f6aa9f5d1ed442bbe136118290922559a9)

packages/backpack-web/src/bpk-component-badge/src/accessibility-test.tsx
vivitt: added coverage for blank prop login [`24cdb2f`](https://github.com/Skyscanner/backpack/pull/4896/changes/24cdb2f6aa9f5d1ed442bbe136118290922559a9) : 
- no blank means no target or rel
- blank=true sets target="_blank" and rel="noopener noreferrer"
- blank doesn't appear as a DOM attribute
- consumer rel is merged correctly

packages/backpack-web/src/bpk-component-badge/src/BpkBadge.tsx
vivitt: moved type="button" after the spread so it always overwrites whatever comes from caller

gc-skyscanner (dismissed):

gc-skyscanner (dismissed):

gc-skyscanner (approved):

PR #4908, opened 2026-07-17
Render component README in every Storybook docs page

gert-janvercauteren:
## What

Adds `parameters.docs.page` to all 112 component story files (`.tsx` and `.js`) so that each component's `README.md` is rendered inline in the Storybook **Docs** tab.

Removes the duplicate `<Title />` block — the README already contains the component heading, so rendering both was redundant.

## How

Each story now imports its `README.md` and renders it via `<Markdown>{readme}</Markdown>` between the page title and the arg types table:

```tsx
parameters: {
  docs: {
    page: () => (
      <>
        <Markdown>{readme}</Markdown>
        <ArgTypes exclude={['zoomEnabled']} />
      </>
    ),
  },
},
```

Nested V2/V3 components (e.g. `BpkModalV2`, `BpkCheckboxV2`, `BpkComparisonTable`) use their own local `README.md` where one exists, falling back to the top-level component README otherwise.

## Test plan

- [ ] Open Storybook and navigate to any component's **Docs** tab — the README content should now appear above the arg types table
- [ ] Confirm no duplicate title (the README heading replaces the old `<Title />`)


packages/backpack-web/src/bpk-component-image/src/BpkImage.stories.js
```diff
@@ -181,13 +183,10 @@ const meta = {
     docs: {
       page: () => (
         <>
-          <Title />
+          <Markdown>{readme}</Markdown>
           <ArgTypes exclude={['zoomEnabled']} />
-          <Markdown>
-            {`**Note:** All [standard img attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#Attributes) including \`srcSet\` are also supported.`}
```
gc-skyscanner, line 187: We removed this, but seems it doesn't exist in the readme, is it ok? or ,could we keep it?(we keep the original markdown in `aria-live`)

packages/backpack-web/src/bpk-component-input/src/BpkInput.stories.tsx
```diff
@@ -444,13 +446,9 @@ const meta = {
     docs: {
       page: () => (
         <>
-          <Title />
+
+          <Markdown>{readme}</Markdown>
           <ArgTypes exclude={['zoomEnabled']} />
-          <Markdown>
-            {
-            `**Note:** Additionally, all native \`input\` attributes such as \`placeholder\` and \`onChange\` are supported.`
```
gc-skyscanner, line 451: same here https://github.com/Skyscanner/backpack/pull/4908/changes#r3602217969,  we told consumers about the HTML-native prop passthrough. Before merging, verify these notes exist in the respective README.md files; if they don’t, add them there. If they’re silently lost, consumers searching the Docs tab will no longer find this information. (seems we are keeping the original markdown in `aria-live`)

.github/workflows/release.yml
```diff
@@ -107,3 +138,16 @@ jobs:
...
           SUPERNOVA_API_KEY: ${{ secrets.SUPERNOVA_API_KEY }}
           DESIGN_SYSTEM_ID: ${{ secrets.SUPERNOVA_DESIGN_SYSTEM_ID }}
+
+      - name: Import Storybook to Supernova
+        run: |
+          npx @supernovaio/cli storybook-import \
+            --from ./libs/backpack-storybook-host/dist-storybook \
+            --designSystemId $DESIGN_SYSTEM_ID \
+            --brandId 80f68ac7-b11d-4c97-a4dc-f01dc458da83 \
+            --sourceId f88b7fa5-26af-44b7-a8dc-87593f336c3c \
```
gc-skyscanner: `--brandId 80f68ac7-b11d-4c97-a4dc-f01dc458da83` and `--sourceId f88b7fa5-26af-44b7-a8dc-87593f336c3c` are embedded as plaintext in the workflow. While these are not authentication secrets, they are environment-specific identifiers that, if they ever change (brand rename, source migration), require a code change rather than a config change. Prefer storing them as repo secrets or environment variables (e.g. `${{ secrets.SUPERNOVA_BRAND_ID }}`), consistent with how `DESIGN_SYSTEM_ID` is already handled

packages/backpack-web/src/bpk-component-accordion/src/BpkAccordion.stories.tsx
```diff
@@ -307,6 +311,17 @@ const meta = {
   subcomponents: {
     BpkAccordionItem,
   },
+  parameters: {
+    docs: {
+      page: () => (
+        <>
+
```
gc-skyscanner: nit: black line in all  112 story page

packages/backpack-web/src/bpk-component-input/src/BpkInput.stories.tsx
gert-janvercauteren: Added to README

packages/backpack-web/src/bpk-component-image/src/BpkImage.stories.js
gert-janvercauteren: Added to README

gc-skyscanner (approved):

PR #4910, opened 2026-07-20
[BD-11077][BpkChipGroup] add renderChip render prop for popover anchoring

LuoTuxiu:
## Summary

- Wrap `BpkSelectableChip` and `BpkDropdownChip` in `forwardRef` so `BpkPopover` can obtain the chip's `<button>` DOM node for anchoring
- Add `renderChip?: (props: ChipRenderProps) => ReactElement | null` to `ChipItem` in `BpkMultiSelectChipGroup` — lets callers replace per-chip rendering (e.g. wrapping in `BpkPopover`) while the group still provides `selected`, `chipStyle`, `accessibilityLabel`, `onClick`, and `index`
- Export `ChipRenderProps` from the package barrel; add a `WithPopover` Storybook story demonstrating per-chip popover anchoring; remove the old standalone popover demo

new

## Test plan

- [x] `pnpm exec jest packages/backpack-web/src/bpk-component-chip packages/backpack-web/src/bpk-component-chip-group` — 103 tests, all pass
- [x] `pnpm tsc -b packages/backpack-web/tsconfig.lib.json` — no errors
- [x] Storybook `WithPopover` story: click "Flights" → popover anchors below Flights chip; click "Trains" → popover anchors below Trains chip (per-chip tracking verified)
- [x] Pre-existing `on-image` story renders correctly (no regression)


packages/backpack-web/src/bpk-component-chip-group/src/BpkChipGroupPopoverExample.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+            id="chip-group-popover"
+            isOpen={activeLabel !== null}
+            label={activeLabel ?? ''}
+            labelAsTitle
+            closeButtonLabel="Close"
+            onClose={() => setActiveLabel(null)}
+            target={<div className={getClassName('bpk-chip-group-popover-example__anchor')} />}
+            placement="bottom-start"
+            showArrow={false}
+          >
```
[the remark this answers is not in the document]
LuoTuxiu: Updated.

packages/backpack-web/src/bpk-component-chip-group/src/BpkChipGroupPopoverExample.stories.module.scss
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  &__image-bg {
+    padding: tokens.bpk-spacing-base();
+    background-image: url('https://content.skyscnr.com/96508dbac15a2895b0147dc7e7f9ad30/canadian-rockies-canada.jpg');
+    background-size: cover;
+    background-position: center;
+    border-radius: tokens.$bpk-border-radius-md;
+    margin-bottom: tokens.bpk-spacing-base();
+  }
```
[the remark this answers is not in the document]
LuoTuxiu: Removed.

packages/backpack-web/src/bpk-component-chip-group/src/BpkChipGroupPopoverExample.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+            BpkPopover's target must be a React element it can clone and attach
+            a ref to. We use a zero-size div as the anchor — it sits at the
+            start of the rail. Because BpkMultiSelectChipGroup renders chips
+            inside a scroll container and chips don't expose refs, this group-
+            level anchor is the only viable positioning point. The popover
+            always appears at the left edge of the group, not below the clicked chip.
+          */}
+          <BpkPopover
+            key={activeLabel ?? 'closed'}
+            id="chip-group-popover"
```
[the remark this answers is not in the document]
LuoTuxiu: Removed.

packages/backpack-web/src/bpk-component-chip-group/src/BpkChipGroupPopoverExample.stories.module.scss
```diff
@@ -0,0 +1,77 @@
...
+
+  &__sentinel {
+    // Zero-size invisible anchor at the left edge of the rail.
+    // BpkPopover attaches its positioning ref here — the popover will always
+    // appear at this position, never at the individual chip's position.
+    display: inline-block;
+    width: 0;
+    height: tokens.bpk-spacing-xl();
+    vertical-align: middle;
+  }
```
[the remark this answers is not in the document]
LuoTuxiu: Removed.

packages/backpack-web/src/bpk-component-chip-group/src/BpkMultiSelectChipGroup.tsx
```diff
@@ -113,21 +122,39 @@ const Chip = ({
...
+    }
+  };
+
+  if (renderChip) {
+    return renderChip({
+      selected: selected ?? false,
+      chipStyle,
+      accessibilityLabel: accessibilityLabel || text,
+      onClick: handleClick,
+      index: chipIndex,
```
jasonhuang-dev: It's best to pass all chip props to the `renderChip` function to prevent omissions. You can define a variable to ensure that the props passed to `CHIP_COMPONENT_MAP[component]` and `renderChip` are the same.

```
const chipProps = {
      selected: selected ?? false,
      type: chipStyle,
      accessibilityLabel: accessibilityLabel || text,
      onClick: handleClick,
      role: ariaMultiselectable ? 'checkbox' : 'radio'
      leadingAccessoryView
      ...rest
}

if (renderChip) {
     return renderChip(chipProps);
}

const Component = CHIP_COMPONENT_MAP[component];

return (
     <Component {...chipProps}>{text}</Component>
)
```

packages/backpack-web/src/bpk-component-chip/src/BpkDropdownChip.tsx
```diff
@@ -29,14 +31,15 @@ import { CHIP_TYPES } from './commonTypes';
...
+const BpkDropdownChip = forwardRef<HTMLButtonElement, Props>(({
   disabled = false,
   leadingAccessoryView = null,
   selected = false,
   type = CHIP_TYPES.default,
   ...rest
-}: Props) => (
+}, ref) => (
   <BpkSelectableChip
+    ref={ref}
```
jasonhuang-dev: any reason to add ref here?

packages/backpack-web/src/bpk-component-chip-group/src/BpkMultiSelectChipGroup.tsx
```diff
@@ -69,8 +69,17 @@ export type SingleSelectChipItem = {
   [rest: string]: any; // Inexact rest. See decisions/inexact-rest.md
 };
 
+export type ChipRenderProps = {
```
wpe1arhy: **`renderChip` drops the group's a11y role.** The default path sets `role={ariaMultiselectable ? 'checkbox' : 'radio'}`, but the `renderChip` path passes no role and `ChipRenderProps` doesn't expose one — so callers can't reproduce it. The container is still `radiogroup`/`group`, so its children need the matching role. Worst in single-select (`ariaMultiselectable={false}` → `radiogroup`) where a custom chip stays non-`radio`.

Suggest adding `role: 'checkbox' | 'radio'` to `ChipRenderProps` (derived from `ariaMultiselectable`) + a single-select test asserting `role="radio"`.

packages/backpack-web/src/bpk-component-chip/src/BpkDropdownChip.tsx
LuoTuxiu: Popover need a ref to control the target. But i found we can use `BpkVessel`. So i revert the change at this file.

packages/backpack-web/src/bpk-component-chip-group/src/BpkMultiSelectChipGroup.tsx
LuoTuxiu: Good catch! Added `role: 'checkbox' | 'radio'` to `ChipRenderProps` (derived from `ariaMultiselectable`), along with a test asserting `role="radio"` in single-select context. Fixed in da9aca35e.

packages/backpack-web/src/bpk-component-chip-group/src/BpkMultiSelectChipGroup.tsx
LuoTuxiu: Good suggestion — extracted a shared `chipRenderProps` variable that both the `renderChip` path and the default `Component` path use, so future additions to `ChipRenderProps` automatically apply to both. Fixed in da9aca35e.

wpe1arhy (dismissed):
LGTM

jasonhuang-dev (dismissed):

packages/backpack-web/src/bpk-component-chip-group/src/BpkMultiSelectChipGroup.tsx
```diff
@@ -71,11 +71,20 @@ export type SingleSelectChipItem = {
 
 export type ChipItem = {
   component?: ChipComponentType;
+  renderChip?: (props: ChipRenderProps) => ReactElement | null;
   onClick?: (selected: boolean, index: number) => void;
   selected?: boolean;
   hidden?: boolean;
 } & SingleSelectChipItem;
 
+export type ChipRenderProps = Omit<ChipItem, 'component' | 'renderChip' | 'onClick' | 'accessibilityLabel' | 'leadingAccessoryView'> & {
```
wpe1arhy: `ChipRenderProps` now derives from `Omit<ChipItem, ...>`, but `ChipItem` carries the `[rest: string]: any` index signature via `SingleSelectChipItem`, and `Omit` preserves that index signature. As a result `ChipRenderProps` effectively accepts any key (all resolved as `any`), so it provides almost no type safety — the story below references `chipStyle`/`index`, which no longer exist on this type, yet tsc doesn't catch it for exactly this reason. Suggest making this a standalone type with only the explicit fields so `renderChip` consumers get real type checking.

packages/backpack-web/src/bpk-component-chip-group/src/BpkChipGroup.stories.tsx
```diff
@@ -447,6 +449,68 @@ export const ExampleStateManagement = {
   render: () => <StateManagement />,
 };
 
+const POPOVER_CHIP_LABELS = ['Flights', 'Hotels', 'Car hire', 'Trains'];
+
+const makePopoverRenderChip =
+  (label: string) =>
+  ({ accessibilityLabel, chipStyle, index, onClick, selected }: ChipRenderProps) =>
```
wpe1arhy: This destructures `chipStyle` and `index`, but the refactored `ChipRenderProps` only provides `type` — `chipStyle`/`index` no longer exist (only the inexact rest signature lets this pass tsc). At runtime: `index` is `undefined`, so all four popovers get `id="popover-chip-undefined"` (line 459) → duplicate DOM ids; and `chipStyle` is `undefined`, so `BpkDropdownChip` (line 471) falls back to the default type and ignores the group's `chipStyle`. Since this story is the demo for anchoring a popover to a specific chip, it needs to switch to `type` and use a stable id source (either add `index` back to `ChipRenderProps`, or derive the id from `text`/label).

packages/backpack-web/src/bpk-component-chip-group/src/BpkMultiSelectChipGroup.tsx
```diff
@@ -113,25 +122,40 @@ const Chip = ({
...
-        if (onClick) {
-          onClick(!selected, chipIndex);
-        }
-      }}
-      role={ariaMultiselectable ? 'checkbox' : 'radio'}
-      leadingAccessoryView={leadingAccessoryView}
-      {...rest}
-    >
+  return (
+    <Component {...chipRenderProps}>
```
wpe1arhy: **(Optional)** `{...chipRenderProps}` also spreads `text`, which `BpkSelectableChip` doesn't destructure, so it leaks as an invalid `<button text="...">` attribute. Suggest excluding `text` here. Non-blocking — DOM noise only, no functional impact.

packages/backpack-web/src/bpk-component-chip-group/src/BpkMultiSelectChipGroup.tsx
LuoTuxiu: Good catch — introduced a `ChipBaseProps` type that holds the explicit fields (`text`, `accessibilityLabel`, `leadingAccessoryView`, `selected`, `hidden`). Both `SingleSelectChipItem` (which carries the index signature) and `ChipRenderProps` now extend `ChipBaseProps` independently, so `ChipRenderProps` gets the shared fields without inheriting `[rest: string]: any`. Fixed in 6461ffe95.

packages/backpack-web/src/bpk-component-chip-group/src/BpkChipGroup.stories.tsx
LuoTuxiu: Fixed — story now uses `type` (instead of `chipStyle`) and `label` as the stable id source (instead of `index`). Fixed in b07d71403.

packages/backpack-web/src/bpk-component-chip-group/src/BpkMultiSelectChipGroup.tsx
LuoTuxiu: Fixed — `text` is now destructured out before spreading `chipRenderProps` onto `Component`, so it no longer leaks as a DOM attribute. Fixed in b07d71403.

wpe1arhy (approved):
LGTM

PR #4897, opened 2026-07-14
[IRN-6860][BpkCollapsible] Add useBpkCollapsibleContext hook and asChild on RootProvider

steviehailey-skyscanner:
JIRA Link: https://skyscanner.atlassian.net/browse/IRN-6860

## What has changed, and why?

**Component:** `bpk-component-collapsible`

**Class of change:** new API surface, no behaviour change to existing usage

Adds two capabilities to `BpkCollapsible` that are required groundwork for the FLEXP-661 Expandables MVP fare-card integration.

**`useBpkCollapsibleContext()`** — a consumer-side reader hook. Any component nested inside `BpkCollapsible.Root` or `BpkCollapsible.RootProvider` can now call this hook to read the current machine state (`open`, `visible`, `disabled`, `setOpen`) without prop-drilling. Follows the same pattern as `useCollapsibleContext` from Ark internally, but exposed under a stable BPK name. Precedent: `useCheckboxCardContext` in `bpk-component-checkbox-card`.

**`asChild` on `RootProvider`** — when `true`, the provider merges its props onto the single child element instead of injecting a wrapper `<div>`. This is load-bearing for pixel parity in the fare-card layout, where the provider needs to wrap an existing item element without adding a DOM node between it and its parent. Precedent: `BpkModalV3Trigger` already ships `asChild`.

Both are demonstrated in new Storybook stories (`ContextReaderHook`, `AsChildOnRootProvider`).

---

Remember to include the following changes:

- [x] Ensure the PR title includes the name of the component you are changing so it's clear in the release notes for consumers of the changes in the version e.g `[Clover-123][BpkButton] Updating the colour`
- [ ] `README.md` (If you have created a new component)
- [ ] Component `README.md`
- [x] Tests
- [ ] Accessibility tests
    - The following checks were performed:
        - [ ] Ability to navigate using a [keyboard only](https://webaim.org/techniques/keyboard/)
        - [ ] Zoom functionality ([Deque University explanation](https://dequeuniversity.com/checklists/web/text)):
            - [ ] The page SHOULD be functional AND readable when only the text is magnified to 200% of its initial size
            - [ ] Pages must reflow as zoom increases up to 400% so that content continues to be presented in only one column i.e. Content MUST NOT require scrolling in two directions (both vertically and horizontally)
        - [ ] Ability to navigate using a [screen reader only](https://webaim.org/articles/screenreader_testing/)
- [x] Storybook examples created/updated
[...]

PR #4913, opened 2026-07-20
[BpkContextMenu] Context Menu Stories and Implementation

tranculent:
## Summary

Adds a new `BpkContextMenu` compound component wrapping Ark UI's `Menu` primitive, with accessible keyboard navigation, focus management, ARIA semantics, and floating positioning out of the box.

## For Backpack reviewers

**New component: `BpkContextMenu`**

Namespaced compound API:

| Part | Purpose |
|------|---------|
| `.Root` | State container — open state, keyboard navigation, selection dispatch. `unmountOnExit` removes the portal from the DOM after the close animation. |
| `.Trigger` | Generic trigger. Renders a plain `<button>` by default; `asChild` (requires `forwardRef` child) merges trigger props onto a custom element. |
| `.SaveTrigger` | Pre-styled circular heart-icon trigger for the save-to-list pattern. No consumer styling required. |
| `.Content` | Floating menu surface. Portalled, `overflow-x: hidden` so long names truncate with ellipsis. |
| `.Item` | Selectable row — `endIcon`, `destructive` variant, `onSelect` (fires for pointer **and** keyboard via Ark's state machine — not a DOM `onClick`). |
| `.ItemGroup` | Visual grouping container. Renders `role="group"` (plain `<div>` intentional — `Menu.ItemGroup` would emit a dangling `aria-labelledby` without a paired label). |
| `.Separator` | Horizontal rule between groups. |
| `.TriggerItem` | Row that opens a nested sub-menu on hover / arrow-right. |

**Architecture decisions**
- `BpkContextMenu.SaveTrigger` owns the heart-button visual entirely within this package — no dependency on `BpkSaveButton` as an Ark `asChild` trigger
- Focus ring suppression on close is handled by `useMenuTriggerFocusGuard` (shared hook on both `Trigger` and `SaveTrigger`): tracks pointer vs keyboard interactions via capture-phase `pointerdown`/`keydown` listeners; suppresses `:focus-visible` after pointer-initiated close, preserves it for Tab/Escape keyboard navigation; also checks `relatedTarget` when focus returns from inside `[role="menu"]`
- `unmountOnExit` on `Menu.Root` removes the content from the DOM after the close animation, preventing the invisible positioner from blocking pointer events
- Ghost hover on open (cursor physically over item when menu appears) suppressed with a `bpk-context-menu-item-hover-guard` keyframe animation — CSS animations take cascade precedence over `:hover`, so items stay transparent for the ~200ms open animation, then normal hover resumes
- All colour tokens migrated to CSS custom properties (`var(--bpk-surface-elevated, ...)`, `var(--bpk-text-primary, ...)`, etc.) for light/dark mode support
- `z-index` uses `tokens.$bpk-zindex-popover` (900), consistent with tooltips and popovers
- Bare `:hover` replaced with `@include utils.bpk-hover` to gate hover behind `.bpk-no-touch-support`

**`BpkSaveButton` — no functional changes**

The initial draft composed `BpkSaveButton` as an Ark `asChild` trigger, which required adding `forwardRef`, `onClick`, and focus-ring overrides. That approach was abandoned in favour of `BpkContextMenu.SaveTrigger`. `BpkSaveButton` is unchanged from its pre-PR state.

**ESLint**: `Menu.Trigger` and `Menu.TriggerItem` added to the `forbid-component-props` allowlist.

Storybook + unit + accessibility tests included. `ResizeObserver` stubbed in tests (Floating UI's `autoUpdate` uses it; JSDOM doesn't ship it). axe scans `document.body` to catch portalled content.

## Consumer usage

```tsx
// has lists → context menu
<BpkContextMenu.Root onSelect={({ value }) => handleSelect(value)}>
  <BpkContextMenu.SaveTrigger aria-label="Save to trip" />
  <BpkContextMenu.Content>
    <BpkContextMenu.ItemGroup>
      {trips.map(trip => (
        <BpkContextMenu.Item key={trip.id} value={trip.id}>
          {trip.name}
        </BpkContextMenu.Item>
      ))}
    </BpkContextMenu.ItemGroup>
    <BpkContextMenu.Separator />
    <BpkContextMenu.ItemGroup>
      <BpkContextMenu.Item value="new-trip" endIcon={<PlusIcon />}>
        Plan a new trip
      </BpkContextMenu.Item>
    </BpkContextMenu.ItemGroup>
  </BpkContextMenu.Content>
</BpkContextMenu.Root>

// no lists → plain BpkSaveButton, unchanged
<BpkSaveButton
  checked={isSaved}
  accessibilityLabel={isSaved ? 'Remove' : 'Save'}
  onCheckedChange={handleToggle}
/>
```

image

image

image

---

- [x] Component `README.md`
- [x] Tests
- [x] Accessibility tests
- [x] Storybook examples


packages/backpack-web/src/bpk-component-card-button/src/BpkSaveButton.tsx
```diff
@@ -55,21 +55,35 @@ type Props = {
...
+  hoverEffect?: boolean;
 };
 
 const AlignedHeartIcon = BpkHeartIcon;
 const AlignedHeartOutlineIcon = BpkHeartOutlineIcon;
 const AlignedHeartIconSm = BpkHeartIconSm;
 const AlignedHeartOutlineIconSm = BpkHeartOutlineIconSm;
 
-const BpkSaveButton = ({
+const BpkSaveButton = forwardRef<HTMLButtonElement, Props>(({
```
tranculent: Needed so that ArkUi can use BpkSaveButton as a trigger. It uses the <button> it renders as a reference - without it, we can't use `asChild` property of Ark UI's `Menu.Trigger` and would result in nested <buttons> which is invalid HTML. 

I looked through repos where BpkSaveButton is used and it looks like it's a safe change. I also went through BpkSaveButton's stories and they all work without any regressions.

packages/backpack-web/src/bpk-component-card-button/src/BpkSaveButton.tsx
```diff
@@ -88,25 +104,31 @@ const BpkSaveButton = ({
       )}
       {...getDataComponentAttribute('SaveButton')}
       onClick={(e: MouseEvent) => {
-        onCheckedChange(e);
-        if (!checked) {
-          setPlayAnim(true);
+        onClick?.(e);
```
tranculent: Ark's menu-toggle handler (opens/closes the menu)

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu-test.tsx
```diff
@@ -0,0 +1,215 @@
...
+  onSelect,
+  open = true,
+}: {
+  onSelect?: (details: { value: string }) => void;
+  open?: boolean;
+} = {}) => (
+  <BpkContextMenu.Root open={open} onSelect={onSelect}>
+    <BpkContextMenu.Trigger>
+      <button type="button">Open menu</button>
+    </BpkContextMenu.Trigger>
```
Supremeyh, line 49: [BpkContextMenu-test.tsx:33–34](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu-test.tsx#L33-L34) wraps a `<button>` inside `<BpkContextMenu.Trigger>` without `asChild` (same pattern repeated at lines 77–78 and 98–99):

```tsx
<BpkContextMenu.Trigger>
  <button type="button">Open menu</button>
</BpkContextMenu.Trigger>
```

`Menu.Trigger` renders its own `<button>` around the child, producing `<button><button>…</button></button>` — invalid HTML. Same issue in [accessibility-test.tsx:29–32](packages/backpack-web/src/bpk-component-context-menu/src/accessibility-test.tsx#L29-L32) and [accessibility-test.tsx:46–49](packages/backpack-web/src/bpk-component-context-menu/src/accessibility-test.tsx#L46-L49). Fix: either add `asChild` or wrap a non-button element (e.g. a `<span>` or just an icon).

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+      : 'default';
+
+  return (
+    <Menu.Item
+      value={value}
+      disabled={disabled}
+      onClick={onClick}
+      className={getClassName(
+        'bpk-context-menu__item',
+        `bpk-context-menu__item--${variantModifier}`,
```
Supremeyh: [BpkContextMenuItem.tsx:60–72](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuItem.tsx#L60-L72) always appends either `bpk-context-menu__item--default` or `bpk-context-menu__item--destructive`, but [BpkContextMenu.module.scss:159](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss#L159) only defines `--destructive`. The `--default` class is emitted into the DOM but has no rules. Either define it or skip appending the class when `variant === 'default'`.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuItemGroup.tsx
```diff
@@ -0,0 +1,37 @@
...
+const getClassName = cssModules(STYLES);
+
+export type BpkContextMenuItemGroupProps = {
+  children: ReactNode;
+};
+
+const BpkContextMenuItemGroup = ({
+  children,
+}: BpkContextMenuItemGroupProps) => (
+  <div className={getClassName('bpk-context-menu__item-group')}>{children}</div>
```
Supremeyh: [BpkContextMenuItemGroup.tsx:34](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuItemGroup.tsx#L34) renders a plain `<div>`. Ark's `Menu.ItemGroup` emits `role="group"` with an `aria-labelledby` link to a `Menu.ItemGroupLabel`, allowing screen readers to announce group boundaries. A plain `<div>` provides none of that. Either use `Menu.ItemGroup` and expose a `label` prop, or document clearly why the semantic variant is intentionally omitted.

packages/backpack-web/src/bpk-component-context-menu/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  BpkContextMenuSelectionDetails,
+  BpkContextMenuOpenChangeDetails,
+} from './src/common-types';
+export type { BpkContextMenuRootProps } from './src/BpkContextMenuRoot';
+export type { BpkContextMenuTriggerProps } from './src/BpkContextMenuTrigger';
+export type { BpkContextMenuContentProps } from './src/BpkContextMenuContent';
+export type { BpkContextMenuItemProps } from './src/BpkContextMenuItem';
+export type { BpkContextMenuSeparatorProps } from './src/BpkContextMenuSeparator';
+export type { BpkContextMenuItemGroupProps } from './src/BpkContextMenuItemGroup';
+export type { BpkContextMenuSubTriggerItemProps } from './src/BpkContextMenuSubTriggerItem';
```
Supremeyh: [index.ts:36](packages/backpack-web/src/bpk-component-context-menu/index.ts#L36) exports `BpkContextMenuSubTriggerItemProps`, but the component is surfaced as `BpkContextMenu.TriggerItem`. Rename the prop type to `BpkContextMenuTriggerItemProps` for consistency.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuSeparator.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { Menu } from '@ark-ui/react';
+
+import { cssModules } from '../../bpk-react-utils';
+
+import STYLES from './BpkContextMenu.module.scss';
+
+const getClassName = cssModules(STYLES);
+
+export type BpkContextMenuSeparatorProps = Record<string, never>;
```
Supremeyh: [BpkContextMenuSeparator.tsx:27](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuSeparator.tsx#L27) prevents passing any attribute including `aria-label` or `role`. Use `{}` or omit the explicit type.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,218 @@
...
+  svg {
+    fill: currentcolor;
+  }
+}
+
+// Positioner is a bare wrapper that Floating UI (via Ark/Zag) positions.
+// It must not carry padding or width — those live on the content surface —
+// otherwise Floating UI's measurements will include the wrong box.
+.bpk-context-menu__positioner {
+  z-index: 1;
```
Supremeyh: [BpkContextMenu.module.scss:79](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss#L79) — most portalled surfaces (modals, toasts) use much higher values. The menu will render behind stacking contexts created by ancestor transforms or other portals. Worth check the token scale (`tokens.$bpk-z-index-*`) and use the appropriate value.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,218 @@
...
+  padding: tokens.bpk-spacing-sm() 0;
+  flex-direction: column;
+  border-radius: tokens.$bpk-border-radius-lg;
+  outline: none;
+  background-color: tokens.$bpk-canvas-day;
+  box-shadow: tokens.$bpk-box-shadow-lg;
+  overflow-y: auto;
+
+  // 8.5 items: (8.5 × 2.5rem item-height) = 21.25rem.
+  max-block-size: 21.25rem;
```
Supremeyh: `max-block-size: 21.25rem` is a magic number, [BpkContextMenu.module.scss:96–97](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss#L96-L97) — the comment explains it (`8.5 × 2.5rem item height`) but both `21.25rem` and the implied `2.5rem` item height are hard to track if the item padding ever changes. Extract as a SCSS variable.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,218 @@
...
+// The content surface — this is the visible "card". Rounded corners, elevation,
+// canvas background. Renders a min/max width so short and long trip names both
+// look consistent.
+.bpk-context-menu__content {
+  display: flex;
+  padding: tokens.bpk-spacing-sm() 0;
+  flex-direction: column;
+  border-radius: tokens.$bpk-border-radius-lg;
+  outline: none;
+  background-color: tokens.$bpk-canvas-day;
```
Supremeyh: All colour references use bare `*-day` tokens (e.g. `$bpk-canvas-day` at [BpkContextMenu.module.scss:92](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss#L92), `$bpk-text-primary-day` at line 129). Given the active CSS-vars migration in this repo (see recent `bpk-component-modal` and `bpk-component-info-banner` PRs), shipping without it is fine as long as it's tracked in a follow-up ticket — please link one here.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.stories.tsx
```diff
@@ -0,0 +1,295 @@
...
+          <BpkContextMenu.Item
+            value="remove"
+            variant={CONTEXT_MENU_ITEM_VARIANTS.destructive}
+          >
+            Remove
+          </BpkContextMenu.Item>
+          <BpkContextMenu.Root>
+            <BpkContextMenu.TriggerItem endIcon={<ChevronRightIcon />}>
+              Move
+            </BpkContextMenu.TriggerItem>
```
Supremeyh, line 98: nit: [BpkContextMenu.stories.tsx:92–95](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.stories.tsx#L92-L95) — the nested `BpkContextMenu.Root` inside `BpkContextMenu.Content` in `WithDestructiveItemExample` is the correct sub-menu pattern but could confuse future readers without a comment.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.stories.tsx
```diff
@@ -0,0 +1,295 @@
...
+          <BpkContextMenu.ItemGroup>
+            <BpkContextMenu.Item
+              value="remove"
+              variant={CONTEXT_MENU_ITEM_VARIANTS.destructive}
+            >
+              Remove
+            </BpkContextMenu.Item>
+            <BpkContextMenu.Item value="move" endIcon={<ChevronRightIcon />}>
+              Move
+            </BpkContextMenu.Item>
```
Supremeyh: nit: - [BpkContextMenu.stories.tsx:287–288](packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.stories.tsx#L287-L288) — `VisualTestDestructive` uses a plain `Item` with a chevron icon for "Move" while the interactive `WithDestructiveItemExample` story correctly uses `TriggerItem`. Inconsistency could mislead readers.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuTrigger.tsx
```diff
@@ -0,0 +1,67 @@
...
+   * contains only an icon with no visible text.
+   */
+  'aria-label'?: string;
+  /**
+   * When true, delegates trigger behaviour to the child element rather than
+   * rendering a Backpack-owned button. The child must be a DOM element or a
+   * forwardRef component. Use this to compose with an existing product button
+   * (e.g. BpkSaveButton) that owns its own styling and state.
+   */
+  asChild?: boolean;
```
Supremeyh: `BpkContextMenuTrigger` exposes `asChild` but the README/JSDoc doesn't warn that the child must be a `forwardRef` component for `asChild` to work. Worth a note.

toveadamsson:
Regarding the long list UI, is it intentional that Plan a new trip and Quick Save aren't visible initially? 
Screenshot 2026-07-27 at 14 00 19

toveadamsson:
Can we include documents in the descriptions that will be useful for external review, as to why certain decisions were made and the figma design as well 🙏

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
tranculent: I removed the $bpk-canvas-day to avoid follow up work. Thanks for letting me know 🙏

tranculent:
> Regarding the long list UI, is it intentional that Plan a new trip and Quick Save aren't visible initially? Screenshot 2026-07-27 at 14 00 19

this is just a story but the way this Ark UI component is exposed to consumers is it lets them craft the menu as they wish. So technically it is possible to just render the Plan a new trip and quick save block on top without any issues. Also, this hasn't been confirmed by design yet.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,228 @@
...
+// The content surface — this is the visible "card". Rounded corners, elevation,
+// canvas background. Renders a min/max width so short and long trip names both
+// look consistent.
+.bpk-context-menu__content {
+  display: flex;
+  padding: tokens.bpk-spacing-sm() 0;
+  flex-direction: column;
+  border-radius: tokens.$bpk-border-radius-lg;
+  outline: none;
+  background-color: var(--bpk-canvas, #{tokens.$bpk-canvas-day});
```
gert-janvercauteren: --bpk-canvas does not exist, it should be --bpk-canvas-default
```suggestion
  background-color: var(--bpk-canvas-default, #{tokens.$bpk-canvas-day});
```

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,228 @@
...
+    var(--bpk-canvas-contrast, #{tokens.$bpk-canvas-contrast-day});
+}
+
+.bpk-context-menu__separator {
+  display: block;
+  margin: tokens.bpk-spacing-sm() 0;
+  border: 0;
+  block-size: 0;
+  border-block-start: tokens.$bpk-border-size-sm solid
+    var(--bpk-line, #{tokens.$bpk-line-day});
```
gert-janvercauteren: --bpk-line does not exist, can you double check? 
I think it's --bpk-other-line-default, but not 100% sure

packages/backpack-web/src/bpk-component-card-button/src/BpkSaveButton.tsx
```diff
@@ -117,4 +117,4 @@ const BpkSaveButton = ({
   );
 };
 
-export default BpkSaveButton;
\ No newline at end of file
+export default BpkSaveButton;
```
Supremeyh: could we revert this unrelated change

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,228 @@
...
+$context-menu-max-width: 20rem;
+
+// Measured rendered height of one item row (vertical padding + line-height).
+// Update this if item padding or typography changes.
+$context-menu-item-height: 2.6rem;
+$context-menu-max-visible-items: 8.5;
+
+// Default trigger button — resets browser button styles so consumers get a
+// clean base to build on via className. Consumers own the visual design.
+.bpk-context-menu__trigger {
```
Supremeyh, line 34: In the dark mode, the style seems a little weird, could you double check? 
image

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,228 @@
...
+  &:focus {
+    outline: none;
+  }
+
+  &:focus-visible:not([data-pointer-focus]) {
+    @include utils.bpk-focus-indicator;
+  }
+
+  svg {
+    fill: currentcolor;
```
Supremeyh, line 77: The heart icon appeared black in dark mode because `.bpk-context-menu__save-trigger` had no `color` property — `fill: currentcolor` on the `svg` was inheriting from the ChakraProvider/body default (`#161616`), not from a theme-aware token.

How about add `color: var(--bpk-text-primary, #{tokens.$bpk-text-primary-day})` to `.bpk-context-menu__save-trigger` and keep `svg { fill: currentcolor }`. This pattern (drive icon colour via `color`, not via `fill` directly):
- A future `variant` prop can simply override `color` to change the icon without touching `fill`.
- It stays consistent with how other Backpack components handle icon colouring.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,228 @@
...
+  color: inherit;
+
+  svg {
+    fill: currentcolor;
+  }
+}
+
+.bpk-context-menu__item-group {
+  --bpk-context-menu-item-divider: #{tokens.$bpk-border-size-sm} solid
+    var(--bpk-canvas-contrast, #{tokens.$bpk-canvas-contrast-day});
```
Supremeyh, line 207: ` --bpk-canvas-contrast` is `#010913` in dark, identical to the surface → invisible

it should be `var(--bpk-other-line-default, #{tokens.$bpk-line-day});`

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu-test.tsx
```diff
@@ -0,0 +1,215 @@
...
+      renderMenuWithSaveTrigger();
+      expect(screen.getByRole('button', { name: 'Save to trip' })).toBeVisible();
+    });
+
+    it('has the ContextMenuSaveTrigger data-component attribute', () => {
+      renderMenuWithSaveTrigger();
+      expect(
+        document.querySelector('[data-component="ContextMenuSaveTrigger"]'),
+      ).toBeInTheDocument();
+    });
```
Supremeyh: **Wrong data attribute — test will throw at runtime** *(recurring pattern)*

`getDataComponentAttribute` emits `data-backpack-ds-component`, not `data-component` ([source](https://github.com/Skyscanner/backpack/blob/main/packages/backpack-web/src/bpk-react-utils/src/getDataComponentAttribute.ts#L38)). `document.querySelector('[data-component="ContextMenuSaveTrigger"]')` always returns `null`, so `expect(null).toBeInTheDocument()` throws a TypeError — this test is broken.

Fix:
```tsx
document.querySelector('[data-backpack-ds-component="ContextMenuSaveTrigger"]')
```

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu-test.tsx
```diff
@@ -0,0 +1,215 @@
...
+      expect(trigger).not.toHaveAttribute('data-pointer-focus');
+    });
+  });
+
+  it('does not accept className from consumers', () => {
+    // @ts-expect-error — className is intentionally not part of the public API
+    render(<BpkContextMenu.Root className="custom-classname" open />);
+    const root = document.querySelector('[data-component="ContextMenu"]');
+    expect(root?.className).not.toContain('custom-classname');
+  });
```
Supremeyh, line 285: **Wrong data attribute — test passes vacuously** *(recurring pattern)*

Same `data-component` vs `data-backpack-ds-component` mismatch. `querySelector` returns `null`, so `null?.className` is `undefined`, and `expect(undefined).not.toContain('custom-classname')` passes — but it is testing nothing. The className restriction is not actually verified.

Fix:
```tsx
const root = document.querySelector('[data-backpack-ds-component="ContextMenu"]');
expect(root?.className).not.toContain('custom-classname');
```

packages/backpack-web/src/bpk-component-context-menu/src/accessibility-test.tsx
```diff
@@ -0,0 +1,72 @@
...
+            value="remove"
+            variant={CONTEXT_MENU_ITEM_VARIANTS.destructive}
+          >
+            Remove
+          </BpkContextMenu.Item>
+        </BpkContextMenu.Content>
+      </BpkContextMenu.Root>,
+    );
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
```
Supremeyh, line 121: **`axe(container)` misses portalled menu content** *(Constitution IV — NON-NEGOTIABLE)*

`BpkContextMenuContent` renders inside Ark's `<Portal>`, which appends to `document.body` — outside the `container` element returned by `render()`. The `[role="menu"]` subtree and all `[role="menuitem"]` elements are invisible to jest-axe here. This open-state test effectively only scans the trigger button.

Fix:
```tsx
const results = await axe(document.body);
```

packages/backpack-web/src/bpk-component-context-menu/src/accessibility-test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+});
```
Supremeyh, line 123: **No axe test for `SaveTrigger` variant** *(Constitution IV)*

`BpkContextMenuSaveTrigger` is an icon-only button whose sole accessible name comes from a required `aria-label` prop — the highest-risk a11y surface in this component. There is no test verifying it passes axe in either state.

Please add a second `describe` block:
```tsx
describe('with SaveTrigger', () => {
  it('should not have accessibility issues (closed)', async () => {
    const { container } = render(
      <BpkContextMenu.Root>
        <BpkContextMenu.SaveTrigger aria-label="Save to trip" />
        <BpkContextMenu.Content>
          <BpkContextMenu.Item value="item">Item</BpkContextMenu.Item>
        </BpkContextMenu.Content>
      </BpkContextMenu.Root>,
    );
    expect(await axe(container)).toHaveNoViolations();
  });
});
```

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,228 @@
...
+  gap: tokens.bpk-spacing-base();
+
+  @include typography.bpk-body-default;
+
+  &:hover:not([data-disabled]) {
+    background-color: var(
+      --bpk-canvas-contrast,
+      #{tokens.$bpk-canvas-contrast-day}
+    );
+  }
```
Supremeyh: **Bare `:hover` bypasses the touch-device guard** *(Constitution III)*

The Backpack convention (`BpkCardV2`, `BpkChip`, `BpkButton`) is `@include utils.bpk-hover { … }`, which gates the rule behind `.bpk-no-touch-support` to prevent sticky hover states on touch devices. Bare `:hover` fires on touch.

Fix (requires adding `@use '../../bpk-mixins/utils'` if not already imported):
```scss
@include utils.bpk-hover {
  &:not([data-disabled]) {
    background-color: var(--bpk-canvas-contrast, #{tokens.$bpk-canvas-contrast-day});
  }
}
```

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
Supremeyh: **`--bpk-canvas` does not exist + wrong semantic var**

Two issues on this line:
1. `--bpk-canvas` is not defined in either theme file — `var()` always falls back to the SASS value (`#ffffff`), so the menu surface is white in dark mode.
2. Even corrected to `--bpk-canvas-default`, the semantic is wrong. Every other floating surface in the repo — `BpkPopover`, `BpkCardV2`, `BpkDrawer`, `BpkDialog` — uses `--bpk-surface-default`, which has a distinct dark-mode value (`#131d2b`) vs `--bpk-canvas-default` (`#010913`).

Fix:
```scss
background-color: var(--bpk-surface-default, #{tokens.$bpk-surface-default-day});
```

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuTrigger.tsx
```diff
@@ -0,0 +1,71 @@
...
+   * handler) onto the child element instead of rendering a Backpack-owned
+   * wrapper button.
+   *
+   * **Requirement:** the child must be a native DOM element or a component
+   * wrapped in React.forwardRef that spreads `...rest` onto its root node.
+   * Plain function components without forwardRef will silently drop the merged
+   * props and the menu will not open.
+   */
+  asChild?: boolean;
+};
```
Supremeyh: **`asChild` needs a discriminated union — wrong `children` type** *(recurring pattern from BpkCollapsible PRs #4897, #4925)*

When `asChild: true`, Ark's Slot calls `Children.only()` and throws (or silently renders null) when given strings, fragments, or multiple children. Declaring `children: ReactNode` allows these at the type level.

The fix applied in BpkCollapsible:
```tsx
type BpkContextMenuTriggerProps =
  | { asChild: true; children: ReactElement; 'aria-label'?: string }
  | { asChild?: false; children: ReactNode; 'aria-label'?: string };
```
This catches invalid usage at compile time instead of silently at runtime.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu-test.tsx
```diff
@@ -0,0 +1,215 @@
...
+});
```
Supremeyh, line 310: **No unit test for `BpkContextMenu.TriggerItem`** *(Constitution VIII)*

`BpkContextMenu.TriggerItem` is exported as part of the public API and used in `WithDestructiveItemExample` and `VisualTestDestructive`. There are zero unit tests for it — no render test, no `endIcon` slot test.

Please add at minimum:
```tsx
it('renders TriggerItem with endIcon', () => {
  render(
    <BpkContextMenu.Root open>
      <BpkContextMenu.Trigger><span>Open</span></BpkContextMenu.Trigger>
      <BpkContextMenu.Content>
        <BpkContextMenu.Root>
          <BpkContextMenu.TriggerItem endIcon={<span data-testid="chevron">›</span>}>
            Move
          </BpkContextMenu.TriggerItem>
          <BpkContextMenu.Content>
            <BpkContextMenu.Item value="dest">Destination</BpkContextMenu.Item>
          </BpkContextMenu.Content>
        </BpkContextMenu.Root>
      </BpkContextMenu.Content>
    </BpkContextMenu.Root>,
  );
  expect(screen.getByTestId('chevron')).toBeVisible();
});
```

Supremeyh (commented):
In the light mode, there is hover state background for each item,  it seems there is blank space background between first/last item and it's border, is that expected? And does it also need hover state background  on dark mode?

image

tranculent:
Dark Mode:
image

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,234 @@
...
+  color: inherit;
+
+  svg {
+    fill: currentcolor;
+  }
+}
+
+.bpk-context-menu__item-group {
+  --bpk-context-menu-item-divider: #{tokens.$bpk-border-size-sm} solid
+    var(--bpk-canvas-contrast, #{tokens.$bpk-canvas-contrast-day});
```
Supremeyh, line 207: should it be `--bpk-other-line-default` for semantic?

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenuItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+  value: string;
+  /**
+   * Optional trailing element rendered on the end side of the row.
+   * Typically an icon (e.g. `+` for "Plan a new trip", bookmark for
+   * "Quick save", chevron for a nested menu entry point).
+   */
+  endIcon?: ReactNode;
+  variant?: BpkContextMenuItemVariant;
+  disabled?: boolean;
+  onClick?: () => void;
```
Supremeyh: `BpkContextMenu.Item` exposes an item-level `onClick`, and line 63 forwards it to `Menu.Item` as a DOM click handler. That only covers pointer/mouse activation. Ark handles keyboard activation from the menu content with `Enter` / `Space` by dispatching its own `menu:select` event and calling the item `onSelect` listener; it does not fire this DOM `onClick` path.

That means a consumer using the documented item-specific handler will get different behaviour for mouse and keyboard users: root `onSelect` fires for both, but the item's own handler only fires for click. Please expose/forward Ark's `onSelect` for item-specific handlers, or rename this API so consumers do not rely on it for selection. A useful regression test would activate a highlighted item with keyboard and assert the item-level handler is called.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  &:focus-visible {
+    @include utils.bpk-focus-indicator;
+  }
+}
+
+.bpk-context-menu__item--destructive {
+  color: var(--bpk-status-danger-spot, #{tokens.$bpk-status-danger-spot-day});
+}
```
Supremeyh: `BpkContextMenu.Item` allows `disabled` and `variant={CONTEXT_MENU_ITEM_VARIANTS.destructive}` at the same time, but the destructive rule is declared after the disabled rule. Because both selectors have the same specificity, line 175 overrides the disabled colour from lines 160-162, so a disabled destructive action still looks enabled/dangerous instead of disabled.

Please gate the destructive colour behind `:not([data-disabled])`, or add a later disabled override. It would also be worth adding a unit/style assertion for the disabled destructive combination so this does not regress.

packages/backpack-web/src/bpk-component-context-menu/src/BpkContextMenu.module.scss
tranculent: no, it doesn't work right - it's too dark. Is --bpk-canvas-contrast to be migrated or it can be used?

Supremeyh (approved):
lgtm, thanks your contribution, worth update PR description with latest changes

PR #4916, opened 2026-07-21
[CLOV-1868] Part 3: add public Backpack Adoption NX plugin

xiaogliu:
## Summary

- Adds the public `@skyscanner/backpack-adoption-nx-plugin` npm package with per-project `analyze` and aggregate `report` Nx executors.
- Bundles the private analyzer from `libs/backpack-adoption-analyzer` into compiled executor artifacts; consumers install no analyzer package.
- Adds validation and manual release automation for the public npm package.

## Release model

- The plugin is published with `--access public` from a manual, non-dry-run dispatch on `main`.
- The release job uses npm trusted publishing (GitHub Actions OIDC), not a long-lived npm publish token. npm must be configured for `Skyscanner/backpack` and `backpack-adoption-nx-plugin-release.yml` before the first automated release.
- The `minor` label represents the initial `0.1.0` release.

## Review scope

This PR is rebased on `main` after Parts 1 and 2 merged. Its diff contains only the Nx plugin, its release workflow, and the minimal workspace configuration needed to bundle the existing analyzer.

## Validation

- The workflow validates lint, typecheck, tests, build, and `npm pack --dry-run` on PRs.
- The prior release-workflow validation passed; the rebased workflow run is the final CI verification.

### DIFF
--- a/.github/workflows/backpack-adoption-nx-plugin-release.yml
+++ b/.github/workflows/backpack-adoption-nx-plugin-release.yml
@@ -113,12 +113,11 @@ jobs:
 
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version-file: .nvmrc
+          # Node 24 includes npm 11, required for npm trusted publishing OIDC.
+          node-version: 24.14.0
           registry-url: https://registry.npmjs.org
           cache: pnpm
 
-      # npm trusted publishing requires npm 11.5.1 or newer for OIDC support.
-      - run: npm install --global npm@11.5.1
       - run: pnpm install --frozen-lockfile
       - run: npx nx run backpack-adoption-nx-plugin:build
       - name: Stamp release version
--- a/.github/workflows/backpack-adoption-nx-plugin-release.yml
+++ b/.github/workflows/backpack-adoption-nx-plugin-release.yml
@@ -59,6 +59,15 @@ jobs:
           release_type: ${{ inputs.release_type || 'minor' }}
           tag_prefix: backpack-adoption-nx-plugin
 
+      - name: Ensure release tag does not already exist
+        env:
+          RELEASE_TAG: ${{ steps.version.outputs.release_tag }}
+        run: |
+          if git rev-parse "refs/tags/$RELEASE_TAG" >/dev/null 2>&1; then
+            echo "::error::Tag $RELEASE_TAG already exists"
+            exit 1
+          fi
+
       - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
 
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
@@ -107,6 +116,15 @@ jobs:
 
       - run: pnpm install --frozen-lockfile
       - run: npx nx run backpack-adoption-nx-plugin:build
+      - name: Recheck release tag before publishing
+        env:
+          RELEASE_TAG: ${{ needs.Validate.outputs.release_tag }}
+        run: |
+          git fetch --force --tags origin
+          if git rev-parse "refs/tags/$RELEASE_TAG" >/dev/null 2>&1; then
+            echo "::error::Tag $RELEASE_TAG already exists"
+            exit 1
+          fi
       - name: Stamp release version
         env:
           RELEASE_VERSION: ${{ needs.Validate.outputs.release_version }}
--- a/.github/workflows/backpack-adoption-nx-plugin-release.yml
+++ b/.github/workflows/backpack-adoption-nx-plugin-release.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/backpack-adoption-nx-plugin-release.yml'
-      - 'packages/backpack-adoption-analyzer/**'
+      - 'libs/backpack-adoption-analyzer/**'
       - 'packages/backpack-adoption-nx-plugin/**'
       - 'pnpm-lock.yaml'
       - 'tsconfig.json'
@@ -37,7 +37,6 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      id-token: write
     outputs:
       release_tag: ${{ steps.version.outputs.release_tag }}
       release_version: ${{ steps.version.outputs.release_version }}
@@ -96,6 +95,7 @@ jobs:
     if: ${{ github.event_name == 'workflow_dispatch' && !inputs.dry_run && github.ref == 'refs/heads/main' }}
     permissions:
       contents: read
+      id-token: write
     steps:
       - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
         id: app-token
@@ -117,6 +117,8 @@ jobs:
           registry-url: https://registry.npmjs.org
           cache: pnpm
 
+      # npm trusted publishing requires npm 11.5.1 or newer for OIDC support.
+      - run: npm install --global npm@11.5.1
       - run: pnpm install --frozen-lockfile
       - run: npx nx run backpack-adoption-nx-plugin:build
       - name: Stamp release version
--- a/.github/workflows/backpack-adoption-nx-plugin-release.yml
+++ b/.github/workflows/backpack-adoption-nx-plugin-release.yml
@@ -1,6 +1,13 @@
 name: Backpack Adoption NX Plugin Release
 
 on:
+  pull_request:
+    paths:
+      - '.github/workflows/backpack-adoption-nx-plugin-release.yml'
+      - 'packages/backpack-adoption-analyzer/**'
+      - 'packages/backpack-adoption-nx-plugin/**'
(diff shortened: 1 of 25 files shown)


.github/workflows/backpack-adoption-nx-plugin-release.yml
```diff
@@ -0,0 +1,141 @@
...
+      - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version-file: .nvmrc
+          registry-url: https://registry.npmjs.org
+          cache: pnpm
+
+      # npm trusted publishing requires npm 11.5.1 or newer for OIDC support.
+      - run: npm install --global npm@11.5.1
```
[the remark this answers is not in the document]
xiaogliu: Addressed in 74b8c7436. The Publish job now uses Node 24.14.0, which includes an npm 11 release compatible with npm trusted-publishing OIDC. This removes the ad-hoc global package installation and keeps the npm version provisioned by the pinned setup-node action.

packages/backpack-adoption-nx-plugin/src/executors/analyze/executor.ts
```diff
@@ -0,0 +1,87 @@
...
+  DEFAULT_ADOPTION_GUARD_THRESHOLD,
+  DEFAULT_PATTERN,
+  analyzeRepository,
+  evaluateGuard,
+} from "@skyscanner/backpack-adoption-analyzer";
+
+import { DEFAULT_RESULTS_FILE_NAME, writeProjectResult } from "../results-store.ts";
+
+import type { ExecutorContext, PromiseExecutor } from "@nx/devkit";
+import type { AnalyzeExecutorSchema } from "./schema.ts";
```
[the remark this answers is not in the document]
xiaogliu: Addressed in 74b8c7436. The import is now extensionless (`./schema`), so TypeScript resolves the existing `schema.d.ts`; the package and workspace `allowImportingTsExtensions` settings were removed. Typecheck passes.

packages/backpack-adoption-nx-plugin/src/executors/report/executor.ts
```diff
@@ -0,0 +1,97 @@
...
+ * limitations under the License.
+ */
+import { joinPathFragments, logger } from "@nx/devkit";
+import { combineGuardStatuses } from "@skyscanner/backpack-adoption-analyzer";
+
+import { DEFAULT_RESULTS_FILE_NAME, readProjectResult } from "../results-store.ts";
+
+import type { GuardResult } from "@skyscanner/backpack-adoption-analyzer";
+import type { ExecutorContext, PromiseExecutor } from "@nx/devkit";
+import type { ReportExecutorSchema } from "./schema.ts";
```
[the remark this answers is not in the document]
xiaogliu: Addressed in 74b8c7436. The report executor now imports `./schema` without a `.ts` extension, resolving the existing declaration file through normal TypeScript resolution. Typecheck passes.

packages/backpack-adoption-nx-plugin/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// This package is consumed via its NX executors (see executors.json),
+// referenced by name in consumers' project.json — not via JS imports. This
+// barrel only exists to satisfy package.json's `main`/`exports` fields.
+export type { AnalyzeExecutorSchema } from "./executors/analyze/schema.ts";
+export type { ReportExecutorSchema } from "./executors/report/schema.ts";
```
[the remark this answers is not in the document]
xiaogliu: Addressed in 74b8c7436. The barrel exports now use extensionless schema paths, matching the `schema.d.ts` files and removing the need for explicit TypeScript-extension imports. Typecheck passes.

packages/backpack-adoption-nx-plugin/src/executors/results-store.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const readProjectResult = (
+  resultsPath: string,
+): ProjectAdoptionResult | null => {
+  try {
+    return JSON.parse(readFileSync(resultsPath, "utf8")) as ProjectAdoptionResult;
+  } catch {
+    return null;
+  }
+};
```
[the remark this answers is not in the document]
xiaogliu: Addressed in 74b8c7436. `readProjectResult` now returns `null` only for an absent file (`ENOENT`) or invalid JSON (`SyntaxError`); it rethrows other I/O errors such as permission failures so CI can surface them.

packages/backpack-adoption-nx-plugin/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@skyscanner/backpack-adoption-nx-plugin",
+  "version": "0.1.0",
```
wpe1arhy: 🔴 **The first `0.1.0` release will fail at the "Stamp release version" step.**

This is already `0.1.0`, and on the first dispatch there are no `backpack-adoption-nx-plugin/v*` tags, so `compute-next-semver` with `minor` returns `0.1.0`. The Publish job then runs `npm version 0.1.0 --no-git-tag-version` on a package already at `0.1.0`, which exits non-zero:

```
npm error Version not changed, might want --allow-same-version
```

So the very first release breaks before it ever publishes. Two options:
- add `--allow-same-version` to the `npm version` call in the workflow; or
- set this to `0.0.0` so the computed `0.1.0` is always a real bump (cleaner — it decouples the committed version from the tag-derived release version).

.github/workflows/backpack-adoption-nx-plugin-release.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - run: pnpm install --frozen-lockfile
+      - run: npx nx run backpack-adoption-nx-plugin:build
+      - name: Stamp release version
+        env:
+          RELEASE_VERSION: ${{ needs.Validate.outputs.release_version }}
+        run: npm version "$RELEASE_VERSION" --no-git-tag-version --ignore-scripts
+        working-directory: packages/backpack-adoption-nx-plugin
+      - name: Publish public npm package
+        run: npm publish --ignore-scripts --access public --provenance
```
wpe1arhy, line 134: 🟡 **Publish-before-tag ordering with no tag pre-existence check.**

The guard workflow guards tag races in both Validate and CreateTags. Here neither exists, and it runs `npm publish` (irreversible) before creating the tag. If the later tag push fails (a concurrent/retried dispatch where the tag already exists, or a transient push error), you end up published-but-untagged, and a re-run then fails because the npm version is immutable.

Consider mirroring the guard workflow and adding a "release tag does not already exist" check in Validate, so a duplicate is caught before the irreversible publish.

packages/backpack-adoption-nx-plugin/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "registry": "https://registry.npmjs.org/"
+  },
+  "scripts": {
+    "build": "node scripts/build.js",
+    "lint": "eslint src jest.config.js --ext .js,.ts",
+    "test": "jest --config jest.config.js",
+    "typecheck": "tsc --project tsconfig.json"
+  },
+  "dependencies": {
+    "@nx/devkit": "22.7.1"
```
wpe1arhy: 🟡 **`@nx/devkit` is pinned hard to `22.7.1` while `nx` is a loose peer (`>=20`).**

`@nx/devkit` is marked `external` in the esbuild bundle, so it's a real installed runtime dependency. A consumer on `nx@20`/`21` gets `@nx/devkit@22.7.1` alongside their older `nx` — but devkit is expected to track the nx major.

Was making `@nx/devkit` a peerDependency too (range aligned with `nx`) considered, so it resolves to the consumer's nx version rather than forcing 22.7.1?

wpe1arhy (commented):
Overall LGTM — tests and docs are solid; the main one is that the first release will fail outright.

packages/backpack-adoption-nx-plugin/package.json
xiaogliu: Addressed in 93831406b. The committed package version is now 0.0.0, so the first minor release computes and stamps 0.1.0 as a genuine version change.

packages/backpack-adoption-nx-plugin/package.json
xiaogliu: Addressed in 93831406b. @nx/devkit is now a >=20 peer dependency aligned with the existing nx >=20 peer, so consumers resolve the devkit version compatible with their Nx installation instead of receiving a forced 22.7.1 copy.

.github/workflows/backpack-adoption-nx-plugin-release.yml
xiaogliu: Addressed in 93831406b. Validate now rejects an existing computed release tag, and Publish fetches tags and repeats that check immediately before publishing. The workflow main-ref concurrency group also serializes manual release dispatches.

wpe1arhy (approved):
LGTM

PR #4975, opened 2026-07-28
[IRN-6854] Fix card-list rail pagination on short final page

jimmycook:
## Summary

Fixes a pagination bug in `BpkCardList` row/rail mode.

## Problem

When the final page of a rail holds **fewer cards than `initiallyShownCards`** (odd totals — e.g. 3 cards shown 2-wide), physically scrolling to the end does **not** advance the pagination indicator.

Cause: in `usePageScrollSync`, the page index is derived from `Math.floor(firstVisibleIndex / initiallyShownCards)`. On a short final page there isn't enough content to scroll that page's first card to the viewport start, so `firstVisibleIndex` never reaches it and the indicator stays on the previous page.

## Change

- `BpkCardListRowRail/utils.tsx` — in `usePageScrollSync`, treat a visible last card as the final page; otherwise the existing calculation is used.
- `BpkCardListRowRail/utils-test.tsx` — unit test for the short-final-page case.
- `BpkCardList.stories.tsx` — two repro stories:
  - `RowToRailShortFinalPage` (2-wide, 3 cards)
  - `RowToRailShortFinalPageWider` (4-wide, 10 cards)

## How to review

Storybook → `bpk-component-card-list` → the two `RowToRailShortFinalPage*` stories. On desktop, physically scroll to the last card; the indicator now tracks to the final page. `pnpm jest packages/backpack-web/src/bpk-component-card-list` is green (61 tests).

## Notes

Draft — no changelog entry yet, wanted a sanity check on the approach before polishing.

### DIFF
--- a/packages/backpack-web/src/bpk-component-card-list/src/BpkCardListRowRail/utils-test.tsx
+++ b/packages/backpack-web/src/bpk-component-card-list/src/BpkCardListRowRail/utils-test.tsx
@@ -584,6 +584,47 @@ describe('usePageScrollSync', () => {
       expect(mockSetCurrentIndex).not.toHaveBeenCalled();
     });
 
+    it('should snap to the last page when the final (short) page cannot scroll its first card to the start', () => {
+      // 3 cards, 2 shown per page => page 0 = [0,1], page 1 = [2].
+      // The final page holds fewer cards than initiallyShownCards, so scrolling to
+      // the end leaves card 2 alongside card 1 (firstVisibleIndex stays at 1).
+      // Because the last card is visible, we should still land on page 1.
+      const shortCardRefs = { current: [] as Array<HTMLDivElement | null> };
+      Array.from({ length: 3 }).forEach((_, index) => {
+        const { mockDiv } = makeMockDiv(index);
+        mockDiv.scrollIntoView = jest.fn();
+        shortCardRefs.current.push(mockDiv);
+      });
+
+      const { rerender } = renderHook(
+        ({ visibilityList }) =>
+          usePageScrollSync({
+            cardRefs: shortCardRefs,
+            container: mockContainer,
+            currentIndex: 0,
+            enabled: true,
+            initiallyShownCards: 2,
+            setCurrentIndex: mockSetCurrentIndex,
+            visibilityList,
+          }),
+        {
+          initialProps: {
+            visibilityList: [1, 1, 0],
+          },
+        },
+      );
+
+      // User initiates scroll via wheel
+      act(() => {
+        mockContainer.dispatchEvent(new Event('wheel'));
+      });
+
+      // Scrolled to the end: cards 1 and 2 visible, card 0 out of view
+      rerender({ visibilityList: [0, 1, 1] });
+
+      expect(mockSetCurrentIndex).toHaveBeenCalledWith(1);
+    });
+
     it('should handle partial visibility at page boundaries by using first visible card', () => {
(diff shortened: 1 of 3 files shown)


packages/backpack-web/src/bpk-component-card-list/src/BpkCardListRowRail/utils.tsx
```diff
@@ -209,7 +209,13 @@ export const usePageScrollSync = ({
...
     if (firstVisibleIndex === -1) return;
 
-    const newPageIndex = Math.floor(firstVisibleIndex / initiallyShownCards);
+    // A short final page can't scroll its first card to the viewport start, so
+    // firstVisibleIndex stays on the previous page. Treat a visible last card as
+    // the final page.
+    const newPageIndex =
+      visibilityList[visibilityList.length - 1] === 1
+        ? Math.ceil(visibilityList.length / initiallyShownCards) - 1
+        : Math.floor(firstVisibleIndex / initiallyShownCards);
```
Supremeyh, line 221: This fires unconditionally whenever the last card is in the viewport **and** `isUserScrollingRef.current` is true. On a wide viewport that fits all cards simultaneously, any `wheel` or `touchstart` event sets `isUserScrollingRef.current = true`, then `visibilityList[last] === 1` immediately triggers the override — forcing the indicator to the last page even though the user has never scrolled the rail.

packages/backpack-web/src/bpk-component-card-list/src/BpkCardListRowRail/utils-test.tsx
```diff
@@ -584,6 +584,47 @@ describe('usePageScrollSync', () => {
       expect(mockSetCurrentIndex).not.toHaveBeenCalled();
     });
 
+    it('should snap to the last page when the final (short) page cannot scroll its first card to the start', () => {
```
Supremeyh, line 587: The added test (`utils-test.tsx` line 584) covers only the intended happy path (short final page, user scrolled to end → indicator advances). There is no test asserting the indicator does **not** jump when all cards are visible but the user has not scrolled the rail to the end.

Without a negative test, the false positive scenario in Issue 1 has no regression guard. Once the logic is tightened (per Issue 1), a negative test locks in the correct boundary and prevents future regressions.

**Suggested addition** (after fixing Issue 1):
```ts
it('should NOT snap to the last page when all cards are visible in a wide viewport', () => {
  // 4 cards, 2 shown per page, wide viewport: all visible from the start.
  // A wheel event should not force the indicator to page 1.
  const allVisibleList = [1, 1, 1, 1];
  // ... render hook with currentIndex: 0, initiallyShownCards: 2
  act(() => { mockContainer.dispatchEvent(new Event('wheel')); });
  rerender({ visibilityList: allVisibleList });
  expect(mockSetCurrentIndex).not.toHaveBeenCalled();
});

steviehailey-skyscanner: test case added

packages/backpack-web/src/bpk-component-card-list/src/BpkCardListRowRail/utils.tsx
steviehailey-skyscanner: Last page guard added

steviehailey-skyscanner (approved):
lgtm

Supremeyh (approved):

PR #562, opened 2018-03-05
[BPK-1388] Aligned native button link styling to agreed standard

matthewdavidson:
- Removed one extraneous view on android
- Removed height restriction, allowing text to wrap
- Added test for theming

PR #4977, opened 2026-07-29
[CLOV-1903] Fix Dependabot alerts - lodash vulnerable to Code Injection

Supremeyh:
## What

Addresses Dependabot alert [#562](https://github.com/Skyscanner/backpack/security/dependabot/562) — lodash CVE-2026-4800 (GHSA-r5fr-rjxr-66jc).

`@figma/code-connect@1.4.9` ships `lodash@4.18.1` natively, resolving the vulnerability without a workspace override.

## Changes

- `package.json` — bumped `@figma/code-connect` from `^1.4.3` to `^1.4.9`
- `pnpm-lock.yaml` — updated lockfile; `lodash@4.17.23` is gone, `@figma/code-connect` now resolves `lodash@4.18.1` directly
- `pnpm-workspace.yaml` — no override needed; unchanged from main

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,6 +6,7 @@ settings:
 
 overrides:
   braces: 3.0.3
+  lodash: ^4.18.1
 
 importers:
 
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
```diff
@@ -6,6 +6,7 @@ settings:
 
 overrides:
   braces: 3.0.3
+  lodash: ^4.18.1
```
IrinaWei: Would it be worth considering updating @figma/code-connect to 1.4.9 instead? That version already depends on lodash@4.18.1, and our existing ^1.4.3 range should allow the upgrade. This might let us address the vulnerability without adding another workspace-wide override. What do you think?

Supremeyh (commented):
Re [r3673028759](https://github.com/Skyscanner/backpack/pull/4977#discussion_r3673028759): Great suggestion — `@figma/code-connect@1.4.9` does ship `lodash@4.18.1` directly (verified via GitHub API), which would make the workspace override unnecessary.

IrinaWei (approved):

PR #4978, opened 2026-07-29
[CLOV-1815][BpkText] Add underline and line clamp controls

IrinaWei:
## Description

Adds text decoration and truncation controls to `BpkText`.

- Adds an `underline` boolean prop, including support for combining underline and strikethrough.
- Adds a `truncate` boolean prop for no-wrap, single-line ellipsis truncation.
- Adds a `lineClamp` prop for configurable single- or multi-line ellipsis truncation.
- Accepts only positive integers for `lineClamp`; zero, negative, fractional, `NaN`, and infinite values are ignored.
- Gives a valid `lineClamp` precedence when both truncation props are supplied.
- Preserves consumer-provided inline styles while using a CSS custom property to configure the clamp count.
- Allows long unbroken content to wrap within a clamped element instead of expanding its available width.
- Adds Storybook examples comparing full text, `truncate`, and `lineClamp` at the same width, including flex-layout and unbroken-text cases.
- Adds unit, accessibility, visual-test, and README coverage for the new behaviour.

Jira: [CLOV-1815](https://skyscanner.atlassian.net/browse/CLOV-1815)

### underline
image

### truncate
image

### lineclamp and comparison between truncate
image

## Testing

- `pnpm exec jest packages/backpack-web/src/bpk-component-text/src/BpkText-test.tsx packages/backpack-web/src/bpk-component-text/src/accessibility-test.tsx --runInBand --watchman=false`
- `pnpm exec eslint packages/backpack-web/src/bpk-component-text/src/BpkText.tsx packages/backpack-web/src/bpk-component-text/src/BpkText-test.tsx packages/backpack-web/src/bpk-component-text/src/BpkText.stories.tsx packages/backpack-web/src/bpk-component-text/src/accessibility-test.tsx`
- `pnpm exec stylelint packages/backpack-web/src/bpk-component-text/src/BpkText.module.scss packages/backpack-web/src/bpk-component-text/src/BpkText.stories.module.scss`
- `pnpm nx run backpack-web:typecheck`
- `pnpm nx run backpack-web:build-sass`

## Checklist

- [x] PR title includes the component name.
- [x] Component `README.md` updated.
- [x] Unit and accessibility tests added.
- [x] Storybook and visual-test examples added.
- [x] Existing default typography and style pass-through behaviour preserved.
- [ ] Manual keyboard, screen reader, and zoom checks.
- [x] Migration guide not required because the change is additive.

[CLOV-1815]: https://skyscanner.atlassian.net/browse/CLOV-1815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/backpack-web/src/bpk-component-text/README.md
```diff
@@ -95,12 +95,56 @@ export default () => (
...
+```
+
+### Truncation Props
+
+Use `truncate` for no-wrap, single-line truncation, or pass a positive integer
+to `lineClamp` to truncate after a specific number of lines. Both need a
+constrained available width. Invalid `lineClamp` values are ignored, and a
+valid `lineClamp` takes precedence when both props are provided.
+
+```javascript
```
Supremeyh: same here

packages/backpack-web/src/bpk-component-text/README.md
```diff
@@ -95,12 +95,56 @@ export default () => (
 );
 ```
 
+### Underline Prop
+
+The `underline` prop renders text with a basic underline. It defaults to
+`false`. The underline uses the current text color, including when the `color`
+prop or dark mode is used.
+
+```javascript
```
Supremeyh: it should use `tsx` instead

packages/backpack-web/src/bpk-component-text/src/BpkText.tsx
```diff
@@ -101,7 +101,17 @@ type Props = {
...
+   * Maximum number of lines to display before truncating the text with an
+   * ellipsis. Must be a positive integer.
+   */
+  lineClamp?: number | null;
   strikethrough?: boolean;
+  /**
+   * Truncates overflowing text to a single line with an ellipsis.
+   */
+  truncate?: boolean;
+  underline?: boolean;
```
Supremeyh, line 115: `lineClamp`  and `truncate`  both carry TSDoc block comments. `underline` added in the same diff hunk has none. Constitution V requires JSDoc/TSDoc for all public props. Add a `/** Renders the text with an underline. */` comment consistent with the `truncate` pattern.

packages/backpack-web/src/bpk-component-text/src/BpkText.stories.module.scss
```diff
@@ -22,4 +22,28 @@
...
+  &_flex-container {
+    justify-content: space-between;
+    align-items: center;
+    gap: tokens.bpk-spacing-sm();
+  }
+}
+
+:where(.bpk-stories-text_flex-container) {
+  display: flex;
 }
```
Supremeyh: The rule `:where(.bpk-stories-text_flex-container) { display: flex; }` uses a literal, unhashed class name. CSS Modules compiles all local class names to `[local]-[hash:base64:5]` at build time, so this selector never matches the runtime class. The `display: flex` declaration is dead code, meaning `justify-content`, `align-items`, and `gap` inside `&_flex-container` (lines 40–44) have no effect — the flex container story is silently broken. Additionally, the `:where()` rule is emitted as a global, unscoped selector.

**Fix:** move `display: flex` into the `&_flex-container` block and remove the `:where()` rule entirely:

```scss
&_flex-container {
  display: flex;           // ← add here
  justify-content: space-between;
  align-items: center;
  gap: tokens.bpk-spacing-sm();
}
```

packages/backpack-web/src/bpk-component-text/src/BpkText.stories.tsx
```diff
@@ -303,6 +415,14 @@ export const Strikethrough = {
   render: () => <StrikethroughExample />,
 };
 
+export const Underline = {
```
Supremeyh, line 426: `Underline`  and `Strikethrough` have standalone named story exports. `truncate` is only demonstrated as a comparison row inside `LineClampExample`, making the prop invisible in isolation in Storybook. A `TruncateExample` component + `export const Truncate` would follow the pattern set by every other BpkText prop.

packages/backpack-web/src/bpk-component-text/src/BpkText.module.scss
```diff
@@ -182,6 +182,36 @@ $text-colors: (
...
+    -webkit-box-orient: vertical;
+    // stylelint-disable-next-line property-no-vendor-prefix
+    -webkit-line-clamp: var(--bpk-text-line-clamp);
+
+    // Retain the standard property for vendor-prefix tooling and future
+    // compatibility. Current browser behaviour relies on the legacy WebKit
+    // declarations above; revisit the display model when standard line-clamp
+    // support is broadly available.
+    line-clamp: var(--bpk-text-line-clamp);
+    overflow-wrap: anywhere;
```
Supremeyh, line 212: here added overflow-wrap: anywhere (allowing for line breaks at any character), which changed the text arrangement behavior of the element, but the README did not mention it. It might come as a surprise to consumers who already have custom line-breaking rules

it can be suggested to add a note in the README to inform consumers that they need to pay attention to the minimum width shrinkage behavior when using Lineclamps on flex/grid sub-items

packages/backpack-web/src/bpk-component-text/src/BpkText.tsx
```diff
@@ -110,28 +120,56 @@ const BpkText = ({
...
+    typeof lineClamp === 'number' &&
+    Number.isInteger(lineClamp) &&
+    lineClamp > 0;
+  const shouldTruncate = truncate && !shouldClamp;
+  const lineClampStyle = shouldClamp
+    ? ({
+        ...style,
+        '--bpk-text-line-clamp': lineClamp,
+      } as CSSProperties)
+    : style;
```
Supremeyh, line 144: '--bpk-text-line-clamp': The lineClamp passes a number, not a string

CSS custom properties in the DOM layer only accepts a string, the React by internal element. The style.css. SetProperty (name, value) set the custom property, will reduce the number implicit coerce into a string, so no problem when running. However, the CSSProperties type does not accept the number value, so the as CSSProperties strong conversion was used to suppress the TS error and cover up the type mismatch.

A more explicit way of writing


```
const lineClampStyle = shouldClamp
  ? ({
    ... style,
    '--bpk-text-line-clamp': '${lineClamp}', // explicitly converts to string
    } as CSSProperties)
  : style;
```

Supremeyh (commented):
Three new public props (`underline`, `truncate`, `lineClamp`) and one new CSS custom property (`--bpk-text-line-clamp`) are introduced. New optional props are a backwards-compatible addition and require a `minor` bump. The current `patch` label will mis-categorise the changelog entry and produce incorrect version increments for consumers.


And it worth add some snapshots in PR description

Supremeyh (approved):

PR #4979, opened 2026-07-29
Fix layout validators to reject negative percentages in spacing props

Faye-Xiao:
## Summary

- `isPercentage` in `tokens.ts` used `-?` in its regex, unintentionally allowing negative percentages to pass validation for `padding`, `margin`, and `gap` props
- Negative percentages are not meaningful CSS spacing values and should be rejected
- `isValidPositionValue` and `isValidSizeValue` intentionally support negative values (e.g. `top: '-50%'` for translate offsets), so they now inline their own `/^-?\d+(\.\d+)?%$/` regex instead of delegating to `isPercentage`

**Files changed:**
- `tokens.ts` — remove `-?` from `isPercentage`; inline negative-% regex in `isValidPositionValue` and `isValidSizeValue`
- `tokenUtils-test.ts` — add tests asserting negative % is rejected for `padding`, `margin`, `gap` and accepted for position props

## Test plan

- [ ] `pnpm run jest --testPathPattern="tokenUtils-test"` — 41 tests pass
- [ ] `eslint` on changed files — no errors
- [ ] Negative % for `padding`/`margin`/`gap` is dropped with a console warning
- [ ] Negative % for `top`/`left`/`right`/`bottom` continues to pass through unchanged

### DIFF
--- a/packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
+++ b/packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
@@ -26,6 +26,10 @@ import {
 import { BpkSpacing } from './tokens';
 
 describe('processBpkProps', () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
   it('converts spacing tokens to rem values', () => {
     const result = processBpkProps({ padding: BpkSpacing.MD });
 
--- a/packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
+++ b/packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
@@ -436,6 +436,12 @@ describe('processBpkProps', () => {
     expect(result.left).toBe('-100%');
   });
 
+  it('passes through non-negative percentage for padding (regression guard)', () => {
+    const result = processBpkProps({ padding: '50%' });
+
+    expect(result.padding).toBe('50%');
+  });
+
   it('rejects negative percentage for padding and drops the prop', () => {
     const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
 
@@ -446,7 +452,8 @@ describe('processBpkProps', () => {
 
     expect(result.padding).toBeUndefined();
     expect(result.paddingTop).toBeUndefined();
-    expect(warnSpy).toHaveBeenCalled();
+    expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('-10%'));
+    expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('-5%'));
     warnSpy.mockRestore();
   });
 
@@ -460,19 +467,25 @@ describe('processBpkProps', () => {
 
     expect(result.margin).toBeUndefined();
     expect(result.marginTop).toBeUndefined();
-    expect(warnSpy).toHaveBeenCalled();
+    expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('-20%'));
+    expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('-5%'));
     warnSpy.mockRestore();
   });
 
-  it('rejects negative percentage for gap and drops the prop', () => {
+  it('rejects negative percentage for gap, rowGap and columnGap and drops those props', () => {
     const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
 
     const result = processBpkProps({
       gap: '-10%' as any,
+      rowGap: '-5%' as any,
+      columnGap: '-5%' as any,
     });
 
     expect(result.gap).toBeUndefined();
-    expect(warnSpy).toHaveBeenCalled();
+    expect(result.rowGap).toBeUndefined();
+    expect(result.columnGap).toBeUndefined();
+    expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('-10%'));
+    expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('-5%'));
     warnSpy.mockRestore();
   });
 });
--- a/packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
+++ b/packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
@@ -422,7 +422,7 @@ describe('processBpkProps', () => {
     expect(result.scrollMarginBottom).toBe('0');
   });
 
-  it('passes through negative percentages for position props (e.g. top: -50% for vertical centering)', () => {
+  it('passes through negative percentages for position props (e.g. top: -50% for translate offsets)', () => {
     const result = processBpkProps({
       top: '-50%',
       left: '-100%',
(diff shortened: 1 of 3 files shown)


packages/backpack-web/src/bpk-component-layout/src/tokens.ts
```diff
@@ -200,25 +202,27 @@ export function isValidMarginValue(value: string): boolean {
  */
 export function isValidSizeValue(value: string): boolean {
   return (
```
Supremeyh, line 208: `isValidSizeValue` comment is misleading (likely a pre-existing bug surfaced here)

```
/^-?\d+(\.\d+)?rem$/.test(value) || // rem values (negative allowed for transforms)
/^-?\d+(\.\d+)?%$/.test(value) ||   // percentage values (negative allowed)
```
`isValidSizeValue` governs width, height, minWidth, maxWidth, etc. Negative values for these properties are invalid CSS — a negative width or minWidth is meaningless. The comment "negative allowed for transforms" doesn't make sense in this context (transform uses isValidTransformValue, not this function). This looks like a copy-paste from the position validator. Worth either:

- Dropping the -? from both lines here (if negative size was never intentional), or
- Adding a note confirming this is intentional with a concrete use-case.

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
```diff
@@ -200,25 +202,27 @@ export function isValidMarginValue(value: string): boolean {
...
-    /^-?\d+(\.\d+)?rem$/.test(value) || // rem values
-    isPercentage(value) || // percentage values
+    /^-?\d+(\.\d+)?rem$/.test(value) || // rem values (negative allowed for transforms)
+    /^-?\d+(\.\d+)?%$/.test(value) || // percentage values (negative allowed)
```
Supremeyh: The `regex /^-?\d+(\.\d+)?%$/` now appears in two places. A named export like:


`const NEGATIVE_PERCENTAGE_RE = /^-?\d+(\.\d+)?%$/;`
would remove the duplication and make the intent self-documenting. Low priority, but worth doing while the change is open.

packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
```diff
@@ -431,6 +431,46 @@ describe('processBpkProps', () => {
...
+  });
+
+  it('rejects negative percentage for gap and drops the prop', () => {
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
+
+    const result = processBpkProps({
+      gap: '-10%' as any,
+    });
+
+    expect(result.gap).toBeUndefined();
```
Supremeyh, line 484: BpkFlex exposes rowGap and columnGap as separate gap props. The new "rejects negative percentage for gap" test only covers gap. If those props go through a different validator path, the fix might not cover them. Worth adding:


```
const result = processBpkProps({ gap: '-10%', rowGap: '-5%', columnGap: '-5%' } as any);
expect(result.rowGap).toBeUndefined();
expect(result.columnGap).toBeUndefined();
```

packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
```diff
@@ -431,6 +431,46 @@ describe('processBpkProps', () => {
...
+    expect(warnSpy).toHaveBeenCalled();
```
Supremeyh: All three new tests assert only expect(warnSpy).toHaveBeenCalled(). If the warning message changes to something uninformative (or is removed), this passes silently. Prefer:


expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('-10%'));
or at minimum toHaveBeenCalledTimes(1) per rejected prop so the count is explicit.

packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
```diff
@@ -422,7 +422,7 @@ describe('processBpkProps', () => {
     expect(result.scrollMarginBottom).toBe('0');
   });
 
-  it('passes through negative percentages for position props (e.g. top: -50% for vertical centering)', () => {
+  it('passes through negative percentages for position props (e.g. top: -50% for translate offsets)', () => {
```
Supremeyh, line 429: There's no test confirming that padding: '50%' still passes after the change. Given this is the exact behavior the fix changes, one positive assertion is cheap insurance.

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
```diff
@@ -200,25 +202,27 @@ export function isValidMarginValue(value: string): boolean {
...
     value === 'auto' ||
     value === 'full' ||
     value === 'fit-content'
```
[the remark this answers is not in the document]
Faye-Xiao: Fixed in the follow-up commit — `isValidSizeValue` now uses `/^\d+(\.\d+)?rem$/` (no `-?`) and delegates to `isPercentage` (non-negative) instead of inlining the signed regex.

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
```diff
@@ -200,25 +202,27 @@ export function isValidMarginValue(value: string): boolean {
  */
 export function isValidSizeValue(value: string): boolean {
   return (
-    /^-?\d+(\.\d+)?rem$/.test(value) || // rem values
-    isPercentage(value) || // percentage values
+    /^-?\d+(\.\d+)?rem$/.test(value) || // rem values (negative allowed for transforms)
```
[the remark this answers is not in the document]
Faye-Xiao: Fixed — the misleading comment and the signed regex in `isValidSizeValue` were both removed in the follow-up commit.

packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
```diff
@@ -431,6 +431,46 @@ describe('processBpkProps', () => {
...
+    const result = processBpkProps({
+      padding: '-10%' as any,
+      paddingTop: '-5%' as any,
+    });
+
+    expect(result.padding).toBeUndefined();
+    expect(result.paddingTop).toBeUndefined();
+    expect(warnSpy).toHaveBeenCalled();
+    warnSpy.mockRestore();
+  });
```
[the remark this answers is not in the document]
Faye-Xiao: Fixed — added `afterEach(() => jest.restoreAllMocks())` to the `processBpkProps` describe block so spy cleanup is guaranteed even if an assertion throws.

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
Faye-Xiao: Fixed — `isValidSizeValue` now uses `/^\d+(\.\d+)?rem$/` (no `-?`) and `isPercentage` (non-negative). Negative size values were never intentional; this was a copy-paste from the position validator.

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
Faye-Xiao: Done — extracted as `NEGATIVE_PERCENTAGE_RE` and used in `isValidPositionValue` to eliminate the duplication.

packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
Faye-Xiao: Done — expanded the gap test to cover `rowGap` and `columnGap` as well.

packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
Faye-Xiao: Done — assertions now use `toHaveBeenCalledWith(expect.stringContaining(value))` so they verify the invalid value appears in the warning message.

packages/backpack-web/src/bpk-component-layout/src/tokenUtils-test.ts
Faye-Xiao: Done — added a dedicated test: `padding: '50%'` passes through as `'50%'` after the change.

packages/backpack-web/src/bpk-component-layout/src/tokens.ts
```diff
@@ -161,21 +161,27 @@ export type BpkResponsiveValue<T> =
   | T
   | Partial<Record<BpkBreakpointToken | 'base', T>>;
 
+// Matches percentage strings that include a leading minus sign, e.g. '-50%', '-0.5%'.
+// Used only by position validators where negative offsets are intentional.
+const NEGATIVE_PERCENTAGE_RE = /^-?\d+(\.\d+)?%$/;
```
Supremeyh: how about rename to `SIGNED_PERCENTAGE_RE` as `-?` let this regex match both positive and negative percentages simultaneously (both 50% and -50% can pass)

Also comment should update
```
// Matches both positive and negative percentage strings, e.g. '50%', '-50%', '-0.5%'.
// Used by position validators where negative offsets are intentional.
const SIGNED_PERCENTAGE_RE = /^-?\d+(\.\d+)?%$/;

```

Faye-Xiao: Good point — renamed to `SIGNED_PERCENTAGE_RE` and updated the comment accordingly.

Supremeyh (approved):

